PR #121, opened 2025-10-28
feat: response api

PierreLeGuen:


crates/api/src/lib.rs
```diff
@@ -57,6 +57,7 @@ pub struct AuthComponents {
 pub struct DomainServices {
     pub conversation_service: Arc<services::ConversationService>,
     pub response_service: Arc<services::ResponseService>,
+    pub response_items_repository: Arc<dyn services::responses::ports::ResponseItemRepositoryTrait>,
```
Evrard-Nil: Why not use a service?

Evrard-Nil (approved):

crates/api/src/lib.rs
PierreLeGuen: The service is response service.

PR #122, opened 2025-10-29
chore: reproducible build

### LINKED ISSUE #77: Enhancement: reproducible docker build
To improve the verifiability of NEAR AI Cloud (as required by some partners), ideally we should make the `cloud-api` docker image reproducible, which means with a given docker image digest on docker registry, anyone could easily find the provenance of the GitHub commit hash for the digest, and can consistently reproduce the build process to generate this docker image digest whenever needed.

Some relevant work can be found in https://github.com/nearai-cloud/nearai-cloud-server/pull/79

think-in-universe:
Resolves #77 

The docker image digest is now reproducible across machines with the command:

```bash
# build the image locally
bash docker-build.sh

# build and push the image
bash docker-build --push nearaidev/cloud-api:<tag>
```

The CI will build and publish docker image for code pushes to `main` branch or creation of release tags (starts with `v`, e.g. `v0.1.2`).

| github ref | image tag | 
| -- | -- |
| branch: `main` | `latest` |
| tag: `v0.1.2` | `0.1.2` |

Some enhancements are completed to make it fully reproducible with artifact attestation:

- [x] 1. Pin the apt package versions;
- [x] 2. Enhance Rust & cargo configs
- [x] 3. Add the GitHub workflow to publish reproducible image digests to docker hub.
- [x] 4. Integrate with GitHub artifact attestation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - chore/repro-build
     tags:
       - 'v*'
 


crates/database/src/migrations/mod.rs
```diff
@@ -1,19 +1,21 @@
 use crate::pool::DbPool;
 use anyhow::{Context, Result};
-use refinery::embed_migrations;
+use refinery::load_sql_migrations;
 use tracing::info;
 
-// Embed migrations from the migrations folder
-embed_migrations!("src/migrations/sql");
```
think-in-universe: @PierreLeGuen We switched from embedding migration to loading SQL migrations because the orders of the SQL files in the binary are non-deterministic across machines. Please help review whether there'll be issues with this change.

crates/database/src/migrations/sql/V0001__initial_schema.sql
think-in-universe: @PierreLeGuen  We also renamed the SQL files to make them ordered by the sequence, but it actually doesn't work with `embed_migrations`. We can change back the names if you think this may bring troubles.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.88.0"
```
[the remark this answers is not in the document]
think-in-universe: Nothing specific. I just keep the Rust version consistent with the Dockerfile.

PR #134, opened 2025-11-03
feat: return model aliases in `/model/list`

alexplash:
GET endpoints for models now returns model aliases.

### DIFF
--- a/crates/database/src/repositories/model.rs
+++ b/crates/database/src/repositories/model.rs
@@ -51,7 +51,7 @@ impl ModelRepository {
                     m.context_length, m.verifiable, m.is_active, m.created_at, m.updated_at,
                     COALESCE(array_agg(a.alias_name) FILTER (WHERE a.alias_name IS NOT NULL), '{}') AS aliases
                 FROM models m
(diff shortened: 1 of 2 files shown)


crates/database/src/repositories/model.rs
```diff
@@ -140,16 +143,35 @@ impl ModelRepository {
             .query(
                 r#"
                 SELECT 
-                    id, model_name, model_display_name, model_description, model_icon,
-                    input_cost_per_token, output_cost_per_token,
-                    context_length, verifiable, is_active, created_at, updated_at
-                FROM models 
-                WHERE model_name = $1 AND is_active = true
+                    m.id, m.model_name, m.model_display_name, m.model_description, m.model_icon,
```
PierreLeGuen: This function is supposed to only get model from its canonical name, not from aliases. What about using this function? https://github.com/nearai/cloud-api/blob/0d59aa6a7115e6b0985583610e0cffc63169742f/crates/database/src/repositories/model.rs#L500

think-in-universe:
Hey Alex @alexplash, could you please have a look at the comments from Copilot?

PR #187, opened 2025-11-18
fix: only include model providers with valid attestation report

think-in-universe:
The PR includes the fixes of:

1. only include the models with valid `/attestation/report` response;
2. The `/attestation/report` endpoint should not return the model attestation reports that contains the error response, e.g.

```json
    {
      "error": {
        "code": null,
        "message": "Client error '400 Bad Request' for url 'http://localhost/GetQuote'\nFor more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
        "param": null,
        "type": "HTTPStatusError"
      }
    },
```

### DIFF
--- a/crates/inference_providers/src/vllm/mod.rs
+++ b/crates/inference_providers/src/vllm/mod.rs
@@ -98,7 +98,7 @@ impl InferenceProvider for VLlmProvider {
             url.push_str(&format!("&nonce={nonce}"));
         }
 
-        if let Some(signing_address) = signing_address {
+        if let Some(ref signing_address) = signing_address {
             url.push_str(&format!("&signing_address={signing_address}"));
         }
 
@@ -115,7 +115,7 @@ impl InferenceProvider for VLlmProvider {
         // Handle 404 responses (expected when signing_address doesn't match)
         if response.status() == 404 {
             return Err(AttestationError::SigningAddressNotFound(
-                response.status().to_string(),
+                signing_address.unwrap_or_default().to_string(),
             ));
         }
 


crates/services/src/inference_provider_pool/mod.rs
```diff
@@ -206,12 +206,51 @@ impl InferenceProviderPool {
...
+                    Ok(report) => {
+                        // Check if the response contains an error field
+                        if report.contains_key("error") {
+                            let error_msg = report
+                                .get("error")
+                                .and_then(|v| v.as_str())
+                                .unwrap_or("unknown error");
+                            tracing::warn!(
+                                model = %model_name,
+                                url = %url,
```
Evrard-Nil: Please avoid logging model's ip in INFO logs and above

crates/services/src/inference_provider_pool/mod.rs
```diff
@@ -473,9 +512,22 @@ impl InferenceProviderPool {
                     .await
                 {
                     Ok(mut attestation) => {
-                        // Remove 'all_attestations' field if present
-                        attestation.remove("all_attestations");
-                        model_attestations.push(attestation);
+                        // Check if the response contains an error field
+                        if attestation.contains_key("error") {
```
hanakannzashi: Just a question, why error attestation data will be returned with 200 status code?

think-in-universe: Good catch. The error handling of the `get_attestation_report` function is problematic. I have fixed it.

think-in-universe:
@PierreLeGuen @Evrard-Nil @hanakannzashi  I have fixed the error handling of `get_attestation_report` function in `crates/inference_providers/src/vllm/mod.rs`. Please help review again. Thanks!

PR #218, opened 2025-11-28
feat: add VPC authentication routes and signature verification

PierreLeGuen:
- Introduced new VPC authentication routes, including a login endpoint that verifies client signatures and generates access tokens.
- Implemented HMAC-SHA256 signature verification for VPC requests, enhancing security.
- Added comprehensive end-to-end tests for VPC login functionality, covering success cases and various failure scenarios.
- Updated dependencies to include `hmac` and `sha2` for cryptographic operations.
- Refactored the AttestationService to support VPC signature verification and manage shared secrets.

### DIFF
--- a/crates/api/tests/e2e_vpc_login.rs
+++ b/crates/api/tests/e2e_vpc_login.rs
@@ -11,20 +11,23 @@ use std::path::PathBuf;
 // VPC Login Tests
 // ============================================
 
-/// Generate a valid VPC signature for testing
-fn generate_vpc_signature(timestamp: i64, secret: &str) -> String {
-    let message = timestamp.to_string();
-    let mut mac =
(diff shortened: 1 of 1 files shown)


crates/services/src/attestation/mod.rs
```diff
@@ -44,6 +49,13 @@ impl AttestationService {
     ) -> Self {
         // Load VPC info once during initialization
         let vpc_info = load_vpc_info();
+
+        // Load VPC shared secret from environment
+        let vpc_shared_secret = std::env::var("VPC_SHARED_SECRET").ok();
```
think-in-universe: Should we load VPC shared secret from `VPC_SHARED_SECRET_FILE`?

PR #223, opened 2025-12-03
fix: non-streaming responses to store signatures

### LINKED ISSUE #211: Signatures are not stored when calling `responses` without stream
After calling `POST /responses` API with `stream: false`, we can not get any signatures from `GET /signature/{response_id}` API, it's because signatures are not stored

danielwpz:
Fixes #211

### DIFF
--- a/crates/api/tests/common/mod.rs
+++ b/crates/api/tests/common/mod.rs
@@ -12,11 +12,11 @@ use std::sync::Arc;
 use tokio::sync::OnceCell;
 
 #[cfg(test)]
+use ed25519_dalek::{Signature as Ed25519Signature, VerifyingKey as Ed25519VerifyingKey};
+#[cfg(test)]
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):

crates/api/src/routes/responses.rs
```diff
@@ -472,6 +472,27 @@ pub async fn create_response(
                     "Created response {} for key {}",
                     response.id, api_key.api_key.created_by_user_id.0
                 );
+
+                // Store signature for non-streaming response
+                let response_id = response.id.clone();
+                let response_json = serde_json::to_string(&response).unwrap_or_default();
```
PierreLeGuen: Let's return this json to the users.

PierreLeGuen (changes requested):

PR #261, opened 2025-12-12
Add billing related api and add more metadata to the models api

henrypark133:
There are two more things we need to provide to hugging face:

1. modify /v1/models such that it includes context length and pricing
2. we have to create a new post api for them to query for billing at an inference id level

The format we expect is as follows:

```
{
    {
      "id": "model-id-0",
      "object": "model",
      "created": 1686935002,
      "owned_by": "organization-owner",
      /// [...] other fields
      "pricing": {
        "input": 0.2, /// Price in US dollars per million input tokens
        "output": 2, /// Price in US dollars per million output tokens
      },
      "context_length": 200000, /// Supported context length in tokens
    },
}
```

The response is also JSON-encoded. The response contains an array of objects specifying the request’s ID and its cost in nano-USD (10^-9 USD).

Copied
 200 OK
Content-Type: application/json

{
    "requests": [
        { "requestId": "deadbeef0", "costNanoUsd": 100 },
        { "requestId": "deadbeef1", "costNanoUsd": 100 },
        { "requestId": "deadbeef2", "costNanoUsd": 100 },
        { "requestId": "deadbeef3", "costNanoUsd": 100 }
    ]
}
This API endpoint will be requested by our system every minute, in batches of up to 10,000 requests.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -984,7 +984,7 @@ pub fn build_billing_routes(
         .with_state(billing_state)
         .layer(from_fn_with_state(
             auth_state_middleware.clone(),
-            auth_middleware_with_api_key,
+            middleware::auth::auth_middleware_with_workspace_context,
         ))
 }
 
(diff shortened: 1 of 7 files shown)


henrypark133:
Removed explicit id logging and added organization level query filter

PR #265, opened 2025-12-14
Add context aware routing.

Evrard-Nil:


crates/inference_providers/src/vllm/mod.rs
```diff
@@ -317,4 +317,38 @@ impl InferenceProvider for VLlmProvider {
         let sse_stream = SSEParser::new(response.bytes_stream(), false);
         Ok(Box::pin(sse_stream))
     }
+
+    /// Tokenize chat messages to get token count
+    async fn tokenize_chat(
```
henrypark133: Can we add a comment here saying that tokenize should work as expected for input text however for each models, it automatically adds [EOS], etc tags to it so the token size might differ slightly and that is to be expected.

PR #274, opened 2025-12-16
Client Disconnect Cost Tracking

henrypark133:
v2 of this PR: https://github.com/nearai/cloud-api/pull/234

  - Enabled continuous_usage_stats - vLLM now sends cumulative token counts in every streaming chunk, not just the final one
  - Track usage on disconnect - Added Drop handler that records usage when clients disconnect mid-stream, using the last received token counts
  - Disconnect-safe response saving - Save partial assistant responses to database before emitting events, ensuring data persists even when client disconnects
  - Added disconnect simulation for testing - Mock provider now supports simulating client disconnects to verify partial data is saved correctly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2047,6 +2047,7 @@ dependencies = [
  "config",
  "deadpool",
  "deadpool-postgres",
+ "dotenvy",
  "native-tls",
  "postgres-native-tls",
  "postgres-types",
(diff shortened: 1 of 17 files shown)


henrypark133:
Addressed comments by Claude for potential race conditions. The other ones are very minor and need to simplify tokenization for mocks.

henrypark133:
Thanks Pierre!

PR #313, opened 2025-12-26
fix: add retry tracking for Brave Search tool calls to reduce false alerts

PierreLeGuen:
LLM-generated tool calls were producing invalid Brave Search query
parameters, causing 422 errors that immediately triggered Datadog alerts.
The errors were already being fed back to the LLM for self-correction, but
each failure was logged at error level.

This change implements retry-aware error handling:
- First 3 failures: Log at warn level (silent for alerts), feed error back to LLM
- After 3 failures: Log at error level (triggers Datadog alerts)
- Reset counter on successful web search

Changes:
- Add web_search_failure_count field to ProcessStreamContext to track consecutive failures
- Implement conditional logging in execute_and_emit_tool_call based on retry count
- Reset failure counter when web search succeeds
- Downgrade Brave API error logging from error to warn level

This allows the LLM to self-correct without triggering alerts for transient
parameter validation issues, while still alerting on persistent failures.

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -845,18 +845,16 @@ impl InferenceProviderPool {
             .await?;
 
         // Store chat_id mapping for sticky routing by peeking at the first event
+        // Must be synchronous to ensure attestation service can find the provider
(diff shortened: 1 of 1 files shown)


think-in-universe (approved):
LGTM!

PR #325, opened 2025-12-31
feat(rate-limit): add dynamic per-organization concurrent request limits

PierreLeGuen:
## Summary

- Add configurable concurrent request limits per organization, replacing the hardcoded value of 64
- Limits are cached with 5-minute TTL to avoid database spam
- Add admin API endpoints for managing limits
- No migration needed (reuses existing `organizations.rate_limit` column)

## Changes

- Add `OrganizationConcurrentLimitRepository` trait for completions service
- Implement cached lookup for per-org limits in `CompletionServiceImpl`
- Add admin API endpoints:
  - `GET /v1/admin/organizations/{id}/concurrent-limit`
  - `PATCH /v1/admin/organizations/{id}/concurrent-limit`
- Add e2e tests for the new endpoints (9 tests)

## Behavior

- **Default**: 64 concurrent requests per org per model (unchanged)
- **Custom**: Set via admin API, stored in `organizations.rate_limit` column
- **Cache**: 5-minute TTL - changes propagate within 5 minutes
- **Fallback**: Uses default 64 if limit is NULL, 0, or on DB error

## API Usage

```bash
# Get current limit
GET /v1/admin/organizations/{org_id}/concurrent-limit

# Update limit (admin only)
PATCH /v1/admin/organizations/{org_id}/concurrent-limit
{ "concurrentLimit": 128 }

# Reset to default
PATCH /v1/admin/organizations/{org_id}/concurrent-limit
{ "concurrentLimit": null }
```

### DIFF
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -18,8 +18,6 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
 
-const DEFAULT_CONCURRENT_LIMIT: u32 = 64;
-
 const FINALIZE_TIMEOUT_SECS: u64 = 5;
 
 type FinalizeFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1070,11 +1070,11 @@ pub fn build_admin_routes(
             axum::routing::patch(update_organization_limits),
         )
         .route(
-            "/admin/organizations/{organization_id}/limits/history",
+            "/admin/organizations/{org_id}/limits/history",
             axum::routing::get(get_organization_limits_history),
         )
         .route(
-            "/admin/organizations/{organization_id}/concurrent-limit",
+            "/admin/organizations/{org_id}/concurrent-limit",
             axum::routing::patch(update_organization_concurrent_limit)
                 .get(get_organization_concurrent_limit),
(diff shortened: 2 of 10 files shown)


crates/services/src/completions/mod.rs
```diff
@@ -432,9 +446,34 @@ impl CompletionServiceImpl {
...
+            .get_with(organization_id, async move {
+                match repo.get_concurrent_limit(organization_id).await {
+                    Ok(Some(limit)) if limit > 0 => limit,
+                    Ok(_) => default_limit, // Use default if NULL or 0
+                    Err(e) => {
+                        tracing::warn!(
+                            organization_id = %organization_id,
+                            error = %e,
+                            "Failed to fetch org concurrent limit, using default"
+                        );
```
henrypark133: Not sure how many different organization we have right now, but this warn log might be somewhat noisy if there is no limit set for a large subset of the organization that have somewhat consistent traffic.

I wonder if this will set off alarms or if datadog is only configured to alarm on error logs.

crates/services/src/completions/mod.rs
```diff
@@ -544,13 +586,13 @@ impl CompletionServiceImpl {
 
         loop {
             let current = counter.load(Ordering::Acquire);
-            if current >= self.concurrent_limit {
+            if current >= limit {
```
think-in-universe: Please notice that the actual concurrent limit can be doubled if we're deploying two cloud-api instances.

PR #349, opened 2026-01-09
feat(admin): add GET /admin/organizations endpoint

PierreLeGuen:
## Summary

Add a dedicated admin endpoint to list all organizations with pagination, fixing the admin UI bug where it showed incorrect organization counts.

## Problem

The admin UI organizations page was fetching users via `/admin/users?include_organizations=true` and extracting organizations client-side. This caused:
- Incorrect total count (showed extracted orgs count from current page, not actual total)
- Improper pagination (paginated by users, not organizations)
- UI showed "Showing 1 to 100 of 100 organizations" even when there were more

## Solution

Add `GET /v1/admin/organizations` endpoint that:
- Lists all organizations with proper pagination (limit/offset)
- Returns accurate total count of organizations
- Includes spend limits and usage data per organization
- Uses `tokio::join!` for parallel query execution

## Changes

- `crates/services/src/admin/ports.rs`: Add `AdminOrganizationInfo` struct and traits
- `crates/database/src/repositories/admin_composite.rs`: Implement repository methods
- `crates/services/src/admin/mod.rs`: Implement service method with parallel queries
- `crates/api/src/routes/admin.rs`: Add endpoint with limit/offset params
- `crates/api/src/models.rs`: Add response models
- `crates/api/src/lib.rs`: Register route

## Design Decisions

- **No search parameter**: Server-side search deferred to separate PR to properly handle SQL escaping and indexing
- **Eventual consistency**: List and count queries run in parallel without transaction - acceptable for admin UI (self-corrects on refresh)
- **Existing indexes**: Queries covered by existing indexes on `organizations(is_active)` and `organization_limits_history(organization_id, ...)`

## Testing

```bash
cargo check --package api  # passes
```

## Related

Admin UI PR: https://github.com/nearai/nearai-cloud-admin-ui/pull/9 (deploy after this)

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -816,8 +816,7 @@ pub async fn list_users(
     tag = "Admin",
     params(
         ("limit" = Option<i64>, Query, description = "Maximum number of organizations to return (default: 100)"),
-        ("offset" = Option<i64>, Query, description = "Number of organizations to skip (default: 0)"),
-        ("search_by_name" = Option<String>, Query, description = "Filter organizations by name (case-insensitive partial match)")
+        ("offset" = Option<i64>, Query, description = "Number of organizations to skip (default: 0)")
     ),
     responses(
         (status = 200, description = "Organizations retrieved successfully", body = ListOrganizationsAdminResponse),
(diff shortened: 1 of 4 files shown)


PierreLeGuen:
## Addressed PR Feedback

All review comments have been addressed in commit 8a4800f:

1. **Sequential DB calls** → Now using `tokio::join!` to execute list and count queries in parallel
2. **LIKE pattern not escaped / Missing index / Duplicated query logic** → Removed `search_by_name` entirely from this PR (deferred to dedicated search PR with proper escaping and indexing)

The code is now simpler with a single clean query path.

PierreLeGuen:
@claude review

PR #396, opened 2026-01-28
External models tools calls

henrypark133:
1. Tool calling support for Anthropic & Gemini - Added full tool/function calling support including  
  request conversion, response parsing, and streaming tool_use events                                  
  2. Refactored external backends - Split anthropic.rs and gemini.rs into mod.rs (HTTP) + converter.rs 
  (format conversion) for better maintainability                                                       
  3. Added thought_signature field - Required for Gemini 3 models to correctly handle tool calls in    
  multi-turn conversations                                                                             
  4. Extended CompletionMessage - Added tool_call_id and tool_calls fields to properly track tool calls
   in message history (required by all providers)                                                      
  5. Fixed agent loop tool handling - Assistant messages now correctly include tool_calls, and tool    
  result messages include matching tool_call_id (required for OpenAI/Anthropic/Gemini APIs)            
  6. New chunk_builder module - Shared helpers for creating OpenAI-compatible streaming chunks,        
  reducing duplication across backends

### DIFF
--- a/crates/services/src/responses/service.rs
+++ b/crates/services/src/responses/service.rs
@@ -1172,11 +1172,18 @@ impl ResponseServiceImpl {
                     })
                     .collect();
 
+            // Defensive: only set tool_calls if non-empty (some providers reject empty arrays)
+            let tool_calls = if completion_tool_calls.is_empty() {
+                None
+            } else {
+                Some(completion_tool_calls.clone())
+            };
+
             messages.push(CompletionMessage {
                 role: "assistant".to_string(),
                 content: stream_result.text.clone(),
                 tool_call_id: None,
-                tool_calls: Some(completion_tool_calls.clone()),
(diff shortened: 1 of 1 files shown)


crates/services/src/completions/mod.rs
```diff
@@ -560,17 +580,36 @@ impl CompletionServiceImpl {
...
+                            thought_signature: tc.thought_signature.clone(),
+                        })
+                        .collect()
+                });
+
+                ChatMessage {
+                    role: match msg.role.as_str() {
+                        "system" => MessageRole::System,
+                        "assistant" => MessageRole::Assistant,
+                        "tool" => MessageRole::Tool,
```
PierreLeGuen: Would it create a conflict for when we set tool to "user" in the above converter?

PR #408, opened 2026-02-02
Improvements for external models

henrypark133:
- add interval refresh for external model updates
- fix issue with sonnet chat completions


henrypark133:
1. model provider pool should run before startup so I don't think this is an issue 
2. postgresql calls currently don't have timeouts. I don't think this is a big concern.

nickpismenkov (approved):

PR #417, opened 2026-02-05
fix: conversation message should include request-level metadata

think-in-universe:
The implementation in #375 works well for item-level metadata, but didn't consider using the request-level metadata  of Responses API, which is actually [what chat-api has implemented](https://github.com/nearai/chat-api/blob/a816fb9de890129fb40cfcbdc00a66fd329b00cc/crates/api/src/routes/api.rs#L1100-L1106). 

- The issue is found when we're testing this from chat-api in this PR: https://github.com/nearai/chat-api/pull/145
- Add the missing tests and fix this issue.

### DIFF
--- a/crates/api/tests/e2e_message_metadata.rs
+++ b/crates/api/tests/e2e_message_metadata.rs
@@ -591,3 +591,106 @@ async fn test_conversation_items_without_metadata() {
         panic!("Expected Message item");
     }
 }
+
+/// Test that item-level metadata takes precedence over request-level metadata
+#[tokio::test]
+async fn test_conversation_items_metadata_precedence() {
+    let (server, _, _mock, _db, _guard) = setup_test_server_with_pool().await;
+    let org = setup_org_with_credits(&server, 10000000000i64).await; // $10.00 USD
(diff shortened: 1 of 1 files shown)


danielwpz (approved):

crates/services/src/responses/service.rs
```diff
@@ -1578,6 +1583,9 @@ impl ResponseServiceImpl {
                         }
                     };
 
+                    // Use item-level metadata if present, otherwise fall back to request metadata
+                    let item_metadata = metadata.or_else(|| request_metadata_clone.clone());
```
Evrard-Nil: Should we do some merging? I.e. inherit metadat?

Evrard-Nil (commented):
Looks pretty good to me a few nits and questions

crates/services/src/responses/service.rs
think-in-universe: I think it's OK to override the request metadata by input item metadata instead of merging. Now we actually only used request metadata in chat-api for conversation sharing.

PR #421, opened 2026-02-08
Feat/record usage api

henrypark133:
- Refactored image usage recording to share helper logic and updated usage service traits/mocks to return UsageLogEntry and support API-driven recording.
- Added a new authenticated /v1/usage endpoint with request/response types, validation, pricing lookup, and usage recording that returns the created log entry. 

---

> [!NOTE]
> **Medium Risk**
> Touches billing/usage accounting and adds a new write endpoint plus a DB uniqueness constraint; mistakes could cause under/over-charging or rejected inserts, though changes are scoped and covered by new E2E tests.
> 
> **Overview**
> Adds a new authenticated `POST /v1/usage` endpoint (and OpenAPI schema) to let clients record chat-completion and image-generation usage directly, with server-side model pricing lookup, request validation, and a tagged-union response.
> 
> Makes usage recording **idempotent per organization** by hashing the caller-supplied `id` into `inference_id`, enforcing it with a new partial unique index, and updating the usage repository/service flow to return the existing record (and skip balance updates + cost metrics) on duplicates; also refactors image usage recording in completion routes to reuse shared helper/timeout+async-retry logic and updates mocks/tests accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c971212acbadddeb4fe4f82c87aaeff2d937106e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/database/src/models.rs
+++ b/crates/database/src/models.rs
@@ -600,6 +600,7 @@ pub struct OrganizationUsageLog {
     pub response_id: Option<ResponseId>,
     /// Number of images generated (for image generation requests)
     pub image_count: Option<i32>,
+    pub was_inserted: bool,
 }
 
 /// Organization balance summary - cached aggregate spending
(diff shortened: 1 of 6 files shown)


henrypark133:
> Issue: The endpoint has no duplicate detection for usage records. Unlike the inference pipeline which generates unique inference_id values for each request, this API allows clients to submit the same usage data multiple times, potentially leading to:

Made external id required so every new usage record has some idempotency protection

think-in-universe:
I think we need to require admin auth for `/v1/usage` endpoint. 

Otherwise any user can change their usage history any time which may introduce some data inconsistency.

PR #529, opened 2026-04-06
feat: TDX attestation verification + TLS SPKI fingerprint pinning

Evrard-Nil:
## Summary

Adds cryptographic verification of inference backend attestation reports and TLS certificate pinning. Previously, cloud-api trusted attestation JSON from backends without verification. Now every backend must prove it runs in a verified TDX enclave with the expected OS image before its TLS certificate is trusted.

## Verification chain (per backend)

| Step | What | How |
|------|------|-----|
| 1. TDX quote | Proves response came from real Intel TDX hardware | `dcap-qvl` verifies Intel signature chain + collateral |
| 2. Debug check | Rejects debug-mode TDX (host can read memory) | Checks `td_attributes` bit 0 |
| 3. Report data binding | Proves TLS fingerprint is bound to the attested enclave | Verifies `report_data[0:32] == SHA256(signing_address \|\| tls_fingerprint)` and `report_data[32:64] == nonce` |
| 4. RTMR3 replay | Proves event log is authentic (not forged) | Replays SHA-384 hash chain from event log entries with `imr == 3`, compares against quote's `rt_mr3` |
| 5. OS image hash | Proves CVM runs expected dstack image | Extracts `os-image-hash` from RTMR3-verified event log, checks against `ALLOWED_IMAGE_HASHES` env var |
| 6. GPU evidence | Proves GPUs are in confidential computing mode | Submits `nvidia_payload` to NVIDIA NRAS API, checks `x-nvidia-overall-att-result` |

## TLS certificate pinning

- Custom `rustls::ServerCertVerifier` (`SpkiFingerprintVerifier`) wraps WebPKI verification and additionally checks the server certificate's SPKI SHA-256 against verified fingerprints
- Rejection happens at the TLS handshake layer — **before any HTTP request data is sent**
- Both the shared `reqwest::Client` and per-completion dedicated clients share the same fingerprint set via `Arc<RwLock<HashSet<String>>>`
- **Fail-closed**: if no attestation can be verified, a sentinel fingerprint exits bootstrap mode and rejects all connections until a future refresh succeeds

## Fingerprint discovery

- **10 parallel** attestation calls per model during provider registration to discover TLS certificates from multiple backends behind L4 load balancing
- Each unique `tls_cert_fingerprint` is independently verified (full chain above) before pinning
- Fingerprints are **cumulative** — only added, never removed. Reused providers across refresh cycles keep all previously verified fingerprints
- Bootstrap mode: empty fingerprint set accepts any valid cert (needed for initial attestation fetch)
- Provider refresh interval changed from 15 minutes to **5 minutes** (discovers new backend certs faster)

## RTMR3 replay (no OS image download)

Instead of using `dstack-verifier` (which downloads OS images to compute expected boot measurements), we replay only RTMR3 from the event log:

1. Parse event log JSON array from attestation response (handles both string and array formats)
2. For each event with `imr == 3`: `RTMR3 = SHA-384(RTMR3 || event.digest)`
3. Compare replayed RTMR3 against the TDX quote's `rt_mr3`
4. If match → event log is authentic → extract `os-image-hash` and `compose-hash`
5. Check `os-image-hash` against allowlist

This is cryptographically sound: the Intel-signed TDX quote contains `rt_mr3`. If our replay matches, every event in the log (including `os-image-hash`) is verified without downloading any OS image artifacts.

## Environment variables

| Variable | Purpose | Default |
|----------|---------|---------|
| `ALLOWED_IMAGE_HASHES` | Comma-separated OS image hash allowlist (hex). Empty = skip check | empty |
| `PCCS_URL` | Intel PCCS URL for TDX collateral fetching | Phala Network PCCS |
| `EXTERNAL_PROVIDER_REFRESH_INTERVAL` | Provider refresh interval in seconds | `300` (was `900`) |

## Integration test results (real production backends)

```
test test_verify_glm5_attestation ... ok        # TDX UpToDate, RTMR3 match, GPU PASS
test test_verify_qwen35_attestation ... ok       # TDX UpToDate, RTMR3 match, GPU PASS
test test_image_hash_rejection ... ok            # Wrong hash rejected after RTMR3 verify
test test_spki_fingerprint_verifier ... ok       # Bootstrap accepts, wrong FP rejects at TLS
```

## Files changed

| File | Change |
|------|--------|
| `crates/inference_providers/src/spki_verifier.rs` | **New** — `SpkiFingerprintVerifier`, SPKI computation, rustls config builder |
| `crates/services/src/attestation/verification.rs` | **New** — `AttestationVerifier` with TDX, RTMR3 replay, image hash, GPU verification |
| `crates/services/tests/attestation_verification_test.rs` | **New** — Integration tests against GLM-5 and Qwen3.5 production backends |
| `crates/inference_providers/src/vllm/mod.rs` | `VLlmProvider` uses custom TLS verifier, fingerprint management methods |
| `crates/services/src/inference_provider_pool/mod.rs` | 10x parallel attestation discovery, verification integration, fail-closed, 30s timeout |
| `crates/config/src/types.rs` | Default refresh interval 900s → 300s |
| `crates/inference_providers/Cargo.toml` | Added `rustls`, `rustls-native-certs`, `x509-parser`, `hex` |
| `crates/services/Cargo.toml` | Added `dcap-qvl` |

## Test plan

- [x] `cargo test --lib --bins` — 193 unit tests pass
- [x] `cargo clippy` — no warnings
- [x] Integration tests pass against GLM-5 and Qwen3.5 production backends
- [x] RTMR3 replay verified against real TDX quotes
- [x] NVIDIA NRAS GPU verification returns PASS
- [x] TLS SPKI pinning: bootstrap accepts, wrong fingerprint rejects
- [x] Image hash allowlist: correct hash passes, wrong hash rejected
- [ ] Deploy to staging with `ALLOWED_IMAGE_HASHES=9b69bb1698bacbb6985409a2c272bcb892e09cdcea63d5399c6768b67d3ff677`
- [ ] Verify logs show "TLS SPKI fingerprints pinned from attestation discovery"
- [ ] Verify inference requests succeed through pinned connections

### DIFF
--- a/crates/services/src/attestation/verification.rs
+++ b/crates/services/src/attestation/verification.rs
@@ -37,15 +37,21 @@ struct EventLogData {
     compose_hash: Option<String>,
 }
 
+/// dstack runtime event type constant (0x08000001).
+const DSTACK_RUNTIME_EVENT_TYPE: u32 = 0x08000001;
+
 /// An entry from the TDX event log.
 #[derive(Debug, Clone, serde::Deserialize)]
 struct EventLogEntry {
     /// SHA-384 digest of the event (hex-encoded).
     digest: String,
+    /// Event type (0x08000001 for dstack runtime events).
+    #[serde(default)]
+    event_type: u32,
     /// Event name (e.g., "os-image-hash", "compose-hash", "app-id").
     #[serde(default)]
     event: String,
-    /// Event payload (hex-encoded or plaintext depending on event type).
+    /// Event payload (hex-encoded raw bytes).
     #[serde(default)]
     event_payload: String,
     /// Which RTMR this event extends (0-3).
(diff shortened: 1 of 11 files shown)


Evrard-Nil:
@claude review

Evrard-Nil:
@claude review

crates/services/src/attestation/verification.rs
```diff
@@ -0,0 +1,540 @@
...
+        let quote_bytes = hex::decode(quote_hex).map_err(|e| {
+            AttestationVerificationError::InvalidFormat(format!("intel_quote hex decode: {e}"))
+        })?;
+
+        let verified_report =
+            dcap_qvl::collateral::get_collateral_and_verify(&quote_bytes, self.pccs_url.as_deref())
+                .await
+                .map_err(|e| {
+                    AttestationVerificationError::TdxVerificationFailed(format!("{e:#}"))
+                })?;
```
hanakannzashi, line 154: Should we check report status and ensure it is `UpToDate`? Refer from [verifiable-ai-sdk](https://github.com/nearai/verifiable-ai-sdk/pull/1/changes#diff-5875b266b61da03b31a680db66dfdfcb0f02c1b5d7e520f678307be60dabaf4aR25)

think-in-universe (approved):
LGTM

We can verify `mr_config` as well, but I don't think this is a blocker and can be implemented in follow-up PRs.

crates/services/src/attestation/verification.rs
```diff
@@ -0,0 +1,567 @@
...
+                ))
+            })?
+        };
+
+        // Replay RTMR3: SHA-384 chain of ALL events with imr == 3.
+        // This includes both boot-time events (system-preparing, app-id, compose-hash,
+        // instance-id, boot-mr-done) and runtime events (mr-kms, os-image-hash,
+        // key-provider, storage-fs, system-ready).
+        let mut rtmr3 = [0u8; 48];
+        for event in &events {
```
hanakannzashi, line 368: At this point we replay RTMR3 from the server-supplied digests, but we later trust `event_payload` to derive `os_image_hash` / `compose_hash` without first validating that each digest was actually computed from `event_type || ":" || event || ":" || event_payload`. That leaves a gap where an attacker can keep the original digests so RTMR3 still matches the quote, while swapping the JSON payload string to an allowed value and bypassing the image-hash gate. We should validate each event before replaying or before consuming the payloads, similar to the in-repo RA-TLS verifier's `event.validate()` step.

crates/services/src/inference_provider_pool/mod.rs
```diff
@@ -1378,33 +1390,133 @@ impl InferenceProviderPool {
...
+                        // mode and rejects all TLS until a future refresh succeeds.
+                        vllm_provider.block_connections();
+                        tracing::warn!(
+                            model = %model_name,
+                            url = %url,
+                            "No TLS fingerprints pinned — provider will reject connections until attestation succeeds"
+                        );
+                    }
+
+                    // Step 2: Fetch signing public keys WITH TLS pinning active.
```
hanakannzashi: This comment says step 2 runs "WITH TLS pinning active", but we are still using the same `VLlmProvider` / pooled `reqwest::Client` that performed the bootstrap discovery requests. Because the SPKI verifier only runs during the TLS handshake, any sockets opened in bootstrap mode remain reusable after `add_verified_fingerprint()` / `block_connections()`, so signing-key fetches and later traffic can still ride pre-pin connections. To make this guarantee true, we need to drain/rebuild the pool or switch to a fresh client/provider after pinning before doing step 2.

crates/services/src/attestation/verification.rs
Evrard-Nil: Done: For runtime events, now recompute SHA384(event_type_le || ":" || name || ":" || payload) and verifies against stored digest before trusting any payload values.

crates/services/src/inference_provider_pool/mod.rs
Evrard-Nil: Done: After pinning, create a new VLlmProvider via new_with_fingerprint_state(), share the same FingerprintState Arc but has a new reqwest Client with an empty connection pool. All signing key fetches open new TLS connections verified against pinned fingerprints.

PR #530, opened 2026-04-07
feat: prefix-aware routing for inference cache optimization

Evrard-Nil:
## Summary

Routes requests with similar prompt prefixes to the same inference backend, maximizing vLLM/SGLang prefix cache hits. Built on top of #529 (merged).

**Problem**: cloud-api creates a new `reqwest::Client` per chat completion. With L4 TLS passthrough (model-proxy), each new client opens a fresh TLS connection routed to a random backend CVM. Requests sharing the same system prompt scatter across backends, missing the prefix cache entirely.

**Solution**: Replace per-completion dedicated clients with a pool of persistent "bucket" clients. A message-level trie maps conversation prefixes to buckets. Same system prompt → same bucket → same TLS connection → same backend → prefix cache hit.

### 1. Message-level trie (`prefix_router.rs`)

Each trie level = one message (hash of `role_tag + content`). On each request:
1. Hash the first N messages (up to `PREFIX_MAX_MESSAGES`, default 8)
2. Traverse the trie following existing edges (read lock, fast path)
3. If the first message (system prompt) is new, take write lock and assign a fresh bucket

**Bucket inheritance**: only root's direct children get fresh buckets. All deeper nodes inherit the parent's bucket. This means requests sharing a system prompt but with different user messages all route to the same bucket:

```
Root (bucket 0)
├── hash("system: You are a Python expert") → bucket 3
│   ├── hash("user: What is a list?") → bucket 3 (inherited)
│   └── hash("user: Explain decorators") → bucket 3 (inherited)
├── hash("system: You are a Rust expert") → bucket 5
│   └── hash("user: What is ownership?") → bucket 5 (inherited)
```

**Read-lock fast path**: since deeper nodes inherit parent buckets, `lookup_readonly` returns the correct bucket even for partially-unseen message sequences. Write lock is only needed when a brand new system prompt appears.

### 2. Bucket clients (`VLlmProvider`)

- `NUM_PREFIX_BUCKETS` (default 64, max 1024) persistent `reqwest::Client` instances
- Each: `pool_max_idle_per_host(1)` — one persistent TLS connection per bucket
- HTTP/2 multiplexing: concurrent same-prefix requests share one connection → guaranteed same backend
- `SharedTlsRoots`: root cert store (~150KB) loaded once and shared via `Arc` across all bucket clients + the general-purpose client. Without sharing, 64 buckets would duplicate ~10MB of root certs.

### 3. Signature pinning

Replaces per-completion dedicated client pattern:

| | Before | After |
|--|--------|-------|
| Completion | `create_dedicated_client()` → new Client | `bucket_clients[prefix_router.route(messages)]` |
| Pending | `pending_clients[request_hash] = Client` | `pending_buckets[request_hash] = bucket_id` |
| Pin | move Client to `signature_clients[chat_id]` | move bucket_id to `signature_buckets[chat_id]` |
| Signature fetch | `signature_clients[chat_id].get(...)` | `bucket_clients[signature_buckets[chat_id]].get(...)` |
| Cleanup | drop Client | remove mapping (bucket client persists) |

## Baseline measurements (staging)

| Model | Cache hit rate | Backend spread | Notes |
|-------|---------------|----------------|-------|
| GLM-5-FP8 | 55% (lucky) / 0% (scattered) | 1 | SGLang 64-token cache page, 90% of tokens cached when same backend |
| Qwen3.5-122B | 0% | 1 | vLLM 2096-token block, prompt too short; needs `--enable-prefix-caching` |

With prefix routing, GLM-5 should consistently hit 90%+ cache rate regardless of concurrent client count, since requests with the same system prompt always hit the same backend.

## Files changed

| File | Change |
|------|--------|
| `crates/inference_providers/src/vllm/prefix_router.rs` | **New** — `PrefixRouter` trie, message hashing, bucket assignment |
| `crates/inference_providers/src/vllm/mod.rs` | Replace `pending_clients`/`signature_clients` with bucket routing, use `SharedTlsRoots` |
| `crates/inference_providers/src/spki_verifier.rs` | Add `SharedTlsRoots` for sharing root cert store across clients |
| `scripts/eval_prefix_cache.py` | **New** — evaluation script for measuring prefix cache hit rates |

## Environment variables

| Variable | Default | Max | Purpose |
|----------|---------|-----|---------|
| `NUM_PREFIX_BUCKETS` | 64 | 1024 | Persistent connections per provider |
| `PREFIX_MAX_MESSAGES` | 8 | — | Trie depth (messages considered for routing) |

## Test plan

- [x] `cargo test --lib --bins` — 193 tests pass (includes prefix router unit tests)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — no warnings
- [x] CI: Lint ✅ Test Suite ✅ security_audit ✅
- [ ] Deploy to staging, run `API_KEY=... scripts/eval_prefix_cache.py` on GLM-5
- [ ] Compare cache hit rate vs baseline
- [ ] Verify signature fetch works (same backend via bucket client)

### DIFF
--- a/crates/inference_providers/src/vllm/mod.rs
+++ b/crates/inference_providers/src/vllm/mod.rs
@@ -1,9 +1,12 @@
+mod prefix_router;
+
 use crate::spki_verifier::FingerprintState;
 use crate::{
     models::StreamOptions, spki_verifier, sse_parser::new_sse_parser, ImageEditError,
     ImageGenerationError, RerankError, ScoreError, *,
 };
 use async_trait::async_trait;
+use prefix_router::PrefixRouter;
 use reqwest::{header::HeaderValue, Client};
 use serde::Serialize;
 use std::collections::HashMap;
@@ -71,19 +74,21 @@ impl VLlmConfig {
 /// Supports both chat completions and text completions with streaming.
 pub struct VLlmProvider {
     config: VLlmConfig,
+    /// General-purpose client for non-completion requests (attestation, models, etc.)
     client: Client,
-    /// Dedicated per-completion clients for connection-pinned signature fetching.
-    /// With L4 TLS passthrough load balancing, each TLS connection is pinned to a
-    /// specific backend. Signature fetches must reuse the same connection that served
-    /// the completion, otherwise they hit a different backend and get 404.
-    ///
-    /// Flow: completion creates a dedicated Client → stored in `pending_clients[request_hash]`
-    /// → pool peeks chat_id from stream → calls `pin_chat_connection(request_hash, chat_id)`
-    /// → moves client to `signature_clients[chat_id]` → `get_signature` uses it.
-    pending_clients: Arc<std::sync::Mutex<HashMap<String, Client>>>,
-    signature_clients: Arc<std::sync::Mutex<HashMap<String, Client>>>,
+    /// Persistent clients indexed by prefix bucket ID. Each maintains a single
+    /// connection pinned to a specific backend via L4 TLS passthrough. Requests
+    /// with the same prompt prefix route to the same bucket → same backend →
+    /// prefix cache hit.
+    bucket_clients: Vec<Client>,
+    /// Prefix router: message-level trie mapping conversation prefixes to bucket IDs.
+    prefix_router: Arc<PrefixRouter>,
+    /// Maps request_hash → bucket_id during streaming (before chat_id is known).
+    pending_buckets: Arc<std::sync::Mutex<HashMap<String, usize>>>,
+    /// Maps chat_id → bucket_id for signature fetching on the correct backend.
+    signature_buckets: Arc<std::sync::Mutex<HashMap<String, usize>>>,
     /// TLS fingerprint verification state (Bootstrap → Pinned or Blocked).
-    /// Shared across the main client and all dedicated per-completion clients.
+    /// Shared across the main client and all bucket clients.
     fingerprint_state: Arc<std::sync::RwLock<FingerprintState>>,
 }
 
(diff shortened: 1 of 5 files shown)


Evrard-Nil:
@claude review

crates/inference_providers/src/vllm/mod.rs
```diff
@@ -304,18 +309,19 @@ impl InferenceProvider for VLlmProvider {
...
+        let bucket_id = self
+            .signature_buckets
             .lock()
             .unwrap_or_else(|e| e.into_inner())
             .get(chat_id)
-            .cloned();
-        let client = dedicated_client.as_ref().unwrap_or(&self.client);
+            .copied();
+        let client = match bucket_id {
+            Some(id) => &self.bucket_clients[id],
```
hanakannzashi, line 342: Bucket-level client mapping does not guarantee signature fetches reuse the same TLS connection. The implementation maps chat_id -> bucket_id and then uses a shared bucket client in get_signature, but pool_max_idle_per_host(1) only limits idle connections, not active ones. Under concurrency (especially with HTTP/1.1), a bucket client may open multiple connections to different backends, so signature requests can hit a different backend and intermittently return 404.

Evrard-Nil: Good catch. The primary mitigation is HTTP/2 multiplexing, CVM nginx negotiates h2 via ALPN, so concurrent requests on the same bucket share one connection (same backend). For the HTTP/1.1 fallback edge case, `get_signature` now retries on 404: first tries the bucket client, and if it gets 404 (wrong backend), retries on the general-purpose client. Also enabled `http2_adaptive_window(true)` on bucket clients.

PR #534, opened 2026-04-11
fix: re-fetch signing keys for reused providers missing pubkey mappings

Evrard-Nil:
## Summary

- On each provider refresh cycle, check if any reused provider (URL unchanged) is missing from `pubkey_to_providers`. If so, re-fetch its signing keys via the attestation endpoint.
- This is a self-healing recovery path: if the initial pubkey fetch fails during startup (transient network error, backend unreachable), subsequent refresh cycles will retry instead of leaving the keys permanently lost.

## Root cause

Confirmed via Datadog logs on `cpu02-cloud-api-prod`:

- **Two `cloud-api` containers** running simultaneously (container IDs `6f5bdef9...` and `45b9e685...`)
- Container A: **18 pubkeys** (all models, including GLM-5 ECDSA `59e5d3f7...`)
- Container B: **16 pubkeys** (missing GLM-5 ECDSA + Ed25519 keys)
- Requests hitting Container B with a GLM-5 pubkey → `Arc::ptr_eq` intersection is empty → **HTTP 421** "The encryption key is no longer valid"
- 10 real 421 errors found between 02:47–03:09 UTC today, all from container `5805230f...` on cpu02-prod, all for `zai-org/GLM-5-FP8` ECDSA

The bug: `load_inference_url_models` only fetches signing keys for **newly created** providers. When a provider is reused (URL unchanged in `inference_url_providers` cache), its existing pubkey mappings are preserved — but if they were never successfully fetched in the first place, they stay permanently missing.

## Test plan

- [x] `cargo test --lib --bins` — 193 tests pass
- [x] `cargo clippy` — no warnings
- [ ] Deploy to staging, verify `pubkey_to_providers state after update` logs show recovery for any missing keys
- [ ] Monitor Datadog for `"Reused providers missing pubkey mappings"` WARN log to confirm the recovery path fires
- [ ] Verify no more 421 errors after deploy

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -1540,7 +1540,61 @@ impl InferenceProviderPool {
         let mut pub_key_updates: Vec<(String, Arc<InferenceProviderTrait>)> = Vec::new();
         let mut new_url_cache: HashMap<String, Arc<InferenceProviderTrait>> = HashMap::new();
 
-        // Reused providers (URL unchanged — keep warm TLS connections)
+        // Reused providers (URL unchanged — keep warm TLS connections).
+        // Check if any reused provider is missing pubkey mappings and re-fetch if so.
+        // This can happen when the initial pubkey fetch failed (e.g., transient network
+        // error during startup) — since reused providers skip pubkey discovery, the keys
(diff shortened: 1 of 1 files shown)


Evrard-Nil:
@claude review

lloydmak99 (approved):

PR #544, opened 2026-04-17
feat: TCP diversity + cumulative attestation discovery

Evrard-Nil:
Follow-up to #543. Staging testing of that PR showed the permanent provider-block deadlock was gone (big win) but revealed two residuals that this PR addresses:

1. **Every model registered with `pinned_fingerprints=1`** regardless of how many backends sit behind the L4 LB. reqwest negotiates HTTP/2 over its warm connection pool, so N staggered calls on one `Client` multiplex onto one TCP connection, hash to one L4 backend, and only that one fingerprint gets pinned. When model-proxy later routes a request to a different backend, TLS is rejected → transient 502s while #537's path evicts and recreates.
2. **GLM-5 and GLM-5.1 sometimes registered with `pub_keys=0`** — the separate post-pin signing-key fetch was doing a second pair of round-trips on the already-pinned connection and occasionally failing, breaking E2EE routing even when non-E2EE traffic worked.

### 1. TCP diversity — fresh `reqwest::Client` per discovery call

Each discovery future now creates its own `VLlmProvider` (and its own `reqwest::Client`) sharing the same `FingerprintState` Arc with peers and the serving provider. Separate clients mean separate TCP handshakes, each hashing independently through L4 — over 5 calls we expect to cover all or most backends behind the LB.

### 2. Extract pubkeys during discovery

The attestation report already contains `signing_public_key` for the requested algo. The discovery futures now alternate `ecdsa` / `ed25519` across the N calls and harvest pubkeys from every verified response. Pubkeys are derived from the TEE compose hash so they're identical across all of a model's backends — first verified response per algo wins. This removes the separate `fetch_signing_public_keys_for_both_algorithms` round-trip from the hot path for the common case.

### 3. Cumulative discovery on reused providers

Every refresh cycle now runs `CUMULATIVE_DISCOVERY_CALLS = 2` fresh-TCP discovery calls against every reused provider. Verified fingerprints accumulate into the shared `FingerprintState` (the `Pinned` set only grows), and pubkeys backfill any provider whose mapping was lost. Within a handful of cycles this covers every backend even when initial discovery only hit one.

### 4. Eviction of `Blocked` state

A reused provider whose `FingerprintState` has transitioned to `Blocked` can't recover in-place — the SPKI verifier rejects every handshake, so cumulative discovery would fail 100%. The refresh now detects this at the start of the reused path and evicts immediately from all three tracking maps (URL cache, fingerprint state map, `model_to_providers`). Next refresh recreates with a fresh `Bootstrap` state.

### 5. Legacy fallback preserved

Reused providers without tracked fingerprint state (test-injected `MockProvider`s, or any pre-upgrade state on the rolling deploy's first cycle) fall through to the existing `fetch_signing_public_keys_for_both_algorithms` path. #537's regression test passes unchanged.

### 6. Logging

- Per discovery call at `debug`: algo, attempt index, `elapsed_ms`, outcome (success/error/timeout).
- Per refresh cycle at `info`: `successful_calls`, `failed_calls`, `new_fingerprints`, `total_pinned`, `pubkey_algos`.
- Cumulative cycle: `debug` if no new fingerprints added, `info` when the pinned set expands.

## Test plan

- [x] `cargo test -p services --lib` — 196 pass (+1 vs main: `test_reused_provider_with_blocked_fingerprint_state_is_evicted`)
- [x] `cargo clippy -p services --lib -- -D warnings` — clean
- [ ] Staging deploy: Datadog `Registered inference_url model` shows `pinned_fingerprints > 1` for multi-backend models (GLM-5, Qwen3.5) after a few refresh cycles.
- [ ] Staging deploy: GLM-5 E2EE works on both cpu01 and cpu02 (was `pub_keys=0` on cpu01) — request with `X-Model-Pub-Key` returns 200, not 502.
- [ ] Staging deploy: over 30+ minutes, confirm no recurring `Provider failed to return attestation report after retries, excluding from pool` warnings for healthy-backend models.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6200,6 +6200,7 @@ dependencies = [
  "rmcp",
  "serde",
  "serde_json",
+ "serde_urlencoded",
  "sha2 0.10.9",
  "sha3",
  "thiserror 2.0.18",
--- a/crates/services/Cargo.toml
+++ b/crates/services/Cargo.toml
@@ -10,6 +10,7 @@ description.workspace = true
 async-trait = "0.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+serde_urlencoded = "0.7"
 uuid = { version = "1.22", features = ["v4", "v5", "serde"] }
 opentelemetry = { version = "0.31", features = ["metrics"] }
 opentelemetry_sdk = { version = "0.31", features = ["rt-tokio", "metrics"] }
--- a/crates/services/src/attestation/verification.rs
+++ b/crates/services/src/attestation/verification.rs
@@ -16,6 +16,15 @@ const NVIDIA_NRAS_URL: &str = "https://nras.attestation.nvidia.com/v3/attest/gpu
 /// this modest to avoid piling attestation work on inference backends.
 pub const ATTESTATION_DISCOVERY_PARALLELISM: usize = 5;
 
+/// Number of cumulative attestation calls per reused provider on each refresh.
+///
+/// Each cycle adds a small number of fresh-TCP discovery calls to a reused
+/// provider, which accumulates new backend fingerprints into the shared
+/// `FingerprintState`. Over several cycles this covers every backend behind
+/// the L4 LB, even when the initial discovery only hit one. Kept small so
+/// steady-state refresh load stays low.
+pub const CUMULATIVE_DISCOVERY_CALLS: usize = 2;
+
 /// Result of verifying an attestation report from an inference backend.
 #[derive(Debug, Clone)]
 pub struct VerifiedAttestation {
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -1,7 +1,7 @@
 use crate::attestation::AttestationVerifier;
 use crate::common::encryption_headers;
 use config::ExternalProvidersConfig;
-use inference_providers::spki_verifier::FingerprintState;
+use inference_providers::spki_verifier::{FingerprintState, SharedTlsRoots};
 use inference_providers::{
     models::{AttestationError, CompletionError},
     AudioTranscriptionError, AudioTranscriptionParams, AudioTranscriptionResponse,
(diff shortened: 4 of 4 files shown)


Evrard-Nil:
Addressed all three items in 11849ca:

1. **Sequential cumulative discovery (critical)** — classified each reused provider upfront (`Blocked` / `Discover` / `LegacyRefetch` / `Skip`) and drove the two work buckets through `buffer_unordered(10)` concurrently via `tokio::join!`. Hit the well-known HRTB inference bug that `.collect::<Vec<_>>()` trips on streams of tuples containing `Arc<dyn Trait>`, so collection is an explicit `while let Some(x) = stream.next().await` loop.

2. **Unverified reports pushed to `attestation_reports`** — fixed by removing the field entirely (next item).

3. **`attestation_reports` field always discarded** — dropped from `DiscoveryOutcome`. Every caller was discarding it; `#[allow(dead_code)]` was the giveaway. Removed allocation and the verify/unverify conflation at once.

196/196 tests pass, fmt + clippy clean. CI rerunning.

(Also: `security_audit` failing on `rustls-webpki` RUSTSEC-2026-0098/0099 is pre-existing on main — the same failure landed on #543 and it was merged.)

Evrard-Nil:
Addressed the Copilot + Gemini reviews in ed6117f. Summary:

**Correctness**

- **Self-blocking (Gemini, critical)** — discovery calls shared the caller's `FingerprintState` Arc, so the first call's pin would reject every peer to a different backend at the TLS-verifier callback. Explains why staging always showed `pinned_fingerprints=1`. Each call now runs with its own isolated `Bootstrap` state; verified fingerprints are merged into the shared accumulator *after* the HTTP call, outside the per-call future.
- **`known_ptrs` mutation race (Copilot, critical)** — the set was mutated as we iterated `reused`, so multi-model-per-provider scenarios misclassified the second model as already mapped. Split into an immutable `mapped_ptrs` snapshot for membership checks and a separate `processed_ptrs` set for per-provider work dedup.
- **Per-algo pubkey completeness (Copilot)** — replaced the binary "has ANY mapping → skip" gate with a per-`(pubkey, provider_ptr)` dedup set so a partial mapping doesn't permanently block the missing algo's backfill.
- **Stale state-map entries (Copilot, Gemini)** — the states map is now pruned to `new_url_cache.keys()` at the end of every refresh.
- **Evicted providers left in `pubkey_to_providers` (Copilot)** — eviction tracks provider pointers and prunes matching entries from the pubkey map in the same atomic write.

**Efficiency**

- **645 reqwest clients per discovery pass (Gemini, high)** — `VLlmProvider` spins up 128 bucket clients per instance; with 5 discovery calls that's ~645 clients per model per refresh just to make 5 GETs. Discovery now builds a minimal `reqwest::Client` directly from a `SharedTlsRoots` (loaded once, stored on the pool) + custom `SpkiFingerprintVerifier`. ~1 client per call.

**Left for follow-up**

- **Per-URL deduplication of discovery across models sharing one URL (Gemini, medium)** — current infra has 1 model per URL, so low value right now. Will revisit if that changes.

Tests: 196/196. fmt + clippy clean.

PR #553, opened 2026-04-29
fix: split vLLM provider timeouts; don't retry timed-out completions

Evrard-Nil:
## Summary

Reasoning models (GLM-5.1, Qwen3.5) routinely take several minutes per non-streaming request. Today, `VLlmProvider` has a single hardcoded **90s** timeout (`crates/inference_providers/src/vllm/mod.rs`), and the pool retries connection/timeout errors **4×** with backoff. The result: a slow completion fails at ~360s with a generic `"The model is currently unavailable"` instead of either finishing or surfacing a clear timeout. Streaming wasn't affected because its timeout only covers TTFB.

This PR makes three changes together:

### 1. Per-op timeouts (configurable via env)

`VLlmConfig` now carries two independent timeouts:

- `completion_timeout_seconds` (default **600s**, env `VLLM_PROVIDER_COMPLETION_TIMEOUT`) — inference paths: chat/text completion, audio, image, embeddings, rerank, score.
- `control_timeout_seconds` (default **90s**, env `VLLM_PROVIDER_CONTROL_TIMEOUT`) — metadata/TTFB paths: models list, attestation report, signature fetch, streaming TTFB.

The legacy positional `timeout_seconds` argument to `VLlmConfig::new` now sets the completion timeout only — control is read independently from env. Existing callers (which pass `None`) get both env-driven values automatically.

### 2. Timed-out completions don't get retried

Added `CompletionError::Timeout { operation, timeout_seconds }`. The non-streaming chat completion path detects `reqwest::Error::is_timeout()` and maps it to this variant. The pool's retry classifier (`retry_with_fallback`) explicitly short-circuits on `Timeout` — and also on string-form `"operation timed out"` / `"timed out after"` inside the legacy `CompletionError::CompletionError`. Connection errors (`connection refused`, `reset`, `broken pipe`) still retry.

Why: a re-send hits the same backend with the same prompt; the model is presumably still chewing on the first request, and retrying a 600s timeout 4× would burn 40 minutes for no gain.

### 3. Cleaner client error

`map_provider_error` maps `Timeout` to HTTP **504** with `"The request timed out waiting for the model to respond. Please try again."` — instead of the sanitized 502 `"The model is currently unavailable."`

## Test plan

- [x] `cargo build --workspace --tests` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test --lib --bins --workspace` — **377 tests pass, 0 failures**, including 9 new tests:
  - `VLlmConfig` env parsing (default / set / positional override / unparseable / negative-clamp)
  - `Timeout` Display content
  - Pool: `Timeout` variant non-retryable
  - Pool: string-form `"operation timed out"` non-retryable
  - `map_provider_error`: `Timeout` → 504
- [ ] Set `VLLM_PROVIDER_COMPLETION_TIMEOUT=600` on cpu01/cpu02 in the cvm-ansible-playbooks deploy and roll out
- [ ] Re-run the failing GLM-5.1 non-streaming repro that previously errored at 363s — should now complete (model finishes ~5 min) or 504 cleanly after 600s with no 4× retry tail

## Behavior change

Before: non-streaming GLM-5.1 with no `max_tokens` fails at ~360s with 502 `"model currently unavailable"`.

After (with `VLLM_PROVIDER_COMPLETION_TIMEOUT=600` deployed):
- If inference completes in <600s → succeeds.
- If it doesn't → fails once at 600s with **504** `"The request timed out waiting for the model to respond."` — no retry amplification.

Defaults are applied even without env override, so the deploy step is just a tunability nicety.


Evrard-Nil:
> a 600s timeout currently triggers a second 600s request before the pool's no-retry logic sees `CompletionError::Timeout`

Caught it — pushed `fee275b`.

Verified the test fails without the guard: 2.006s end-to-end on the unresponsive listener (two timeout cycles) vs. ~1s with the fix. Test asserts both `elapsed < 1700ms` and `accept_count == 1` so a regression on either dimension would catch it.

On the minor field-visibility point — the integration tests in `tests/integration_tests.rs` construct `VLlmConfig` via struct literal, which would break under `pub(crate)`. Leaving it `pub` for now and treating `.completion_timeout()` / `.control_timeout()` as the recommended accessors. Happy to revisit if you'd rather move the integration test off the literal form.

PierreLeGuen (approved):

PR #554, opened 2026-04-29
fix: bump reqwest read_timeout to match completion_timeout (follow-up to #553)

Evrard-Nil:
## Summary

Follow-up to #553. Staging verification showed the per-call timeout fix is live (clean 504, no retry, correct error message), but the request was clipped at **301s** instead of the configured 600s.

**Root cause**: reqwest's `read_timeout` is a per-chunk idle timeout. For non-streaming chat completion the connection is silent the entire inference time — the server computes, then sends the body in one shot. The hardcoded `read_timeout(300)` on bucket clients fires before the per-request `.timeout(600)` budget — and reqwest reports it as `is_timeout()`, so it correctly maps to `CompletionError::Timeout`, just at the wrong deadline.

**Fix**: tie `read_timeout` to the configured completion budget in the two places bucket clients are built:
- `inference_providers::vllm::VLlmProvider::build` (legacy eager-bucket path)
- `services::inference_provider_pool::PoolBackendVerifier::create_verified_client` (production verified-bucket path)

The general-purpose (non-completion) client now uses `control_timeout`, matching its actual usage (models list, attestation, signature fetch).

## Staging evidence (before this PR)

Datadog logs from cpu01-cloud-api-staging at 14:50:00Z right after #553 deployed:

```
ERROR  Provider per-call timeout during chat completion
       timeout_seconds=600 inference_op=chat_completion
WARN   error_detail="Request timed out after 600s during chat_completion"
       attempt=1 retry=0
ERROR  All providers failed for model
       total_attempts=1 providers_tried=1
```

Curl wall time: **301s**. So #553's no-retry + clean-error parts work; only the deadline was clipped by `read_timeout`.

## Test plan

- [x] `cargo build --workspace` clean
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test --lib --bins -p inference_providers -p services` — all 299 tests pass (no regressions)
- [ ] Re-deploy to staging and re-run the failing GLM-5.1 repro: should now run for the full 600s budget (or the model finishes, whichever comes first)

## Why no new unit test

The bug only manifests when `read_timeout < completion_timeout`. After this fix they're tied, so a unit test would need to exercise the production verifier path (which lives in the `services` crate). End-to-end verification on staging is the natural place; #553 already added the regression test for the no-retry behavior.

### DIFF
--- a/crates/inference_providers/src/models.rs
+++ b/crates/inference_providers/src/models.rs
@@ -729,9 +729,12 @@ pub enum CompletionError {
     /// Distinct from connection errors: the request was sent and the server is
     /// (probably) still working — retrying the same request would just hit the
     /// same wall, so the pool treats this as non-retryable.
+    ///
+    /// `operation` is owned (not `&'static str`) so the `Deserialize` derive on
+    /// this enum stays usable from arbitrary input lifetimes.
     #[error("Request timed out after {timeout_seconds}s during {operation}")]
     Timeout {
-        operation: &'static str,
+        operation: String,
         timeout_seconds: u64,
     },
 }
--- a/crates/inference_providers/src/vllm/mod.rs
+++ b/crates/inference_providers/src/vllm/mod.rs
@@ -463,15 +463,21 @@ impl VLlmProvider {
         client_override: Option<&Client>,
     ) -> Result<reqwest::Response, CompletionError> {
         let client = client_override.unwrap_or(&self.client);
+        let ttfb_timeout_secs = self.config.control_timeout_seconds.max(0) as u64;
         let response = tokio::time::timeout(
             self.config.control_timeout(),
             client.post(url).headers(headers).json(params).send(),
         )
         .await
-        .map_err(|_| CompletionError::HttpError {
-            status_code: 504,
-            message: "Timed out waiting for response headers from inference backend".to_string(),
-            is_external: false,
+        // TTFB stalls indicate the same backend is stuck — surface as
+        // `Timeout` (non-retryable in the pool) for consistency with the
+        // non-streaming path. Pre-`Timeout` this was an `HttpError 504` and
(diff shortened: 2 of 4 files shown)


Evrard-Nil:
Pushed `c78ba20` addressing all four review threads:

**@gemini-code-assist** — connect-timeouts now stay retryable in three places:
1. `map_send_err`: `is_timeout && !is_connect` for `Timeout`, else fallthrough.
2. Bucket-clear arm guard: `(!is_timeout || is_connect)` admits connect-timeouts back into the retry path.
3. Pool string-classifier: messages containing both "timed out" and "connect" remain retryable.

Added `test_connect_timeout_string_is_retryable` — proves the pool retries 4 rounds for a connect-timeout-shaped string error.

(I left the `read_timeout` on the general-purpose client at `control_timeout`. That client serves models list / attestation / signature fetch / streaming TTFB only — `prepare_encryption_headers` aside, the slow inference paths (rerank, score, embeddings, audio) all run through the bucket clients which already use `completion_timeout`. If you saw a path I missed, point me at it and I'll move it.)

**@Copilot** — `Timeout.operation` is now `String`. The `&'static str` would have silently broken `Deserialize` for the enum. Updated all four call sites + tests.

Test FD leak: acceptor task now owns the held sockets and `acceptor.abort()` at test end drops them cleanly.

**@claude (code review)** — streaming TTFB timeout now maps to `CompletionError::Timeout { operation: "chat_completion_stream", ... }` instead of retryable `HttpError 504`. Makes streaming TTFB stalls fail once cleanly, consistent with the non-streaming path. TODO comment added on `with_clean_timeout_env` for the `env::set_var` Rust 1.81+ migration.

300 unit tests pass; clippy + fmt clean.

PR #559, opened 2026-05-01
docs: add local-development & manual-testing guide

Evrard-Nil:
## Summary
- New `docs/local-development.md`: end-to-end walkthrough of running cloud-api locally and exercising the API with `curl` — mock auth flow (`Bearer rt_…` + `User-Agent: Mock User Agent`), org → workspace → API key path, and pointing local cloud-api at production or local inference backends.
- README now links the new doc next to the existing `make dev` quick-start.
- Calls out the common gotchas I keep tripping on: API-key cache delay, `/v1/attestation/report` 503 locally, `inferenceUrl` camelCase requirement, seed dir relative path.

No code changes.

## Test plan
- [ ] `make dev` from repo root, then run the §4 walkthrough — confirm the seeded admin can create an org, workspace, and API key.
- [ ] Hit `/v1/models` with the new key and confirm 401 → 200 within ~10s (cache delay).
- [ ] Open `/docs` in a browser, confirm Scalar UI loads.
- [ ] (Optional) Set `MODEL_DISCOVERY_SERVER_URL` to a real backend per §5 and confirm `/v1/chat/completions` streams.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ A Rust-based cloud API for AI model inference, conversation management, and orga
    - Seeds the database with development data
    - Starts the API server on http://localhost:3000
 
+For an end-to-end walkthrough — mock auth, creating an org/workspace/API key,
+and exercising chat completions with `curl` — see
+[`docs/local-development.md`](docs/local-development.md).
+
 ## Testing
 
 ### Prerequisites for Testing
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -24,10 +24,10 @@ real inference backends.
 docker compose up -d postgres
 
 cp env.example .env
-# The defaults in env.example match docker-compose.yml. The only change
-# you need for a hello-world session is to enable mock auth:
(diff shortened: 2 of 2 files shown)


Evrard-Nil:
Thanks for the review — both findings were valid, addressed in d7b979f:

1. **§8 repro scripts** — confirmed only `test_signature.sh` is on `main`; the other three live on my machine but were never committed. Trimmed §8 to just `test_signature.sh` and fixed the invocation (it takes positional `base_url`/`api_key`, not env overrides).
2. **`AUTH_MOCK` default** — rewrote the §2 callout to be explicit that `env.example` ships with `AUTH_MOCK=false` and the reader needs to flip it.

Also picked up the non-blocking §5 Option B hint pointing at `crates/database/src/migrations/sql/`.

Left the §5 Option A `cloud-api.near.ai` URL as-is — happy to add an env-caveat sentence if you'd like.

lloydmak99 (approved):

PR #560, opened 2026-05-01
fix: H2 keepalive on bucket clients to prevent signature 404s

Evrard-Nil:
## Summary

Multi-host models (GLM-5, GLM-5.1, Qwen3.5) see ~170 `Failed to store chat signature` errors per day in prod. They all collapse to the same race:

1. cloud-api routes a completion to backend **X** via the bucket's pinned H2 connection.
2. The bucket sits idle between chats.
3. CVM nginx hits `keepalive_timeout` (default 75s) and sends GOAWAY.
4. Next request opens a fresh TCP, model-proxy's L4 LB picks backend **Y**.
5. inference-proxy on **Y** has no `chat_id` in its in-memory cache → `404 "Chat id not found or expired"`.

Per the inference-proxy `ChatCache` (`inference-proxy/src/cache.rs`), signatures are not replicated across backends — stickiness is the only correctness guarantee, and it relies on the H2 connection staying up.

Server-side counts from Datadog (24h, prod):

| Model | UUID | Hosts | Failures/24h |
|---|---|---|---|
| GLM-5.1 | `5e07b80f…` | 3 | **111** |
| GLM-5 | `2c164ce3…` | 2 | **46** |
| Qwen3.5-122B | `300868ee…` | 2 | **10** |

Single-host models don't hit this — only multi-host. That confirms the root cause is the L4 LB picking a different backend on reconnect.

## Change

Add H2 keepalive PINGs to the bucket clients in both creation paths:

- **Legacy eager path**: `crates/inference_providers/src/vllm/mod.rs` (used in tests / non-TEE).
- **Verified lazy path**: `crates/services/src/inference_provider_pool/mod.rs::PoolBackendVerifier::create_verified_client` (the prod path).

Settings:

```rust
.http2_keep_alive_interval(Duration::from_secs(30))
.http2_keep_alive_timeout(Duration::from_secs(10))
.http2_keep_alive_while_idle(true)   // critical — PINGs even with 0 streams
.tcp_keepalive(Duration::from_secs(30))
.pool_idle_timeout(None)             // liveness comes from PINGs
```

`http2_keep_alive_while_idle(true)` is the key one for our pattern. Without it, PINGs only fire when a stream is active. Between chats the bucket has 0 streams → no PINGs → nginx kills the connection at the 75s idle timeout. With it, PINGs fire every 30s and reset nginx's idle timer indefinitely.

## Companion change

A matching PR in `cvm-compose-files` bumps `keepalive_timeout` and `keepalive_requests` on the multi-host models' nginx configs so the server side cooperates (and so a single H2 connection isn't capped at the default 1000 requests).

## Test plan

- [x] `cargo check -p inference_providers -p services` — clean
- [x] `cargo clippy -p inference_providers -p services` — clean
- [x] `cargo test -p inference_providers --lib` — 100 passed
- [x] `cargo test -p services --lib` — 202 passed
- [ ] Deploy to staging, run a long idle test against `glm-5-1.completions.near.ai` (>5 min between completion and signature) and confirm no 404
- [ ] After prod deploy, re-aggregate `@fields.message:"Failed to store chat signature"` over 24h and confirm drop

## Risk

Low. The new options only affect existing reqwest clients — they don't change call sites, signatures, or behavior on the happy path. Worst case if H2 PINGs misbehave: connections close more aggressively (PING timeout 10s), forcing reverification — same path we hit today on connection drop, just earlier.

### DIFF
--- a/crates/inference_providers/src/bucket_keepalive.rs
+++ b/crates/inference_providers/src/bucket_keepalive.rs
@@ -1,0 +1,243 @@
+//! Shared HTTP/2 keepalive configuration for bucket-style sticky clients.
+//!
+//! Multi-host inference models (e.g. GLM-5, GLM-5.1, Qwen3.5) sit behind
+//! `model-proxy`'s L4 TLS passthrough, which load-balances per TCP connection.
+//! `cloud-api` uses one `reqwest::Client` per prefix bucket and relies on the
+//! H2 multiplexed connection staying up so completion + signature fetch land
+//! on the same backend. If the connection drops, the L4 LB may pick a
+//! different backend and the signature fetch returns 404 "Chat id not found".
+//!
+//! These settings keep the connection alive across long idle gaps:
+//! - PINGs every 30s (and crucially `while_idle`, so PINGs fire with 0 streams)
+//! - Close fast (10s timeout) on a dead connection so re-verification can run
+//! - TCP keepalive as a backstop for silently-dropped paths
+//! - `pool_idle_timeout(None)` — liveness comes from PINGs, not a wall-clock
+//!   timer that fires regardless of health
+//!
+//! Both creation paths (legacy eager in `vllm/mod.rs` and verified lazy in
+//! `services::inference_provider_pool::PoolBackendVerifier`) call [`apply`]
+//! so the configuration cannot drift between them.
+
+use std::time::Duration;
+
+use reqwest::ClientBuilder;
+
+/// HTTP/2 PING interval. Must be shorter than the server's `keepalive_timeout`
+/// (CVM nginx is set to `1h` in cvm-compose-files; nginx default is 75s).
+pub const H2_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(30);
+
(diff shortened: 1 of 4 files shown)


Evrard-Nil:
Addressed both review threads in ca793ab:

- Extracted the duplicated H2 keepalive options into `inference_providers::bucket_keepalive::apply()` (with public `H2_KEEPALIVE_INTERVAL`, `H2_KEEPALIVE_TIMEOUT`, `TCP_KEEPALIVE` constants). Both call sites — `vllm/mod.rs` and `PoolBackendVerifier::create_verified_client` — now go through the same helper, so the configuration cannot drift.
- Added a guard test `idle_bucket_client_sends_h2_ping` that addresses the load-bearing-flag concern (claude-review): it stands up a minimal raw-frame TCP server speaking h2c, drives one request via `apply_with()`, then waits for the connection to go idle (0 streams) and asserts a PING frame arrives. Without `http2_keep_alive_while_idle(true)`, no PING fires when there are no open streams and the test times out.

lloydmak99 (approved):

PR #561, opened 2026-05-01
debug: log NVIDIA NRAS per-check failures on verdict=FAIL

Evrard-Nil:
## Summary

When NRAS rejected GPU evidence, we surfaced only `"NVIDIA attestation verdict: FAIL (expected PASS)"` — discarding the JWT claims that name *which* sub-check failed and the GPU/driver/VBIOS versions involved. Today's prod incident (~242 verdict-FAIL events between 11:46 and 14:42 UTC, across 7 different models on multiple GPU hosts) was effectively undiagnosable from the logs alone — we couldn't tell whether NVIDIA pushed a new RIM that doesn't cover our drivers, our cert chain broke, or evidence was malformed under load.

After this change, the same scenario produces:

- A structured `tracing::warn!` with `verdict`, `failed_checks` (e.g. `["x-nvidia-gpu-driver-rim-fetched", "x-nvidia-gpu-vbios-rim-version-match"]`), `gpu_arch`, `driver_version`, `vbios_version`, and the raw `x-nvidia-attestation-detailed-result` object.
- An error string that propagates into `provider_message` with the failed-check list inline, so the existing Datadog `status:error` query already shows it without a separate facet.

NRAS JWT claims contain only GPU/firmware metadata (no customer data), so this is privacy-safe under the cloud-api logging rules.

## Implementation notes

- Refactored `extract_nvidia_verdict` → `parse_nras_jwt`, which now returns `(verdict, claims_map)` instead of just the verdict string.
- `nras_failed_checks` walks both flat `x-nvidia-*: false` top-level claims **and** the nested `x-nvidia-attestation-detailed-result` object — different NRAS versions emit one or the other.
- 6 unit tests cover: PASS-boolean, FAIL-string, missing-verdict error, flat-bool collection, nested-object collection, and the all-pass empty case.

## Test plan

- [x] `cargo build -p services`
- [x] `cargo test -p services --lib attestation::verification` (6/6 pass)
- [x] `cargo clippy -p services --lib --tests` (no warnings)
- [ ] Wait for next NRAS verdict=FAIL recurrence and confirm the new structured log shows the per-check names

### DIFF
--- a/crates/services/src/attestation/verification.rs
+++ b/crates/services/src/attestation/verification.rs
@@ -561,20 +561,65 @@ impl AttestationVerifier {
             })?;
 
         // Decode JWT payload (second segment, base64url-encoded)
-        let verdict = extract_nvidia_verdict(jwt_token)?;
+        let parsed = parse_nras_jwt(jwt_token)?;
 
-        if verdict != "PASS" {
-            return Err(AttestationVerificationError::GpuVerificationFailed(
-                format!("NVIDIA attestation verdict: {verdict} (expected PASS)"),
-            ));
+        if parsed.verdict != "PASS" {
+            let failed_checks = nras_failed_checks(&parsed.claims);
+            let gpu_arch = nras_str_claim(&parsed.claims, "x-nvidia-gpu-arch");
+            let driver_version = nras_str_claim(&parsed.claims, "x-nvidia-gpu-driver-version");
+            let vbios_version = nras_str_claim(&parsed.claims, "x-nvidia-gpu-vbios-version");
+            let detailed = parsed
+                .claims
+                .get("x-nvidia-attestation-detailed-result")
+                .map(|v| v.to_string())
+                .unwrap_or_default();
+
+            // Log full diagnostic context once. Caller already wraps the
+            // returned error in a higher-level "Failed to create verified
+            // client" message, so the structured fields here are the only
+            // place the per-claim breakdown surfaces.
+            tracing::warn!(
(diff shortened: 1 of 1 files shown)


Evrard-Nil:
Pushed 0dea536 addressing all the review feedback:

- **gemini** (avoid map clone + no `_` catch-all on enums): `parse_nras_jwt` now pattern-matches `Value::Object(map)` to take ownership directly; sibling variants enumerated explicitly. Same treatment for the verdict match.
- **Copilot + claude-bot** (broad `contains("detailed-result")`): switched to exact `key == "x-nvidia-attestation-detailed-result"`. New test `nras_failed_checks_ignores_other_objects_named_with_detailed_substring` pins the precision.
- **Copilot** (eager `to_string()` on detailed JSON): dropped — the log field is now `detailed_result = ?option_ref` so tracing Debug-formats lazily, no allocation when filtered out.
- **claude-bot** (`dedup` was a no-op given the prefixed format): nested entries now emit the bare check name, so flat and nested forms actually collapse. New test `nras_failed_checks_dedupes_flat_and_nested_overlap` covers the case where NRAS emits both.
- **claude-bot** (`nras_str_claim` allocates): now returns `Option<&str>`.

8/8 unit tests pass, fmt + clippy clean.

The `security_audit` failure on the previous run is unrelated — two `hickory-proto` advisories (RUSTSEC-2026-0118, -0119) were published 2026-05-01 and hit any PR opened against `main` today. Needs a separate deps bump to clear.

PR #575, opened 2026-05-06
feat: include user's role in organization API responses

### LINKED ISSUE #436: GET /v1/organizations/{id} should return the user's role
## Problem

The `GET /v1/organizations/{org_id}` endpoint returns `OrganizationResponse` which does not include the requesting user's role in the organization. The same issue exists for `GET /v1/organizations` (list).

The frontend `Organization` type expects a `role` field (`"owner" | "admin" | "member"`), but it's never populated by these endpoints. As a workaround, the frontend currently looks up the role from the `/v1/users/me` response (`UserOrganizationResponse`), which does include roles.

## Expected

Both endpoints should return the user's membership role, similar to how `/v1/users/me` constructs `UserOrganizationResponse` by calling `get_user_role()` for each organization.

## Relevant code

- `OrganizationResponse` struct: `crates/api/src/models.rs` (missing `role` field)
- `UserOrganizationResponse` struct: `crates/api/src/models.rs` (has `role` field — reference)
- GET org endpoint: `crates/api/src/routes/organizations.rs`
- `get_user_role()`: `crates/services/src/organization/mod.rs`
- Conversion: `crates/api/src/conversions.rs` (`services_org_to_api_org` — doesn't set role)

PierreLeGuen:
## Summary
- Adds `role` (`owner`/`admin`/`member`) to `OrganizationResponse` for organization create/list/get/update responses.
- Uses a role-aware organization listing query so `GET /v1/organizations` returns membership roles without an N+1 role lookup loop or silent partial pages.
- Returns a server error if the caller role cannot be loaded after `PUT /v1/organizations/{id}` instead of guessing `owner`.
- Keeps OAuth behavior out of this PR after rebasing onto current `main`; OAuth review comments are stale against the current diff.

Closes #436

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo check -p services -p database -p api`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test -p services organization --lib`
- [x] `cargo test -p services auth --lib`
- [x] `cargo test -p api --lib`
- [x] `cargo test -p database --lib`
- [x] `git diff --check`
- [ ] E2E create/list/get/update role-response scenarios not run locally; covered by API/service/database unit coverage and CI.

### DIFF
--- a/crates/api/src/routes/organizations.rs
+++ b/crates/api/src/routes/organizations.rs
@@ -69,7 +69,7 @@ pub async fn list_organizations(
     match app_state
         .organization_service
         .list_organizations_for_user(
-            user_id,
+            user_id.clone(),
             params.limit,
             params.offset,
             params.order_by.map(From::from),
@@ -79,10 +79,19 @@ pub async fn list_organizations(
     {
         Ok(organizations) => {
             debug!("Found {} organizations for user", organizations.len());
-            let org_responses: Vec<OrganizationResponse> = organizations
-                .into_iter()
-                .map(crate::conversions::services_org_to_api_org)
-                .collect();
+            let mut org_responses: Vec<OrganizationResponse> = Vec::new();
+            for org in organizations {
+                let role = match app_state
+                    .organization_service
+                    .get_user_role(org.id.clone(), user_id.clone())
+                    .await
+                {
+                    Ok(Some(role)) => crate::conversions::services_role_to_api_role(role),
+                    _ => continue,
+                };
+                org_responses
+                    .push(crate::conversions::services_org_to_api_org(org, role));
+            }
 
             Ok(Json(ListOrganizationsResponse {
                 organizations: org_responses,
@@ -216,7 +225,10 @@ pub async fn create_organization(
                 }
             }
 
-            Ok(Json(crate::conversions::services_org_to_api_org(org)))
+            Ok(Json(crate::conversions::services_org_to_api_org(
+                org,
+                crate::models::MemberRole::Owner,
+            )))
         }
         Err(OrganizationError::InvalidParams(msg)) => {
             debug!("Invalid organization creation params: {}", msg);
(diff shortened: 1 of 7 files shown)


crates/api/src/routes/organizations.rs
```diff
@@ -331,12 +346,12 @@ pub async fn get_organization(
                 )),
             ))
         }
-        Err(e) => {
-            error!("Failed to check organization membership: {}", e);
+        Err(_) => {
```
PierreLeGuen: Why ignoring the error?

crates/api/src/routes/organizations.rs
```diff
@@ -281,46 +293,49 @@ pub async fn get_organization(
...
+        Err(OrganizationError::NotFound) => {
+            return Err((
+                StatusCode::NOT_FOUND,
+                Json(ErrorResponse::new(
+                    "Organization not found".to_string(),
+                    "not_found".to_string(),
+                )),
+            ));
+        }
+        Err(_) => {
```
PierreLeGuen: Why not using the error?

Evrard-Nil:
@PierreLeGuen pls fix lint

Evrard-Nil (approved):

PierreLeGuen:
Review triage after rebasing and patching:

- Fixed the `list_organizations` N+1 and silent-drop issues in `c1a30ab`: the route now calls `list_organizations_with_roles_for_user`, backed by one repository query that returns each organization and the caller's membership role together.
- Fixed the `update_organization` privilege-misrepresentation issue in `c1a30ab`: if the caller role cannot be reloaded after update, the endpoint returns `500` instead of defaulting to `owner`.
- The OAuth/login comments are stale for the current PR diff after rebase. This branch no longer changes `crates/database/src/repositories/user.rs` or OAuth lookup behavior; the only auth-module diff is test-stub plumbing for the organization service trait.
- The remaining double-fetch note on `get_organization` is a performance cleanup, not a correctness/merge blocker. I left it out of this PR to keep the fix scoped to returning role data safely.

Local validation passed:
- `cargo fmt --check`
- `cargo check -p services -p database -p api`
- `cargo test -p services organization --lib`
- `cargo test -p services auth --lib`
- `cargo test -p api --lib`
- `cargo test -p database --lib`
- `git diff --check`

PierreLeGuen:
@claude review

PierreLeGuen:
The new owner-join concern was relevant. Fixed in `c898b6a`:

- The role-aware list query now uses a `LEFT JOIN LATERAL (...) LIMIT 1` for the owner row, so anomalous multiple-owner rows do not duplicate organizations or break pagination.
- Missing-owner rows are no longer silently excluded by the join; they are surfaced as a data-conversion error, matching the previous behavior of `db_to_domain_organization`.

Local validation passed again:
- `cargo fmt --check`
- `cargo check -p database -p services -p api`
- `cargo test -p database --lib`
- `git diff --check`

PR #585, opened 2026-05-12
feat: x-auto-redact for chat completions

Evrard-Nil:
## Summary

Opt-in PII redaction on `/v1/chat/completions`. The user enables it with either:

- HTTP header: `x-auto-redact: on`
- Body field: `"auto_redact": true`

Both spellings are equivalent. The body field is stripped before forwarding so strict-schema providers (Anthropic) don't 422.

## How it works

1. The handler invokes the privacy-filter model via `pool.privacy_classify` (the endpoint added in #584) on the user's message text.
2. PII spans are replaced with stable placeholders: `<email1>`, `<phone1>`, `<account1>`, etc. Same literal PII within one request reuses the same placeholder (dedup). Ordinals are monotonic per category. If the user's own text already contains a `<categoryN>`-shaped literal, the allocator skips that ordinal to avoid collision.
3. The provider — vLLM or 3p — only ever sees the redacted form. **Provider-agnostic by design.**
4. The response is un-redacted before being returned to the client:
   - **Non-streaming**: walks `choices[*].message.content`, `reasoning_content`, and `reasoning`. Note: when auto-redact is on we re-serialize the response (drops the raw_bytes signing path — documented as an explicit tradeoff for the privacy guarantee).
   - **Streaming**: wraps each SSE chunk with a sliding-window unredacter. A 16-byte tail buffer holds any in-progress `<…>` token across chunks so a placeholder split mid-flight is never emitted partial.

## Fail-closed behavior

If the privacy-filter model is unreachable, errors, or isn't registered in the cloud-api models DB, the handler returns 503 with `\"type\": \"auto_redact_unavailable\"` rather than silently degrading to send raw PII upstream. This honors the privacy guarantee the user opted into.

## Out of scope (deferred)

- **/v1/completions** — the legacy handler currently returns 501. The new `maybe_redact` helper is generic over `ServiceCompletionRequest` so wiring it in is a 5-line copy when that handler is enabled.
- **/v1/responses** — depends on a separate conversation-history persistence decision (design memo discussed: store original PII in DB, redact fresh on each provider call).
- **Category opt-out filter** — all-or-nothing for v1. Easy to add via header `x-auto-redact-categories: …` if needed.
- **`raw_bytes` signing parity** with auto-redact on — the response is re-serialized post-unredact, breaking byte-for-byte signature verification. The user explicitly opted into munging the response; we accept this. (Signature verification still works when auto-redact is off.)
- **Token billing on the redacted form** is already what happens since the provider sees the redacted prompt and reports its own token usage — slightly cheaper for PII-heavy prompts.

## Module layout

\`\`\`
crates/services/src/auto_redact/
  mod.rs              — public API: is_enabled, redact_messages, strip_body_field
  placeholders.rs     — RedactionMap (bidirectional, collision-safe minting)
  detect.rs           — call privacy-filter via pool.privacy_classify
  apply.rs            — walk CompletionMessage content (string + parts arrays)
  stream_unredact.rs  — sliding 16-byte tail for SSE
\`\`\`

## Tests

- **34 unit tests** in services/auto_redact covering placeholder dedup, collision avoidance, span application, stream chunk-splitting, UTF-8 boundaries, hallucinated placeholders, etc.
- **7 e2e tests** in crates/api/tests/e2e_all/auto_redact.rs covering:
  - Header off → raw PII reaches provider
  - Header \"on\" → provider sees `<email1>`, client gets original back
  - Body field `auto_redact: true` is equivalent to header
  - Body field is stripped from \`extra\` before forwarding
  - Streaming: placeholder split across SSE chunks is reassembled
  - Fail-closed when PII model is unavailable (503 `auto_redact_unavailable`)
  - Header `off` is a true no-op
  - Multi-PII per message (email + phone + SSN) round-trips correctly

The MockProvider's `privacy_classify_raw` does shape-based detection (email, SSN, phone regex) so e2e tests exercise the full pipeline without a live privacy-filter model.

## Test plan

- [x] \`cargo check --workspace --all-targets\`
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\`
- [x] \`cargo fmt --all -- --check\`
- [x] \`cargo test --lib --bins -p services -p inference_providers -p api\` (258 unit pass)
- [x] \`cargo test --test e2e_all auto_redact -- --test-threads 1\` (7 e2e pass)
- [ ] Staging E2E: enable on a chat completion against a 3p model (e.g. anthropic/claude-sonnet-4-5) and verify provider logs don't contain the PII

Builds on #584. The privacy-filter model must be registered in the cloud-api models DB (already done on staging + prod after #584).

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -10,6 +10,7 @@ use axum::{
     response::{IntoResponse, Json as ResponseJson, Response},
 };
 use futures::stream::StreamExt;
+use services::auto_redact::{self, AutoRedactError, RedactionMap, StreamUnredact};
 use services::common::encryption_headers as service_encryption_headers;
 use services::completions::{
     hash_inference_id_to_uuid,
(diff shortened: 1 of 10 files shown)


Evrard-Nil:
Pushed `9107ea7` addressing the bot review. Walk-through:

## Security-critical / blocking

**`redact_one` silently passed through on malformed input** (gemini critical, Copilot):
Now returns `AutoRedactError::Internal` on:
- Span offsets that don't land on UTF-8 char boundaries (e.g. mid-multibyte-sequence)
- `span.end > text.len()` or inverted bounds
- Final cursor not on a char boundary

Propagated through `redact_messages` so the handler responds 500 rather than letting raw PII through. Two new unit tests cover the failure paths.

**Streaming state shared across choices** (gemini high):
Streaming un-redact now keyed by `choice.index` (`HashMap<i64, StreamUnredact>`) with separate maps for `content`, `reasoning_content`, and `reasoning`. For `n>1` completions or models that emit `content` + `reasoning` concurrently, each (choice, field) gets its own sliding tail buffer.

## High-value

**Stream tail never flushed at end-of-stream** (Copilot):
Added `build_flush_chunks()` that drains each per-(choice, field) state at `[DONE]` time. Any bytes still in a sliding tail (e.g. upstream cut off mid-`<email1>`) get emitted as a synthetic final SSE chunk before `[DONE]` with the same `id`/`model`/`created` as the first real chunk. Empty fields are skipped so the synthetic chunk only fires when there's something to emit.

**PII reaches logs at debug level** (Copilot):
The `tracing::debug!("Completion stream event: ...")` line is now suppressed when `auto_redact_enabled`. After un-redact the chunk content is the user's original PII; logging it would defeat the privacy invariant the user opted into. CLAUDE.md's "no customer content in logs, even at debug" rule applies here.

## Worth-addressing

**`detect.rs` overwriting spans for duplicate index** (gemini medium): Changed `out[item.index] = ...` to `out[item.index].extend(...)`.

**Re-serialize even when map is empty** (Copilot): `auto_redact_enabled` is now `requested && !map.is_empty()`. A request that opts in but contains no PII keeps the raw-bytes/signing path and the existing debug log.

## Nits

- Removed dead `docs/auto-redact.md` reference from mod.rs doc comment.
- `MAX_PLACEHOLDER_LEN` doc example now matches what's actually minted (`<address9999>`, not `<account_number999>`).
- Dropped `Span.text` (was only set from the detector response and never read).

## Mock side-effect

`MockProvider::privacy_classify_raw` keeps `usage.input_tokens=10` (the original constant) for backward compat with the `test_privacy_classify_costs_deducted` test from #584. Only `spans` is computed from the input.

## Test pass

- 260 unit pass (added 2 fail-closed tests in apply.rs)
- 8 auto_redact e2e pass (added `auto_redact_skips_response_munging_when_no_pii_detected`)
- 7 privacy_classify e2e pass (the failing one from CI is now green)
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo fmt --check` clean

PierreLeGuen (approved):

PR #586, opened 2026-05-12
fix(gemini): tolerate truncated content on MAX_TOKENS responses

lloydmak99:
## Summary
- `GeminiContent` rejected Google's truncated `content` objects when `finishReason: MAX_TOKENS` hit before any output tokens, surfacing to clients as 502 "The model is currently unavailable."
- Added `#[serde(default)]` to `role` and `parts` so both observed payload shapes parse cleanly.
- Made `GeminiCandidate.content` an `Option` so the safety-block / terminal-only shape (where the field is absent entirely) also parses.
- Extracted the response conversion to `convert_to_openai_response()` so the user-visible contract (content/finish_reason mapping) is exercised end-to-end in unit tests.
- `tracing::warn!` when a candidate yields no usable content and `finish_reason != MAX_TOKENS`, so any future upstream regression that silently returns empty `STOP`/`SAFETY` responses is visible in Datadog instead of being lost in the relaxed-schema fallthrough.
- Doc on `GeminiContent` rewritten to describe behaviour, not enumerate model variants.

## Why

`google/gemini-3-pro` on prod cloud-api was failing ~30% of the time (more at low `max_tokens`). Logged on `cpu01-cloud-api-prod` / `cpu02-cloud-api-prod`:

```
WARN  Provider failed, will try next provider if available
  error_detail: "Failed to perform completion: Failed to parse response:
                 missing field `role` at line 4 column 19"
  model_id: "google/gemini-3-pro"
  retry_decision: "non_retryable_no_keyword_match"
```

Direct calls to Google's v1beta endpoint reproduced the truncated payloads on the first attempt:

```
gemini-3-flash-preview → {"candidates":[{"content":{}, "finishReason":"MAX_TOKENS", ...}]}
gemini-2.5-flash       → {"candidates":[{"content":{"role":"model"}, "finishReason":"MAX_TOKENS", ...}]}
```

Reviewer flagged a third variant where `content` is absent entirely (safety-blocked responses). Now covered by `Option<GeminiContent>`.

## Behaviour change

- **Before**: parse error → cloud-api drops the request and renders 502 to client; not retried because parser errors aren't in the retry-keyword set.
- **After**: parse succeeds. `extract_response_content(&[])` returns `(None, None)`, producing an OpenAI-canonical `content: null, finish_reason: "length"` response — the shape clients already expect on a MAX_TOKENS-with-no-output completion. `SAFETY` finish reasons round-trip identically with `content: null`.
- **Silent-regression guard**: if Google ever returns empty content with `finish_reason != MAX_TOKENS`, a `tracing::warn!` fires with the model id and finish reason (no request/response content — privacy rule compliance).

## Test plan
- [x] `cargo test -p inference_providers --lib gemini` — **19/19 pass** including:
  - 2 captured-payload regression tests (`content: {}`, `content: {"role": "model"}`)
  - 2 new schema tests (`content: null`, `content` field absent)
  - 3 end-to-end conversion tests through `convert_to_openai_response` (MAX_TOKENS-empty, SAFETY-missing, normal-STOP)
- [x] `cargo check --all-targets` clean across the workspace
- [ ] **After merge + staging deploy**: re-run a **30× reproduction** (was 10× — bumped on review feedback for a tighter confidence interval) of `google/gemini-3-pro` at `max_tokens: 32`. Prior baseline 7/10; acceptance ≥ 28/30 succeed.
- [ ] **After prod promote**: confirm `test_chat_completion[gemini-3-pro]` and `test_tool_calling[gemini-3-pro]` drop off the Allure Product-defects list over 4 consecutive infra-tests runs.

## Streaming path

The SSE parser (`GeminiEventParser` in `converter.rs`) consumes the same `GeminiResponse` / `GeminiCandidate` types, so the schema relaxation applies there too. The Option-handling change to the SSE consumer at `converter.rs:428` is in this PR; the silent-regression warn is currently only in the non-streaming path. If we observe an analogous regression on streaming, we can add a parallel warn in the parser state machine.

## Related
- Diagnosis: nearai/infra#108
- Parent: nearai/infra#104 (infra-tests pass rate to 95%)

### DIFF
--- a/crates/inference_providers/src/external/gemini/converter.rs
+++ b/crates/inference_providers/src/external/gemini/converter.rs
@@ -641,5 +641,8 @@ mod tests {
         assert_eq!(response.candidates.len(), 1);
         assert_eq!(response.candidates[0].content.role, "model");
         assert!(response.candidates[0].content.parts.is_empty());
+        let (text, tool_calls) = extract_response_content(&response.candidates[0].content.parts);
+        assert!(text.is_none());
+        assert!(tool_calls.is_none());
     }
 }
--- a/crates/inference_providers/src/external/gemini/converter.rs
+++ b/crates/inference_providers/src/external/gemini/converter.rs
@@ -81,12 +81,11 @@ pub struct GeminiFunctionResponse {
 
 /// Gemini content format
 ///
-/// `role` and `parts` are marked `#[serde(default)]` because Google's
-/// `generateContent` response omits one or both fields when generation ends
-/// with `finishReason: MAX_TOKENS` before producing any output tokens
-/// (observed in `gemini-3-flash-preview` returning `content: {}` and
-/// `gemini-2.5-flash` returning `content: {"role": "model"}`). The strict
-/// schema rejected these payloads and surfaced as 502s to clients.
+/// Both `role` and `parts` are marked `#[serde(default)]` because Google's
+/// `generateContent` response may omit either field when no usable output
+/// was produced (typically when `finishReason` is `MAX_TOKENS`, `SAFETY`, or
+/// `RECITATION`). Defaulting to empty values lets the parser succeed and the
+/// downstream code treat it as an empty completion.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct GeminiContent {
     #[serde(default)]
@@ -150,10 +149,16 @@ pub struct GeminiRequest {
 // =============================================================================
 
 /// Gemini response candidate
+///
+/// `content` is `Option` because Google omits the field entirely on some
+/// terminal-only responses (e.g. safety-blocked candidates that return only
+/// `finishReason`). Consumers must treat a missing `content` the same as a
+/// content with empty `parts`.
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct GeminiCandidate {
-    pub content: GeminiContent,
+    #[serde(default)]
+    pub content: Option<GeminiContent>,
     pub finish_reason: Option<String>,
 }
 
@@ -424,8 +429,13 @@ impl SSEEventParser for GeminiEventParser {
         let is_first = state.chunk_index == 0;
         state.chunk_index += 1;
 
-        // Extract text and function calls from parts
-        let (text, tool_calls) = extract_response_content(&candidate.content.parts);
+        // Extract text and function calls from parts. `content` may be absent
+        // entirely (e.g. safety-blocked candidates); treat it as empty parts.
+        let parts: &[GeminiPart] = candidate
+            .content
+            .as_ref()
+            .map_or(&[], |c| c.parts.as_slice());
+        let (text, tool_calls) = extract_response_content(parts);
 
         // Determine finish reason
         let has_function_call = tool_calls.is_some();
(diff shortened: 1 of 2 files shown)


lloydmak99:
Addressed in 9c2d534:

- **Test polish** ✅ — `test_parse_response_with_role_only_content_on_max_tokens` now calls `extract_response_content` and asserts `(None, None)`, matching its sibling. `cargo test -p inference_providers --lib gemini` → 14/14 still pass.
- **Missing `content` field on `GeminiCandidate`** — deferring per your own guidance. We have no captured payload exhibiting it; same-class behaviour but distinct schema position. Will revisit if it shows up in prod logs.

lloydmak99:
Addressed in ffb2758. Mapping to each point:

1. **`GeminiCandidate.content` may be missing entirely** ✅ — now `Option<GeminiContent>` with `#[serde(default)]`. Two new schema tests cover `content: null` (explicit) and `content` field absent (safety-block shape). Both the non-streaming converter (`mod.rs`) and the SSE parser (`converter.rs:428`) handle the Option uniformly via `candidate.content.as_ref().map_or(&[], |c| c.parts.as_slice())`.

2. **E2E test missing** ✅ — extracted the inline conversion in `chat_completion` to `convert_to_openai_response(gemini_response, model)`. Three end-to-end tests now exercise the full contract: MAX_TOKENS-empty-content → `content: null, finish_reason: "length"`; SAFETY-missing-content → `content: null` with mapped finish reason; normal-STOP → populated content. Locks in the user-visible behaviour, not just the parser.

3. **Streaming path** — already covered by the shared types (`GeminiContent` defaults + `GeminiCandidate.content` Option), and the SSE consumer was updated in the same change. Called out explicitly in the updated PR description. Not adding a parallel `tracing::warn!` in the streaming state machine speculatively — will do that if/when we observe the same class of upstream regression on the streaming path.

4. **Doc bit-rot** ✅ — rewrote the `GeminiContent` doc to describe the behaviour ("either field may be omitted when no usable output is produced, typically MAX_TOKENS/SAFETY/RECITATION") instead of enumerating `gemini-3-flash-preview` and `gemini-2.5-flash`. Won't decay when a new variant appears.

5. **30× repro for confidence** ✅ — updated the test plan in the PR description (was 10× / acceptance 9/10; now 30× / acceptance ≥28/30). Agree the prior margin was thin.

6. **Silent-semantic-change risk** ✅ — added `tracing::warn!` in `convert_to_openai_response` when a candidate yields no usable content (parts empty / content missing / no tool calls) **and** `finish_reason != "MAX_TOKENS"`. Under the old strict schema that combination would have surfaced as a 502; under the relaxed schema it could silently bill prompt tokens and return `content: null`. The warn keeps any future upstream regression visible in Datadog. Logs only `model` and `finish_reason` — no request/response content (privacy rule compliance).

Tests: 19/19 pass on `cargo test -p inference_providers --lib gemini`. `cargo check --all-targets` clean.

Evrard-Nil (approved):
LGTM. Reviewed code + ran local verification:

**Static**
- `cargo test -p inference_providers --lib gemini` — 19/19 pass
- `cargo check --all-targets` clean
- `cargo clippy -p inference_providers --lib --no-deps` clean
- Grepped for stale `candidate.content.parts` access in production code — none; all production paths route through `Option<GeminiContent>` via `map_or(&[], |c| c.parts.as_slice())`

**Live API**
Wrote an example that hits `gemini-2.5-flash:generateContent` at `maxOutputTokens=32` and parses each byte stream with both the PR-#586 schema and a reconstructed pre-PR strict schema. 5/5 real Google responses came back as `{"content": {"role": "model"}, "finishReason": "MAX_TOKENS"}` — every one:
- parses cleanly with the new schema → OpenAI canonical `finish_reason: "length"`, empty content
- would have failed with `missing field 'parts'` under the old schema (i.e. surfaced as 502)

This matches the production failure mode in the PR description.

**Notes**
- gemini-code-assist's tool_call_id comment is a false positive — `#[serde(default)]` is on `GeminiContent::{role, parts}`, not on any tool_call_id field.
- Streaming path omits the silent-regression `warn!`; called out in PR description as deliberate scope. Reasonable.

PR #589, opened 2026-05-13
perf(api): enable response compression and add Cache-Control on public GETs

Evrard-Nil:
## Why

Production measurements show `/v1/model/list` returns ~31 KB uncompressed even when the client sends `Accept-Encoding: gzip` — with gzip it drops to ~5 KB. Combined with the >100 ms RTT cost for users far from cpu01, every catalog refetch is unnecessarily slow. Public model/services endpoints also have no caching directives, so every browser/SDK re-fetches in full on every call.

### Task A — Response compression (gzip + brotli)

- Added `compression-gzip` + `compression-br` features to `tower-http` in `crates/api/Cargo.toml`.
- Wrapped the main router with `tower_http::compression::CompressionLayer::new()` as the outermost layer in `routes()` (after CORS and metrics, so it applies to all responses).

**SSE / streaming endpoints (NOT excluded — they are auto-skipped):**

`CompressionLayer` inspects the response `Content-Type` and skips known streaming types, including `text/event-stream`. Both `/v1/chat/completions` (when `stream=true`) and `/v1/responses` set `Content-Type: text/event-stream` for the SSE path, so compression is a no-op there — clients continue to receive token deltas with normal latency.

**Attestation / signature endpoints:**

The body-hash middleware (`crates/api/src/middleware/body_hash.rs`) hashes the *request* body, not the response body. The `/v1/signature/{chat_id}` and `/v1/attestation/report` endpoints return JSON responses containing signatures over inference inputs/outputs computed inside services — they do NOT sign the HTTP response payload. So compressing the response is transparent to verification. No paths needed to be excluded from compression.

### Task B — Cache-Control on public, anonymous GETs

Implemented via a small `SetResponseHeaderLayer`-based helper (`cache_control_layer`) applied per route subgroup in `lib.rs`. `if_not_present` is used so handlers retain the ability to override per request if they ever need to.

| Endpoint(s) | TTL | Rationale |
|-------------|-----|-----------|
| `GET /v1/model/list`, `GET /v1/model/{name}` | `public, max-age=30, stale-while-revalidate=120` | Public, anonymous, identical for every caller. Changes only when an admin updates the catalog. SWR lets CDNs/browsers serve cached copies instantly while refreshing in the background. |
| `GET /v1/services`, `GET /v1/services/{name}` | `public, max-age=30, stale-while-revalidate=120` | Same shape — public, anonymous, admin-write speed (used by `chat-api` to fetch `web_search` pricing). |
| `GET /v1/health` | `public, max-age=5` | Short window absorbs thundering-herd from monitors/uptime probes; outages still surface within ~5 s. |
| `GET /docs`, `GET /api-docs/openapi.json` | `public, max-age=300, stale-while-revalidate=3600` | Spec + Scalar UI HTML change only on deploy. Five-minute fresh window with one-hour SWR. |

### Endpoints audited and intentionally NOT cached

- `/v1/users/me`, conversations, responses, workspaces, organizations, files, usage, billing, admin/* — authenticated and/or user-specific.
- `/v1/check_api_key` (gateway) — POST.
- `/login` — interactive OAuth flow page, response varies by request.
- Anything inside the auth/management plane.

## Verification

- `cargo build` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --lib --bins` — all 458 tests pass (260 in `api`, 117 in `services`, etc.)
- Full e2e (`cargo test --test e2e_all`) requires a local Postgres; not run here, but the changes are layer-only and do not touch any handler logic, repositories, or service traits.

## Test plan

- [ ] Hit `https://cloud-api.near.ai/v1/model/list` with `curl -H 'Accept-Encoding: gzip' -i` after deploy — expect `Content-Encoding: gzip` (or `br`) and `Cache-Control: public, max-age=30, stale-while-revalidate=120`.
- [ ] Hit `https://cloud-api.near.ai/v1/health` after deploy — expect `Cache-Control: public, max-age=5`.
- [ ] Hit `https://cloud-api.near.ai/api-docs/openapi.json -H 'Accept-Encoding: gzip'` — expect compression and `max-age=300`.
- [ ] Confirm a streaming `/v1/chat/completions` request with `stream=true` returns `Content-Type: text/event-stream` and NO `Content-Encoding` header (CompressionLayer should auto-skip SSE).
- [ ] Confirm `/v1/signature/{chat_id}` still verifies end-to-end with the existing verifier flow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aead"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -44,6 +50,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
 ]
 
 [[package]]
@@ -806,6 +827,18 @@ name = "asn1_der"
 version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4858a9d740c5007a9069007c3b4e91152d0506f13c1b31dd49051fd537656156"
+
+[[package]]
+name = "async-compression"
+version = "0.4.37"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d10e4f991a553474232bc0a31799f6d24b034a84c0971d80d2e2f78b2e576e40"
+dependencies = [
+ "compression-codecs",
+ "compression-core",
+ "pin-project-lite",
+ "tokio",
+]
 
 [[package]]
 name = "async-lock"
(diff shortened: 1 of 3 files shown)


Evrard-Nil:
Addressed the claude-review concern about `Cache-Control` being attached to error responses.

`SetResponseHeaderLayer::if_not_present` ran on every response, so a transient 500 on `list_models` (or a 404 on `get_model_by_name`, or any failure on `/v1/services`, `/v1/health`, `/docs`) would carry `Cache-Control: public, max-age=…, stale-while-revalidate=…` and could be pinned by Cloudflare "Cache Everything" / Fastly / some browsers for the full TTL.

Replaced the helper with an `axum::middleware::from_fn_with_state`-based middleware that only inserts `Cache-Control` when:
- the response is a success (`status().is_success()`), and
- the handler did not already set a `Cache-Control` header

Same call sites, same header values, same TTLs — only the success-gating semantics change. New unit tests cover all four invariants (2xx gets header, 2xx with handler-set header is not overridden, 4xx has no header, 5xx has no header).

Pushed as f0468d3.

PR #590, opened 2026-05-13
perf(api): drop /v1/model/list pagination, cache the model list

Evrard-Nil:
## Why

The public `/v1/model/list` endpoint ran **two sequential DB queries** on every anonymous request — a `COUNT(*)` and a `JOIN ... GROUP BY` on `models` x `model_aliases`. The catalog has roughly 34 models, so pagination served no purpose; it just added latency and load.

## What changed

- **Drop pagination on `/v1/model/list`.** Removed `limit` / `offset` query params, removed the `validate_limit_offset` check, removed the count query, removed `ModelListQuery`.
- **Cache the list in-process.** Added a `moka::future::Cache<&'static str, Arc<Vec<ModelWithPricing>>>` field on `ModelsServiceImpl` (capacity 1, TTL 30s) so most hits skip Postgres entirely. Pattern mirrors the existing `api_key_cache` in `services::auth`.
- **Invalidate on admin writes.** `AdminServiceImpl` now holds an `Arc<dyn ModelsServiceTrait>` and calls `invalidate_models_cache()` after every successful `batch_upsert_models`, `delete_model`, and `deprecate_model`.
- **Repository cleanup.** Simplified `ModelRepository::get_all_active_models` to drop `LIMIT`/`OFFSET`. Removed `get_all_active_models_count` (no remaining callers). Other admin paths use the separate `AdminCompositeRepository::list_models` and are unaffected.

## Files

- `crates/api/src/routes/models.rs` — handler signature, OpenAPI annotations
- `crates/api/src/models.rs` — `ModelListResponse` drops `limit` and `offset`, keeps `total = models.len()`
- `crates/api/src/routes/completions.rs` — adjusted caller of `get_models_with_pricing`
- `crates/api/src/lib.rs` — passes `models_service` into `build_admin_routes`
- `crates/services/src/models/{mod.rs,ports.rs}` — cache, `invalidate_models_cache` on trait
- `crates/services/src/admin/mod.rs` — invalidation hooks after model writes
- `crates/database/src/repositories/model.rs` — drop pagination, drop count

## Compatibility note for clients

- The `limit` / `offset` query parameters on `/v1/model/list` are now **silently ignored**.
- The response body drops the `limit` and `offset` fields. `models` and `total` are unchanged. `total` is now always equal to `models.len()`.
- After admin model edits there is at most ~30s of staleness on cached responses already in flight; admin writes always invalidate the cache.

## Test plan

- [x] `cargo build`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --lib --bins` (260+ tests pass)
- [ ] e2e: hit `/v1/model/list` against a running server, verify response shape matches `{ models, total }`
- [ ] e2e: admin `PATCH /v1/admin/models` followed by `/v1/model/list` returns the new pricing within < 1s (verifying invalidation)
- [ ] Load: confirm < 1 DB round-trip per ~30s window of public `/v1/model/list` traffic

## Note on parallel PR

There is a parallel PR in flight adding `CompressionLayer` and `Cache-Control` headers to `crates/api/src/routes/models.rs` and `crates/api/src/lib.rs`. This PR scopes its `models.rs` changes to the handler signature and OpenAPI annotations to minimize conflict — resolution at merge time should be mechanical.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -820,6 +820,7 @@ pub fn build_app_with_config(
         config.clone(),
         app_state.inference_provider_pool.clone(),
         analytics_service,
+        domain_services.models_service.clone(),
     );
 
     let invitation_routes =
@@ -1355,6 +1356,7 @@ pub fn build_admin_routes(
     config: Arc<ApiConfig>,
     inference_provider_pool: Arc<services::inference_provider_pool::InferenceProviderPool>,
     analytics_service: Arc<services::admin::AnalyticsService>,
+    models_service: Arc<services::models::ModelsServiceImpl>,
 ) -> Router {
     use crate::middleware::admin_middleware;
     use crate::routes::admin::{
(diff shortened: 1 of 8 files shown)


PierreLeGuen (approved):
LGTM, thanks!

PR #599, opened 2026-05-18
feat(auto_redact): natural-looking dummies + secret category + span merge

Evrard-Nil:
## Summary

Three changes that close the loop on observed staging behavior of #585/#588:

1. **Realistic-shaped dummies replace `<categoryN>` brackets.** Models recognize `<email1>` as template syntax: Anthropic refused PII-handling prompts wholesale and gpt-oss stripped the brackets before echoing the value into tool-call arguments (breaking the un-redact round trip). Each category now mints a value in its own natural shape:

| Category | New dummy |
|---|---|
| `private_email` | `redacted{N}@example.com` |
| `private_phone` | `+1-555-01{NN:02}` (RFC 2606 fictional) |
| `account_number` | `000-00-{N:04}` (invalid SSN area) |
| `private_address` | `{N}00 Redacted Way` |
| `private_name` | `Redacted{N:03}` |
| `secret` | `redacted_secret_{N:06}` |
| catch-all | `redacted_pii_{N}` |

All formats are drawn from RFC-reserved or otherwise-invalid ranges so they can't collide with real third-party PII.

2. **Adjacent same-category spans merge.** privacy-filter often returns one PII as multiple contiguous spans (`alice@example.com` → `"alice@example"` + `".com"`). Without merging, the model received two placeholders for one logical entity, causing two parallel tool calls (one per placeholder) and weirder paraphrasing. `redact_one` now folds spans with the same category where `prev.end == curr.start` before minting.

3. **`secret` category mapped.** privacy-filter's `secret` category (passwords, API tokens — caught reliably in live tests) now mints `redacted_secret_{N}` instead of the generic `redacted_pii_{N}`.

## Side effects

- **Un-redact is now exact-string substitution** over the per-request map, iterated longest-first. No regex needed. `PLACEHOLDER_RE` is removed.
- **Collision avoidance is haystack-based.** The mint loop bumps the ordinal until a candidate doesn't appear as a substring of any input fragment. Replaces the regex-based `<categoryN>` reservation that no longer matches the new shapes.
- **Streaming sliding-tail is dynamic.** Size = `max_dummy_len` from the map, capped at `MAX_PLACEHOLDER_LEN = 64`. No more relying on `<…>` boundary characters.

## Tests

- 46 unit pass: placeholder formats, JSON-escape with the new shapes, span merge (same-cat / different-cat / with-gap), streaming with natural dummies char-by-char split, longest-first un-redact ordering.
- 20 e2e pass (`auto_redact` + `auto_redact_adversarial`). The collision-avoidance test was reframed: the user's dummy-shaped literal is now treated as PII by the detector and gets its own non-colliding dummy via ordinal bump.
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo fmt --check` clean

## Test plan

- [x] Unit + e2e against MockProvider
- [ ] After merge: deploy to staging, re-run the Anthropic + Qwen + gpt-oss tool-call round trips that previously surfaced bracket-stripping/two-tool-calls — expect the natural dummies to round-trip cleanly without a "verbatim preservation" system prompt.

Builds on #588.


Evrard-Nil:
Pushed \`b25a028\` (+ \`223381f\` cleanup) addressing the four real-correctness items.

**1. Same-category overlap merge (gemini security-high)**
\`merge_adjacent_same_category\` renamed to \`merge_overlapping_same_category\`; condition is now \`prev.end >= span.start\` so both contiguous AND overlapping same-category spans union. Different-category overlaps still fall through to \`redact_one\`'s cursor-drop (rare, ambiguous category winner). New unit test asserts (8..15)+(12..19) → one dummy covering 8..19 — without the fix, the trailing \`.com\` would leak.

**2. Phone format inside RFC range (claude #1/#2)**
Ordinals 1..=100 mint \`+1-555-01{NN:02}\` strictly inside the RFC 2606 fictional range. Past 100, fall back to \`redacted_pii_N\` so we never produce a real-NANP-shape number.

**3. Streaming flush on \`finish_reason\` (Copilot #1)**
New \`StreamUnredact::drain\` that processes + empties the tail in one call. \`unredact_chunk_in_place\` detects \`choice.finish_reason.is_some()\` and uses \`drain\` instead of \`process\` for all per-(choice, field) state. The end-of-stream \`build_flush_chunks\` stays as defense-in-depth.

**4. Collision test now exercises response un-redact (Copilot #2)**
Mock now responds with both forwarded dummies; assertions verify the un-redact map restores both originals (including the user's literal \`redacted1@example.com\`) and that neither minted dummy leaks to the client.

Plus: cleaned up stale comments in \`lookup_or_mint\` (the "reserved literal" concept was removed in this PR but a comment still mentioned it).

**Skipped (out of scope for blockers):**
- gemini-medium haystack-per-fragment perf optimization — premature; K (PII count per request) and N (text fragment count) are small in practice. Worth revisiting if profiling shows it.
- claude notes about \`unredact\` complexity (O(K·N·L)) and \`merge_overlapping_same_category\` accepting unsorted input — both fine in practice; the sort-tolerance is now reflected in the function's doc.

Test counts: 47 unit (+1 overlap-merge), 20 e2e auto_redact pass, clippy + fmt clean.

Re: 223381f — accidentally staged a stray review tree under \`./tmp/\` along with the fix commit; untracked it in a follow-up.

PierreLeGuen:
The finish-reason drain still misses the common empty-delta finish chunk case. `finalize` is set when `choice.finish_reason` is present, but we only call `drain` for fields present on that same chunk. For OpenAI/vLLM-style final chunks like `delta: {}` + `finish_reason: "stop"`, any held content/reasoning/tool-call tail remains in `states` and is emitted later by `build_flush_chunks`, after the finish-reason event. Clients that stop on `finish_reason` will still miss the restored bytes.

We need to drain the per-choice states when `finish_reason` is seen even if the finish chunk has no delta fields, or otherwise attach the drain output to the finish chunk before serializing it.

Evrard-Nil:
@PierreLeGuen Filed a fix at #633 targeting this branch. Drains every remaining per-choice state into the finish chunk's delta in place (creating absent fields / `ToolCallDelta` entries as needed), so the held tail never reaches `build_flush_chunks`. Two unit tests reproduce the `delta: {}` + `finish_reason: stop` and `… + finish_reason: tool_calls` cases — both fail without the change.

PierreLeGuen (approved):
Reviewed the natural dummy redaction flow, span merge behavior, streaming unredact changes, and passing CI. Looks good.

PR #604, opened 2026-05-18
ci: date+sha stamped staging/prod docker tags for rollback

lloydmak99:
## Motivation

`:prod` and `:staging` are mutable docker tags — every promotion overwrites them. Once promoted, the previous prod image is only addressable by digest pulled from GH release notes, which makes rollbacks painful (manual digest lookup → manual extra-vars override on the ansible playbook).

This PR adds **immutable date+sha stamped tags** alongside the moving `:prod` / `:staging` tags, so rollback consumers in [`nearai/cvm-ansible-playbooks`](https://github.com/nearai/cvm-ansible-playbooks) can roll back by tag name without digging up digests.

Paired with: rollback workflows in `nearai/cvm-ansible-playbooks` (separate PR).

### `build.yml`
- When pushing `:staging` on main, also push `staging-YYYYMMDD-<shortsha>` using `GITHUB_SHA`.
- Skipped during Tue/Wed staging freeze (same as `:staging` itself).

### `promote.yml`
- After copying source → `:prod`, search Docker Hub for a `staging-YYYYMMDD-<shortsha>` tag whose digest matches the just-promoted digest, to recover the source git short-sha.
- If no match found (e.g. promoted by arbitrary digest input), fall back to first 7 hex chars of the image digest as a uniqueness suffix. This guarantees the stamped tag is unique even on unusual promotion paths.
- Push `prod-YYYYMMDD-<shortsha>` from the resolved digest (not from `:prod`) to avoid registry eventual-consistency races.
- GH release tag mirrors the stamped docker tag (`prod-YYYYMMDD-<shortsha>`) for easy correlation between docker image and GH release.
- Release creation is now idempotent (`gh release view ... || create`), so manual same-day re-runs don't fail on duplicate tag. `--latest` flag is only set on create, not edit.

## Notes

- **Bootstrap**: the first prod rollback after merge has no `prod-*` stamped tag history. Either wait for 2 promotions to accumulate, or manually `skopeo tag` the current `:prod` once.
- **Tag retention**: not addressed in this PR. Staging accumulates a tag per build; revisit cleanup if/when Docker Hub tag list becomes noisy.
- No changes to image content, signing, or attestation flow.

## Testing

YAML linted. To verify end-to-end after merge:
1. Trigger build.yml on main → expect `:staging` and `staging-<today>-<sha>` both pushed.
2. Trigger promote.yml manually → expect `:prod` and `prod-<today>-<sha>` both pushed, GH release `prod-<today>-<sha>` created.
3. Verify both stamped tags resolve to the same digest as the moving tag.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,8 +105,14 @@ jobs:
         if: env.ALSO_TAG_STAGING == 'true'
         run: |
           STAGING_REF="${{ env.REGISTRY }}/${{ env.IMAGE_REPOSITORY }}:staging"
+          # Immutable date+sha staging tag for rollback (format: staging-YYYYMMDD-<shortsha>)
+          SHORT_SHA=$(echo "${GITHUB_SHA}" | cut -c1-7)
+          STAMP=$(date -u +%Y%m%d)
+          STAGING_STAMPED_REF="${{ env.REGISTRY }}/${{ env.IMAGE_REPOSITORY }}:staging-${STAMP}-${SHORT_SHA}"
           skopeo copy oci-archive:./oci.tar "docker://${STAGING_REF}"
+          skopeo copy oci-archive:./oci.tar "docker://${STAGING_STAMPED_REF}"
           echo "STAGING_REFERENCE=${STAGING_REF}" >> "$GITHUB_ENV"
+          echo "STAGING_STAMPED_REFERENCE=${STAGING_STAMPED_REF}" >> "$GITHUB_ENV"
 
       - name: Trigger staging deployment
         if: env.ALSO_TAG_STAGING == 'true'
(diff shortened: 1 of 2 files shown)


lloydmak99:
Thanks for the careful review. Addressed 1–4 in commit 05bde1e:

**1. curl timeout** — added `--max-time 30 --retry 2 --retry-delay 2` on the pagination loop.

**2. Pagination max-iteration safeguard** — added `MAX_PAGES=50` counter with a `::warning::` log if reached. At 100 tags/page, 50 pages = 5000 tags — well above any plausible legitimate count.

**3. Hardcoded image name** — extracted `IMAGE_NAME: cloud-api` to the workflow env block. The Docker Hub API URL now uses `${{ env.IMAGE_NAME }}` like the rest of the file uses `${{ env.IMAGE }}`.

**4. Inspect-after-write race** — biggest fix. Restructured to resolve the source digest BEFORE copying to `:prod`:
   - If `inputs.digest` is provided, use it directly (zero registry round-trips).
   - Otherwise, `skopeo inspect docker://...:staging` once before the copy.

   Then `:prod` and the stamped `prod-YYYYMMDD-<sha>` tag both `skopeo copy @${DIGEST} → ...`, so they're provably identical bytes. The order is now: resolve → search-for-sha → copy-to-prod → copy-to-stamped. No reads of `:prod` after writing it anywhere in the workflow.

**5. Heredoc indentation** — intentionally left as-is. The notes render correctly in GH releases (markdown ignores leading spaces in front of `-` bullets), and switching to `<<-EOF` requires tabs which YAML lint will fight. Not worth the churn.

Evrard-Nil (approved):

PR #611, opened 2026-05-19
fix(embeddings): propagate upstream HTTP status instead of 502 'try again later'

### LINKED ISSUE #605: Embeddings: dimensions param returns 502 with generic 'try again later' instead of 400
## Summary

Passing the OpenAI-spec `dimensions` parameter to `/v1/embeddings` consistently returns:

```
HTTP 502
{"error":{"message":"Embeddings request failed. Please try again later.","type":"server_error","param":null,"code":null}}
```

Even with `dimensions` equal to the model's native dim (1024), so the truncation path itself isn't being exercised — the request shape alone is rejected. The error UX tells users to retry; they will, and nothing will change.

## Repro

```bash

# Works
curl -s https://cloud-api.near.ai/v1/embeddings \
  -d '{"model":"Qwen/Qwen3-Embedding-0.6B","input":"hello"}'

# 502, repeatable across attempts
curl -s https://cloud-api.near.ai/v1/embeddings \
  -d '{"model":"Qwen/Qwen3-Embedding-0.6B","input":"hello","dimensions":256}'

# Also 502 with dimensions=512 and dimensions=1024 (== native dim)
```

## Two distinct fixes

1. **Error normalization (this repo)**: a malformed/unsupported request should return `400` with a specific message ("`dimensions` is not supported for this model" or whatever the upstream actually says), not `502 server_error / try again later`. The current wrapper hides the real cause and trains users into useless retries.
2. **Make `dimensions` actually work (upstream)**: `Qwen/Qwen3-Embedding-0.6B` is documented to support Matryoshka representation learning. The current vLLM config in [`cvm-compose-files small-models.yaml:618-626`](https://github.com/nearai/cvm-compose-files/blob/main/small-models.yaml#L618-L626) runs with `--runner pooling` only — no MRL flag — so the upstream rejects the parameter. If we want to support MRL, that's a separate compose-file change.

For this issue, the actionable bit is #1 — surface real upstream errors with the right status code.

---

Evrard-Nil:
## Summary
Surface the real upstream HTTP status and message for embeddings errors, so users see why their request was rejected instead of a misleading 502.

## Reproduced before
\`\`\`
POST /v1/embeddings  {"model":"Qwen/Qwen3-Embedding-0.6B","input":"hi","dimensions":256}
→ HTTP 502 {"error":{"message":"Embeddings request failed. Please try again later.","type":"server_error",...}}
\`\`\`

## Root cause
Three bugs combined to mask every error as 502:

1. **Pool** (\`inference_provider_pool::embeddings\`) collapsed \`EmbeddingError::HttpError\` into \`RequestFailed\`, discarding the upstream status code.
2. **vLLM provider** (\`embeddings_raw\`) stored the raw upstream JSON body in the message field. Other endpoints already use \`extract_error_message\`; embeddings did not.
3. **Route handler** hardcoded \"Embeddings request failed. Please try again later.\" for every \`ProviderError\`, overriding the real upstream message.

## What changes
- Pool preserves \`HttpError\` variant when present (sanitized message kept).
- \`vllm::embeddings_raw\` runs upstream JSON through \`extract_error_message\`.
- Route returns upstream status + message for 4xx; keeps the generic message for 5xx.
- \`extract_error_message\` learns vLLM's flat \`{\"object\":\"error\",\"message\":\"...\",\"type\":\"...\"}\` shape (previously fell through to raw body — affected other endpoints too).

## After the fix
\`\`\`
POST /v1/embeddings  {... \"dimensions\":256}
→ HTTP 400 {\"error\":{\"message\":\"dimensions is not supported for this model\",\"type\":\"invalid_request_error\",...}}
\`\`\`

## Out of scope
- Making \`dimensions\` actually work for Qwen3-Embedding (would need an MRL flag in cvm-compose-files; see #605 §2).
- Same error-handling pattern in \`audio_transcription\` (same hardcoded message bug) — left for a follow-up.

## Test plan
- [x] \`cargo test --lib --bins\` — 301 services + 6 inference_providers + others, all pass
- [x] New pool tests cover HttpError preservation and RequestFailed pass-through
- [x] 6 new \`extract_error_message\` tests cover vLLM flat, OpenAI nested, FastAPI detail, fallback paths
- [x] \`cargo fmt --all -- --check\` clean
- [x] \`cargo clippy --lib --tests --bins -- -D warnings\` clean

Closes #605

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -2905,16 +2905,30 @@ pub async fn embeddings(
                 }
                 services::completions::ports::CompletionError::ProviderError {
                     status_code,
-                    ..
+                    message,
                 } => {
-                    tracing::error!("Embeddings provider error");
                     let http_status = StatusCode::from_u16(status_code)
                         .unwrap_or(StatusCode::INTERNAL_SERVER_ERROR);
-                    (
-                        http_status,
-                        "server_error",
-                        "Embeddings request failed. Please try again later.".to_string(),
-                    )
+                    // For client errors (4xx), surface the upstream message so users
+                    // see *why* their request was rejected (e.g. "dimensions parameter
+                    // is not supported for this model") instead of a generic "try
+                    // again later" that trains them into useless retries.
+                    // For 5xx, keep a generic message — the upstream body is more
+                    // likely to be a stack trace or internal detail.
+                    if http_status.is_client_error() {
+                        tracing::warn!(
+                            status = status_code,
+                            "Embeddings rejected by upstream with client error"
+                        );
+                        (http_status, "invalid_request_error", message)
+                    } else {
+                        tracing::error!(status = status_code, "Embeddings provider error");
+                        (
+                            http_status,
+                            "server_error",
+                            "Embeddings request failed. Please try again later.".to_string(),
+                        )
+                    }
                 }
                 services::completions::ports::CompletionError::InvalidModel(msg) => {
                     tracing::warn!("Embeddings model not found");
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -1266,6 +1266,92 @@ mod tests {
     }
 
     #[test]
+    fn test_classify_provider_error_404_surfaces_message() {
+        let (status, error_type, message) =
+            classify_provider_error(404, "model 'foo' not found".to_string());
+        assert_eq!(status, StatusCode::NOT_FOUND);
+        assert_eq!(error_type, "not_found_error");
+        assert_eq!(message, "model 'foo' not found");
+    }
+
+    #[test]
+    fn test_classify_provider_error_429_surfaces_message() {
+        let (status, error_type, message) =
+            classify_provider_error(429, "too many concurrent requests".to_string());
+        assert_eq!(status, StatusCode::TOO_MANY_REQUESTS);
+        assert_eq!(error_type, "rate_limit_error");
+        assert_eq!(message, "too many concurrent requests");
+    }
+
+    #[test]
+    fn test_classify_provider_error_400_surfaces_message() {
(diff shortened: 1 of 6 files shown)


Evrard-Nil:
Updated to address review feedback:

**1. Granular 4xx classification** (gemini-code-assist, claude bot)

Replaced the blanket `all 4xx → invalid_request_error` mapping with a small `classify_provider_error` helper:

- `404` → `not_found_error` (upstream message surfaced)
- `429` → `rate_limit_error` (upstream message surfaced)
- `400` / `422` / other 4xx → preserve status with `invalid_request_error`
- `401` / `403` / `407` → **`500 server_error` with generic message**. These mean *our* credentials/config to the backend are wrong (vLLM/SGLang auth, proxy auth) — the client did nothing to cause it. Echoing the upstream auth error would be both misleading to the user and a small information leak. Logged at `error!` so we get alerted.
- `5xx` (and unknown) → `500 server_error`, generic message (avoid leaking stack traces / internal hostnames).

Added 8 unit tests covering each branch, including assertions that 401 bodies containing credentials never reach the surfaced message.

**2. Out of scope (not addressed in this PR):**

- `gemini-code-assist` on `vllm/mod.rs:118` — applying `extract_error_message` consistently across *all* vLLM provider methods. Same pattern bug, but each endpoint has its own handler; making this PR fix all of them blurs the scope. Worth a follow-up touching `chat_completions_raw`, `audio_transcription`, etc. consistently.
- `gemini-code-assist` on `inference_provider_pool/mod.rs:2054` — breaking the provider iteration loop on a definitive 4xx. The pool's retry semantics are shared across all endpoints; changing that should be its own PR with broader test coverage of which statuses are retryable per endpoint type.

Rebased onto latest main to resolve the conflict in `inference_provider_pool/mod.rs` (the fast-path tests from #626 landed in the same `mod tests` block as the new embeddings tests).

PR #617, opened 2026-05-20
Add Brave context search tool

PierreLeGuen:
## Summary
- add an opt-in `web_context_search` Responses tool backed by Brave LLM Context API
- keep classic `web_search` and MCP search behavior unchanged
- allow `BRAVE_LLM_CONTEXT_API_KEY` for the context endpoint, falling back to `BRAVE_SEARCH_PRO_API_KEY`
- reuse the existing web source/citation path for context results while keeping logs keyed by the requested tool
- add sanitized production logs for search execution and Brave requests without logging query text, URLs, titles, snippets, raw responses, or upstream error bodies
- add e2e coverage with mocked inference and mocked context search, so CI does not need Brave credentials

## Tests
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p services -p api`
- `cargo test -p services responses::tools`
- `cargo test -p api --lib`
- `DATABASE_HOST=localhost DATABASE_PORT=5432 DATABASE_USERNAME=postgres DATABASE_PASSWORD=postgres DATABASE_TLS_ENABLED=false TEST_DATABASE_NAME=platform_api_e2e DEV=true RUST_LOG=debug cargo test -p api --test e2e_all test_non_streaming_web_context_search_with_mock_provider -- --nocapture`


PierreLeGuen:
Added e2e coverage for the Responses `web_context_search` tool in `crates/api/tests/e2e_all/web_context_search.rs`. It uses mocked inference plus a mocked context-search provider, so CI does not need Brave credentials.\n\nLocal checks run:\n- `cargo fmt --check`\n- `git diff --check`\n- `cargo test -p api --test e2e_all test_non_streaming_web_context_search_with_mock_provider --no-run`\n\nThe full runtime e2e path still needs the repo's local Postgres test setup; on my machine the direct run was blocked by Postgres connection refused.

PierreLeGuen:
Runtime e2e is now verified locally against Docker Postgres.\n\nCommand run:\n`DATABASE_HOST=localhost DATABASE_PORT=5432 DATABASE_USERNAME=postgres DATABASE_PASSWORD=postgres DATABASE_TLS_ENABLED=false TEST_DATABASE_NAME=platform_api_e2e DEV=true RUST_LOG=debug cargo test -p api --test e2e_all test_non_streaming_web_context_search_with_mock_provider -- --nocapture`\n\nResult: passed (1 passed, 0 failed). GitHub CI is also green for security_audit, Lint, and Test Suite; the PR is still blocked only on required review.

PierreLeGuen:
Addressed the review feedback in `1aff536` and resolved the two inline threads.

What changed:
- Simplified Brave query parameter construction; I applied it to both classic and context search so the local style does not drift.
- Initialized context-search params with defaults up front, then only overrode fields present in the tool call.
- Made `balanced` an explicit threshold mode.
- Hoisted duplicated result-stat and web-search error-category helpers.
- Added `BRAVE_LLM_CONTEXT_API_KEY` with fallback to `BRAVE_SEARCH_PRO_API_KEY`, so prod can split subscriptions if Brave requires it without breaking existing config.
- Added short comments for the intentional `web_search_call` event reuse, title/URL-only context-result dropping, and why `serde_json::Error` logging does not leak response bodies.

Critical read: I kept the response item/event type as `web_search_call` intentionally for client/citation compatibility. If we later need product analytics to distinguish context search from classic search in persisted output, that should be a schema/API decision rather than a hidden field added in this PR. Server logs already include `tool_name = web_context_search` / `endpoint = llm_context`.

Fresh local checks after the patch:
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p services -p api`
- `cargo test -p services responses::tools`
- `cargo test -p api --lib`
- focused runtime e2e against Docker Postgres: passed (1 passed, 0 failed)

Evrard-Nil (approved):

Evrard-Nil (approved):
Approved. Clean addition with good test coverage and thoughtful log sanitization. Non-blocking nits noted in review: minor duplication of threshold-mode validation between the executor and brave.rs, no streaming e2e for web_context_search, and the same response.web_search_call.* event names are reused (deliberate, but clients can't distinguish context vs classic search via the event stream).

PR #618, opened 2026-05-20
fix(admin): invalidate org concurrent-limit cache on PATCH

Evrard-Nil:
## Summary

The per-org concurrent-request limit was already enforced end-to-end (`try_acquire_concurrent_slot` in `crates/services/src/completions/mod.rs:914`), but admin PATCHes to `/v1/admin/organizations/{org_id}/concurrent-limit` only wrote the DB row — they did not invalidate `CompletionServiceImpl.org_concurrent_limits`, which caches the value with a 5-minute TTL (`ORG_LIMIT_CACHE_TTL_SECS = 300`, mod.rs:484).

Operators setting a lower limit on a noisy tenant observed *"no effect"* because the cached default (64) stayed live for up to 5 min after each change. Same gap that the models cache had — fixed there with `models_service.invalidate_models_cache()` after each write.

## Change

Mirror that pattern for concurrent limits:

- Add `invalidate_org_concurrent_limit(org_id)` to `CompletionServiceTrait` (`completions/ports.rs`). Implementation just calls `self.org_concurrent_limits.invalidate(&org_id).await` (`completions/mod.rs`).
- Thread `Arc<dyn CompletionServiceTrait>` into `AdminServiceImpl::new` (`services/admin/mod.rs`) and call it after the repository write in `update_organization_concurrent_limit`.
- Update `build_admin_routes` (`api/src/lib.rs`) to forward `completion_service` to the admin service constructor.

## Test plan

- [x] `cargo check`, `cargo fmt --check`, `cargo clippy --all-targets`
- [x] `cargo test --lib --bins -p services concurrent` — both existing concurrent-limit tests pass.
- [ ] Manual verification on staging after merge: `PATCH /v1/admin/organizations/{org_id}/concurrent-limit` with `{"concurrent_limit": 2}` then immediately fire 3 parallel chat completions on a model; expect the 3rd to 429 without waiting out the 5-min TTL.

## Operational context

GLM-5.1 backend was queue-saturated under one tenant's traffic on 2026-05-19; admin tried to throttle the tenant via the existing concurrent-limit endpoint but observed no effect during the incident window. Root-caused to this cache miss. Pairs with a separate cvm-compose-files change adding `--max-queued-requests 8` to SGLang as defense in depth.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -834,6 +834,7 @@ pub fn build_app_with_config(
         app_state.inference_provider_pool.clone(),
         analytics_service,
         domain_services.models_service.clone(),
+        domain_services.completion_service.clone(),
     );
 
     let invitation_routes =
@@ -1437,6 +1438,7 @@ pub fn build_admin_routes(
     inference_provider_pool: Arc<services::inference_provider_pool::InferenceProviderPool>,
     analytics_service: Arc<services::admin::AnalyticsService>,
     models_service: Arc<services::models::ModelsServiceImpl>,
+    completion_service: Arc<services::CompletionServiceImpl>,
 ) -> Router {
     use crate::middleware::admin_middleware;
     use crate::routes::admin::{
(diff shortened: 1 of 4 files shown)


Evrard-Nil:
Addressed in 271a9b0:
- **Copilot — missing test for invalidate path**: added `test_org_concurrent_limit_cache_invalidates` (services/completions/mod.rs in the existing test module). Builds the same `moka::Cache<Uuid, u32>` shape as `org_concurrent_limits` with the production 300s TTL, populates via `get_with` (load-on-miss), confirms a second `get_with` returns the cached value, calls `invalidate(&org_id)`, and asserts the next `get_with` re-runs the loader. Same level of integration as the existing `test_concurrent_limit_state` — exercises the exact contract the new `CompletionServiceTrait::invalidate_org_concurrent_limit` method relies on.

- **Gemini — multi-instance cache divergence**: confirmed limitation, not addressing in this PR. cloud-api runs as multiple replicas (cpu01 + cpu02 prod, same for staging) each with their own moka cache; an admin PATCH against one instance leaves the other cached for up to 5 min. For the original incident shape ("Alex's limit didn't take effect immediately") this is dramatically better than the current "no effect for up to 5 min on either" — but if we want true cluster-wide immediate consistency, the right move is either a shared cache (Redis) or a tiny pubsub on a Postgres LISTEN channel keyed by `org_concurrent_limit_changed`. Worth a follow-up issue if you want me to file one.

PR #620, opened 2026-05-20
feat(http): return 529 when all inference backends are overloaded

Evrard-Nil:
## Summary

cloud-api currently returns **HTTP 503** when `retry_with_fallback` has exhausted every inference backend for a model with 5xx errors — most commonly SGLang's `--max-queued-requests` 503s under sustained load (cf. cvm-compose-files **v0.0.164** which added `--max-queued-requests 8` to GLM-5.1).

That conflates two distinct client signals:
- "This dependency is down" — per-handler 503 (e.g. PII detector unavailable, auth service unreachable).
- "Every backend for this model is overloaded; retry with backoff later."

This change surfaces the second case as **HTTP 529 "Site Is Overloaded"** — non-standard but widely used by Cloudflare and Anthropic for exactly this scenario. Clients already retrying on 5xx with backoff are unaffected; clients with status-specific handling can now distinguish "deploy issue, page someone" from "transient capacity pressure, back off and retry."

## Why this is safe to flip without a more nuanced detector

`retry_with_fallback` already walks **every** provider in the model pool with exponential backoff (3 rounds × N providers, 500ms→4s for connection/5xx, 1s→8s for 429) before surfacing `ServiceOverloaded`. By the time that variant reaches the route layer, we genuinely have tried every backend. Adding extra per-attempt status inspection would be redundant.

## Changes

- `routes/common.rs`: new `status_overloaded()` helper returning `StatusCode::from_u16(529)`. Centralized `map_domain_error_to_status` now returns 529 for `ServiceOverloaded`. Existing test updated to assert `.as_u16() == 529`.
- `routes/completions.rs`: five per-handler `ServiceOverloaded` arms (audio transcription, rerank, embeddings, privacy classify, score) call `status_overloaded()` instead of `StatusCode::SERVICE_UNAVAILABLE`. Message updated from "The service is temporarily overloaded" to "All inference backends are overloaded" — same advice ("retry with exponential backoff"), just clearer about scope.

Chat completions and Responses APIs already route through `map_domain_error_to_status`, so they inherit the change for free.

## Test plan

- [x] `cargo check`, `cargo fmt --check`, `cargo clippy --all-targets`
- [x] `cargo test --lib --bins -p api service_overloaded` — updated test passes
- [ ] Manual staging verification: with `--max-queued-requests 8` on GLM-5.1 backends, drive >8× concurrent requests at a single backend (e.g. via a single sticky pubkey) and confirm cloud-api returns 529 once retry_with_fallback exhausts every replica, not 503.

## Related

- cvm-compose-files **v0.0.164** — `--max-queued-requests 8` for GLM-5.1 (produces the upstream 503s).
- PR #618 — cache invalidation for the per-org concurrent-limit admin PATCH.
- No conflict with PR #611 (touches `ProviderError` arm of embeddings handler + `extract_error_message`).

### DIFF
--- a/crates/api/src/routes/common.rs
+++ b/crates/api/src/routes/common.rs
@@ -4,6 +4,16 @@ use services::completions::CompletionError;
 use services::organization::OrganizationError;
 use uuid::Uuid;
 
+/// HTTP 529 "Site Is Overloaded" — non-standard but widely used (Cloudflare,
+/// Anthropic) for "all backends exhausted; please retry with exponential
+/// backoff." Surfaced when `retry_with_fallback` has tried every inference
+/// backend for a model and they all rejected with 5xx — typically SGLang
+/// `--max-queued-requests` 503s under load. 503 stays reserved for narrower
+/// per-handler "this dependency is down" cases.
+pub fn status_overloaded() -> StatusCode {
+    StatusCode::from_u16(529).expect("529 is a valid HTTP status code")
+}
+
 /// Map domain errors to HTTP status codes
 pub fn map_domain_error_to_status(error: &CompletionError) -> StatusCode {
     match error {
@@ -14,7 +24,7 @@ pub fn map_domain_error_to_status(error: &CompletionError) -> StatusCode {
         CompletionError::ProviderError { status_code, .. } => {
             StatusCode::from_u16(*status_code).unwrap_or(StatusCode::INTERNAL_SERVER_ERROR)
         }
-        CompletionError::ServiceOverloaded(_) => StatusCode::SERVICE_UNAVAILABLE,
+        CompletionError::ServiceOverloaded(_) => status_overloaded(),
         CompletionError::InternalError(_) => StatusCode::INTERNAL_SERVER_ERROR,
     }
 }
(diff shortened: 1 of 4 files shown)


Evrard-Nil:
Addressed in 9a81574:
- **Copilot — e2e tests asserting 503**: updated `provider_errors.rs` (both `test_provider_error_503_propagated` and `test_provider_error_message_preserved_in_streaming` now assert 529) and `general.rs` (high-context fallback accepts 529 alongside 503).
- **Copilot — utoipa docs missing 529**: added the 529 entry to nine `#[utoipa::path]` response lists (chat_completions, completions, image_generations, audio_transcriptions, image_edits, rerank, embeddings, privacy_classify, score) so the generated OpenAPI matches runtime.
- **Gemini — pass `msg` from `ServiceOverloaded(msg)` through to the response**: declining for now. In the rerank/embeddings/score paths (`completions/mod.rs` ~1520/1568/1672), the variant is constructed with the *raw upstream message* from the inference backend, which can contain backend-internal detail. The 503-from-retry-with-fallback path (`map_provider_error` line 710) already discards that and uses a generic — so the variant message is non-uniform. Passing it through unconditionally would leak upstream detail through these endpoints. The route-layer hardcoded message is intentional defense-in-depth for privacy/abstraction. Happy to consolidate later by sanitizing/replacing at the construction sites if there's appetite — but I'd keep it out of this PR.

PR #623, opened 2026-05-20
feat(privacy): add POST /v1/privacy/redact endpoint

Evrard-Nil:
## Summary

Adds `POST /v1/privacy/redact`: a one-shot text sanitizer that runs the privacy-filter and returns each input rewritten with realistic-looking dummies in place of detected PII. The original PII is **never echoed back** — only the redacted form. Useful as a pre-step before sending prompts to a third-party LLM.

## Request / response

Request mirrors `/v1/privacy/classify`:

```json
POST /v1/privacy/redact
{
  "model": "openai/privacy-filter",
  "input": "Email alice@example.com please",
  "threshold": 0.5
}
```

Response is redacted text only — no original PII, no spans:

```json
{
  "model": "openai/privacy-filter",
  "data": [
    { "index": 0, "redacted": "Email redacted1@example.com please", "usage": { "input_tokens": 10 } }
  ]
}
```

`input` accepts either a string or an array of strings; the array case preserves order.

## How it works

1. Handler parses body, normalizes `input` to `Vec<String>`.
2. Calls `try_privacy_classify` against the same `openai/privacy-filter` model — inherits concurrency limit, error mapping, and provider routing for free.
3. Parses the upstream span response and applies redactions locally via the new `auto_redact::apply_detected_spans` helper (reuses `redact_one` + `RedactionMap`).
4. Records usage under `InferenceType::PrivacyClassify` since the billable inference call is identical (one input_tokens charge per input). If we ever want to distinguish redact-vs-classify in analytics, we can add a new variant + migration later.
5. Returns `data[*].redacted` + per-entry `usage.input_tokens`.

## Fail-closed behavior

- Malformed spans / non-UTF-8 boundaries from the upstream → 500, never pass the raw text through.
- Detector unreachable → 503.
- Same anomaly checks as classify (zero-tokens warning, max-tokens cap).

## Stacking

Based on `feat/auto-redact-natural-placeholders` (#599) because the endpoint relies on:
- `redact_one(text, spans, map, haystack)` (4-arg form with haystack collision avoidance)
- Realistic-dummy `RedactionMap::lookup_or_mint` with `would_collide` closure

Once #599 lands on main, I'll rebase this onto main.

## Test plan

- [x] `cargo build -p api` clean
- [x] `cargo clippy -p api -p services --tests` clean
- [x] `cargo fmt --check` clean
- [x] `cargo test -p services auto_redact` — 47/47 pass
- [ ] E2E tests (need Postgres; will run in CI): basic email, array order preservation, no-PII passthrough, 404 unknown model, 401 missing key, 400 missing model / empty input / non-string element, 413 body cap, billing parity with classify

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -19,7 +19,7 @@ use crate::{
         billing::{get_billing_costs, BillingRouteState},
         completions::{
             audio_transcriptions, chat_completions, embeddings, image_edits, image_generations,
-            models, privacy_classify, rerank, score,
+            models, privacy_classify, privacy_redact, rerank, score,
         },
         conversations,
         health::health_check,
@@ -1013,6 +1013,12 @@ pub fn build_completion_routes(
         .route(
             "/privacy/classify",
             post(privacy_classify).layer(DefaultBodyLimit::max(PRIVACY_CLASSIFY_MAX_BODY_SIZE)),
+        )
+        // /privacy/redact runs a classify call under the hood, so the same
+        // 256 KB cap applies.
+        .route(
+            "/privacy/redact",
+            post(privacy_redact).layer(DefaultBodyLimit::max(PRIVACY_CLASSIFY_MAX_BODY_SIZE)),
         )
         .layer(DefaultBodyLimit::max(AUDIO_TRANSCRIPTION_MAX_BODY_SIZE))
         .with_state(app_state.clone())
--- a/crates/api/src/openapi.rs
+++ b/crates/api/src/openapi.rs
@@ -50,6 +50,7 @@ use utoipa::{Modify, OpenApi};
         crate::routes::completions::rerank,
         crate::routes::completions::score,
         crate::routes::completions::privacy_classify,
+        crate::routes::completions::privacy_redact,
         // crate::routes::completions::completions,
         crate::routes::completions::models,
         // Model endpoints (public model catalog)
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -3275,6 +3275,24 @@ pub async fn privacy_redact(
             .into_response();
     }
 
+    if let Some(t) = parsed.threshold {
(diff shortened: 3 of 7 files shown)


Evrard-Nil:
Thanks for the reviews — addressed the four blockers/cleanups inline in 4d16a40:

| # | Issue | Fix |
|---|---|---|
| 1 | Double-billing on duplicate `index` entries | Dedupe `parsed_usage.data` by `index` into a `HashMap<usize, UsageFields>` before summing. Last-write-wins, so the billed amount and the client-visible `usage.input_tokens` stay in sync. |
| 2 | Dead `DetectorUnavailable` arm in the apply branch | Collapsed the match — `apply_detected_spans` is purely in-process (JSON parse + UTF-8 checks), so a single `Err(_) => 500` is correct and self-documenting. |
| 4 | Threshold default divergence from `/privacy/classify` | Don't inject `DEFAULT_THRESHOLD` upstream; pass the field through verbatim only if the client sent one. Removed the now-unused `DEFAULT_THRESHOLD_PUBLIC` re-export from `auto_redact`. |
| 5 | `threshold` not range-checked | Added boundary validation: must be a finite `f64` in `[0.0, 1.0]`. NaN rejected explicitly (`!(0.0..=1.0).contains(&NaN)` is `true`, but we want the error message to be specific). Added e2e tests for `-0.1`, `1.5`, `NaN`, and `0.0`/`0.5`/`1.0`. |

Deferred (acknowledged, fine as follow-ups):
- **#3** (duplication with `privacy_classify` ~140 LoC): I agree this should be extracted into a shared helper (`record_privacy_filter_usage(...)` returning `Result<i32, Response>`). I'll do it once #599 lands and I rebase onto main, so the helper extraction shows up cleanly against current code.
- **#6** (double-deserialize): keeping as-is for 256 KB bodies. Not a hot path.

Tests workflow is gated on `pull_request -> [main, develop]`, which is why only `security_audit` + `claude-review` ran on this stacked PR. The full suite will run once I rebase onto main after #599 merges.

Issue #27, opened 2025-10-11
Enhancement: list organizations and workspaces by creation time in ascending or descending order


think-in-universe:
### Problem

Currently we'd like to fetch the first workspace in the first organization, i.e. the default workspace in the default organization, so we can create API key for users in this default workspace. 

But the current endpoints for listing organizations and workspaces doesn't allow rank results by creation time in ascending or descending order. So if someone created workspaces via the APIs somehow, it might be harder for frontend to find the default workspace (the first created workspace) if there're a large number of workspaces.

### Solution

1. Allow querying organizations (GET `/organizations`) and workspaces (GET `/organizations/{org_id}/workspaces`) by creation time in ascending or descending order;
2. Ideally also support pagination for listing organization endpoint (GET `/organizations`)

PR #637, opened 2026-05-21
feat(vllm): rotation-SNI fallback on 5xx + recognize in-stream error chunks

Evrard-Nil:
## Summary

When the sticky bucket-client (H2-pinned to one backend) returns 5xx/429, the retry path used to hand the same request back to the same overloaded backend: `retry_with_fallback` only iterates over `model_to_providers[model_id]` (usually one URL per model), and `prefix_router.route(...)` is deterministic in the prompt hash. SGLang queue-full aborts thus produced 3 retries × 7.5s of backoff against the same SGLang and surfaced as **HTTP 529** to the client.

This PR walks every healthy backend by index via model-proxy's rotation SNI (`<canonical>-i<N>.<base>`, [model-proxy PR #27](https://github.com/nearai/model-proxy/pull/27)) before propagating the error. With H2 pooling disabled on the per-attempt client, every attempt lands on a distinct backend by construction.

## Failure chain (pre-PR)

Documented in the [#infra thread](https://jasnahworkspace.slack.com/archives/C0B25RF6HQE/p1779313724474859) and PR #131 / #620:

```
cloud-api  chat_completion (sticky bucket → backend X via H2)
  ↓ 503 from SGLang queue full
retry_with_fallback (1 provider, 3 retries, 7.5s backoff)
  ↓ same bucket → same H2 → same backend X → still 503
ServiceOverloaded → HTTP 529 to client
```

What we want:

```
cloud-api  chat_completion (sticky bucket → backend X via H2)
  ↓ 503
rotation-SNI fallback: try i=0..backend_count
  ↓ i=0: backend X (still 503)  → next
  ↓ i=1: backend Y (200 OK)     → return
```

### `chat_completion` (non-streaming)

After the canonical bucket-client attempt returns 5xx/429, fall through to `try_chat_completion_rotation`. Iterates `0..rotation_count()`, builds a fresh single-shot reqwest::Client per index (no pool reuse, no H2 sharing), targets `https://<canonical>-i<N>.<base>/v1/chat/completions`. First 200 wins; the rotation index is recorded in `signature_rotation` so subsequent `get_signature` calls hit the same backend.

### `chat_completion_stream` (streaming)

inference-proxy's `SseTransformer` forwards the SGLang queue-full chunk verbatim (HTTP 200 + `data: {"error":{"code":503,...}}`). The OpenAI SSE parser previously surfaced this as `InvalidResponse("Failed to parse event")`, hiding the upstream status. Two changes:

- `OpenAIEventParser::parse_event` now promotes `{"error":{"code":N,...}}` chunks to typed `HttpError { status_code: N }`.
- The stream-completion path **peeks the first SSE event** before exposing the stream to the caller; if it's a retryable `HttpError`, the stream is dropped and `try_chat_completion_stream_rotation` runs. Because peek happens *before* any byte hits the client, retrying the whole stream is byte-safe.

### Plumbing

- `rotation` module moved from `services::inference_provider_pool` → `inference_providers` (pub). Both discovery and traffic-time fallback now need the URL helpers. Tests moved with the file.
- New `InferenceProvider::set_backend_count` trait method (default no-op, overridden by `VLlmProvider`). Discovery's new-provider path and cumulative refresh path both call it, so the cache is seeded on startup and kept fresh per refresh cycle.
- Rotation fan-out is clamped to `MAX_ROTATION_FANOUT = 256` (mirrors the discovery cap) so a bogus `/backends/count` reading can't burn unbounded fresh-TCP handshakes.

### Signature retrieval

`get_signature` checks `signature_rotation` first. If the chat was served by a rotation index, it builds a one-shot rotation client targeting that exact backend *before* the existing bucket/general fallback. Without this, the signature endpoint would 404 against every other backend.

## What's preserved

- **Prefix-cache stickiness**: the canonical bucket is still the first attempt — rotation only kicks in on 5xx/429. Cache-hit-friendly requests pay zero extra cost.
- **Connection-error fallback**: the existing `is_connection_error` → `clear_bucket` → retry path runs before the rotation gate, so transient TLS hiccups still benefit from the bucket-level refresh.
- **Non-rotation URLs unaffected**: for URLs that don't fit `<canonical>.<multi-label-base>` (`localhost`, IP literals), `rotation_parts` stays `None`, `rotation_count()` returns 0, and the canonical 5xx propagates as before.

## Related

- inference-proxy PR #131 — assembler-path queue-full → HTTP 503 (non-streaming).
- cloud-api PR #620 — `ServiceOverloaded` → HTTP 529 to the client.
- cloud-api PR #629 — SSE error frames are valid JSON for OpenAI clients.
- model-proxy PR #27 — rotation SNI + `/backends/count`.

## Test plan

- [x] 152 `inference_providers` lib tests (141 pre-existing + 11 new):
  - SSE parser: queue-full error chunk → `HttpError { status_code: 503 }`; missing `code` → 502; `error: null` not misclassified.
  - VLlmProvider: rotation gate matches the pool's `retryable_http_5xx` taxonomy; rotation disabled for one-label URLs; rotation URL shape & index; `MAX_ROTATION_FANOUT` clamp; `pending_rotation → signature_rotation` promotion; orphan cleanup with empty chat_id; `unpin_chat_connection` clears the rotation map.
- [x] 302 `services` lib tests (no regressions from moving the rotation module + adding the trait method).
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all -- --check` — clean.
- [ ] Manual: drive GLM-5.1 above `--max-queued-requests`, confirm cloud-api now returns 200 from a non-saturated backend instead of 529.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3812,6 +3812,7 @@ dependencies = [
  "tokio-stream",
  "tokio-test",
  "tracing",
+ "url",
  "uuid",
  "x509-parser 0.18.1",
 ]
--- a/crates/inference_providers/Cargo.toml
+++ b/crates/inference_providers/Cargo.toml
@@ -29,6 +29,7 @@ dstack-sdk-types = "0.1.2"
 serde_urlencoded = "0.7.1"
 regex = "1"
 sha2 = "0.11"
+url = "2.5"
 
 # Dev dependencies for testing and examples
 [dev-dependencies]
--- a/crates/inference_providers/src/external/openai_compatible.rs
+++ b/crates/inference_providers/src/external/openai_compatible.rs
@@ -11,7 +11,7 @@
 
 use super::backend::{BackendConfig, ExternalBackend};
 use crate::{
-    models::StreamOptions, sse_parser::new_sse_parser, AudioTranscriptionError,
+    models::StreamOptions, sse_parser::new_external_sse_parser, AudioTranscriptionError,
     AudioTranscriptionParams, AudioTranscriptionResponse, ChatCompletionParams,
     ChatCompletionResponse, ChatCompletionResponseWithBytes, CompletionError, ImageGenerationError,
     ImageGenerationParams, ImageGenerationResponse, ImageGenerationResponseWithBytes,
@@ -129,8 +129,11 @@ impl ExternalBackend for OpenAiCompatibleBackend {
             });
         }
 
-        // Use the SSE parser to handle the stream
-        let sse_stream = new_sse_parser(response.bytes_stream(), true);
+        // Use the SSE parser to handle the stream. External upstream → any
+        // in-stream `{"error":{...}}` frame surfaces as
+        // `HttpError { is_external: true }` so `map_provider_error` applies
+        // the third-party taxonomy (e.g. 404 → `ProviderError 502`).
+        let sse_stream = new_external_sse_parser(response.bytes_stream(), true);
         Ok(Box::pin(sse_stream))
     }
 
--- a/crates/inference_providers/src/lib.rs
+++ b/crates/inference_providers/src/lib.rs
@@ -87,7 +87,9 @@ pub use models::{
     ScoreError, ScoreParams, ScoreResponse, ScoreResult, ScoreUsage, StreamChunk, StreamOptions,
     TokenUsage, ToolChoice, ToolDefinition, TranscriptionSegment, TranscriptionWord,
 };
-pub use sse_parser::{new_sse_parser, BufferedSSEParser, SSEEvent, SSEEventParser, SSEParser};
+pub use sse_parser::{
+    new_external_sse_parser, new_sse_parser, BufferedSSEParser, SSEEvent, SSEEventParser, SSEParser,
+};
 pub use vllm::{VLlmConfig, VLlmProvider};
 
 // Chunk builder for external provider parsers
--- a/crates/inference_providers/src/lib.rs
+++ b/crates/inference_providers/src/lib.rs
@@ -60,6 +60,7 @@ pub mod chunk_builder;
 pub mod external;
 pub mod mock;
 pub mod models;
+pub mod rotation;
 pub mod spki_verifier;
 pub mod sse_parser;
 pub mod vllm;
@@ -262,6 +263,13 @@ pub trait InferenceProvider {
 
     /// Clean up the dedicated client for a chat_id after signature fetching.
     fn unpin_chat_connection(&self, _chat_id: &str) {}
+
+    /// Update the provider's view of how many healthy backends sit behind its
+    /// canonical SNI. Discovery calls this after each cycle so the traffic-
+    /// time fallback path knows how many rotation-SNI indices to iterate when
+    /// the sticky backend returns a 5xx. Default is a no-op — only providers
+    /// that participate in model-proxy rotation (vLLM) override it.
+    fn set_backend_count(&self, _count: usize) {}
 
     async fn get_attestation_report(
         &self,
--- a/crates/inference_providers/src/rotation.rs
+++ b/crates/inference_providers/src/rotation.rs
@@ -16,6 +16,14 @@ use serde::Deserialize;
 use tracing::debug;
 use url::Url;
 
+/// Hard cap on rotation fan-out, shared between the discovery cycle (which
+/// caps `/backends/count` before issuing per-index attestation probes) and
+/// the traffic-time fallback path (which caps how many indices we walk on a
+/// 5xx). Far above any realistic backend pool; hitting it means a registry
+/// reading is bogus and we're defending against unbounded fresh-TCP
+/// handshakes per request.
+pub const MAX_FANOUT: usize = 256;
+
 /// Pieces of an inference URL that the rotation path needs.
 ///
 /// `host` is the original SNI/Host the provider was registered with
--- a/crates/inference_providers/src/rotation.rs
+++ b/crates/inference_providers/src/rotation.rs
@@ -1,11 +1,12 @@
-//! Rotation-SNI helpers for backend-deterministic attestation discovery.
+//! Rotation-SNI helpers for backend-deterministic routing.
 //!
 //! Model-proxy publishes a synthetic SNI scheme `<canonical>-i<N>.<base>` that
 //! routes a fresh-TCP connection to backend `N % healthy_count`, bypassing the
 //! least-connections LB that otherwise collapses our probes onto a stable
-//! subset. Cloud-api combines this with `GET /backends/count?domain=<host>` to
-//! learn the live healthy count per cycle and iterate every backend in one
-//! pass.
+//! subset. Combined with `GET /backends/count?domain=<host>` to learn the live
+//! healthy count, callers can iterate every backend in one pass — used by both
+//! attestation discovery (deterministic per-backend probing) and traffic-time
+//! fallback (retry on a different backend when the sticky one returns 5xx).
 //!
 //! See model-proxy PR #27.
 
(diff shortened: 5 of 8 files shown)


Evrard-Nil:
Thanks for the careful review. Addressed in commit cbc765a:

**#1 (streaming rotation transport errors)** — fixed. The stream-side rotation now mirrors the non-streaming sibling: 4xx (other than 429) returns immediately (every backend would reject the same way), and every other error — including `Timeout` from `send_streaming_request`'s TTFB guard and generic `CompletionError` for TLS/TCP failures — tries the next index. The TTFB-stall case was exactly the kind of stuck-backend signal rotation should recover from, so this was a real correctness gap.

**#2 (is_external hardcoded)** — fixed. `OpenAIParserState` gained an `is_external: bool` field; new `new_external_sse_parser` constructor (also re-exported from `lib.rs`) tags state with `is_external: true`. `external::openai_compatible::chat_completion_stream` switches over; internal vLLM stays on `new_sse_parser` (defaults to `is_external: false`). New unit test asserts a third-party streaming 404 now surfaces with `is_external: true` so it maps to `ProviderError 502` rather than `InvalidModel`.

**#3 (peek-on-happy-path latency) + #4 (PR 629 SSE-frame regression for cold start)** — fixed together by gating the peek behind `rotation_count() > 0`. Both concerns shared the same root cause: peeking unconditionally when rotation couldn't actually help. Now:
- Rotation capable (`rotation_count() > 0`): peek runs, error-chunk → rotation fallback as before.
- Non-rotation (cold start before first discovery cycle, `localhost`, IP literals): pre-PR behavior preserved — return immediately after headers, let the route layer's `sse_error_frame` (PR #629) surface any in-stream error to OpenAI clients.

**#5 (pending-map cleanup contract)** — already handled. `inference_provider_pool::chat_completion_stream` always calls `provider.pin_chat_connection(&request_hash, "")` on the orphan path (when peek fails or yields no chat_id), and `pin_chat_connection` now also drains `pending_rotation` for the empty-chat_id case without leaking an entry. Test `pin_chat_connection_with_empty_chat_id_drains_pending_without_writing_signature` covers this.

**Gemini's note on client reuse**: kept the one-shot rotation client per attempt deliberately — `pool_max_idle_per_host(0)` is essential because we explicitly *don't* want H2 connection reuse across indices (the entire point is each attempt landing on a fresh, distinct backend via a fresh TCP handshake). A reused client would let the LB collapse subsequent retries onto whichever backend happened to keep the connection alive.

Re-running CI on cbc765a now.

Evrard-Nil:
Follow-up on cbc765a

The four findings from my earlier review are all correctly addressed. A few small observations to flag, none blocking.

## Verified fixes

**#1 (streaming retry on transport errors)** — `try_chat_completion_stream_rotation` now matches the non-streaming sibling. 4xx-other-than-429 returns immediately, everything else (HttpError 5xx/429, `Timeout` from `send_streaming_request`'s TTFB guard, generic `CompletionError` for TLS/TCP/transport) tries the next index. The wildcard `_ =>` arm in the inner match is the right shape — explicit enough to be reviewable, broad enough to catch future `CompletionError` variants. (`vllm/mod.rs:968-995`)

**#2 (`is_external` threading)** — `OpenAIParserState.is_external` is now propagated into the typed `HttpError` (`sse_parser.rs:206, 269`). `new_external_sse_parser` is the only public path that sets `true`; `openai_compatible::chat_completion_stream` is the only caller. Test `test_external_sse_parser_tags_error_chunks_as_external` asserts the 404 → `is_external: true` mapping. End-to-end through `map_provider_error` (`completions/mod.rs:666-685`) the third-party 404 lands on `ProviderError 502` and our own 404 still lands on `InvalidModel`.

**#3 / #4 (peek gating)** — the `match canonical_send { Ok(response) if self.rotation_count() == 0 => ... }` arm at `vllm/mod.rs:1387` cleanly preserves both pre-PR happy-path TTFB (no peek) and PR #629's `sse_error_frame` for first-chunk errors on non-rotation URLs (cold start, `localhost`, IP literals). Decision tree in the comment block at 1369-1385 is clear.

## Minor observations (non-blocking)

1. **Pool retry × rotation fan-out compounds attempts.** When every backend is unhealthy, `retry_with_fallback`'s 4 rounds × (1 canonical + N rotation) attempts amplifies fast. Per-attempt cost is low (queue-full responses return quickly with `connect_timeout(5s)`), but at the limit (N=256, capped) this is many handshakes. Today's realistic N is ~10, so this is operational rather than urgent. If it ever shows up in latency tail, the cheapest fix is sticking a "rotation already exhausted this round" flag in the request and short-circuiting.

2. **`canonical_err` is overwritten by the first iteration's `last_error = e`.** If every rotation attempt fails with a non-`HttpError` (TCP RST, TLS failure, Timeout), the final returned error type differs from the canonical 503 → `classify_retry_decision` ends up labelling this as `retryable_connection_keyword` instead of `retryable_http_5xx`. Same outcome (retry → 529), but the trace category is misleading. Easy preservation: keep `canonical_err` as a fallback when the rotation pass yielded only non-HttpError failures.

3. **408 is treated as non-retryable inside the rotation loop** (`is_rotation_retryable_status` covers only 429 + 5xx) but the pool's `classify_retry_decision` treats 408 as "next-provider-worthy". Tiny taxonomy mismatch; safe to leave since rotation indices share the same provider implementation. Mentioning for the next time someone touches the gate.

4. **Canonical SNI may land on the same physical backend as `-i0` rotation.** Least-connections LB doesn't always collapse them, but it can. Detectable only by what model-proxy exposes (it doesn't expose backend identity to the client today). Not worth adding signal for now.

5. **`build_rotation_client()` rebuilds the full `ClientConfig` per attempt** (root store + crypto provider + verifier). With `pool_max_idle_per_host(0)` the Client itself is cheap, but the config clone is non-trivial. Could cache the config behind an `OnceCell` since fingerprint_state is already an `Arc<RwLock<...>>`. Not material at current rotation counts.

6. **`MAX_ROTATION_FANOUT = 256` is defined twice** (`vllm/mod.rs:764` and `inference_provider_pool/mod.rs:931`). Discovery's cap already prevents `set_backend_count(>256)`, so the second cap inside `rotation_count()` is defense-in-depth. Worth a follow-up to extract to `rotation::MAX_FANOUT`.

LGTM. Ship it once manual GLM-5.1 verification clears the checkbox.

Evrard-Nil:
Follow-up landed in e4a70ea:

**#2 (canonical_err preservation)** — `try_chat_completion_rotation` and the streaming sibling now only overwrite `last_error` when the new error is itself an `HttpError`. Transport-level failures (TCP RST, TLS handshake, `Timeout` from the TTFB guard, generic `CompletionError`) are logged but don't poison the trace category. If every rotation index produces only transport errors, `canonical_err` (always 5xx/429 by call-site construction) survives as the returned error, so `classify_retry_decision` keeps labelling it `retryable_http_5xx` (or `_429`/`_408`) instead of falling back to `retryable_connection_keyword`.

**#3 (408 in rotation gate)** — `is_rotation_retryable_status` now matches the pool's `classify_retry_decision`: 408 + 429 + 5xx. Test renamed and broadened: `rotation_retryable_status_covers_5xx_429_and_408` asserts 408 → true plus extra negative coverage (401, 422 → false).

**#6 (de-duplicate `MAX_ROTATION_FANOUT`)** — hoisted to `inference_providers::rotation::MAX_FANOUT`. Both call sites (discovery cap + `VLlmProvider::rotation_count` clamp) reference it. Single source of truth.

Rebased onto main (picked up #611). 159 inference_providers + 304 services lib tests pass. Clippy + fmt clean.

Leaving #1 (compound retry attempts), #4 (canonical/i0 LB collision), and #5 (`ClientConfig` caching via `OnceCell`) for follow-up — none are urgent at current N≈10, and the right time to address them is when latency tail or attestation-load metrics actually surface them.

PR #639, opened 2026-05-21
ci: rollback workflow for cloud-api prod

lloydmak99:
## Motivation

Per review feedback on [nearai/cvm-ansible-playbooks#233](https://github.com/nearai/cvm-ansible-playbooks/pull/233#issuecomment-4489474439): rollback belongs next to `promote.yml`, not in the ansible-playbooks repo. This PR adds the rollback as the symmetric counterpart of [`promote.yml`](https://github.com/nearai/cloud-api/blob/main/.github/workflows/promote.yml) — same retag + audit-tag + release pattern, run in reverse, against a previously-promoted `prod-YYYYMMDD-<shortsha>` immutable tag (created by promote.yml since #604).

Paired PR [nearai/cvm-ansible-playbooks#233](https://github.com/nearai/cvm-ansible-playbooks/pull/233) remains a valid alternative path — keeping both lets emergency rollbacks fire from whichever repo the operator happens to be in.

## Changes

New workflow: `.github/workflows/rollback.yml`.

### `workflow_dispatch` inputs
- `target_tag` *(optional)* — e.g. `prod-20260101-abc1234`. If empty, auto-resolves to the most recent `prod-*` tag whose digest differs from current `:prod`, sorted by Docker Hub `last_updated`. Same-day promotions tiebreak by push time, not by short-sha lexicographic order (which would be wrong).
- `confirm` *(required)* — must equal `ROLLBACK`. Typos surface as a red ❌ workflow run with a clear error, not silently-skipped jobs.

### Jobs
1. **`preflight`** *(ubuntu-latest)* — validates `confirm`. Fails fast before consuming any retag work.
2. **`rollback-prod`** *(ubuntu-latest, `environment: production`)* — uses skopeo + jq to:
   - Capture current `:prod` digest as the "before" state for the workflow summary and release notes.
   - Resolve target tag and digest (auto or explicit). Validates tag format, refuses if target digest == current digest.
   - Re-tag the resolved digest → `:prod` and → audit tag `prod-rollback-YYYYMMDDHHMMSS-from-<orig-stamp>`. Both copies use `@sha256:...` (not the source tag) to avoid registry eventual-consistency races.
   - Create idempotent GH release (`gh release view || edit/create`) mirroring the audit tag. **Not marked `--latest`** — `:prod` has moved backwards, so the GH `latest` pointer should keep pointing at the most recent forward promotion's release.
3. **`trigger-redeploy`** *(ubuntu-latest)* — dispatches `Update Cloud API Prod` in nearai/cvm-ansible-playbooks via `gh workflow run` using `REPOSITORY_DISPATCH_TOKEN` (the same token `build.yml` already uses for cross-repo dispatch into that repo).

## Design choices

- **Why call `update_cloud_api_prod.yml`'s `workflow_dispatch` rather than `repository_dispatch`?** The reviewer's suggested `repository_dispatch` shape would require a companion change in `cvm-ansible-playbooks` to register a new event type. `Update Cloud API Prod` already exposes `workflow_dispatch` and the underlying playbook re-reads `:prod` from Docker Hub, which we just re-pointed at the rolled-back digest. So this workflow is self-contained on the cloud-api side, and PR #233 can land or not land independently.
- **Why `environment: production` only on the retag job?** Matches `promote.yml`. The trigger-redeploy job is a thin dispatch on a token, not the actual fleet write, so it doesn't need the production gate.
- **Why not pass the digest as an extra-var to the playbook?** The existing `update_cloud_api_prod.yml` doesn't expose inputs, and the user-requested constraint is no edits to that repo. Docker Hub eventual-consistency on `:prod` lookup is small relative to GH Actions queue latency between retag completing and the dispatched workflow starting, but it's a known minor race. Future hardening could add a `digest` input to `update_cloud_api_prod.yml`.
- **Why a separate audit tag in addition to the GH release?** The release is correlated to a docker tag, but consumers tooling against Docker Hub (e.g. an ansible playbook variable) needs a docker-side handle. `prod-rollback-YYYYMMDDHHMMSS-from-<orig>` is provably unique even if a same-day promotion happens to share a stamp.

## Pre-merge checklist

- [ ] `REPOSITORY_DISPATCH_TOKEN` is present in this repo (already used by `build.yml` for the same target repo). Verify it has `actions:write` (or equivalent fine-grained permission) on `nearai/cvm-ansible-playbooks` — `repository_dispatch` and `workflow_dispatch` are separate scopes.
- [ ] At least one `prod-*` stamped tag exists on Docker Hub *other* than the one currently published as `:prod`. (Two promotions since paired PR cloud-api#604 landed will satisfy this.) Until then, `target_tag` must be supplied explicitly.

## Testing

YAML linted. End-to-end test plan after merge:

1. **Sanity (no-op)**: trigger with `confirm=ROLLBAC` (typo) → expect red ❌ workflow, preflight job fails with the typo echoed.
2. **Sanity (no-op)**: trigger with `target_tag=` (current prod stamp) → expect rollback-prod job fails with "already matches current :prod digest".
3. **Auto-resolve**: trigger with empty `target_tag` + `confirm=ROLLBACK`. Expect:
   - Resolves to the second-most-recent `prod-*` tag (by `last_updated`).
   - Re-tags `:prod` and `prod-rollback-...` audit tag.
   - GH release `prod-rollback-...` created (not marked latest).
   - `Update Cloud API Prod` workflow run starts in nearai/cvm-ansible-playbooks.
   - `https://cloud-api.near.ai/v1/health` returns 200 throughout (blue-green failover inherited from the existing upgrade path).


lloydmak99:
Thanks for the thorough review. All issues addressed in commits [8f5769b](https://github.com/nearai/cloud-api/commit/8f5769b), [3169ef5](https://github.com/nearai/cloud-api/commit/3169ef5), and [70245bd](https://github.com/nearai/cloud-api/commit/70245bd):

**⚠️ Shell injection (blocker)** — Fixed. Both `inputs.confirm` and `inputs.target_tag` are now passed through step-level `env:` vars (`CONFIRM` and `TARGET_TAG_INPUT`) and referenced as `$VAR` in shell. The `::error::` message no longer re-interpolates the confirm value.

**⚠️ No concurrency guard (blocker)** — Fixed. Added `concurrency: group: prod-image-mutation / cancel-in-progress: false` to both `rollback.yml` and `promote.yml` so they share the same serialization lock. A rollback will queue behind an in-flight Thursday promotion (and vice versa) rather than racing on `:prod`.

**Nit: skopeo inspect error message** — Fixed. Added `2>/dev/null || true` to the initial `:prod` inspect, matching the explicit error handling used a few lines later for `TARGET_DIGEST`.

**Nit: fire-and-forget redeploy** — Deferred per your suggestion. The operator-attended assumption holds for emergency rollbacks.

Evrard-Nil:
I think the tracing changes are unrelated?

Evrard-Nil (approved):

PR #644, opened 2026-05-22
fix(vllm): raise control_timeout default from 90s to 300s

Evrard-Nil:
## Summary

Bump `VLlmConfig::DEFAULT_CONTROL_TIMEOUT_SECS` from **90s** to **300s**.

The control timeout in `inference_providers/src/vllm/mod.rs` gates three call paths:

1. **TTFB on streaming chat completions** (`send_streaming_request`, around line 793) — wraps `client.post(...).send()` so the timeout applies only to receiving response headers, not body streaming.
2. **`/v1/attestation/report` fetches** (`get_attestation_report`, around line 1343) — full request timeout.
3. Models-list, signature lookups, and other control-plane operations.

### Why bump

- **GLM-5.1, Qwen3.5** are reasoning models. Under load, vLLM/sglang can queue a request behind a long thinking phase and delay HTTP response headers by minutes. The 90s control_timeout fires as `CompletionError::Timeout` before the model emits anything — surfaces to clients as a 504 / failure even though the backend is healthy.
- **Attestation fetches** trigger TDX quote generation + GPU evidence collection (NVIDIA `cc_admin.collect_gpu_evidence_remote` via Python subprocess, serialized behind a host-level mutex in `inference-proxy`). Under contention this regularly crosses 90s. We've been seeing intermittent `503 Service Unavailable` on `/v1/attestation/report?model=zai-org/GLM-5.1-FP8` with body:
  > `"error sending request for url (https://glm-5-1.completions.near.ai/v1/attestation/report?...)"`
- **300s** matches `cloud_api_helpers.TIMEOUT` used by `infra-tests`, and is the value reasoning tests already cite (`nearai/infra#121`). It sits well below the 600s `completion_timeout` cap so a sustained backend stall still surfaces, just no longer at 90s.

### Backward compat

- The constant remains overridable per-deployment via `VLLM_PROVIDER_CONTROL_TIMEOUT`.
- Existing unit tests reference `VLlmConfig::DEFAULT_CONTROL_TIMEOUT_SECS` symbolically (not the hard-coded 90), so they automatically track the new default. `cargo check -p inference_providers` is clean.

## Test plan

- [x] `cargo check -p inference_providers`
- [ ] CI: full `cargo test` and clippy
- [ ] Post-deploy: re-run the attestation burst probe against `cloud-api.near.ai/v1/attestation/report?model=zai-org/GLM-5.1-FP8` and confirm no transient 503s under the prior reproduction (4/500 fails at conc=8 cold-start)

### DIFF
--- a/crates/inference_providers/src/vllm/mod.rs
+++ b/crates/inference_providers/src/vllm/mod.rs
@@ -124,8 +124,13 @@ impl VLlmConfig {
     /// on a single non-streaming request; 600s is a comfortable ceiling that
     /// still surfaces genuinely stuck requests.
     pub const DEFAULT_COMPLETION_TIMEOUT_SECS: i64 = 600;
-    /// Default control timeout. Metadata/TTFB ops should resolve quickly.
-    pub const DEFAULT_CONTROL_TIMEOUT_SECS: i64 = 90;
+    /// Default control timeout. Covers TTFB on streaming requests, attestation
+    /// report fetches, models-list and signature lookups. Reasoning models
+    /// (GLM-5.1, Qwen3.5) can delay TTFB by minutes when the backend queues a
+    /// request behind a long thinking phase, and attestation TDX-quote + GPU
+    /// evidence collection can also cross 90s under load. 300s gives enough
+    /// headroom for those without masking a sustained backend stall.
+    pub const DEFAULT_CONTROL_TIMEOUT_SECS: i64 = 300;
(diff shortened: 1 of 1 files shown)


Evrard-Nil:
Addressed reviewer feedback in 76d3922: updated the two stale comments that referenced the old 90s/270s values. Comments-only — no behavior change.

PR #645, opened 2026-05-22
feat(external): support extra_request_body for OpenRouter provider preferences

PierreLeGuen:
## Summary

- Adds `extra_request_body` field to `ProviderConfig::OpenAiCompatible` — a map of JSON fields injected into every outgoing request body
- Enables configuring OpenRouter-specific parameters (provider preferences, fallback behavior, data collection policies) at the model level via `PATCH /v1/admin/models`
- User-supplied request fields take precedence over provider defaults

Example `providerConfig` for an OpenRouter model with preferred provider:
```json
{
  "backend": "openai_compatible",
  "base_url": "https://openrouter.ai/api/v1",
  "model_name": "moonshotai/kimi-k2.6",
  "extra_request_body": {
    "provider": {
      "order": ["Chutes"],
      "allow_fallbacks": true,
      "data_collection": "deny"
    }
  }
}
```

## Test plan

- [x] Unit test: `extra_request_body` deserializes from JSON config
- [x] Unit test: fields injected into request params
- [x] Unit test: user-supplied fields take precedence over provider defaults
- [x] All 307 existing + new unit tests pass
- [x] `cargo fmt` and `cargo clippy` clean
- [ ] Staging: Kimi K2.6 and Qwen3.7 Max deployed via OpenRouter and tested

### DIFF
--- a/crates/inference_providers/src/external/anthropic/mod.rs
+++ b/crates/inference_providers/src/external/anthropic/mod.rs
@@ -276,6 +276,7 @@ mod tests {
             api_key: "test-key".to_string(),
             timeout_seconds: 30,
             extra: std::collections::HashMap::new(),
+            extra_request_body: std::collections::HashMap::new(),
         };
 
         let headers = backend.build_headers(&config).unwrap();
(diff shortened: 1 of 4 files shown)


PierreLeGuen:
Fixed in e6336ac. extra_request_body now deep-merges nested defaults, the helper works on the shared extra map, and the same injection path is applied to the other JSON-body calls that carry extra metadata.

Checked: cargo fmt --check, focused inference_providers tests, external::tests, and clippy for inference_providers.

Evrard-Nil (approved):

PR #648, opened 2026-05-22
Show organization names in invitations

PierreLeGuen:
## Summary
- enrich `/v1/users/me/invitations` responses with `organization_name`
- add a converter for `OrganizationInvitationWithOrgResponse`
- add e2e coverage for the invited-user inbox response

## Tests
- `cargo fmt --all --check`
- `cargo test -p api --test e2e_all test_user_invitations_include_organization_name` (compiled, but local run could not bootstrap because Postgres was not listening on localhost:5432)

### DIFF
--- a/crates/api/src/conversions.rs
+++ b/crates/api/src/conversions.rs
@@ -669,6 +669,17 @@ pub fn services_invitation_to_api(
     }
 }
 
+/// Convert services OrganizationInvitationWithDetails to API OrganizationInvitationWithOrgResponse
+pub fn services_invitation_to_api_with_org(
+    invitation: services::organization::OrganizationInvitationWithDetails,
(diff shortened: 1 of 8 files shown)


PierreLeGuen:
Addressed the review feedback in 4ed2a35:

- Replaced the route-level per-invitation organization lookup with a single repository query that joins invitations to organizations.
- Removed the empty-string fallback path; dangling org rows are not returned by the inner join, and valid invitations keep the real org name.
- Populated `invited_by_display_name` from the inviter user row instead of hard-coding `null`.
- Updated the e2e assertion to cover both `organization_name` and `invited_by_display_name`.

The Gemini review thread for the N+1 issue is resolved.

Evrard-Nil (approved):

PR #661, opened 2026-05-25
feat(models): OpenRouter-compatible fields on /v1/models

Evrard-Nil:
## Summary
Makes `GET /v1/models` conform to [OpenRouter's provider spec](https://openrouter.ai/docs/guides/community/for-providers) so cloud-api can register as an OR provider. Adds the metadata fields OR requires (HF id, quantization, max output length, supported features, supported sampling parameters), emits per-token USD-string pricing alongside the legacy per-million floats, and returns a real `created` timestamp.

- **Schema (V0050)**: 5 nullable/empty-default columns on `models` and `model_history` — admins backfill per model, no data migration needed for existing rows.
- **`/v1/models` response**: new top-level fields (`name`, `description`, `hugging_face_id`, `quantization`, `max_output_length`, flat `input_modalities`/`output_modalities`, `supported_sampling_parameters`, `supported_features`); `pricing` gains OR's `{prompt, completion, image, request, input_cache_read}` as strings while keeping the legacy `{input, output}` floats for back-compat.
- **`/v1/admin/models` PATCH**: accepts the new fields under camelCase keys (`huggingFaceId`, `maxOutputLength`, `supportedSamplingParameters`, …); upsert SQL + history audit preserve them.

Plumbed end-to-end through `Model` / `ModelHistory` (db) → `ModelWithPricing` (service, gains `created_at`) → `ModelInfo` (api), plus the parallel admin path (`UpdateModelAdminRequest`, `ModelPricing`, `ModelHistoryEntry`, `AdminModelInfo`).

## Follow-ups (out of scope)
- Backfill the new columns for the 38 production models via `/v1/admin/models` PATCH before exposing the catalog to OpenRouter.
- Business side: set up auto top-up or invoicing with OR (their other onboarding requirement).

## Test plan
- [x] `cargo check --workspace --tests`
- [x] `cargo clippy --workspace --no-deps`
- [x] `cargo test --lib --bins --workspace` (580 unit tests pass)
- [ ] e2e: bring up Postgres, run migrations, confirm V0050 applies cleanly
- [ ] e2e: PATCH a model via `/v1/admin/models` with new fields, GET `/v1/models`, verify shape matches OR spec
- [ ] e2e: confirm history audit captures the new fields on upsert


PierreLeGuen (commented):
**Findings**

[P2] New admin fields are write-only in admin APIs. `PATCH /v1/admin/models` accepts and persists `huggingFaceId`, `quantization`, `maxOutputLength`, `supportedSamplingParameters`, and `supportedFeatures`, but `ModelMetadata` and `ModelHistoryEntry` do not expose them, and the admin response mappings drop them: `crates/api/src/models.rs`, `crates/api/src/routes/admin.rs`. Operators cannot read back or audit the metadata they just set, so backfilling OpenRouter data through the admin API is mostly blind.

[P2] OpenRouter metadata is accepted without validation. The route forwards the new fields directly into `UpdateModelAdminRequest`, but `validate_model_request` still only checks model name/display/description. Invalid quantization/features/sampling parameters, or non-positive `maxOutputLength`, will be stored and emitted from `/v1/models`, producing a catalog that violates the OpenRouter provider contract: https://openrouter.ai/docs/guides/community/for-providers.

**Verification**

`cargo check -q` passed. A targeted `cargo test -q -p api routes::completions::tests` could not complete because the filesystem ran out of space while writing Rust build/query-cache artifacts; I cleaned the generated artifacts afterward.

PierreLeGuen (approved):
Approved. I rechecked the latest head and the follow-up commits address my review findings: the OpenRouter fields are now visible in readback/history paths, and the admin write path validates the OpenRouter vocabulary before persisting.

PR #674, opened 2026-05-26
Track and resend organization invitation emails

### LINKED ISSUE #578: Admin APIs for invitation email delivery oversight and resend
## Context
Organization invitation emails are now sent from `cloud-api` through Resend and invitation rows track delivery metadata (`email_status`, `email_sent_at`, `email_last_error`, `email_message_id`). We need admin-facing APIs so operators can oversee delivery health and manually recover failed/skipped emails.

## Goals
- Give platform admins visibility into organization invitation email delivery status.
- Allow admins to resend invitation emails when delivery failed, was skipped, or needs manual retry.
- Preserve auditability and avoid exposing invitation tokens unnecessarily.

## Proposed API work
- Add admin-only endpoints under the existing admin route namespace to list invitation email deliveries with filters for organization, recipient email, status, date range, and invitation status.
- Include response fields for organization id/name, invitation id, recipient email, role, invitation status, email status, sent timestamp, last error, Resend email id, inviter id/email, created/expires/responded timestamps.
- Add admin-only resend endpoint for a single invitation, reusing the same Resend sender/template path as normal invitation creation.
- Add optional bulk resend endpoint for filtered failed/skipped/not_attempted pending invitations, with clear per-invitation results.
- Record each resend attempt back onto `organization_invitations` delivery columns and emit structured logs/metrics.
- Enforce safe resend rules: invitation must exist, still be pending, and not expired unless a separate reissue flow is implemented.

## Acceptance criteria
- Platform admin auth is required for all endpoints.
- Listing supports pagination and deterministic ordering.
- Resend returns per-invitation success/failure with sanitized errors.
- Failed resend does not delete or invalidate existing invitations.
- Tests cover permissions, listing filters, single resend success/failure, expired/non-pending resend rejection, and delivery metadata updates.

## Cross references
- Implementation PR: https://github.com/nearai/cloud-api/pull/579
- Admin UI issue: https://github.com/nearai/nearai-cloud-admin-ui/issues/34

PierreLeGuen:
## Summary
- add admin endpoints to list invitation email deliveries and resend pending invitations
- expose delivery and invitation status fields in API responses
- add repository and service support plus coverage for list and resend flows

## Testing
- cargo fmt --check
- cargo check -p services -p database -p api
- cargo test -p services resend_invitation_email --lib
- cargo test -p api --test e2e_all admin_invitation_email_deliveries -- --nocapture (blocked locally: Postgres connection refused)

Closes #578


PierreLeGuen:
Addressed review feedback:

- Cast invitation enum columns to text in list/count filters.
- Mark expired invitations as expired during resend attempts before returning the validation error.
- Kept the admin route builder under the lint argument limit by bundling route dependencies.
- Removed unnecessary joins from the delivery count query.
- Avoided stale resend responses by using the recorded invitation result when available and returning not found if a fallback reload cannot find the invitation.
- Return an explicit error when invitation email delivery is skipped because the invitation URL is not configured.

Local checks:
- cargo fmt --all -- --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo check -p services -p database -p api
- cargo test -p services resend_invitation_email --lib

Evrard-Nil (approved):
Pulled the branch, reviewed end-to-end, and ran the test suite locally.

**Build:** `cargo check -p services -p api -p database` clean.
**Tests:** All 11 invitation tests pass (5 new `resend_invitation_email_*` + 5 existing `create_invitations_*` + 1 email rendering). No regressions.
**Scope check:** Diff vs. correct base (`837c4a5`) is ~1210 additions across 11 files — the `main..pr-674` apparent removal of `thought_signature` is just an artifact of the PR being branched before #649 merged, not a real revert. Worth rebasing to get rid of that confusion.

Overall the change is well-structured. Refactoring `send_invitation_email` from `(bool, Option<String>)` to a `InvitationEmailAttempt` struct that threads back the updated `OrganizationInvitation` is a clean improvement and the resend path reuses it cleanly. Single-query `list_email_deliveries` with the JOIN — good (avoided the N+1 pattern that bit cloud-api#662). Admin-only gating via `AdminUser` extension confirmed. Tokens are NOT exposed in the admin response — verified the response model has no `token` field.

**One important thing worth surfacing as a real concern:**

🔒 **`email_last_error` privacy surface.** `sanitize_error()` (`crates/services/src/email.rs:40`) only collapses whitespace and truncates to 1000 chars — it is not a PII scrubber. If Resend ever returns something like `\"Invalid recipient: foo@bar.com\"` or `\"Domain blocked: example.com\"`, that goes straight into the DB column `email_last_error` and surfaces via `GET /v1/admin/invitation-email-deliveries`. Per the privacy rules in `CLAUDE.md` (\"Never log request/response bodies — Even at debug level, unless you're 100% certain they don't contain customer data\"), this is a soft violation. This isn't a log strictly, but it's persistent and externally visible. Either:
- Tighten `sanitize_error` to strip email-looking patterns and provider-side identifiers, OR
- Document explicitly that this is admin-only inside the TEE perimeter and that acceptable downstream consumers are trusted (and update CLAUDE.md if so).

1. **Subtle behavior change to existing `create_invitations` response.** When `invitations_url` is unset, `email_error` was `None`; now it's `Some(\"Invitation URL is not configured\")`. The unit test at `crates/services/src/organization/mod.rs:2254` was updated to match, but external clients that parsed `email_error == None` as \"success\" will see a new string. Worth a one-liner in the PR description / changelog so consumers know.

2. **No anti-abuse on `/resend`.** An admin (or compromised admin session) can hit POST `…/{id}/resend` in a tight loop. Each call performs a network roundtrip to Resend and writes a DB row. Resend's external rate limit will eventually catch this, but we'd churn `email_last_error` rows. Probably fine because admin-only, but a per-invitation cooldown (e.g. \"can't resend within 30s of the last attempt\") would be defensive.

3. **TOCTOU between status read and send.** `resend_invitation_email_impl` reads status=Pending, then calls the network-bound `send_invitation_email`, then writes results. Between the read and the send, the invitee may have accepted/declined — race is benign (one extra email goes out for an already-accepted invite, status stays Accepted because `record_email_*` doesn't change `status`). Worth a code comment noting that's intentional.

4. **List pagination consistency.** `list_email_deliveries` issues the rows query and the count query via separate pool acquisitions. Under concurrent writes, `total` and `deliveries.len()` can briefly disagree. Acceptable for admin observability; calling it out so it's not a surprise later.

5. **No composite index for the sort.** `ORDER BY i.created_at DESC, i.id DESC` plus org_id filter would benefit from a composite `(organization_id, created_at DESC, id DESC)` index when invitation rows grow. Not needed today.

6. **Email substring search is unindexable.** `email ILIKE '%' || $2 || '%'` prevents the `idx_org_invitations_email` index from being used; full filter scan only. Acceptable for admin audit.

7. **Resend endpoint lacks an e2e test.** The list endpoint has `admin_invitation_email_deliveries.rs`, which is great. The resend endpoint relies on unit tests only. Adding a simple e2e (insert invitation fixture → POST resend → assert email_status changed) would round it out — author's own validation note says local Postgres was refused, so this would also unblock that gate.

LGTM. The two critical things author needs to confirm/address are (a) the `email_last_error` PII surface (CLAUDE.md privacy rule), and (b) the `create_invitations` response behavior change in changelog/migration notes. Everything else is polish. Approving.

PR #676, opened 2026-05-26
fix(completions): pass unknown chat tool types through to upstream

Evrard-Nil:
## Problem

`extract_tools_from_extra` in `crates/services/src/completions/mod.rs` had two interacting bugs:

```rust
let tools = extra.remove("tools").and_then(|v| {
    serde_json::from_value::<Vec<inference_providers::ToolDefinition>>(v).ok()
});
```

1. `extra.remove(\"tools\")` happens **unconditionally**, before parsing.
2. `.ok()` swallows parse errors silently.

Net effect: any chat-completions request whose `tools` array doesn't fit our typed `ToolDefinition` shape (i.e. anything other than `{type:\"function\", function:{name,description,parameters}}`) is silently dropped — removed from `extra` and parsed to `None`. The upstream never sees the tool, and the model never knows it could call it.

This blocked the `web_context_search` agent loop (nearai/inference-proxy#144) from running end-to-end through cloud-api. The inference-proxy expands `{\"type\":\"web_context_search\"}` into a proper function tool inside the CVM, but cloud-api was stripping it before the request ever got there.

Empirical evidence: `POST /v1/chat/completions` with `tools:[{\"type\":\"web_context_search\"}]` against GLM-5.1 returns `prompt_tokens=19` (just the user message). The tool schema alone would add ~100 tokens — so cloud-api forwarded the request without it.

## Fix

Only consume from `extra` on **successful** parse. If parsing fails, the raw value stays in `extra` and flows through to the upstream request body verbatim via the existing `#[serde(flatten)]` on `ChatCompletionParams.extra`. Same treatment for `tool_choice` for symmetry.

This is a permissive-passthrough change at the cloud-api layer: deciding whether a tool type is supported is the upstream's job (vLLM/SGLang via inference-proxy for our models, the provider for Gemini/Anthropic). External providers already only consume typed tools, so they're unaffected — non-function tools simply don't reach them.

## Tests

`cargo test -p services --lib extract_tools` (4 new tests, all pass):

- `extract_tools_consumes_function_shaped_tools` — regression cover for the existing path: standard function tools parse, get consumed from `extra` (so we don't end up with both the typed `tools` field and a flattened `extra.tools`, which would be a duplicate key in the upstream JSON).
- `extract_tools_passes_through_unknown_tool_types` — the `web_context_search` regression: array remains in `extra`.
- `extract_tools_passes_through_mixed_unknown_in_array` — an array mixing function and unknown stays fully opaque. Conservative: don't strip some entries; cloud-api isn't the right layer to half-parse arrays.
- `extract_tools_passes_through_unknown_tool_choice` — symmetric guard for `tool_choice`.

Full suite: `cargo test -p services --lib` → 315 existing + 4 new pass. `cargo clippy --all-targets -- -D warnings` clean. `cargo fmt --check` clean.

## Test plan after deploy

- [ ] Plain (no-E2EE) `POST /v1/chat/completions` to `cloud-api.near.ai` with `tools:[{\"type\":\"web_context_search\"}]` against a GLM-5.1 deployment that has `WEB_CONTEXT_SEARCH_*` env set (gpu23/04/03/26 — done in cvm-conf v0.0.191). Expect: `nearai_tool_result` chunk + `data: [DONE]`.
- [ ] Same request with E2EE headers (`X-Signing-Algo`, `X-Client-Pub-Key`). Expect: encrypted `nearai_tool_result.output` on the wire, decryptable with the client key.
- [ ] Existing chat completion + standard `function`-typed tool — should be byte-identical to today (the regression test covers this in code).

## Related

- nearai/inference-proxy#144 — server-side `web_context_search` loop in the inference-proxy.
- nearai/cvm-compose-files: digest bump + Brave env wiring to all 4 GLM-5.1 hosts (gpu23, gpu04, gpu03, gpu26).

### DIFF
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -555,22 +555,48 @@ impl CompletionServiceImpl {
         }
     }
 
-    /// Extract tools and tool_choice from extra HashMap if present.
-    /// This handles the case where the Responses API places tools in request.extra.
-    /// Returns (tools, tool_choice) and removes them from extra to avoid duplication.
+    /// Extract tools and tool_choice from the extra HashMap if present and
+    /// parseable as the typed `ToolDefinition` / `ToolChoice` shapes.
+    ///
+    /// Returns `(tools, tool_choice)` and only removes the corresponding
+    /// key from `extra` when parsing succeeds. If parsing fails — for
+    /// example because the client sent a non-standard tool type like
+    /// NEAR's `{"type":"web_context_search"}`, which the upstream
+    /// inference-proxy handles natively but doesn't fit the
+    /// function-tool shape — the original value is left in `extra` so it
+    /// flows through to the upstream request body verbatim via the
+    /// `#[serde(flatten)]` on `ChatCompletionParams.extra`. This means
+    /// cloud-api never silently drops tool definitions it doesn't
+    /// recognise; the upstream gets to decide.
     fn extract_tools_from_extra(
         extra: &mut std::collections::HashMap<String, serde_json::Value>,
     ) -> (
         Option<Vec<inference_providers::ToolDefinition>>,
         Option<inference_providers::ToolChoice>,
     ) {
-        let tools = extra.remove("tools").and_then(|v| {
-            serde_json::from_value::<Vec<inference_providers::ToolDefinition>>(v).ok()
-        });
+        let tools = match extra.get("tools").cloned() {
+            Some(raw) => {
+                match serde_json::from_value::<Vec<inference_providers::ToolDefinition>>(raw) {
+                    Ok(parsed) => {
+                        extra.remove("tools");
+                        Some(parsed)
+                    }
+                    Err(_) => None,
+                }
+            }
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Reviewed the live `/v1/chat/completions` service path plus the relevant outbound provider paths.

I do not see a blocking issue. The helper now only consumes `tools` / `tool_choice` when they parse into the typed shape; otherwise the raw values stay in `extra`, which is what the direct pass-through providers serialize. Provider-native adapters still rebuild their own request bodies from typed fields, so this does not unexpectedly enable unsupported raw tool shapes there.

Verified locally:
- `cargo fmt --check`
- `cargo test -p services --lib extract_tools`
- `cargo test -p services --lib`
- `cargo test -p inference_providers --lib strip_reasoning_effort`
- `cargo test -p inference_providers --lib flatten`

GitHub checks are green. Non-blocking: the PR text should be explicit that pass-through providers may now surface upstream validation errors for tool shapes that were previously silently dropped.

Issue #26, opened 2025-10-11
Feature: disable or enable an API key


think-in-universe:
### Problem

It's a common case that the user may want to disable an API key without deleting the key, so she can still track the usage history of the API key even if the key has been disabled. This is supported in the current NEAR AI Cloud. 

<img width="388" height="463" alt="Image" src="https://github.com/user-attachments/assets/b7addf21-4a42-459e-b62f-bf05fb898fdf" />

### Solution

In [update API key endpoint](https://nearai-platform-api-dev-59501de18ae5.herokuapp.com/docs#/Workspaces/update_workspace_api_key), add a boolean parameter `enabled` that can be set to `true` or `false` when needed. If the API key is disabled, it cannot be used for authentication any more.

PR #677, opened 2026-05-26
fix(completions): preserve unknown chat delta fields end-to-end

Evrard-Nil:
## Problem

After #676 merged and deployed to staging, the agent-loop test against `cloud-stg-api.near.ai` showed the model successfully calling `web_context_search` — `prompt_tokens=223` (tool schema reached the model), `finish_reason: tool_calls` on iter 1, `finish_reason: length` on iter 2 (loop continued and was producing a grounded answer when max_tokens hit). But **no `nearai_tool_result` chunks reached the client**, just the model's own `tool_calls` and `content`.

Root cause: same pattern as #676 but on the response side. `inference_providers::models::ChatDelta` is a typed struct:

```rust
pub struct ChatDelta {
    pub role: Option<MessageRole>,
    pub content: Option<String>,
    ... // explicit fields
    pub reasoning: Option<String>,
}
```

No `#[serde(flatten)]` catch-all, no `#[serde(deny_unknown_fields)]`. Per serde's default behavior, unknown fields on deserialize are **silently dropped**. The inference-proxy's server-side agent loop (nearai/inference-proxy#144) emits a synthetic `delta.nearai_tool_result` chunk between iterations. Cloud-api's intercept layer deserializes the chunk into `ChatDelta` (for signing / usage tracking / streaming re-emission), `nearai_tool_result` is dropped, and the re-serialized chunk doesn't carry it.

## Fix

Add a flattened catch-all to `ChatDelta`:

```rust
#[serde(flatten)]
pub extra: std::collections::HashMap<String, serde_json::Value>,
```

Unknown delta fields now round-trip verbatim. Derived `Default` on the struct so the empty catch-all is `HashMap::new()` by default, then updated the handful of explicit `ChatDelta { ... }` literals across the workspace to set `extra: Default::default()`.

This is symmetric with the request-side fix in #676: cloud-api stops being opinionated about JSON shapes it doesn't recognize; the upstream (inference-proxy) gets to decide.

## Tests

New regression: `chat_delta_preserves_unknown_fields_round_trip` in `crates/inference_providers/src/models.rs`. Constructs a chunk with `delta.nearai_tool_result`, deserializes it, asserts the catch-all contains the expected payload, then round-trips the re-serialized JSON to confirm the synthetic field is on the wire.

`cargo test --workspace --lib` clean: 315 + 174 + 106 + 16 + 11 passing (including the new test), 0 failures, 3 ignored (pre-existing). `cargo clippy --all-targets -- -D warnings` clean. `cargo fmt --check` clean.

## Test plan after deploy

```bash
export API_KEY=sk-...
python3 py/encrypted_agent_loop_verifier.py --model zai-org/GLM-5.1-FP8
# or
curl -sS -N -X POST 'https://cloud-api.near.ai/v1/chat/completions' \
  -H "Authorization: Bearer $API_KEY" -H 'Content-Type: application/json' \
  -d '{"model":"zai-org/GLM-5.1-FP8","stream":true,"max_tokens":500,
       "messages":[{"role":"user","content":"How deep is the Mediterranean? Use web_context_search."}],
       "tools":[{"type":"web_context_search"}]}' \
  | grep -E 'nearai_tool_result|finish_reason|\[DONE\]'
```

Expected: at least one `nearai_tool_result` line in the output, followed by a second iteration with `finish_reason: stop` and `[DONE]`. The verifier script (in `nearai-cloud-verifier` PR #26) returns exit 0 on a clean tool round-trip.

## Related

- nearai/cloud-api#676 — request-side: don't strip unknown chat tool types. Merged + deployed; this PR is the matching response-side fix.
- nearai/inference-proxy#144 — server-side agent loop, emits the `nearai_tool_result` chunks.
- nearai/cvm-compose-files v0.0.191 — inference-proxy digest bump + Brave env on all four GLM-5.1 backends.
- nearai/nearai-cloud-verifier#26 — customer-facing demo script (already validates the loop against `completions.near.ai`; will go green against `cloud-api.near.ai` once this PR deploys).


PierreLeGuen (approved):
Reviewed locally; no blocking findings.

PR #679, opened 2026-05-26
Preserve hidden state fields in chat completions

satojandro:
## Summary

Adds support for hidden-state passthrough in chat completions.

- Adds explicit `return_hidden_states` and `layers` request fields to `ChatCompletionParams`
- Adds flattened `extra` catch-all fields to `ChatCompletionChunk` and `ChatCompletionResponse`
- Preserves provider-specific response fields such as `hidden_states` during deserialize/reserialize
- Updates affected struct literals with default `extra` values
- Adds regression coverage for unknown field round-tripping on streamed chunks and full responses

## Testing

- `cargo fmt --check`
- `cargo test -p inference_providers`
- `cargo check`
- `git diff --check`

This mirrors the existing `ChatDelta` catch-all behavior so unknown upstream fields are not silently dropped.

### DIFF
--- a/crates/api/src/conversions.rs
+++ b/crates/api/src/conversions.rs
@@ -114,6 +114,8 @@ impl From<ChatCompletionRequest> for ChatCompletionParams {
             store: None,
             stream_options: None,
             modalities,
+            return_hidden_states: None,
+            layers: None,
             extra,
         }
     }
--- a/crates/api/src/conversions.rs
+++ b/crates/api/src/conversions.rs
@@ -114,8 +114,8 @@ impl From<ChatCompletionRequest> for ChatCompletionParams {
             store: None,
             stream_options: None,
             modalities,
-            return_hidden_states: None,
-            layers: None,
+            return_hidden_states: extra.remove("return_hidden_states").and_then(|v| v.as_bool()),
+            layers: extra.remove("layers").and_then(|v| serde_json::from_value(v).ok()),
             extra,
         }
     }
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -719,6 +719,7 @@ fn build_flush_chunks(states: &mut StreamUnredactStates, template: &ChunkTemplat
             usage: None,
             prompt_token_ids: None,
             modality: None,
+            extra: Default::default(),
         };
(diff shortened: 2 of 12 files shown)


Evrard-Nil:
Thanks for the PR, no backend has --enable-return-hidden-states currently so this would be a no-op. May I ask what's your use case?

satojandro:
Hey mate, thanks for the prompt review and reply. I'm doing mechanistic interpretability research that requires per-layer activation from transformer models. gland and vLLM both support returning hidden stated natively, so the main gap is the proxy layer stripping them them before they reach the client(?)

I understand that this may be currently a no-op, but this was actually suggested to be by Illia, so hope we can find a way to enable it. Happy to have a chat on telegram or call @Evrard-Nil

Evrard-Nil:
Thanks again @satojandro for this. I dug into where sglang actually emits the activations and opened #739 to carry it forward (you're credited as co-author).

One wrinkle worth flagging: sglang attaches `hidden_states` **per-choice** (`choices[].hidden_states`) for non-streaming, and **inside the delta** (`choices[].delta.hidden_states`) for streaming — not at the top level. So the top-level catch-all here wouldn't have captured them on cloud-api's re-serialized response path (verified against sglang `entrypoints/openai/protocol.py`). #739:

- moves the response catch-all to `ChatCompletionResponseChoice` (streaming is already handled by the existing `ChatDelta` catch-all), and
- relies on the existing `ChatCompletionParams.extra` for the request side rather than typed fields — which also resolves the silent-drop-on-mistype concern from the review here.

Still gated on a backend launched with `--enable-return-hidden-states`, but the proxy layer is ready. Leaving this open so you can take a look at #739.

Evrard-Nil:
Hi @satojandro, I was checking `--enable-return-hidden-states` impact on performance and it seems to have a pretty big drawdown in performance so I'd rather not enable it on all the models. Are there specific models you're interested in?

satojandro:
Hi @Evrard-Nil - thanks for looking into this and implementing it. Preference would be DeepSeek v4 Flash, but I am flexible. If it does have a meaningful impact on performance, enable it on the model that would be least disruptive for NEAR AI current operations. Any large model is suitable now for the current round of testing

PR #680, opened 2026-05-27
ci: add Re-tag Image workflow

Evrard-Nil:
## Summary
- Adds `.github/workflows/retag.yml`, a `workflow_dispatch` workflow that points one tag at the image currently published under another tag (or at an explicit `sha256:` digest). E.g. set `:staging` to whatever `:latest` currently is.
- Resolves the `source` to a digest first, then `skopeo copy` from the digest to the `target` tag, so the bytes written are provably identical to what was inspected and immune to a concurrent push racing the source tag mid-run.
- Refuses `target=prod`: the `:prod` tag is owned by `promote.yml` / `rollback.yml`, which serialize (concurrency group `prod-image-mutation`) and audit every prod mutation. This workflow never bypasses that flow.

## Inputs
- `source` — tag (e.g. `latest`, `staging`) or digest (`sha256:...`) to copy from
- `target` — tag to (re)point at the source (e.g. `staging`); `prod` rejected

## Test plan
- [ ] Dispatch with `source=latest`, `target=staging`; confirm `:staging` digest now matches `:latest`
- [ ] Dispatch with `target=prod`; confirm it fails fast with the guard message
- [ ] Dispatch with a bogus `source`; confirm the resolve step errors cleanly

### DIFF
--- a/.github/workflows/retag.yml
+++ b/.github/workflows/retag.yml
@@ -37,9 +37,15 @@ jobs:
           TARGET: ${{ inputs.target }}
         run: |
           set -euo pipefail
-          # :prod is owned by promote.yml / rollback.yml (serialized + audited).
-          if [ "$TARGET" = "prod" ]; then
-            echo "::error::Refusing to target :prod. Use promote.yml or rollback.yml, which serialize and audit prod mutations."
+          # The entire prod-* tag namespace is owned by promote.yml / rollback.yml:
+          # :prod itself, the immutable prod-YYYYMMDD-<shortsha> rollback history,
+          # and prod-rollback-* audit tags. rollback.yml auto-selects the newest
+          # prod-* tag as a rollback candidate, so writing any prod-* tag here —
+          # without the production environment, the prod-image-mutation concurrency
+          # group, or a GitHub release — could pollute that history and later point
+          # a rollback at an unaudited image. Refuse the whole namespace.
+          if echo "$TARGET" | grep -Eq '^prod($|-)'; then
+            echo "::error::Refusing to target the prod-* tag namespace ('${TARGET}'). Use promote.yml or rollback.yml, which serialize and audit prod mutations."
             exit 1
           fi
           # Reject obviously malformed refs so we fail before touching the registry.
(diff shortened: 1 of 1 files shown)


PierreLeGuen (changes requested):
Requesting changes for one production-tag safety issue.

.github/workflows/retag.yml:40-47 only rejects `target=prod`, but `.github/workflows/rollback.yml:112-124` treats `prod-YYYYMMDD-<shortsha>` tags as production rollback history and auto-selects the newest matching Docker Hub tag. This workflow can still write a target like `prod-20260527-deadbee` without the production environment, `prod-image-mutation` concurrency group, or GitHub release audit used by `promote.yml`/`rollback.yml`, which can pollute rollback candidates and later point rollback at an unaudited/non-prod image. Please reject all production-owned tag namespaces here, at least `^prod($|-)`, or otherwise exclude `prod-*` from this retag tool.

Checks: inspected the diff and the existing promote/rollback/build workflows locally; `git diff --check origin/main...HEAD` passed; `gh pr checks 680 --watch=false` shows Lint, Test Suite, claude-review, and security_audit all passing. `actionlint` and a Ruby YAML parser were not installed in the local environment.

Evrard-Nil:
Thanks both — addressed in 8f09128f.

**@PierreLeGuen (blocking, prod namespace):** You're right that blocking only exact `target=prod` was insufficient — `rollback.yml` auto-selects the newest `prod-*` tag as a rollback candidate, so a `prod-YYYYMMDD-<sha>` or `prod-rollback-*` written here (without the `production` environment, the `prod-image-mutation` concurrency group, or a GitHub release) could pollute that history and later point a rollback at an unaudited image. The guard now refuses the entire namespace via `^prod($|-)`. Verified: `prod`, `prod-20260527-deadbee`, `prod-rollback-x` all blocked; `staging`/`latest`/`dev`/`product` allowed.

**@copilot — source validation:** Added a tag-form check for `source` when it isn't a `sha256:` digest, so typos fail fast.

**@copilot — skopeo error masking:** Dropped `2>/dev/null | ... || true`; the resolve step now captures and prints skopeo stderr on failure, so auth/rate-limit/network errors are distinguishable from a genuinely missing tag.

PierreLeGuen (approved):
Reviewed the updated workflow locally. I inspected `.github/workflows/retag.yml` against the existing build/promote/rollback image workflows; the previous production namespace issue is addressed by rejecting `^prod($|-)`, and I do not see a remaining substantive blocker.

Checks run: `git diff --check origin/main...HEAD -- .github/workflows/retag.yml`, `actionlint .github/workflows/retag.yml`, `yq` parse check, and `gh pr checks 680 --watch=false` (Lint, Test Suite, and security_audit all passing). I also ran actionlint across related image workflows; it only reported pre-existing ShellCheck style warnings outside this PR.

PR #681, opened 2026-05-27
feat(models): add OpenRouter top_provider block to /v1/models

Evrard-Nil:
## Summary

OpenRouter's [`/api/v1/models`](https://openrouter.ai/api/v1/models) schema nests **context length** and **max output tokens** under a `top_provider` object:

```json
"top_provider": {
  "context_length": 262144,
  "max_completion_tokens": 8192,
  "is_moderated": false
}
```

The OR for-providers requirement explicitly calls these two out ("Should show the maximum number of output tokens you allow per model, along with context length"). We already expose them as top-level `context_length` / `max_output_length` (PR #661), but an OR ingester reading their schema looks for `top_provider.max_completion_tokens`.

This adds the `top_provider` block **additively** — the existing top-level fields stay — so OR maps our limits with no ambiguity, whether read by a human or programmatically.

- `context_length` mirrors top-level `context_length`
- `max_completion_tokens` mirrors top-level `max_output_length`
- `is_moderated` is always `false` (we don't moderate)

Serializer-only change: the data already exists in the DB (no migration). `TopProvider` registered in the OpenAPI schema list.

## Test plan

- [x] `cargo check -p api`
- [x] `cargo test -p api --lib completions` (24 passed)
- [ ] After deploy: `curl -s https://cloud-api.near.ai/v1/models | jq '.data[0].top_provider'` shows `{context_length, max_completion_tokens, is_moderated}`


PierreLeGuen (approved):
Reviewed the /v1/models response-shape change locally. I inspected the diff in crates/api/src/models.rs, crates/api/src/routes/completions.rs, crates/api/src/openapi.rs, the surrounding model conversion code, OpenAPI registration, related model-list tests, and current GitHub check status.\n\nLocal checks run:\n- cargo fmt --all -- --check\n- cargo test -p api --lib test_openapi\n- cargo test -p api --lib nano_dollars\n- cargo check -p api\n\nGitHub checks for Lint, Test Suite, and security_audit are also green. I did not find substantive issues.

Issue #229, opened 2025-12-05
Feature: NEAR Wallet login


think-in-universe:
### NEAR Wallet Login

Enable NEAR Wallet login for NEAR AI Cloud.

The reference implementation in Private Chat can be helpful:

- backend: https://github.com/nearai/chat-api/pull/68
- frontend: https://github.com/nearai/private-chat/pull/167

### Notes

1. For users logged in with NEAR Wallet, we may need to ask users to set up an email address in user profile, before proceeding payment with Stripe.

PR #682, opened 2026-05-27
[codex] add admin organization balance endpoint

PierreLeGuen:
## Summary

Adds an admin-authenticated organization balance endpoint for billing services:

- `GET /v1/admin/organizations/{org_id}/usage/balance`
- Computes `remaining` from active summed limits minus total spend using the existing balance response shape
- Wires the route into OpenAPI and admin routing
- Documents that `nearai-cloud-ui` owns Stripe/credits synchronization in `CLAUDE.md`

This supports auto top-up in `nearai-cloud-ui`, which needs an admin-readable balance check before charging.

## Validation

- `cargo test --test e2e_all test_admin_get_organization_balance --no-run`

Note: runtime e2e execution still requires the repo's local database test environment.


Evrard-Nil (approved):
Pulled the branch, read the full diff, built it, and attempted the e2e suite locally.

**Build:** `cargo check -p api` clean. `cargo test --test e2e_all test_admin_get_organization_balance --no-run` compiles. I also spun up a local Postgres and tried to *run* the 5 new tests — they fail at server init (`lib.rs:341`) because deriving ed25519 signing keys needs a dstack TEE socket (`/var/run/dstack/...`) that isn't present off-CVM. So it's a TEE-environment limitation, not a test defect; the DB bootstrap + migrations ran fine. CI (which has the dstack mock) will exercise them.

**The change is clean and well-factored.** The core move is extracting the balance computation out of `get_organization_balance` into a shared `compute_organization_balance_response(usage_service, org_id)`, then having two thin callers:
- the existing member route (`get_organization_balance`) — unchanged behavior, calls the shared fn *after* `check_org_membership`;
- the new admin route (`get_admin_organization_balance`) — same computation, gated by the `AdminUser` extension instead of membership.

I diffed the extracted logic line-by-line against the original — it's behavior-preserving. One tiny, harmless normalization: in the `balance=None, limit=Some` branch the response's `organization_id` now comes from `organization_id.to_string()` (canonical UUID) instead of the raw `org_id` path string. For the member route that means a non-canonical input (odd casing) now echoes back canonical — arguably more correct, definitely not a regression.

**Auth boundary is right and well-tested.** This is the security-critical bit — the endpoint deliberately drops the membership check so a billing service can read *any* org's balance with only admin creds. The test suite covers exactly the cases I'd want:
- happy path (session token) → 200
- admin access token (the actual billing-service path) → 200
- non-admin session → **403** (`test_admin_get_organization_balance_rejects_non_admin`)
- malformed UUID → 400
- missing org → 404

That 403 test is the one that matters most, and it's there. `_admin_user` being unused is fine — the extension's presence proves `admin_middleware` ran (same pattern as the other admin routes).

**System fit is correct.** Per the new CLAUDE.md section this PR adds, `nearai-cloud-ui` owns Stripe and already calls `PATCH /v1/admin/organizations/{org_id}/limits` with `NEAR_AI_CLOUD_ADMIN_ACCESS_TOKEN`. This adds the *read* counterpart so the auto-top-up flow (#229) can check remaining balance before charging. Reusing the existing `OrganizationBalanceResponse` shape keeps the two balance endpoints consistent. Good.

**Logging:** the refactor upgrades `.map_err(|_| …)` to `.map_err(|e| { tracing::error!("…: {:?}", e); … })`. The logged errors are usage/limit DB errors (IDs + error categories), no customer content — compliant with the CLAUDE.md privacy rules.

**One thing to flag before merge — this PR collides with #674.** Both refactor `build_admin_routes` from positional args into a struct and both add a field to `AdminAppState`:
- #674: struct `AdminRouteDependencies`, adds `organization_service`
- #682: struct `AdminRouteServices`, adds `usage_service`

Whichever lands first, the second will hit a merge conflict in `lib.rs` + `admin.rs` and needs a trivial rebase to (a) pick one struct name and (b) carry *both* new fields. Not a defect in either PR, but worth coordinating the merge order so the rebase is clean.

**Minor:** the `compute_organization_balance_response` doc comment is absent — a one-liner noting it's the shared helper for both the member and admin balance routes (and that callers own the auth check) would help the next reader understand why auth isn't done inside it.

LGTM. Approving. Recommend resolving the #674/#682 `AdminAppState` overlap at merge time, and confirming CI runs the 5 new e2e tests green (I couldn't run them locally due to the dstack TEE dependency).

PierreLeGuen (commented):
Re-reviewed against `15ce4d29`. Both items I flagged in the prior review are addressed:

- ✅ **DRY:** `compute_organization_balance_response` extracted into `crates/api/src/routes/usage.rs:98-185` and both the user-scoped and admin handlers now delegate to it. The admin handler at `crates/api/src/routes/admin.rs:923-926` is down to a 3-line shim. The user handler at `crates/api/src/routes/usage.rs:313-316` also picks up the improved `error!(\"…: {:?}\", e)` logging that was previously only in the admin variant.
- ✅ **Test coverage:** `test_admin_get_organization_balance_invalid_org_id_returns_400` and `test_admin_get_organization_balance_accepts_admin_access_token` added in `crates/api/tests/e2e_all/general.rs`. The admin-token test correctly exercises the intended call path (create `adm_*` token, hit the endpoint with matching User-Agent) — which is the actual production code path used by the billing service. The misleadingly-named 404 test was renamed to `_missing_org_returns_404`. The happy-path test also picked up `total_requests` and `total_tokens` assertions.

LGTM. The OpenAPI `session_token` annotation pattern is still a project-wide pre-existing issue, but not something to block on here.

PR #688, opened 2026-05-28
inference_provider_pool: don't retry upstream 5xx caused by client media-fetch errors

### LINKED ISSUE #687: cloud-api retries media-fetch upstream 500s — 4× amplification → backend saturation
## Symptom

cloud-api `services::inference_provider_pool` treats upstream `5xx` from the inference proxy as `retryable_http_5xx` and retries the same payload 3 times. When the upstream `5xx` is actually caused by **bad client input** (a multimodal request referencing a broken/unsupported media URL), each such request gets **4 attempts** (1 + 3 retries) of identical work, amplifying load 4× on the backends.

Observed across two incidents on `google/gemma-4-31B-it` and `Qwen/Qwen3.5-122B-A10B` over the past two days. Same retry-amplification pattern; in 2026-05-27 it combined with a SGLang gemma4 engine crash bug (separate cvm-compose-files fix v0.0.196) to produce a saturation outage on both gemma backends.

## Evidence

cloud-api logs (`message: All providers failed for model`, `error_kind: http_5xx`, `retry_decision: retryable_http_5xx`) — same `request_id` appears 4× with the identical underlying `error_detail`:

\`\`\`
2026-05-26 17:07:21 — request id e.g. 9bb71bb8 attempts 1
2026-05-26 17:07:24 — same request, retry 1 → same error_detail
2026-05-26 17:07:26 — retry 2 → same error_detail
2026-05-26 17:07:30 — retry 3 → "All providers failed for model"
\`\`\`

The underlying `error_detail` bodies are always one of these patterns from the inference engine (vLLM or SGLang) trying to fetch/decode a client-supplied media URL:

- \`HTTP error 500: 404, message='Not Found', url='https://www.facebook.com/v24.0/...'\`
- \`HTTP error 500: Internal server error: An exception occurred while loading IMAGE data at index 0: Error while loading data ... 403 Client Error: Forbidden for url: https://external.fsyd16-2.fna.fbcdn.net/...\`
- \`HTTP error 500: Internal server error: An exception occurred while loading VIDEO data at index 0: ... 429 Client Error: Too Many Requests for url: https://www.google.com/sorry/...\` (YouTube)
- \`HTTP error 500: Internal server error: ... cannot identify image file <_io.BytesIO ...>\` (e.g. base64 mp4 sent as \`image_url\`)
- \`HTTP error 500: ... SingleStreamDecoder, ... Failed to open input buffer: Invalid data found when processing input\` (torchcodec on a broken video URL)

These are **permanent client errors** — retrying the same payload cannot succeed.

## Impact

- 4× extra compute per malformed request, latency to the abusive client increased without benefit (it always 5xxes after ~5–8s).
- Under a sustained flood of bad-media requests, the amplification saturates the gemma backends (observed 2026-05-27 ~16:50–17:00 UTC: gpu11 backend queue grew to ~150, both backends timed out user traffic until the queue drained).
- The original 2026-05-26 17:07:30 \"All providers failed\" alert (and many like it) were not infra failures — they were retried bad inputs.

## Proposed fix

In the retry classifier (same module/pattern as the existing 400→\`Client error from provider, not retrying\` path, and PR #611's \`classify_provider_error\` pattern), pattern-match the upstream \`error_detail\` body on these strings and downgrade to **non-retryable client error** even when the upstream HTTP status is 500:

- \`Error while loading data\` / \`An exception occurred while loading (IMAGE|VIDEO) data\`
- \`cannot identify image file\`
- \`Failed to open input buffer\`

(Ideally also surface as 422 to the client so they know it's their input, not our infra. Behind a small allowlist of substrings to avoid masking real backend bugs.)

## Followups (out of scope here)

- Inference engines (vLLM, SGLang) should return 4xx for media-fetch/decode failures, not 500.
- Consider rate-limiting clients that produce sustained malformed-media bursts.

## References

- cvm-compose-files PR #48 (vLLM→SGLang on gemma-4-31B-it, surfaced this clearly)
- cvm-compose-files v0.0.196 (separate SGLang crash fix, `--disable-fast-image-processor`)
- This repo PR #611 (\`classify_provider_error\` pattern for upstream auth errors — same shape of fix)

Evrard-Nil:
Fixes nearai/cloud-api#687.

## Summary

vLLM and SGLang return HTTP **500** when they can't fetch or decode a multimodal media URL the client supplied (broken Facebook CDN image URLs, geo-blocked YouTube videos, base64 mp4 sent as `image_url`, etc.). These are **permanent client-input errors** — the same payload re-runs the same fetch and produces the same failure. But `classify_retry_decision` treated every 5xx the same way → `retryable_http_5xx`, so each malformed request got **4 attempts** (1 + 3 retries) of identical work.

Under sustained bad-media load this 4× amplifies until the backends saturate. Observed twice:

- **2026-05-26 17:07 UTC** — `google/gemma-4-31B-it` + `Qwen/Qwen3.5-122B-A10B`, ~40 `All providers failed for model` events in 20 min, same `request_id` retried 4× each on identical broken URLs (`https://www.facebook.com/v24.0/...` returning 404, `https://www.youtube.com/watch?v=2w_pbUrVZHY` failing torchcodec).
- **2026-05-27 ~16:50 UTC** — gemma backends saturated (gpu11 queue grew to ~150) until the queue drained. cloud-api gemma went `0/5` then recovered.

## Fix

`is_client_media_fetch_error(message)` substring-matches the upstream `error_detail` body for high-confidence client-fetch markers:
- `loading image data` / `loading video data` (sglang + vLLM)
- `cannot identify image file` (PIL `UnidentifiedImageError`)
- `Failed to open input buffer` (torchcodec)
- `, url='http…` combined with `, message='…` (aiohttp wrapper format: `HTTP error 500: 404, message='Not Found', url='https://www.facebook.com/...'`)

In `classify_retry_decision`, the 5xx branch now consults this helper and returns `non_retryable_client_media_error` when it matches. The inner retry loop already gates on `starts_with("retryable_")`, so the label change is sufficient to stop the retry — no other plumbing needed.

Genuine backend 5xx (no media markers — KV-cache evictions, NCCL hiccups, OOM, etc.) still classify as `retryable_http_5xx` and retry as before.

## Same shape as #611

This is the `classify_provider_error` pattern from PR #611 (upstream auth-error masking), applied to the retry decision instead of the response classification. The two complement each other: #611 protects callers from seeing our backend-creds bugs; this protects backends from being retry-amplified by client-input bugs.

## What this does **not** do

- Doesn't change the **response status** the client sees (still 500 from the underlying upstream message). A followup could surface these as 422 with a cleaner message — out of scope to keep the diff minimal.
- Doesn't change `error_kind` (`http_5xx` stays accurate at the engine layer; only `retry_decision` flips).
- Doesn't address the root cause upstream (engines should return 4xx for media-fetch failures) — see referenced issue.

## Test plan

- [x] `cargo test -p services --lib test_classify_retry_decision` passes (added 5 new assertions: 4 verbatim from prod logs, 1 negative case to confirm non-media 500s still retry).
- [x] `cargo check -p services` clean.
- [ ] CI green.
- [ ] On rollout, watch the cloud-api `All providers failed for model` rate for `google/gemma-4-31B-it` and `Qwen/Qwen3.5-122B-A10B` — expect a significant drop in "4-attempts-same-error" patterns. Total RPS to inference backends should fall by ~3× of the malformed-media RPS.

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -1383,14 +1383,29 @@ impl InferenceProviderPool {
     /// the same failure. Treat these as non-retryable so one broken URL
     /// from a client doesn't get amplified into 4x backend work.
     fn is_client_media_fetch_error(message: &str) -> bool {
-        let lower = message.to_lowercase();
-        lower.contains("loading image data")
+        // ASCII-only lowercase: the markers are all ASCII and this path can
+        // run at high volume during a malformed-media incident.
+        let lower = message.to_ascii_lowercase();
+        if lower.contains("loading image data")
             || lower.contains("loading video data")
             || lower.contains("cannot identify image file")
             || lower.contains("failed to open input buffer")
-            // aiohttp wrapper format: "HTTP error 500: 4xx, message='...', url='http..."
-            // — the inference engine collapsed a client-fetch 4xx into a 500.
-            || (lower.contains(", url='http") && lower.contains(", message='"))
+        {
+            return true;
+        }
+        // aiohttp wrapper format observed when the inference engine collapses
+        // a client-fetch 4xx into a 500: `HTTP error 500: 4xx, message='...',
+        // url='http...'`. Constrain to a wrapped 4xx (\\d{2} after the leading
+        // "4") so genuinely retryable wrapped 5xx (e.g. "HTTP error 500: 503,
+        // message='Service Unavailable', url='...'") still retry as before.
+        static AIOHTTP_WRAPPED_4XX: std::sync::OnceLock<Regex> = std::sync::OnceLock::new();
+        let re = AIOHTTP_WRAPPED_4XX.get_or_init(|| {
+            // (?i) is on the original message; lower is already ASCII-lowered,
+            // so an anchored lower-case literal is sufficient and faster.
+            Regex::new(r"http error 500:\s*4\d{2},\s*message='[^']*',\s*url='https?://")
+                .expect("static regex compiles")
+        });
+        re.is_match(&lower)
     }
 
     /// Single source of truth for the retry decision: the inner retry loop
@@ -3569,20 +3584,24 @@ mod tests {
 
         // Upstream 5xx caused by a broken client media URL — must NOT retry
         // (would otherwise amplify load 4x on every broken URL the client sends).
-        // SGLang gemma4 image-load failure (verbatim from prod logs):
+        // Test fixtures use dummy URLs; the matcher only depends on the marker
+        // substrings (`loading IMAGE/VIDEO data`, `cannot identify image file`,
+        // `Failed to open input buffer`, aiohttp wrapper shape).
+
+        // SGLang gemma4 image-load failure shape:
         assert_eq!(
             InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
                 status_code: 500,
-                message: "Internal server error: An exception occurred while loading IMAGE data at index 0: Error while loading data ImageData(url='https://external.fsyd15-2.fna.fbcdn.net/...'): 403 Client Error: Forbidden for url: ...".to_string(),
+                message: "Internal server error: An exception occurred while loading IMAGE data at index 0: Error while loading data ImageData(url='https://example.test/img.jpg'): 403 Client Error: Forbidden for url: ...".to_string(),
                 is_external: false,
             }),
             "non_retryable_client_media_error",
         );
-        // vLLM Qwen3.5-122B torchcodec video-load failure (verbatim from prod logs):
+        // vLLM Qwen3.5-122B torchcodec video-load failure shape:
         assert_eq!(
             InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
                 status_code: 500,
-                message: "Internal server error: An exception occurred while loading VIDEO data at index 0: Error while loading data https://www.youtube.com/watch?v=2w_pbUrVZHY: SingleStreamDecoder, Failed to open input buffer: Invalid data found when processing input".to_string(),
+                message: "Internal server error: An exception occurred while loading VIDEO data at index 0: Error while loading data https://example.test/vid: SingleStreamDecoder, Failed to open input buffer: Invalid data found when processing input".to_string(),
                 is_external: false,
             }),
             "non_retryable_client_media_error",
@@ -3596,20 +3615,41 @@ mod tests {
             }),
             "non_retryable_client_media_error",
         );
-        // aiohttp wrapper format: 500 wrapping a 4xx from a Facebook Graph URL fetch:
+        // aiohttp wrapper format: 500 wrapping a 4xx (client fetch failed):
         assert_eq!(
             InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
                 status_code: 500,
-                message: "HTTP error 500: 404, message='Not Found', url='https://www.facebook.com/v24.0/122181986942783109_1506385084189441'".to_string(),
+                message: "HTTP error 500: 404, message='Not Found', url='https://example.test/img'".to_string(),
                 is_external: false,
             }),
             "non_retryable_client_media_error",
+        );
+        // aiohttp wrapper format: 500 wrapping a 5xx (transient backend) — MUST
+        // remain retryable. The new check requires the wrapped status to be a
+        // 4xx; this guards against the regression PierreLeGuen flagged.
+        assert_eq!(
+            InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
+                status_code: 500,
+                message: "HTTP error 500: 503, message='Service Unavailable', url='https://example.test/backend'".to_string(),
+                is_external: false,
+            }),
+            "retryable_http_5xx",
         );
         // 5xx WITHOUT the media-fetch markers — still retryable (real backend hiccup).
         assert_eq!(
             InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
                 status_code: 500,
                 message: "engine: KV cache full, retract".to_string(),
+                is_external: false,
+            }),
+            "retryable_http_5xx",
+        );
+        // Generic 5xx message that happens to contain a url=... but no
+        // wrapper-shape and no media markers — still retryable.
+        assert_eq!(
+            InferenceProviderPool::classify_retry_decision(&CompletionError::HttpError {
+                status_code: 500,
+                message: "internal: failed to dial postgres url='postgres://...' message='conn refused'".to_string(),
                 is_external: false,
             }),
             "retryable_http_5xx",
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -1376,6 +1376,23 @@ impl InferenceProviderPool {
         }
     }
 
+    /// Inference engines (vLLM, SGLang) return HTTP 500 when they fail to
+    /// fetch or decode a multimodal media URL supplied by the client. The
+    /// upstream status is 5xx but the *cause* is a permanent client-input
+    /// error — retrying the same payload re-runs the same fetch and produces
+    /// the same failure. Treat these as non-retryable so one broken URL
+    /// from a client doesn't get amplified into 4x backend work.
+    fn is_client_media_fetch_error(message: &str) -> bool {
+        let lower = message.to_lowercase();
+        lower.contains("loading image data")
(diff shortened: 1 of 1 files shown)


PierreLeGuen (changes requested):
Requesting changes for one retry-classification regression:

- `crates/services/src/inference_provider_pool/mod.rs:1391`: the aiohttp-wrapper branch is broader than the comment says. It marks any 500 body containing `, message='` and `, url='http` as `non_retryable_client_media_error`, even when the wrapped aiohttp status is another 5xx such as `HTTP error 500: 503, message='Service Unavailable', url='https://...'`. That skips the existing provider retry behavior for transient upstream/backend 5xxs just because the error string includes a URL. Please constrain this branch to wrapped 4xx media-fetch statuses, or otherwise require a stronger media-fetch marker, and add a negative test for wrapped 5xx staying `retryable_http_5xx`.

Local check run: `cargo test -p services test_classify_retry_decision` passed. GitHub lint/security/claude-review checks passed when inspected; the GitHub Test Suite was still pending.

Evrard-Nil:
Thanks all — pushed 0906f30 addressing the convergent feedback in one go.

**@PierreLeGuen** — fixed the aiohttp-wrapper regression. The branch now requires the wrapped status to be a 4xx via regex (`http error 500:\s*4\d{2},\s*message='[^']*',\s*url='https?://`), so `HTTP error 500: 503, message='Service Unavailable', url='...'` stays `retryable_http_5xx`. Added the exact negative test you requested (assertion lines 3601 / 3635 in the diff).

**@gemini-code-assist** — adopted the regex approach almost verbatim; only difference is I lowercase the message once and match against that (instead of `(?i)`) since I'm already paying the lowercase cost for the substring branches.

**@Copilot** —
- Switched to `to_ascii_lowercase()` — markers are all ASCII.
- Redacted the verbatim Facebook ID + YouTube ID in test fixtures to `example.test` dummies; the matcher only depends on the marker substrings, not the host.
- Re your aiohttp tightening note — same fix as PierreLeGuen's, regex with 4xx anchor.

Test count is now 7 assertions (4 positive, 3 negative). The new "500 wrapping 503" and "generic noise with url=/message=" negatives guard against drift back to the old loose substring check.

The Test Suite hang on the first run was in step "Run unit tests" (48m, no completed_at — looks like a step-level hang). My added tests are pure substring/regex matching with no I/O, so unlikely to be the cause. Watching the rerun on this push.

@lloydmak99 @PierreLeGuen ready for another look.

PierreLeGuen (changes requested):
Blocking findings:

- crates/services/src/inference_provider_pool/mod.rs:1405,1712,1735: the aiohttp 500-wrapping-4xx case is tested against the raw message, but the production retry gate classifies `last_error` after `sanitize_completion_error` stores it. `sanitize_error_message` redacts `https://...` to `[URL_REDACTED]`, so the regex requiring `url='https?://` no longer matches and those wrapped client-fetch 4xx errors still retry as `retryable_http_5xx`. Please classify/store the retry decision before sanitizing, or make the matcher/tests cover the sanitized message used by the retry gate.

- crates/services/src/inference_provider_pool/mod.rs:1687-1697,1735: the new non-retryable client-media 5xx path is still counted as a provider health failure before the retry decision is applied. Sustained bad client media can still increment `provider_failure_counts` and demote healthy backends, even though the cause is client input. Please gate the failure counter on the same raw-error retry decision.

Checks run locally:
- `cargo test -p services --lib test_classify_retry_decision` passed.
- `cargo check -p services` passed.
- `cargo fmt --check` failed on formatting in the changed test lines; GitHub Lint is failing too.

Evrard-Nil:
Pushed 5714361 (force-pushed to drop `0e610dc` from copilot-swe-agent[bot] — that commit added only build artifacts under a `false/` directory, no source changes). Branch rebased on current main.

**@PierreLeGuen** — both blocking findings fixed in the new commit:

**1. Sanitize-vs-classify ordering (your line 1405/1712/1735 finding):** you were right — `sanitize_error_message` redacts `url='https?://...'` to `url='[URL_REDACTED]'` *before* the retry gate re-classified, so my regex never matched in production. Refactored `run_with_retries` to compute the retry decision **once on the raw `CompletionError`** and store it in a new `last_retry_decision: Option<&'static str>`. That stored decision now drives:
  - the failure-counter gate (new),
  - the in-loop retry gate (was re-classifying the sanitized form),
  - the terminal `All providers failed` log's `retry_decision` field (was also re-classifying).

The `Provider failed, will try next provider` warn log now also emits `retry_decision` so the rationale is visible per attempt.

I added a regression-pin test: `sanitize_error_message` applied to the wrapped-4xx fixture produces a string that the matcher *no longer* classifies as `non_retryable_client_media_error` — so if anyone moves classification back to the sanitized error, this assertion fails loudly with a comment explaining why.

**2. Failure-counter demotion on client-input 5xx (your 1687-1697/1735 finding):** also right. The increment now sits behind `if is_retryable_error {}` using the same raw-error `retry_decision`. Non-retryable client-media 5xx no longer bumps `provider_failure_counts` — so a flood of broken client URLs can't demote a healthy backend.

**3. fmt:** `cargo fmt --all -- --check` clean.

Local checks all green:
- `cargo test -p services --lib test_classify_retry_decision` ✓ (8 assertions now: 4 positive, 3 negative, 1 sanitize-ordering pin)
- `cargo check -p services` ✓
- `cargo fmt --check` ✓

Re-requesting review.

PierreLeGuen (changes requested):
Blocking finding:

- `crates/services/src/inference_provider_pool/mod.rs:1697`, `crates/services/src/inference_provider_pool/mod.rs:1732`, `crates/services/src/inference_provider_pool/mod.rs:1760`: `non_retryable_client_media_error` is only stored as `last_retry_decision` after the current provider fails, but the inner provider loop keeps going and the outer retry gate uses only the last provider's decision. With multiple providers, a malformed-media 500 from provider A followed by a retryable 5xx/429 from provider B will retry the whole provider list, including provider A, so the PR does not reliably bail out after the first permanent client-media failure. Even when every provider reports the media error, this still sends the same bad client payload to each backend once, unlike the existing immediate return for non-retryable 4xx. Please short-circuit this client-media classification the same way as non-retryable client errors, or otherwise make the retry decision aggregate preserve the non-retryable client-media result, and add a multi-provider test covering media-error then retryable-error ordering.

Checks run locally:
- `cargo fmt --check`
- `cargo test -p services --lib test_classify_retry_decision`
- `cargo test -p services --lib test_5xx_error_retries_with_exponential_backoff`
- `cargo test -p services --lib test_non_connection_error_does_not_retry`
- `cargo test -p services --lib test_timeout_error_does_not_retry`
- `cargo check -p services`

GitHub checks inspected with `gh pr checks 688 --watch=false`: Lint, Test Suite, and security_audit are passing.

Evrard-Nil:
Pushed b8d5c28 addressing the short-circuit blocker.

**@PierreLeGuen** — fixed in the new commit. The handler now mirrors the existing 4xx fast-return pattern: when the raw error's `retry_decision == "non_retryable_client_media_error"`, the function returns the sanitized error immediately. Provider B is never tried, the round never re-runs, and provider A doesn't get hit with the same bad payload again.

The exact test you asked for is in `test_client_media_error_short_circuits_across_providers` — two registered mock providers; the closure returns the media-500 on attempt 1 and a retryable 502 on any subsequent call. With the short-circuit, `attempt_count == 1`. Without it (before this commit), the for-loop would walk into B's 502, the outer gate would see B's retryable label, and the round would loop hitting A again — `attempt_count` would be ~8 (4 rounds × 2 providers).

A note on scope: I deliberately limited the short-circuit to the **client-media** classification, *not* to any `non_retryable_*` label. Other non-retryable categories (inference timeout, no-keyword-match, explicit timeout) reflect per-provider issues where another provider may still succeed; only the client-input-cause categories (the existing 4xx fast-return, and now this client-media 5xx) bail out immediately. Happy to broaden if you'd rather generalize.

Local checks all green:
- `cargo fmt --check` ✓
- `cargo test -p services --lib test_classify_retry_decision` ✓
- `cargo test -p services --lib test_client_media_error_short_circuits_across_providers` ✓
- `cargo check -p services` ✓

PierreLeGuen (approved):
Reviewed the retry decision changes in `crates/services/src/inference_provider_pool/mod.rs`, including raw-error classification before sanitization, the non-retryable client-media 5xx short-circuit, failure-counter gating, and the new/adjacent retry tests.

Local checks run:
- `cargo test -p services inference_provider_pool::tests --features test-mocks`
- `cargo clippy -p services --all-targets --features test-mocks -- -D warnings`

GitHub checks are also green: Lint, Test Suite, and security_audit. I did not find substantive issues to block this PR.

PR #691, opened 2026-05-29
Forward reasoning controls to Anthropic backend

### LINKED ISSUE #690: Expose reasoning controls for Anthropic/Opus models
## Problem

NEAR AI Cloud exposes Anthropic Opus models, but reasoning controls are not currently effective for them through our OpenAI-compatible API surface.

Prod behavior observed:

- `openai/gpt-5.5` rejects an invalid `reasoning_effort`, which confirms GPT/OpenAI-compatible requests forward and validate the field upstream.
- `anthropic/claude-opus-4-7` accepts both valid and invalid `reasoning_effort` values and returns normally.
- `anthropic/claude-opus-4-7` also accepts an Anthropic-style `thinking` object through `/v1/chat/completions`, but returns normally with no evidence that the field is forwarded.

This means Opus appears to run with default behavior only, even when callers ask for an explicit reasoning/thinking level.

## What is ours to fix

The OpenAI-compatible backend preserves flattened extra params, but the Anthropic backend rebuilds an Anthropic Messages request and only maps known fields. It currently does not map reasoning-related inputs from the Cloud API request into Anthropic's request shape.

Relevant code areas:

- `crates/inference_providers/src/external/anthropic/mod.rs`
- `crates/inference_providers/src/external/anthropic/converter.rs`
- `crates/api/src/models.rs`
- `crates/services/src/responses/models.rs`

## Expected behavior

For Anthropic-backed models that support thinking/reasoning, callers should have a supported way to request it, and unsupported/invalid values should not be silently ignored.

Possible acceptance criteria:

- Define the supported public input shape for Anthropic reasoning controls on Cloud API.
- Map supported values into the Anthropic Messages request `thinking` config where applicable.
- Reject or clearly ignore unsupported fields/values instead of silently accepting them.
- Add coverage for Opus/Anthropic request serialization so valid thinking config is forwarded and invalid values are handled intentionally.

## Related note

GPT + function tools on OpenAI `/chat/completions` is a separate upstream API limitation. This issue is specifically about the Anthropic/Opus adapter dropping or ignoring reasoning controls.

PierreLeGuen:
## Problem

NEAR AI Cloud exposes Anthropic Opus models through the OpenAI-compatible `/v1/chat/completions` surface, but reasoning/thinking controls were silently dropped for them.

The OpenAI-compatible backend serializes the whole request (including flattened `extra` fields) straight to the upstream provider, so `reasoning_effort` reaches the provider and is validated upstream. The Anthropic backend instead **rebuilt** an Anthropic Messages request from scratch and only copied a fixed set of known fields — everything in `params.extra` (including `reasoning_effort` and an Anthropic-style `thinking` object) was thrown away. The result: Opus appeared to accept reasoning controls but always ran with default behavior.

## Change

Make the Anthropic backend a **passthrough**, mirroring the OpenAI-compatible backend:

- `converter.rs`: add `#[serde(flatten)] pub extra: HashMap<String, serde_json::Value>` to `AnthropicRequest`.
- `mod.rs`: populate it with `extra: params.extra.clone()` in `build_request`.

The upstream provider validates the forwarded fields and returns its own error on unsupported values, so reasoning controls are no longer silently dropped.

### Resulting behavior

- **Native `thinking: {type, budget_tokens}`** → forwarded verbatim; the provider applies extended thinking. The default `temperature=1.0` is already thinking-compatible, and `build_request` drops `top_p` when temperature is set, so the common case works as-is.
- **`reasoning_effort`** → forwarded; the provider rejects it with a 400, which `map_provider_error` surfaces as `InvalidParams` → HTTP 400 with the provider's message. No more silent 200s.
- **Typed OpenAI-only sampling params** (`frequency_penalty`, etc.) live in named struct fields, never in `extra`, so they are still not forwarded — no regression where we'd send the provider a field it rejects.

## Tests

Added unit tests in `anthropic/mod.rs` covering serialization of `build_request`:
- native `thinking` object is forwarded
- `reasoning_effort` is forwarded
- typed OpenAI-only params do not leak into the request body
- empty `extra` adds no stray fields

`cargo test -p inference_providers anthropic` → 25 passed. `cargo clippy` and `cargo fmt --check` clean.

## Out of scope

- Surfacing thinking *output* as `reasoning_content`/`reasoning` (streaming-parser change).
- Forwarding the typed `reasoning` config on the `/v1/responses` API, which currently does not forward it to the provider at all.

Fixes #690

### DIFF
--- a/crates/inference_providers/src/external/anthropic/converter.rs
+++ b/crates/inference_providers/src/external/anthropic/converter.rs
@@ -88,6 +88,13 @@ pub struct AnthropicRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tool_choice: Option<AnthropicToolChoice>,
     pub stream: bool,
+    /// Passthrough for caller-supplied fields we don't model explicitly
+    /// (e.g. Anthropic's `thinking` reasoning config, or any other extra
+    /// param). Mirrors the OpenAI-compatible backend, which forwards `extra`
+    /// to the upstream provider. Anthropic validates these and returns its own
+    /// error on unsupported values, so we don't silently drop reasoning controls.
+    #[serde(flatten)]
+    pub extra: HashMap<String, serde_json::Value>,
 }
 
 // =============================================================================
--- a/crates/inference_providers/src/external/anthropic/converter.rs
+++ b/crates/inference_providers/src/external/anthropic/converter.rs
@@ -88,11 +88,12 @@ pub struct AnthropicRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tool_choice: Option<AnthropicToolChoice>,
     pub stream: bool,
-    /// Passthrough for caller-supplied fields we don't model explicitly
-    /// (e.g. Anthropic's `thinking` reasoning config, or any other extra
-    /// param). Mirrors the OpenAI-compatible backend, which forwards `extra`
-    /// to the upstream provider. Anthropic validates these and returns its own
-    /// error on unsupported values, so we don't silently drop reasoning controls.
+    /// Allowlisted reasoning-control fields forwarded from the caller's request
+    /// (`thinking`, `reasoning_effort` — see `ANTHROPIC_PASSTHROUGH_KEYS`).
+    /// Flattened to top-level JSON. Populated by `build_request`, which filters
+    /// the request's `extra` map so internal E2EE keys and OpenAI-only fields
+    /// never reach Anthropic. The allowlist guarantees no collision with the
+    /// named fields above, so `flatten` cannot emit duplicate keys.
     #[serde(flatten)]
     pub extra: HashMap<String, serde_json::Value>,
 }
--- a/crates/inference_providers/src/external/anthropic/mod.rs
+++ b/crates/inference_providers/src/external/anthropic/mod.rs
@@ -23,6 +23,25 @@ use reqwest::{header::HeaderValue, Client};
 
 const DEFAULT_ANTHROPIC_VERSION: &str = "2023-06-01";
 
+/// Caller-supplied `extra` keys we forward to Anthropic's Messages API.
+///
+/// This is an allowlist on purpose: `ChatCompletionParams.extra` is an
+/// unbounded catch-all that also holds internal E2EE keys and OpenAI-only
+/// fields, so we only pass through the reasoning controls Anthropic actually
+/// understands (`thinking`) plus `reasoning_effort` (which Anthropic does not
+/// accept and will reject with its own 400, instead of us silently dropping it).
+const ANTHROPIC_PASSTHROUGH_KEYS: &[&str] = &["thinking", "reasoning_effort"];
+
+/// Pick the allowlisted reasoning-control fields out of `extra`.
+fn extract_passthrough(
+    extra: &std::collections::HashMap<String, serde_json::Value>,
+) -> std::collections::HashMap<String, serde_json::Value> {
+    ANTHROPIC_PASSTHROUGH_KEYS
+        .iter()
+        .filter_map(|&key| extra.get(key).map(|value| (key.to_string(), value.clone())))
+        .collect()
+}
+
 /// Anthropic backend - handles HTTP communication with Anthropic's API
 pub struct AnthropicBackend {
     client: Client,
@@ -94,10 +113,14 @@ impl AnthropicBackend {
             tools,
             tool_choice,
             stream,
-            // Passthrough of caller-supplied extra fields (e.g. `thinking`,
-            // `reasoning_effort`). Internal/tracing keys are already stripped
-            // upstream in ExternalProvider before the backend runs.
-            extra: params.extra.clone(),
+            // Forward only the reasoning-control fields from `extra`, not the
+            // whole map. A full passthrough is unsafe here: `extra` also carries
+            // internal E2EE keys (`x_signing_algo`, `x_client_pub_key`, …) that
+            // must never reach Anthropic, OpenAI-only fields that Anthropic
+            // rejects (`max_completion_tokens`, `presence_penalty`,
+            // `frequency_penalty`, …), and could collide with named fields
+            // (`system`, `stop_sequences`) producing duplicate JSON keys.
+            extra: extract_passthrough(&params.extra),
         }
     }
 }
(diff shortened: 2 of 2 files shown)


PierreLeGuen:
Thanks for the reviews — all valid. Addressed in 9f452d76 by switching from a full `extra` passthrough to an **allowlist** (`thinking`, `reasoning_effort`), which fixes every point at once:

- **Internal E2EE keys** (`x_signing_algo`, `x_client_pub_key`, `x_encryption_version`, `x_encrypt_all_fields`) — inserted into `extra` by `insert_encryption_headers` and only `x_model_pub_key` is stripped downstream. The allowlist keeps all of them out of the Anthropic body.
- **OpenAI-only fields** — `max_completion_tokens` (chat) and `presence_penalty`/`frequency_penalty` (from the `/v1/completions`→chat translation) live in `extra`; no longer forwarded, so no new 400s for those.
- **Duplicate JSON keys** — `system`/`stop_sequences` are not in the allowlist, so they can never collide with the named fields. Added a test asserting non-allowlisted keys (incl. these and the E2EE keys) are dropped and that `stop_sequences` comes from `params.stop`.

An allowlist is preferred over a denylist since `extra` is an unbounded, growing surface. `reasoning_effort` is intentionally kept on the list so Anthropic returns its own 400 rather than us silently dropping it.

lloydmak99 (commented):
## Code Review

Four findings from a multi-angle review of the diff, ranked by severity.

---

### 🔴 Critical — E2EE keys leak to Anthropic

**File:** `crates/inference_providers/src/external/anthropic/mod.rs` (line ~97, `extra: params.extra.clone()`)

The PR comment says *"Internal/tracing keys are already stripped upstream in ExternalProvider before the backend runs"* — this is only half-true. `strip_internal_tracing_keys()` removes exactly three tracing keys (`x_request_id`, `x_org_id`, `x_workspace_id`). The five E2EE keys inserted by `insert_encryption_headers()` in `completions.rs` are **never stripped for external/Anthropic backends**:

- `x_client_pub_key`
- `x_signing_algo`
- `x_model_pub_key`
- `x_encryption_version`
- `x_encrypt_all_fields`

The vLLM provider removes them inside `prepare_encryption_headers()`, but that code path is not reached here. After this PR, a request that hits an Anthropic model via E2EE routing will forward the client's public key and cipher metadata as top-level JSON fields to Anthropic's servers. There is even an existing vLLM test (~line 2409) that explicitly guards against this exact leak — the same fix is needed here.

**Fix:** extend `strip_internal_tracing_keys()` to also remove the five E2EE keys (or extract the vLLM stripping logic into a shared `strip_e2ee_keys()` helper called for all external backends).

---

### 🟡 Medium — Blanket clone forwards OpenAI-only fields, breaking existing callers

**File:** `crates/inference_providers/src/external/anthropic/mod.rs` (line ~97)

`params.extra.clone()` is completely unfiltered. Any OpenAI-only field a caller was previously passing silently (e.g., `logprobs`, `top_logprobs`, `seed`, `logit_bias`) will now be serialized as a top-level JSON field and forwarded to Anthropic, which rejects unknown fields with a 400. This is a silent breaking change for any caller that routes to Anthropic models while passing OpenAI-style extra params.

The PR acknowledges `reasoning_effort` → 400 as intentional ("No more silent 200s"), but the same logic applies to every unrecognized field. An allowlist (e.g., only forwarding `thinking` and `reasoning_effort`) would avoid this while still achieving the stated goal.

---

### 🟡 Medium — `thinking` + omitted/wrong `temperature` produces a 400

**File:** `crates/inference_providers/src/external/anthropic/mod.rs` (lines 80–84)

Anthropic requires `temperature=1.0` for extended thinking. Two cases currently produce a 400 silently:

1. **`params.temperature = None`** — `build_request` omits `temperature` entirely (it's `Option<f32>` with `skip_serializing_if`). Anthropic's server-side default is not guaranteed to be 1.0.
2. **`params.temperature = Some(0.7)`** — clamped to 0.7 (valid Anthropic range), but Anthropic rejects the `thinking` field when temperature ≠ 1.0.

The PR description says *"The default temperature=1.0 is already thinking-compatible"* but the default is *omission*, not 1.0. A simple guard would help:

```rust
// If caller supplies a thinking config, coerce temperature to 1.0
let temperature = if extra.contains_key("thinking") {
    Some(1.0)
} else {
    params.temperature.map(|t| t.clamp(0.0, 1.0))
};
```

---

### 🟠 Low — `#[serde(flatten)]` on HashMap can emit duplicate JSON keys

**File:** `crates/inference_providers/src/external/anthropic/converter.rs` (line 91)

When `extra` contains a key that matches a named struct field (e.g., `"stream"`, `"model"`, `"max_tokens"`), serde serializes the named field first and then emits the flatten map's entry — producing duplicate JSON keys. Anthropic's parser uses last-wins semantics, meaning a caller could pass `{"stream": false}` in extra on a streaming call and get a full JSON body back instead of SSE chunks, or pass `{"model": "claude-3-haiku"}` to route to a cheaper model than the one selected by the backend. For the `stream` field specifically this causes an immediate client-side parse error; for `model` it's a silent authorization bypass.

This risk is reduced if finding #2 is addressed with an allowlist, but worth noting independently.

lloydmak99 (approved):
## Updated Review — Following up on 9f452d76

All three of the critical/medium findings are resolved by the allowlist approach. Walking through each:

**🔴 E2EE key leak → ✅ Fixed.**
`extract_passthrough()` only returns keys present in `ANTHROPIC_PASSTHROUGH_KEYS` (`thinking`, `reasoning_effort`). All five E2EE keys (`x_signing_algo`, `x_client_pub_key`, `x_model_pub_key`, `x_encryption_version`, `x_encrypt_all_fields`) are dropped regardless of what `strip_internal_tracing_keys()` does. `test_build_request_drops_non_allowlisted_extra_keys` explicitly asserts four of the five are absent — good belt-and-suspenders.

**🟡 OpenAI-only fields causing 400 → ✅ Fixed.**
`max_completion_tokens`, `presence_penalty`, `frequency_penalty`, `response_format` (and any other OpenAI-only field) are silently dropped because they're not on the allowlist, matching the pre-PR behavior. The test covers the extra-map variants of these too, not just the typed-field variants.

**🟠 Duplicate JSON key collision → ✅ Fixed.**
Neither `thinking` nor `reasoning_effort` is a named field on `AnthropicRequest`, so `#[serde(flatten)]` on the filtered map can never produce duplicate keys. The comment in `converter.rs` documents this invariant explicitly.

**🟡 `thinking` + wrong `temperature` → still unaddressed (intentional).**
If a caller passes `thinking` with an explicit non-1.0 temperature (or omits temperature entirely), Anthropic returns a 400. The PR is consistent about this design choice ("Anthropic validates and returns its own error"), and the failure mode is a clear error rather than silent misbehavior, so I won't block on it. Worth a follow-up to either auto-coerce temperature to 1.0 or document it explicitly in the API surface.

### Other observations

- `extract_passthrough` iterates a 2-element slice per request — negligible cost, fine.
- `test_build_request_does_not_leak_openai_only_params` tests the typed-field path while `test_build_request_drops_non_allowlisted_extra_keys` covers the extra-map path; both are needed and correct.
- `stop_sequences` assertion in the new test (`params.stop` wins over the `extra` collision key) is a nice proof that named-field derivation is not affected by extra content.

**Approved.** The allowlist is the right fix — it directly resolves the security and correctness issues without requiring any changes to `strip_internal_tracing_keys()` or the broader E2EE stripping pipeline.

lloydmak99 (approved):

PR #693, opened 2026-05-29
feat(admin): fire GitHub dispatch on model PATCH (infra#153)

lloydmak99:
## Summary

After a successful `PATCH /v1/admin/models`, cloud-api now spawns a fire-and-forget GitHub `repository_dispatch` event per upserted model id. Downstream automation in `nearai/cvm-ansible-playbooks` listens for the configured `event_type` and kicks off the validate-model-staging workflow.

Part of [infra#153](https://github.com/nearai/infra/issues/153) — the trigger mechanism for the model onboarding chain.

## How it works

- New `services::github_dispatch` module mirrors the `email` service pattern: trait + noop impl + reqwest impl + factory + mockable transport.
- `AdminAppState` gets an `Arc<dyn GitHubDispatcher>` field. Built once at startup from `config.github_dispatch`.
- `batch_upsert_models` handler loops over upserted model ids after registration completes and `tokio::spawn`s the dispatch. Errors are logged at WARN; the PATCH response is unaffected.

## Configuration

Off by default. Enable on **staging cloud-api only** (production keeps it off so promote-driven prod PATCHes don't recursively re-fire the chain):

```
ENABLE_GITHUB_DISPATCH=true
GITHUB_DISPATCH_REPO=nearai/cvm-ansible-playbooks
GITHUB_DISPATCH_EVENT_TYPE=stg_model_loaded
GITHUB_DISPATCH_PAT_FILE=/run/secrets/github_dispatch_pat
```

PAT scope: fine-grained, `actions: write` on cvm-ansible-playbooks only. Rotate every 90 days. Eventual goal: replace with a GitHub App.

When enabled, missing `GITHUB_DISPATCH_REPO` or PAT will fail config validation at startup (not silently degrade). When disabled, the dispatcher is a noop and consumes no resources.

## Test plan

- [x] `cargo test --lib --bins` — 321 passed, 0 failed
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [ ] After merge + deploy to cloud-stg-api: PATCH a test model, verify `repository_dispatch` event arrives at cvm-ansible-playbooks (will land in a no-op workflow until the validate workflow is built in the follow-up PR)
- [ ] Verify failure path: revoke PAT, PATCH a model, confirm PATCH still succeeds with a WARN log entry

## Out of scope (follow-up work)

- `validate-model-staging.yml` workflow in cvm-ansible-playbooks (consumer of this dispatch)
- `promote-model-prod.yml`, `rollback-model-prod.yml`
- New generic tests in `infra-tests` (tool_call_generic, e2ee, signature_retrieval, load_smoke)

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1582,6 +1582,9 @@ pub fn build_admin_routes(
             as Arc<dyn services::completions::CompletionServiceTrait>,
     )) as Arc<dyn services::admin::AdminService + Send + Sync>;
 
+    let github_dispatcher =
+        services::github_dispatch::dispatcher_from_config(&config.github_dispatch);
+
     let admin_app_state = AdminAppState {
         admin_service,
         analytics_service: services.analytics_service,
@@ -1590,6 +1593,7 @@ pub fn build_admin_routes(
         config,
         admin_access_token_repository,
         inference_provider_pool: services.inference_provider_pool,
+        github_dispatcher,
     };
 
     Router::new()
@@ -1828,6 +1832,7 @@ mod tests {
             },
             cors: config::CorsConfig::default(),
             external_providers: config::ExternalProvidersConfig::default(),
+            github_dispatch: config::GitHubDispatchConfig::default(),
         };
 
         // Initialize services
@@ -1928,6 +1933,7 @@ mod tests {
             },
             cors: config::CorsConfig::default(),
             external_providers: config::ExternalProvidersConfig::default(),
+            github_dispatch: config::GitHubDispatchConfig::default(),
         };
 
         let auth_components = init_auth_services(database.clone(), &config);
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -344,22 +344,34 @@ pub async fn batch_upsert_models(
         }
     }
 
-    // Fire GitHub repository_dispatch for each upserted model. Downstream
-    // automation (validate / promote workflows in cvm-ansible-playbooks)
-    // listens for the configured event_type and reacts. Fire-and-forget:
-    // a GitHub outage does not block the PATCH. Only enabled on staging
-    // cloud-api via ENABLE_GITHUB_DISPATCH; production keeps it off so
-    // promote-driven prod PATCHes do not recursively re-fire the chain.
-    for model_id in batch_request.keys() {
+    // Fire GitHub repository_dispatch for each model this PATCH (re)loaded.
+    // Downstream automation (validate / promote workflows in
+    // cvm-ansible-playbooks) listens for the configured event_type and reacts.
+    // Fire-and-forget: a GitHub outage does not block the PATCH. Only enabled
+    // on staging cloud-api via ENABLE_GITHUB_DISPATCH; production keeps it off
+    // so promote-driven prod PATCHes do not recursively re-fire the chain.
+    //
+    // Skip deactivations: a PATCH that sets is_active=false is an unload, not a
+    // load, and must not trigger validate/promote of a model being taken
+    // offline. Dispatch the whole set from a single background task so a large
+    // batch does not burst N concurrent requests at GitHub's dispatch API
+    // (which would trip secondary rate limits and silently drop events).
+    let dispatch_model_ids: Vec<String> = batch_request
+        .iter()
+        .filter(|(_, request)| request.is_active != Some(false))
+        .map(|(model_id, _)| model_id.clone())
+        .collect();
+    if !dispatch_model_ids.is_empty() {
         let dispatcher = app_state.github_dispatcher.clone();
-        let model_id = model_id.clone();
         tokio::spawn(async move {
-            if let Err(e) = dispatcher.dispatch_model_loaded(&model_id).await {
-                tracing::warn!(
-                    error = %e,
-                    model_id = %model_id,
-                    "GitHub dispatch failed; manual workflow trigger may be required"
-                );
+            for model_id in dispatch_model_ids {
+                if let Err(e) = dispatcher.dispatch_model_loaded(&model_id).await {
+                    tracing::warn!(
+                        error = %e,
+                        model_id = %model_id,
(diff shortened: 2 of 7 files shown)


lloydmak99:
## Review comments addressed

Pushed `cd4b6b6` on top of the earlier hardening commit `783ede9`. Status of each finding:

**Already resolved in `783ede9` (no further change):**
- **Gemini — `tokio::time::timeout` only wraps `.send()`, leaving `.text()` unprotected:** the whole exchange (send + error-body read) is now wrapped in `DISPATCH_TIMEOUT` (`github_dispatch/mod.rs:97-131`).
- **Gemini — `truncate` uses O(N) `chars().count()`:** the bespoke `truncate` was removed; error bodies now go through `email::sanitize_error`, matching the mirrored email module.
- **Claude — dispatch fires for deactivations:** PATCHes with `is_active == Some(false)` are now filtered out before dispatch (`admin.rs:359-363`).
- **Claude — unbounded N `tokio::spawn`:** the batch is now dispatched from a single background task that loops sequentially, so a large batch can't burst N concurrent requests at GitHub.

**Fixed in `cd4b6b6`:**
- **Gemini — span not propagated into spawned task:** the fire-and-forget task is now wrapped with `.instrument(tracing::Span::current())` (matching the existing `completions.rs` pattern). Note: admin handlers don't yet establish a correlation span, so this is primarily future-proofing + the correct idiom rather than an immediate behavior change.
- **Claude — silent noop downgrade in `dispatcher_from_config`:** added a `tracing::error!` on the `enabled=true` + missing repo/PAT path so a misconfiguration that slips past config validation surfaces in logs instead of silently disabling the chain.

**Acknowledged, not changed (reviewer marked optional / out of scope):**
- **No test for the `HttpGitHubDispatcher` branch of the factory:** would require refactoring the factory to inject a transport (can't downcast `Arc<dyn ...>`); deferred.
- **PAT held as `String` (no `Zeroize`):** not a regression — same pattern as other secrets here; out of scope for this PR.

`cargo fmt`, `cargo clippy -p services -p api --all-targets -- -D warnings`, and the `github_dispatch` unit tests are clean.

Evrard-Nil (approved):
Pulled and traced the whole module + wiring + tests.

This is well-structured. The trait-mocked transport pattern mirrors the email service (`NoopGitHubDispatcher` / `HttpGitHubDispatcher` / `ReqwestDispatchTransport`), the config validation is fail-loud at startup when enabled and noop-defensive when disabled, and the integration into `batch_upsert_models` has the right small details:

- **Filter `is_active == Some(false)` before dispatch** — a PATCH that takes a model offline should not fire validate/promote. ✓
- **Single background task that loops over the batch**, not N parallel spawns — explicit comment about GitHub's secondary rate limits. Good instinct.
- **`Span::current() → .instrument(...)`** on the fire-and-forget task so dispatch warnings still carry the request log context. Easy to forget, you caught it.
- **`tokio::time::timeout(10s)` wraps both `send().await` AND `text().await`** — the comment explicitly notes "reqwest has no client-level timeout configured, so reading the body of a stalled error response would otherwise hang the spawned task." This is exactly the lesson from compose-manager#20's HTTP-timeout finding, applied proactively. ✓

**Config / safety properties verified:**

- Enabled-only path reads the PAT; disabled instances boot cleanly even if `GITHUB_DISPATCH_PAT_FILE` points at an unmounted secret (test `github_dispatch_disabled_ignores_unreadable_pat_file` covers this). Sensible — a shared env template shouldn't break disabled environments.
- Enabled + missing repo/PAT/malformed repo → startup error with a specific message. Three tests cover the failure modes.
- `Default` impl matches `from_env` event_type fallback (`stg_model_loaded`) — small but important so a `Default`-constructed test config doesn't carry empty event_type.
- `event_type` is fixed per dispatcher instance — currently only the "model loaded" use case. Fine for now; if future automation wants "model unloaded" / "model errored", the trait signature will need to grow.

**Wiring is complete.** Both `build_admin_routes` (production path) and the two test helpers in `lib.rs` (`build_test_app_*`) plus `tests/common/mod.rs::test_config()` all populate `github_dispatch: GitHubDispatchConfig::default()`. No path constructs an `ApiConfig` that would fail the new field. ✓

1. **PAT held as `String` in memory** — `GitHubDispatchConfig.pat: Option<String>` and `HttpGitHubDispatcher.pat: String`. Inside the TEE this is contained, but a memory dump (e.g. crash diagnostics) could leak it. `secrecy::Secret<String>` or `Zeroizing<String>` would zeroize on drop. Not a regression vs how other secrets in this codebase are stored (api keys, oauth tokens) but worth a sweep in a future hardening pass.

2. **Serial batch dispatch** inside the spawned task — N model ids → N sequential 10s-timeout-bound POSTs. For typical onboarding (1–2 models per PATCH) this is fine; if a PATCH ever batches dozens, the loop could run for minutes in the background. The fire-and-forget pattern makes that benign (PATCH already returned), but worth a comment if the typical batch size could grow.

3. **No retry on transient failures** — single attempt, log warn, done. Doc-comment already calls this out: operator falls back to manual `gh workflow run` trigger. Right tradeoff for a non-critical signal: don't queue and risk recursive re-fires after an outage.

4. **Production safety knob is the env flag, not code** — `ENABLE_GITHUB_DISPATCH=true` on staging, off in prod. That's the right place for the kill switch; just calling out that an operator typo on prod (`true` instead of unset) would start firing dispatches into the prod-vs-staging chain. Worth a short note in the deploy/runbook so the order-of-operations is clear when first enabling.

5. **`bearer_auth(pat)`** uses the PAT inline in the Authorization header — same concern as (1), but standard reqwest pattern.

**System fit.** Per the linked `infra#153`, this is the trigger half of the model-onboarding chain: admin PATCH on staging fires `stg_model_loaded` → cvm-ansible-playbooks workflow listens (the `validate-model-staging` workflow is the companion PR `#252`) → after green validate, promote to prod. Off-by-default on prod prevents the recursive re-fire when prod-PATCH happens. Coherent design.

LGTM. Approving.

PR #694, opened 2026-05-30
[codex] Expose inactive admin users

PierreLeGuen:
## Summary

- Include inactive users by default in `/v1/admin/users`, including the `include_organizations=true` path.
- Add admin-only OAuth identity fields (`auth_provider`, `provider_user_id`) to admin user responses.
- Add server-side `search` and `is_active` filters with totals based on the exact filtered result set.
- Add e2e coverage for inactive user lookup and provider identity search.

## Why

Admins could not reliably diagnose login failures caused by hidden inactive user rows or OAuth identity collisions without direct production DB access. This makes those rows discoverable from admin tooling.

## Validation

- `cargo fmt`
- `cargo check -p api`
- `git diff --check`
- `cargo test -p api --test e2e_all test_admin_list_users_finds_inactive_and_oauth_identity_fields -- --nocapture` compiled, then failed during local test DB bootstrap because Postgres was not accepting connections (`Connection refused`).

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -1174,7 +1174,9 @@ pub async fn deprecate_model(
         ("limit" = Option<i64>, Query, description = "Maximum number of users to return (default: 100)"),
         ("offset" = Option<i64>, Query, description = "Number of users to skip (default: 0)"),
         ("include_organizations" = Option<bool>, Query, description = "Whether to include organization information and spend limits for the first organization owned by each user (default: false)"),
-        ("search_by_name" = Option<String>, Query, description = "Filter users by organization name (case-insensitive match). Only effective when include_organizations=true; ignored otherwise.")
+        ("search" = Option<String>, Query, description = "Filter users by email, username, display name, user id, auth provider, or provider user id (case-insensitive partial match)."),
+        ("is_active" = Option<bool>, Query, description = "Filter users by active status. Omit to include active and inactive users."),
+        ("search_by_name" = Option<String>, Query, description = "Filter users by organization name (case-insensitive match). Only effective when include_organizations=true; separate from user search.")
     ),
     responses(
         (status = 200, description = "Users retrieved successfully", body = ListUsersResponse),
@@ -1193,15 +1195,26 @@ pub async fn list_users(
     crate::routes::common::validate_limit_offset(params.limit, params.offset)?;
 
     debug!(
-        "List users request with limit={}, offset={}, include_organizations={}, search_by_name={:?}",
-        params.limit, params.offset, params.include_organizations, params.search_by_name
+        "List users request with limit={}, offset={}, include_organizations={}, search={:?}, is_active={:?}, search_by_name={:?}",
+        params.limit,
+        params.offset,
+        params.include_organizations,
+        params.search,
+        params.is_active,
+        params.search_by_name
     );
 
     let (user_responses, total) = if params.include_organizations {
         // Fetch users with their default organization and spend limit
         let (users_with_orgs, total) = app_state
             .admin_service
-            .list_users_with_organizations(params.limit, params.offset, params.search_by_name)
+            .list_users_with_organizations(
+                params.limit,
+                params.offset,
+                params.search.clone(),
+                params.is_active,
+                params.search_by_name.clone(),
+            )
             .await
             .map_err(|e| {
                 error!("Failed to list users with organizations");
@@ -1253,6 +1266,8 @@ pub async fn list_users(
                     created_at: u.created_at,
                     last_login_at: u.last_login_at,
                     is_active: u.is_active,
+                    auth_provider: u.auth_provider,
+                    provider_user_id: u.provider_user_id,
                     organizations,
                 }
             })
@@ -1263,7 +1278,12 @@ pub async fn list_users(
         // Return users data only
         let (users, total) = app_state
             .admin_service
-            .list_users(params.limit, params.offset)
+            .list_users(
+                params.limit,
+                params.offset,
+                params.search.clone(),
+                params.is_active,
+            )
             .await
             .map_err(|e| {
                 error!("Failed to list users");
@@ -1293,6 +1313,8 @@ pub async fn list_users(
                 created_at: u.created_at,
                 last_login_at: u.last_login_at,
                 is_active: u.is_active,
+                auth_provider: u.auth_provider,
+                provider_user_id: u.provider_user_id,
                 organizations: None,
             })
             .collect();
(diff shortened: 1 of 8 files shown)


crates/database/src/repositories/admin_composite.rs
```diff
@@ -604,10 +621,6 @@ impl AdminRepository for AdminCompositeRepository {
         Ok((result, total_count))
     }
 
-    async fn get_active_user_count(&self) -> Result<i64> {
```
PierreLeGuen, line 607: `get_active_user_count` is removed from the `AdminRepository` trait here, but the underlying inherent method `UserRepository::get_active_user_count` (crates/database/src/repositories/user.rs:206) now has no callers anywhere in the workspace. Worth deleting it in this PR so it doesn't linger as dead code.

PierreLeGuen (commented):
Reviewed the admin user-visibility changes. Logic is sound — inactive users are now included by default, the multi-field search is correctly parameterized and LIKE-escaped (nice dedup into `escape_like_query`), and the org-path count/list use `COUNT(DISTINCT u.id)` / `DISTINCT ON (u.id)` so the new joins don't inflate totals. A few inline notes below; none are blocking.

crates/api/src/routes/admin.rs
```diff
@@ -1193,15 +1195,26 @@ pub async fn list_users(
     crate::routes::common::validate_limit_offset(params.limit, params.offset)?;
 
     debug!(
-        "List users request with limit={}, offset={}, include_organizations={}, search_by_name={:?}",
-        params.limit, params.offset, params.include_organizations, params.search_by_name
+        "List users request with limit={}, offset={}, include_organizations={}, search={:?}, is_active={:?}, search_by_name={:?}",
+        params.limit,
+        params.offset,
+        params.include_organizations,
+        params.search,
```
PierreLeGuen: Minor privacy note: `search` is logged here verbatim, and an admin searching by email / display name puts PII into the log line. It's `debug!` (not emitted at prod info level), so this is low-risk and consistent with the pre-existing `search_by_name` logging — just flagging given the repo's strict logging rules. Consider logging only `search.is_some()` if you want to be conservative.

crates/database/src/repositories/user.rs
```diff
@@ -244,24 +251,102 @@ impl UserRepository {
...
+                .pool
+                .get()
+                .await
+                .context("Failed to get database connection")
+                .map_err(RepositoryError::PoolError)?;
+
+            let count_row = client
+                .query_one(
+                    r#"
+            SELECT COUNT(*) as total_count
```
PierreLeGuen, line 252: Non-blocking perf note: when `search` is set this runs a leading-wildcard `ILIKE ('%' || $ || '%')` across 6 columns, and it's executed twice per request (count + page), neither of which can use a btree index → sequential scan of `users`. Fine at current admin-tooling scale; if the users table grows large this is the first thing that'll get slow. (When `search` is `NULL` the `$::TEXT IS NULL` guard short-circuits, so the no-search path is unaffected.)

Evrard-Nil (approved):
Pulled, read every diff, `cargo check -p api -p services -p database` clean.

This is a focused improvement to the admin user surface. Three things going on, all coherent:

1. **Inactive users now reachable by default.** `list_users` and `list_users_with_organizations` no longer hide rows where `is_active = false`. New `is_active` query param gates either way. ✓
2. **`auth_provider` and `provider_user_id` exposed on admin responses + searchable.** Right surface for diagnosing OAuth identity collisions; lives behind the existing admin auth gate. ✓
3. **`search` param does a single ILIKE-ESCAPE across email/username/display_name/id::TEXT/auth_provider/provider_user_id.** `search_by_name` remains the org-name filter and the docstring explicitly says they're separate. ✓

**Subtle behavior changes worth being aware of:**

- **`total` now matches the *filtered* result set.** The old code returned `get_active_user_count()` regardless of what was actually listed (a real surprise if you ever paged through "include inactive" and the total disagreed with the rows). The new query carries `COUNT(*) OVER ()` or equivalent through the filter and reports the real total. The PR removed `get_active_user_count` from the repo + service trait + composite; nothing else called it. ✓ This is a real change for any UI that was treating `total` as "active users in the system" — it's now "users matching the active filter for this listing." Worth a one-liner in the API changelog so frontend consumers don't read it as a bug.

- **Org-name filter no longer matches users with no org.** Before: `OR o.id IS NULL` made `search_by_name=X` match any user with `o.name LIKE X` *plus* every user with no org at all. After: only users whose org actually matches. That removal makes the filter behave like a filter rather than "filter OR no-org" — almost certainly the intended fix, but it is a behavior change. The new e2e test `org_name_search.total == 0` (when no org matches) confirms the new semantics.

- **Query restructure to a CTE.** `first_org_per_user` CTE wraps the `DISTINCT ON (u.id)` + join, then the WHERE filters apply. Functionally equivalent for the existing filters; the CTE makes it easier to add filters on user-level fields without confusing the planner about which row the WHERE bound to. Postgres plan should be the same (both forms collapse to a hash join + sort + distinct).

**Privacy / ACL sanity:**

- `provider_user_id` is now in admin responses + indexed by search. This is PII (Google sub, GitHub user id, etc.) and CLAUDE.md treats the admin endpoints as the management plane — explicitly not the data plane — so admin visibility is the intended trust boundary. No logging changes that would leak it outside the response body. ✓
- Logged search params use `has_search` / `has_search_by_name` booleans, not the raw search string. Good — search terms can contain PII (emails) and the existing privacy rules say don't log PII at info level. ✓

**Tests.** `test_admin_list_users_finds_inactive_and_oauth_identity_fields` covers:
- inserts an inactive user with explicit `auth_provider=google` + `provider_user_id`,
- searches by email → 1 result, inactive,
- searches by email with `is_active=true` → 0,
- searches by `provider_user_id` with `is_active=false` → 1,
- same with `include_organizations=true` → 1,
- org-name search for an unrelated org → 0.

That's exactly the matrix the PR changes. ✓

**Non-blocking:**

1. **`search` and `search_by_name` are separate** — fine, but worth documenting in the openapi description that an empty-string `search` is treated as "no filter" rather than "find rows with no email." The handler builds `has_search` as `!search.is_empty()` for the log; verify the SQL side also bails on empty (`$2::TEXT IS NULL OR ...` plus `.filter(|s| !s.is_empty())` upstream). If not, an accidental `?search=` would return zero rows.
2. **Search field set could grow without a knob** — eventually `org_id`, `last_login_at` ranges, etc. The query is already 6 ILIKEs `OR`d together; a 7th won't hurt. Just calling out that the LIKE pattern fanout is bounded by `escape_like_query`. ✓
3. **Index coverage.** The `provider_user_id` ILIKE is unindexable as a substring. For thousands of users, the search will do a seq-scan on each request. Probably fine at current scale; a trigram index on `(email, username, display_name, auth_provider, provider_user_id)` would future-proof it. Out of scope for this PR.

LGTM. Approving.

PR #699, opened 2026-06-01
Retry chat-signature fetch on 404 to absorb backend signing race

lloydmak99:
## Problem

`store_chat_signature_from_provider` runs in the completion **hot path** (before `[DONE]`) and fetches the per-chat verification signature from the inference backend. The backend signs in a background task that finalizes *after* the final stream chunk and only then caches the signature — so cloud-api's immediate fetch frequently loses a race and gets `HTTP 404 ("Chat id not found or expired")`.

The error is logged and swallowed, so **the completion still returns fine** — but the chat's signature is never stored, and the response becomes **permanently unverifiable** (the verify path reads only cloud-api's repository and never re-fetches from the provider; after the backend's 1200s cache TTL the signature is gone there too).

Datadog triage: this 404 is **~95%** of the chronic `Failed to store chat signature` errors (`host:cpu02-cloud-api-prod`), running at dozens–hundreds/hr with org-concentrated bursts (a 3,550 burst at 14:00 UTC on 2026-06-01). The affected streams completed normally — **zero** disconnect warnings — confirming a timing/affinity miss rather than aborted streams.

## Fix

`get_signature` previously made a single pass (rotation-SNI → bucket → general LB) and gave up on a 404. This wraps that pass in a **bounded retry**: when *every reachable backend* returns 404 (the signing-race signature), back off briefly and re-fetch — up to 3 attempts with `[100ms, 250ms]` backoff.

- A non-404 status or a transport error (e.g. a **TLS SPKI / attestation mismatch**) stays **definitive** and is never retried.
- Total added hot-path latency is capped at **350ms**, well under the caller's 5s `FINALIZE_TIMEOUT`.
- Successful first attempts are unaffected (zero added latency).
- The backoff schedule lives in a module const with a pure helper, unit-tested for boundedness/termination (a faithful HTTP-level test isn't practical here — the signature clients use TLS-SPKI pinning, so there's no existing mock harness for `get_signature`).

## Testing

- `cargo test -p inference_providers --lib` — 180 passed (incl. new `signature_fetch_backoff_is_bounded_and_terminates`)
- `cargo test -p services --lib attestation` — 20 passed
- `cargo clippy -p inference_providers` — clean

## Scope

Addresses the dominant 404 cause in nearai/infra#160. Does **not** address the separate GLM-5.1 attested-fingerprint drift (the TLS SPKI mismatches) — that's an ops/discovery concern tracked in the same issue.


PierreLeGuen (commented):
I inspected the `get_signature` retry path, the rotation/bucket signature-routing code around it, related attestation/finalization callers, CI metadata, and ran focused local checks.

Non-blocking observation: `crates/inference_providers/src/vllm/mod.rs:38` and `crates/inference_providers/src/vllm/mod.rs:1213` say non-404 statuses and transport errors are definitive, and that the added hot-path latency is capped to the backoff sum. That is true for the bucket/general clients, but rotation-SNI failures at `crates/inference_providers/src/vllm/mod.rs:1238` and `crates/inference_providers/src/vllm/mod.rs:1245` are logged and then the loop can still retry when bucket/general return 404. The behavior preserves the old best-effort fallback, so I do not think it blocks this fix, but the comments should either narrow that contract to bucket/general or the code should track rotation definitive failures if that is the intended behavior.

Local checks passed: `cargo fmt --check`; `cargo test -p inference_providers signature_fetch_backoff_is_bounded_and_terminates`; `cargo test -p inference_providers signature_rotation`; `cargo test -p inference_providers signature_fetch`; `cargo test -p inference_providers unpin_chat_connection_clears_signature_rotation`. GitHub checks at review time: Lint/security/claude-review passed; Test Suite still pending.

lloydmak99:
Thanks for the reviews — addressed in `518ca2e` (and `24d1a51` earlier).

**Rotation-SNI definitive failures now fail fast** (@gemini-code-assist's high-priority point + review note #2): when a chat is pinned to a rotation backend, that backend's response is authoritative — bucket/general can't have the signature. A rotation 404 is still the signing-race signal (retry), but any non-404 status or transport error (e.g. a TLS SPKI/attestation mismatch) now breaks immediately and returns the rotation error verbatim, instead of falling through to a masking bucket/general 404 and getting retried 3×.

Other notes:
- **Latency wording** (#1): comment reworded — the ~350ms is the *backoff* sum only; each request keeps its own `control_timeout`, and the overall fetch is bounded by the caller's `FINALIZE_TIMEOUT`.
- **3× cost on non-race 404s** (#3): accepted as bounded; Datadog shows ~95% are the signing race. The rotation fast-fail above also removes the wasted bucket/general probes when rotation errors definitively.
- **`all_not_found` naming** (#4): left as-is — the inline comment clarifies it means "every client tried returned 404," and it's a no-op distinction with ≤2 clients today.

Also added earlier: a transport error on a *non-final* bucket/general client falls through to the next client rather than aborting the whole fetch.

`cargo clippy -p inference_providers` clean; `cargo test -p inference_providers --lib` 180 pass.

PierreLeGuen (changes requested):
Blocking finding:

- `crates/inference_providers/src/vllm/mod.rs:1237-1289` / `:1349`: when `rotation_target` is present, a rotation 404 still falls through to the bucket/general clients before the retry gate is allowed to back off. The comments correctly state that the rotation backend is authoritative for a rotation-pinned chat and bucket/general cannot have that signature. If the rotation backend returns the transient 404 this PR is trying to absorb, but the non-authoritative bucket/general path returns a 5xx or final transport error, `all_not_found` is cleared or the function returns, so the rotation 404 is not retried and the signature can still be permanently lost. For a rotation-pinned chat, a rotation 404 should be sufficient to retry the rotation target (or skip bucket/general entirely), while non-404/transport errors on the rotation target should continue to fail fast.

Local checks run:
- `cargo test -p inference_providers --lib` (pass, 180 tests)
- `cargo clippy -p inference_providers -- -D warnings` (pass)

lloydmak99:
@PierreLeGuen good catch — fixed in `f0c045c`.

You're right: for a rotation-pinned chat, a retryable rotation 404 was still falling through to the non-authoritative bucket/general clients, and a 5xx / final transport error there would clear `all_not_found` (or return early) and discard the rotation 404 — permanently losing a signature the retry was meant to recover.

Fix: the attempt is now two branches.
- **Rotation pin present** → talk *only* to the rotation backend (it's authoritative; bucket/general can't have the signature). A 404 is the signing-race signal → retry the rotation target; any non-404 status or transport error is definitive → fail fast with that error. Bucket/general are skipped entirely.
- **No pin** → walk bucket → general as before, treating an all-404 sweep as the race signal (with the non-final-client transport fall-through).

Also renamed `all_not_found` → `all_clients_404` per the earlier nit.

`cargo clippy -p inference_providers -- -D warnings` and `cargo test -p inference_providers --lib` (180) both pass.

PierreLeGuen (approved):

PR #706, opened 2026-06-02
fix(models): default input/output modalities to text when architecture is NULL

Evrard-Nil:
## Problem

`GET /v1/models` implements the [OpenRouter provider spec](https://openrouter.ai/docs/guides/community/for-providers), which marks `input_modalities` and `output_modalities` as **REQUIRED** fields. cloud-api derives them from the nullable `architecture` column (added in `V0043__add_model_architecture.sql`), so models whose architecture was never backfilled emit **no modality fields at all** — a spec-compliance gap.

Verified currently missing on 4 of 39 prod models:
- `anthropic/claude-sonnet-4-5`
- `openai/gpt-5.2`
- `openai/gpt-oss-120b`
- `Qwen/Qwen3-30B-A3B-Instruct-2507`

## Fix

`model_with_pricing_to_info` (the `/v1/models` OpenRouter serialization path in `crates/api/src/routes/completions.rs`) now defaults both `input_modalities` and `output_modalities` to `["text"]` whenever the DB modalities are NULL. This populates both the flat OpenRouter fields and the nested `architecture` shape, so the required fields can never be absent. Real per-model values are still set via the admin API and pass through untouched.

No DB migration needed — this is a serialization default only.

Added two unit tests:
- `model_without_architecture_defaults_to_text_modalities` — NULL architecture still emits text/text.
- `model_with_architecture_preserves_real_modalities` — real DB modalities pass through.

`cargo fmt --check`, `cargo clippy -p api`, and the new `cargo test -p api modalities` all pass locally. E2E suite uses MockProvider and doesn't exercise the modality serialization differently — the new unit tests cover the exact serialization path.

## Post-merge backfill checklist (run AFTER deploy, pending user sign-off)

The serialization default makes the listing spec-compliant immediately. To set the *real* modality values for the 4 affected models, run these `PATCH /v1/admin/models` calls (camelCase keys, body shape is `{model_id: {...}}`). Requires an admin session token or `adm_*` access token. **Do not run against prod without explicit sign-off.**

```bash
ADMIN=<admin-session-or-adm-token>

# claude-sonnet-4-5 — matches other Claude models (text+image in, text out)
curl -sS -X PATCH https://cloud-api.near.ai/v1/admin/models \
  -H "Authorization: Bearer $ADMIN" -H "Content-Type: application/json" \
  -d '{"anthropic/claude-sonnet-4-5":{"inputModalities":["text","image"],"outputModalities":["text"],"changeReason":"backfill OpenRouter modalities"}}'

[...]

PR #707, opened 2026-06-02
feat(models): add OpenRouter datacenters model field

Evrard-Nil:
## What

Adds per-model `datacenters` support, completing OpenRouter provider-spec coverage of the model catalog. Per the [OpenRouter provider spec](https://openrouter.ai/docs/guides/community/for-providers), a model may declare the datacenters it runs in as an array of objects with an ISO 3166 Alpha-2 `country_code`, e.g. `"datacenters": [{"country_code": "US"}]`.

This follows the exact conventions established by migration `V0051__add_openrouter_model_fields.sql` (and the in-flight V0052 `is_ready`/`deprecation_date` PR #705): nullable storage, no CHECK constraint, validation at the admin write path, mirrored onto `model_history` for audit.

## How

- **Storage** (`V0052__add_openrouter_datacenters.sql`): new nullable `datacenters TEXT[]` column on `models`, mirrored onto `model_history`. We store only the country codes; the `[{"country_code": ...}]` object wrapper is reconstructed at the serialization layer. Unlike `supported_features` (NOT NULL DEFAULT `'{}'`), `datacenters` is nullable so `GET /v1/models` can omit it entirely when unset — matching OpenRouter, which omits the field rather than emitting an empty array.
- **Write path** (admin `PATCH /v1/admin/models`): accepts `datacenters: [{"country_code": "US"}]`. Validates each code is exactly two ASCII uppercase letters (ISO 3166 Alpha-2); rejects garbage / lowercase / wrong-length with `400 invalid_request`.
- **Read path** (`GET /v1/models`, `GET /v1/model/{name}`, admin list, model history): serialized as `datacenters: [{"country_code": "US"}, ...]`, omitted when NULL/empty.
- Threaded through every layer the V0051 fields touch: DB models, `model.rs` / `admin_composite.rs` repositories (SELECT/INSERT/UPDATE/RETURNING/history), service ports (`models`, `admin`), API models (`ModelInfo`, `ModelMetadata`, `UpdateModelApiRequest`, `ModelHistoryEntry` + new `Datacenter` struct), and route handlers.

## Migration number (V0052 — must merge before #705)

Now uses **V0052** (renumbered from V0053). main is at V0051; neither this PR nor #705 has merged, so V0052 is free.

**Why renumber.** The PR previously claimed refinery "tolerates gaps so the two PRs can land in either order" — that is wrong, and @PierreLeGuen correctly flagged it. `crates/database/src/migrations/mod.rs` runs `refinery::Runner::new(&migrations)` with no overrides, and refinery 0.9.1's `Runner::new` defaults `abort_missing = true` (`refinery-core-0.9.1/src/runner.rs:240`). `verify_migrations` (`traits/mod.rs:69-83`) aborts startup with `MissingVersion` for any filesystem migration that is *not yet applied* whose version is `<= current` (the last-applied version). Increasing gaps are fine (V0051 → V0053 applies cleanly); a *lower-than-applied* unapplied version is not. So if V0053 (this PR) deployed before #705's V0052, #705's later V0052 would be lower-than-applied and abort migrations on startup.

**Resolution (option 1 — decouple).** This PR claims **V0052** and is self-contained: it can merge independently and first, with no cross-PR ordering dependency. **#705 must move its migration from V0052 to V0053** (a higher, brand-new version — always safe regardless of merge order). Coordinated with the #705 author.

The column / struct / SQL *content* is unchanged — only the migration file's version prefix moved (`V0053__… → V0052__…`).

- [ ] #705 renumbers its migration V0052 → V0053 to avoid the collision.

## Tests

- `crates/api/tests/e2e_all/admin_list_models.rs`: round-trip (set `[{US},{FR}]` via upsert, read back on admin list) + validation rejection (`"usa"` and lowercase `"us"` → 400).
- `crates/api/tests/e2e_all/model_history_test.rs`: `datacenters` captured in history across two updates (`[US]` → `[US, DE]`).
- Ran locally against docker postgres: all 3 new tests pass, full `admin_list_models` + `model_history` suites (20 tests) pass, `cargo test --lib --bins` (all) pass, `cargo fmt --check` + `cargo clippy --all-targets` clean.

## Post-merge backfill checklist (do NOT run against prod from this PR)

All NEAR AI self-hosted inference runs in US datacenters (OVH US), so self-hosted models should be backfilled with `[{"country_code":"US"}]`. After merge + deploy, per model:

```bash
curl -X PATCH https://cloud-api.near.ai/v1/admin/models \
  -H "Authorization: Bearer $ADMIN_SESSION_TOKEN" \
  -d '{"<model_name>": {"datacenters": [{"country_code": "US"}]}}'
```

- [ ] Verify on staging (`cloud-stg-api.near.ai`) first.
- [ ] Backfill self-hosted models in prod with `[{"country_code":"US"}]`.
- [ ] External/3rd-party-provider models: set per actual provider region (leave NULL if unknown).
- [ ] Confirm `GET /v1/models` surfaces `datacenters` for backfilled models.


Evrard-Nil:
Thanks for the careful review — both findings were valid and are fixed in 71fb507:

- **`soft_delete_model` history snapshot** (gemini + Copilot): added `datacenters` to the `UPDATE models ... RETURNING` clause so `record_model_history` captures it. Covered by `test_soft_delete_preserves_datacenters_in_history`.
- **`deprecate_model` in-transaction history INSERT** (gemini + Copilot): added the `datacenters` column + value (and the corresponding `datacenters` to that path's `UPDATE ... RETURNING`). Covered by `test_deprecate_preserves_datacenters_in_history`.

Note on the broader gemini comment that the deprecation INSERT also omits the V0051 fields (`hugging_face_id`, `quantization`, `max_output_length`, `supported_sampling_parameters`, `supported_features`): that's a pre-existing gap from #V0051 and out of scope for this PR — I only wired through `datacenters` to keep the diff focused. Happy to file a follow-up to backfill the rest into the deprecation snapshot if desired.

PierreLeGuen (changes requested):
Blocking operational issue:

- `crates/database/src/migrations/sql/V0053__add_openrouter_datacenters.sql:1` adds V0053 while PR #705 is still open with `V0052__add_openrouter_is_ready_deprecation_date.sql`. The PR body says refinery tolerates gaps and these can land in either order, but this repo runs `refinery::Runner::new(&migrations)` with default settings in `crates/database/src/migrations/mod.rs:45`; refinery 0.9 defaults `abort_missing = true`, and treats a filesystem migration with a version lower than the last applied version as missing. If this PR deploys first and applies V0053, then #705 later introduces unapplied V0052, startup migrations will fail with `MissingVersion`. Please make the merge order/renumbering explicit and enforceable before merge, or renumber so no lower-version migration can appear after V0053 has been applied.

Checks run locally: `cargo fmt --check` passed; `cargo test --lib --bins` passed. I also tried `cargo test --test e2e_all datacenters -- --nocapture` against a disposable local Postgres; migrations completed, but the e2e app startup failed in this environment because `/tmp` lacks the dstack socket needed to derive signing keys, so those targeted e2e assertions did not run locally. GitHub CI Test Suite is currently green.

Evrard-Nil:
Thanks @PierreLeGuen — you're right, and the old PR-body claim ("refinery tolerates gaps, either order works") was wrong. I verified the exact behavior in the vendored crate:

- `crates/database/src/migrations/mod.rs:45` calls `refinery::Runner::new(&migrations)` with no overrides.
- `refinery-core-0.9.1/src/runner.rs:240` — `Runner::new` defaults `abort_missing = true` (and `abort_divergent = true`).
- `refinery-core-0.9.1/src/traits/mod.rs:69-83` — `verify_migrations` aborts with `MissingVersion` for any not-yet-applied filesystem migration whose version is `<= current` (last applied). Increasing gaps are fine (V0051 → V0053 applies cleanly); a *lower-than-applied* unapplied version is not — exactly the V0053-then-V0052 hazard you described.

**Resolution: option 1 (decouple), not option 2.** I renumbered this migration **V0053 → V0052** so this PR is self-contained and can merge independently and first, with no cross-PR merge-order dependency. main is at V0051 and neither PR has merged, so V0052 is free.

**Action required on #705:** it must move its migration from V0052 → V0053. V0053 is a higher, brand-new version, so it applies safely regardless of which PR lands first. I've coordinated this with the #705 author and added a checklist item to the PR body. (Option 2 — keep V0053 + "depends on #705" gate — would have coupled this PR to #705, which is still mid-revision, so option 1 is the more robust choice you outlined.)

Column/struct/SQL content is unchanged; only the version prefix moved.

Re your local e2e note: same `/tmp` dstack-socket startup limitation here, but the migration itself runs in the e2e harness startup, and the targeted suites pass locally against docker postgres now at V0052 — `cargo test --test e2e_all -- admin_list_models model_history` → 21 passed (incl. datacenters round-trip, `400` validation rejection, history-records, and soft-delete-preserves-datacenters). `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test --lib --bins` (331) all clean. Rebased onto merged #706; the gemini/Copilot fixes (datacenters in `soft_delete_model` RETURNING + `deprecate_model` history INSERT) survived the rebase. Re-requesting review.

PierreLeGuen (changes requested):
Blocking migration collision remains:

- `crates/database/src/migrations/sql/V0052__add_openrouter_datacenters.sql:1` now claims V0052, but PR #705 is still open with `crates/database/src/migrations/sql/V0052__add_openrouter_is_ready_deprecation_date.sql`. If this review approves #707, it clears the existing changes-requested state and lets this V0052 land while #705 can still later merge a second V0052 file. GitHub will not flag that as a file conflict, and refinery startup will reject the migration set. Please renumber #705 to V0053, or otherwise make the merge-order/version collision impossible, before this PR is approved.

Checks: inspected the diff and surrounding DB/API code; `cargo check --workspace` passed; `cargo test -p api routes::completions::tests:: --lib` passed; local e2e was not run because there is no local Postgres (`pg_isready localhost:5432` reports no response). GitHub Lint, Test Suite, and security_audit are currently passing.

Evrard-Nil:
@PierreLeGuen the V0052 collision is resolved — this was posted just before #705 was renumbered. #705's migration moved **V0052 → V0053** in commit 6cd0803 (2026-06-03 09:28), so #707 now solely owns **V0052** and #705 owns **V0053**. The two are contiguous and apply cleanly in either merge order — no second V0052 file can land. Nothing changes in this PR; re-requesting review.

PierreLeGuen (approved):
Reviewed the datacenters field end to end: API DTOs/serialization, admin validation and mapping, repository SELECT/INSERT/UPDATE/RETURNING paths, model_history snapshots, migration V0052, and the added e2e coverage. I also verified via gh that PR #705 no longer claims V0052; its current diff uses V0053, so the duplicate-version blocker from the previous review is resolved for this PR.

Checks run locally: cargo fmt --all -- --check; cargo check --all-targets; cargo test -p api nano_dollars --lib; cargo test -p api test_openapi_spec_generation --lib; cargo test -p api test_openapi_models_endpoint_is_public --lib. GitHub CI for this PR is green. I did not rerun the DB-backed e2e suite locally; CI already ran the Test Suite successfully.

Issue #573, opened 2026-05-06
Silent model substitution: catalog deepseek-ai/DeepSeek-V3.1 returns Qwen/Qwen3.5-122B-A10B; gateway never checks attested model_name matches request


amiller:
# Issue draft for `nearai/cloud-api`

**Suggested title:** Silent model substitution — catalog `deepseek-ai/DeepSeek-V3.1` returns a `Qwen/Qwen3.5-122B-A10B` backend; gateway never checks attested `model_name` matches the request

> Filed against `nearai/cloud-api`. Independent of #224 (which is about cryptographic
> verification of the backend) — this is about catalog identity ↔ served model.

---

## Summary

Requesting `deepseek-ai/DeepSeek-V3.1` from `cloud-api.near.ai` returns a TDX-attested response from a backend that is actually serving **`Qwen/Qwen3.5-122B-A10B`** — different model, different weights, different KMS-derived signing key. The gateway silently routes the request and returns the wrong model's attestation; clients (and the model itself, when asked) report Qwen, but no protocol-level error or warning surfaces to the caller.

## Reproduction (single curl + jq, anyone with a `NEAR_API_KEY` can run)

**1. Attested model name (TDX-signed by the backend TD):**

```bash
curl -s "https://cloud-api.near.ai/v1/attestation/report?model=deepseek-ai/DeepSeek-V3.1&nonce=$(openssl rand -hex 32)" \
  -H "Authorization: Bearer $NEAR_API_KEY" | jq -r '.model_attestations[0].model_name'
# → Qwen/Qwen3.5-122B-A10B
```

**2. OpenAI-protocol response (the chat-completions API itself reports the substitution):**

```bash
curl -s "https://cloud-api.near.ai/v1/chat/completions" \
  -H "Authorization: Bearer $NEAR_API_KEY" -H "Content-Type: application/json" \
  -d '{"model":"deepseek-ai/DeepSeek-V3.1","messages":[{"role":"user","content":"What model are you?"}],"max_tokens":300}' \
  | jq '{response_model: .model, content: .choices[0].message.content}'
# → response_model: "Qwen/Qwen3.5-122B-A10B"
# → content: "I am Qwen3.5."
```

**3. Public-Git chain (the inner YAML at the deployed commit confirms HF download is Qwen):**

```bash
curl -s "https://cloud-api.near.ai/v1/attestation/report?model=deepseek-ai/DeepSeek-V3.1&nonce=$(openssl rand -hex 32)" \
[...]

PR #708, opened 2026-06-03
Surface model alias substitution: warning field + x-model-alias-resolved header + x-no-aliasing strict mode (#573)

Evrard-Nil:
## Summary

Fixes the *silent* part of #573: requests served through a model alias (e.g. a model deprecated via `POST /v1/admin/models/deprecate`) previously gave the caller no protocol-level signal that the catalog rewrote their model. This PR makes alias substitution **announced by default** and **rejectable on demand**:

1. **`x-model-alias-resolved` response header** (default-on) — every aliased chat completion (streaming + non-streaming) and attestation report now carries `x-model-alias-resolved: <requested> -> <canonical>`.
2. **OpenAI-style `warning` body field** (default-on) — aliased chat completions get a top-level `"warning"` string naming both the alias and the canonical model: in the response object for non-streaming, in the **first SSE chunk** for streaming.
3. **`x-no-aliasing` request header** (opt-in strict mode) — when set (any value except `false`/`0`), a request whose `model` is an alias is rejected with `400 model_alias_rejected` **before any inference happens** — before tokens are billed, and (on `/v1/attestation/report`) before an E2EE client binds a payload to a different model TD's signing key. Strict mode fails closed if the catalog can't be consulted.

Example, with the live alias from #573:

```
$ curl -si https://cloud-api.near.ai/v1/chat/completions -H "Authorization: Bearer $KEY" \
    -d '{"model":"deepseek-ai/DeepSeek-V3.1","messages":[...]}'
HTTP/2 200
x-model-alias-resolved: deepseek-ai/DeepSeek-V3.1 -> Qwen/Qwen3.5-122B-A10B

{"id":"...","model":"Qwen/Qwen3.5-122B-A10B","warning":"The requested model 'deepseek-ai/DeepSeek-V3.1' is an alias of 'Qwen/Qwen3.5-122B-A10B'; ...","choices":[...]}

$ curl -si ... -H "x-no-aliasing: true" -d '{"model":"deepseek-ai/DeepSeek-V3.1",...}'
HTTP/2 400
{"error":{"message":"Model 'deepseek-ai/DeepSeek-V3.1' is an alias of 'Qwen/Qwen3.5-122B-A10B' and the request set x-no-aliasing. Use the canonical model name 'Qwen/Qwen3.5-122B-A10B'.","type":"model_alias_rejected"}}
```

## Design notes

- **Alias detection is catalog-confirmed, not echo-based.** When the served model differs from the requested one, the route resolves the requested name against the catalog and only annotates when the requested name is a registered alias of the served model. External-provider echo differences (e.g. `gpt-4o` answered by a dated snapshot id) do **not** trigger the warning. Detection runs against the cache-backed active-models list (no per-request DB round-trip — `requested != served` is the *common* case for external providers); strict mode keeps the authoritative DB resolve, paid only by clients that opt in.
- **Raw-bytes hash trade-off (non-streaming):** injecting `warning` re-serializes the body, so it no longer byte-matches what the backend TD signed — same trade-off `x-auto-redact` already makes, but here it's default-on *for aliased responses only*. Deliberate: the warning must reach clients that never read headers, and those are exactly the clients hit by silent substitution. Clients that need raw-bytes verification should send the canonical name (the warning tells them which) or use `x-no-aliasing`.
- **E2EE bodies are opaque** and are left untouched; aliased E2EE responses signal via the response header only. The intended strict flow for E2EE clients is `x-no-aliasing` on `/v1/attestation/report`, which refuses *before* key binding.
- **Scope:** `/v1/chat/completions`, `/v1/completions` (legacy text), and `/v1/attestation/report`. `/v1/responses` can get the same treatment as a follow-up if we agree on the approach.
- This complements (does not replace) the discovery-time `model_name` equality check discussed in #573, which guards a different layer (backend self-report vs. discovered name).

## Tests

- Unit tests for header parsing and warning injection (`routes::common::tests`).
- New e2e suite `model_alias_transparency.rs` reproducing the exact #573 production path (create model → `POST /v1/admin/models/deprecate` onto the e2e Qwen model → request the deprecated name):
  - non-streaming: 200 + `warning` + `x-model-alias-resolved`
  - streaming: warning on **first** chunk only + header
  - `x-no-aliasing` (`true`/`1`/bare) → 400 `model_alias_rejected`; explicit `false` still serves
  - canonical-name and non-aliased requests stay byte-identical (no header, no warning)

Closes nothing on its own — #573 stays open for the discovery-time check and the catalog-hygiene question (whether cross-family deprecation aliases should sunset).


PierreLeGuen (changes requested):
Requesting changes for one strict-mode correctness issue.

- `crates/api/src/routes/attestation.rs:250`: the attestation `x-no-aliasing` check fails open on catalog/DB errors. This block ignores any `models_service.resolve_and_get_model()` error and then still calls `attestation_service.get_attestation_report`; that service resolves aliases internally (`crates/services/src/attestation/mod.rs:728`) before fetching the canonical model report. For strict E2EE clients, a transient catalog failure can therefore still bind the request to a different model despite `x-no-aliasing`, while also omitting `x-model-alias-resolved`. The chat path fails closed for the same case. Please distinguish `NotFound` from resolver failures here, fail closed when `x-no-aliasing` is set, and add attestation strict-mode coverage.

Local checks run: `cargo test -p api routes::common::tests`; `cargo test -p api --test e2e_all model_alias_transparency --no-run`. GitHub checks were passing when refreshed.

Evrard-Nil:
Addressed all review feedback in ec6cfcf0:

- **Hot-path DB lookup** (claude review): `check_alias_resolution` now resolves against the cache-backed active-models list (`get_models_with_pricing`, 60s-TTL `models_list_cache`) instead of a per-request `resolve_and_get_model` DB query. The external-provider snapshot-echo path no longer touches the DB or delays TTFT. Strict mode (`x-no-aliasing`) keeps the authoritative DB resolve since it's opt-in and fail-closed. PR description corrected accordingly.
- **Attestation strict mode fails closed** (gemini, copilot): `resolve_and_get_model` errors other than `NotFound` now return 500 when `x-no-aliasing` is set, matching the completions route.
- **`Access-Control-Expose-Headers`** (copilot ×2, claude): both chat paths now emit a single comma-joined value (`Inference-Id, x-model-alias-resolved`); the attestation path appends to any existing expose list instead of replacing it.
- **Header-value panic guard** (claude): the alias header value in completions is now built via `HeaderValue::from_str` and skipped on invalid bytes rather than deferring a panic to `.body().unwrap()`.
- **Attestation e2e coverage** (copilot): added `test_attestation_report_announces_alias` + `test_attestation_report_no_aliasing_rejects`.

Full e2e suite: 515 passed / 0 failed.

Evrard-Nil:
@PierreLeGuen this crossed with ec6cfcf0 (pushed a few minutes after your review) — it fixes exactly this:

- `get_attestation_report` now matches on the resolver result: `Ok` (alias → 400-in-strict / annotate), `Err(NotFound)` falls through to the attestation service's own unknown-model error, and **any other resolver error returns 500 when `x-no-aliasing` is set** — same fail-closed semantics as the chat path.
- Added attestation strict-mode e2e coverage: `test_attestation_report_no_aliasing_rejects` (strict 400 on alias, asserted before any report fetch) and `test_attestation_report_announces_alias` (header on the served path). The resolver-*failure* branch isn't e2e-exercisable without fault injection on the models repo, so that branch is covered by the explicit `Err(_)` match rather than a test — shout if you'd rather have a mock-repo unit test around the handler.

CI is green on ec6cfcf0. Re-requesting review.

PierreLeGuen (changes requested):
Reviewed locally against `origin/main...HEAD`.

Blocking finding:
- `crates/api/src/routes/completions.rs:107-110`: `check_alias_resolution` only surfaces an alias when the response `model` exactly equals the catalog canonical name. That misses real aliases for external-provider models that use `provider_config.model_name` overrides. For example, a catalog model `openai/gpt-5.2` can route upstream as `gpt-5.2` (`crates/inference_providers/src/external/mod.rs:239-245`), and the OpenAI-compatible backend preserves the upstream response model (`crates/inference_providers/src/external/openai_compatible.rs:240-255`). If a client requests an alias of `openai/gpt-5.2`, the completion service resolves it and sends `gpt-5.2`, but this check sees `served == "gpt-5.2"` and fails `m.model_name == served`, so both the warning and `x-model-alias-resolved` header are silently omitted on streaming and non-streaming responses (`completions.rs:1142-1145`, `completions.rs:1404-1408`). Please derive alias-ness from catalog resolution of the requested name, or otherwise map external upstream names back to the catalog canonical before comparing, and add coverage for an alias to an external model with a remote model-name override.

Checks run:
- `cargo test -p api --lib routes::common::tests::test_no_aliasing_requested_present_variants`
- `cargo test -p api --lib routes::common::tests::test_inject_warning_field_object`
- `cargo test -p inference_providers test_create_external_provider_openai_with_model_name_override`

Evrard-Nil:
Fixed in d2a69bf9 — good catch on the `provider_config.model_name` override path.

- `check_alias_resolution` now derives alias-ness from catalog resolution of the **requested** name (find the active model whose `aliases` contain it), mirroring exactly what the completion service resolved — the response echo is no longer compared against the canonical name, so upstream-name overrides can't suppress the warning/header. Echo differences on non-alias requests (`gpt-4o` → dated snapshot) still resolve to `None` since the requested name isn't a registered alias of anything.
- Coverage: added `ResponseTemplate::with_model` to the mock provider (echoes an upstream name instead of the request param, like `external/mod.rs` override behavior) and `test_alias_to_external_model_with_upstream_name_override_warns` — external model with `providerConfig.model_name`, deprecated alias onto it, asserts `response.model == upstream` while the `warning` + `x-model-alias-resolved: <alias> -> <canonical>` still fire.

Local: clippy clean, `inference_providers` 180/180, full e2e 515 passed + the one `admin_analytics::test_admin_platform_org_revenue` failure which reproduces without this branch on a polluted shared test DB (passes on a fresh DB — top-100 revenue window pollution, unrelated).

Re-requesting review.

PierreLeGuen (changes requested):
Blocking issue found:

- `crates/api/src/routes/completions.rs:110`: `check_alias_resolution` returns `None` whenever `requested == served`, but aliases can legitimately resolve to a canonical external model whose `provider_config.model_name` echoes the same string the client requested. Example: alias `gpt-5.2` -> canonical `openai/gpt-5.2`, with upstream/model override `gpt-5.2`. In that case the completion service still substituted the alias before dispatch, but both the streaming path (`crates/api/src/routes/completions.rs:1152`) and non-streaming path (`crates/api/src/routes/completions.rs:1414`) will skip the warning/header, making the alias substitution silent. This also contradicts the PR's test intent that alias-ness is derived from catalog resolution rather than the response echo; the current external override test only covers `served != requested`. Please derive alias-ness from the catalog even when the provider echo equals the requested string, or carry the pre-dispatch resolution result through from the service, and add a regression test where the upstream model override equals the alias.

Checks run locally:
- `git diff --check origin/main...HEAD` passed.
- `cargo test -p api routes::common::tests --lib` passed.
- `cargo test -p api --test e2e_all model_alias_transparency -- --nocapture` compiled, but could not run locally because Postgres was not available (`Connection refused` during test DB bootstrap).

Evrard-Nil:
Fixed in aef003ae — you're right that the override can collapse onto the alias string itself, at which point any echo-gated check is structurally blind.

Detection is now **pre-dispatch and fully echo-independent**: a new `ModelsServiceTrait::resolve_alias_cached(identifier) -> Option<canonical>` resolves the requested name against the cached active-models alias map once, before the streaming/non-streaming split (same `moka` coalesced cache as `get_models_with_pricing`, via a factored `cached_models()` returning the `Arc` — no Vec clone, no DB on the hot path). The response `model` echo is no longer consulted anywhere; the first-chunk-model plumbing in the streaming path is deleted. Strict mode keeps the authoritative DB resolve.

Regression coverage: `test_alias_equal_to_upstream_override_still_warns` — canonical `openai/<bare>` with `providerConfig.model_name = <bare>` AND alias `<bare>`; asserts echo == requested while the `warning`, `x-model-alias-resolved: <bare> -> openai/<bare>`, and strict-mode 400 all still fire.

Local: clippy clean; alias suite 9/9; full e2e 517/0 on a fresh DB; services 331/0; inference_providers 180/0. Re-requesting review.

PierreLeGuen (changes requested):
Requesting changes for one endpoint coverage gap.

- `crates/api/src/routes/completions.rs:1647` / `crates/api/src/routes/completions.rs:1730`: `/v1/completions` still dispatches through `create_chat_completion_stream`/`create_chat_completion`, and that service resolves aliases, but this route never calls `reject_if_aliased` or `resolve_alias_cached` and only exposes `Inference-Id` at `crates/api/src/routes/completions.rs:1708` / `crates/api/src/routes/completions.rs:1752`. An aliased legacy completion can still be silently served, and `x-no-aliasing: true` will not reject it, while chat has the new strict/header path at `crates/api/src/routes/completions.rs:1054`. Since this endpoint uses the same alias-resolution path, this leaves a public model-substitution path outside the new contract. Please apply the same strict-mode and `x-model-alias-resolved`/warning handling here, or explicitly disable aliases for `/v1/completions`.

Checks run locally:
- `cargo fmt --check` passed.
- `cargo test -p api routes::common` passed.
- `cargo test -p api --test e2e_all model_alias_transparency -- --nocapture` could not be validated locally because the test bootstrap DB was unavailable (`Connection refused`).

GitHub checks at review time: Lint passed; Test Suite and security_audit were still pending.

Evrard-Nil:
Covered in a943bd63 — `/v1/completions` now has the identical contract:

- `reject_if_aliased` (strict `x-no-aliasing` → 400 `model_alias_rejected`, fail-closed on resolver errors) and pre-dispatch `resolve_alias_cached` run before dispatch, same as chat
- aliased responses carry `x-model-alias-resolved` (comma-joined into the existing expose list) and the top-level `warning` — in the response object for non-streaming, in the first SSE chunk for streaming
- no raw-bytes trade-off on this endpoint: it already re-serializes, and E2EE/auto-redact are rejected upfront

Coverage: `test_legacy_completions_alias_contract` — non-streaming warning+header, streaming first-chunk-only warning+header, strict 400, and canonical-request stays unannotated.

Local: clippy clean, alias suite 10/10, full e2e 518/0 on a fresh DB. Re-requesting review.

PierreLeGuen (approved):
Reviewed the alias transparency changes across chat completions, legacy completions, attestation, model alias cache plumbing, and the new e2e coverage. I did not find a blocking correctness issue.\n\nLocal checks: cargo fmt --all -- --check; cargo test -p api routes::common::tests. GitHub checks observed passing: Lint, Test Suite, security_audit.

Issue #630, opened 2026-05-21
completions: decide whether SSE error frames belong in the response-bytes hash


Evrard-Nil:
## Context

In `crates/api/src/routes/completions.rs`, the streaming chat-completions path appends every successful SSE chunk and the final `data: [DONE]` to `accumulated_bytes` (the buffer used for the response-bytes hash / signature). The stream-error branch does not.

```rust
// success path
accumulated_inner.lock().await.extend_from_slice(&sse_bytes);
Ok::<Bytes, Infallible>(sse_bytes)

// error path
Ok::<Bytes, Infallible>(sse_error_frame(&e))   // <-- not extended into accumulated_inner
```

This is **pre-existing behavior** (the prior `data: error: <msg>` shape had the same omission) and was flagged by Copilot during review of #629.

## Question

Two possible interpretations of the response-bytes hash:

1. **\"Valid model output only\"** — error frames are cloud-api-injected, not from the model, so excluding them is correct and the signature should not cover them.
2. **\"Everything sent to the client\"** — the hash represents the on-wire response, and the signature should cover error frames too. Otherwise a verifier replaying \`accumulated_bytes\` cannot reproduce what the client received.

Pick one and document it next to \`accumulated_bytes\`. If #2, also extend \`accumulated_inner\` in the error branch.

## Repro / observation

Send a streaming chat-completions request that hits a stream-level error (upstream sglang/vLLM SSE chunk fails JSON decode, or timeout). The bytes returned to the client include the error frame; the bytes hashed for signing do not.

## Related

- PR #629 introduced the new (well-formed JSON) error frame but did not change which bytes go into \`accumulated_inner\`
- Surfaced by Copilot review: https://github.com/nearai/cloud-api/pull/629#discussion_r3280400650

PR #709, opened 2026-06-03
fix(completions): forward streamed SSE bytes verbatim for byte-exact TEE verification

### LINKED ISSUE #701: Gateway re-serializes streamed SSE chunks, breaking byte-exact chat-verification against TEE signatures
## Summary

The documented [chat verification flow](https://docs.near.ai/cloud/verification/chat) (hash the exact response bytes you received, compare with the TEE signature's response hash) **does not work through the gateway** for streaming requests: cloud-api re-serializes upstream SSE chunks before forwarding them, so the bytes the client receives differ from the bytes the model TEE emitted and signed.

## Repro (2026-06-02, production)

Same request body sent twice (`Qwen/Qwen3.5-122B-A10B`, `stream: true`):

**Direct (`qwen35-122b.completions.near.ai`):** received-bytes sha256 `8cb30eef…` == response hash in `/v1/signature/{id}` text `Qwen/Qwen3.5-122B-A10B:2974f24b…:8cb30eef…` ✅ (signature recovers to TEE signing address)

**Gateway (`cloud-api.near.ai`):** request hash in the signature matches the sent body (`2974f24b…` ✅) but the response hash (`c2c78cc3…`) does **not** match the sha256 of the received bytes (`db25207d…`) ❌

Chunk-level diff shows the gateway rewrites upstream chunks:

Upstream TEE chunk:
```
data: {"choices":[{"delta":{"content":"Under","reasoning_content":null},"finish_reason":null,"index":0,"logprobs":null,"matched_stop":null}],"created":...,"id":"...","model":"...","object":"chat.completion.chunk"}
```

Same chunk as forwarded by the gateway (keys reordered, fields stripped, usage injected):
```
data: {"id":"...","object":"chat.completion.chunk","created":...,"model":"...","choices":[{"index":0,"delta":{"content":"Under"}}],"usage":{"prompt_tokens":18,"completion_tokens":1,"total_tokens":19}}
```

## Impact

- A client that talks to the gateway cannot reproduce the signed response hash from the bytes it received, so signature verification fails even for untampered responses.
- Docs previously presented the gateway flow as byte-for-byte verifiable; updated in nearai/docs#29 to steer byte-exact verification to direct completions endpoints, but the gateway-side gap remains.

## Options

1. Forward upstream SSE bytes verbatim on the streaming path (pass-through instead of decode/re-encode) so the TEE-signed hash is reproducible through the gateway.
2. Declare gateway streams non-verifiable byte-for-byte and document direct completions as the only byte-exact path (status quo after nearai/docs#29).
3. Have the gateway return its own transformation proof (e.g. sign its re-serialized bytes inside the gateway TEE and chain to the model signature).

## Related

- #630 (which bytes belong in the response hash — error frames)
- #573 (gateway never checks attested model_name)

Evrard-Nil:
Fixes #701.

## Problem

The documented [chat verification flow](https://docs.near.ai/cloud/verification/chat) doesn't work through the gateway for streaming: cloud-api decoded every upstream SSE chunk into `ChatCompletionChunk` and re-serialized it (`completions.rs` streaming path), so the bytes the client received differed from the bytes the inference TEE emitted and signed — `sha256(received bytes)` could never match the signed response hash. Non-streaming already forwards `raw_bytes` verbatim; this brings streaming to parity.

## Approach (option 1 from #701)

**Lossless parser.** `BufferedSSEParser` gains `SSEEventParser::passthrough_raw()` (true for `OpenAIEventParser`, i.e. our vLLM/SGLang path + external OpenAI-compatible upstreams). For passthrough parsers, control lines — blank separators, `: comments`, non-data SSE fields (`event:`/`id:`), `data: [DONE]`, and any trailing unterminated bytes at stream end — are emitted as chunk-less events (`SSEEvent.chunk: Option<StreamChunk>`, `None` = control) so concatenating every event's `raw_bytes` reproduces the upstream byte stream **exactly**. This mirrors what inference-proxy hashes: the bytes it actually sends, including non-data lines and a final-line flush. Gemini/Anthropic parsers keep the historical skip behavior (their raw bytes are native-format and must be re-serialized).

**Route passthrough.** The chat streaming route forwards `event.raw_bytes` untouched when the event is passthrough-capable and auto-redact is inactive, and suppresses the gateway-minted `[DONE]` when the upstream's own terminator was forwarded. Auto-redact (mutates chunk text) and non-OpenAI providers (need normalization) keep the re-serialization path. Parsing still happens alongside — usage/billing (`InterceptStream`), TTFT/ITL metrics, `Inference-Id`, first-chunk error classification/rotation are all unchanged; none of them needs to rewrite client bytes.

**Not changed:**
- `/v1/completions` (text) reshapes chat chunks into text-completion format by design → stays on re-serialization, just skips control events.
- Mid-stream error frames still surface as a gateway-minted error frame (#630 territory; inference-proxy doesn't cache a signature for streams that don't complete cleanly anyway).
- Clients already received per-chunk `usage` (the gateway requests `continuous_usage_stats` upstream) — no behavior change there.

**Cleanups:** removed `accumulated_bytes` (buffered every streamed response in memory, never read) and `chat_id_state` (write-only) from the chat streaming route.

## Tests

- New parser test `test_sse_parser_passthrough_byte_exact_reassembly`: multi-packet stream with comments, CRLF endings, `event:` fields, `[DONE]`, trailing unterminated bytes — asserts byte-exact reassembly.
- New `test_sse_parser_non_passthrough_parser_skips_control_lines` pins the Gemini/Anthropic-style behavior.
- **Regression guard:** the mock provider now emits wire bytes in a deliberately different-but-equivalent JSON encoding (sorted keys + a field unknown to our structs) and registers signature hashes over those bytes — exactly how inference-proxy signs what it sends. `test_streaming_chat_completion_signature_verification` (asserts `sha256(received) == signed hash` end-to-end) **fails on the re-serialization path (verified by temporarily disabling passthrough) and passes only with verbatim forwarding**.
- Full suite green locally: 667 unit + 508 e2e (docker postgres).

## Verification after deploy

`nearai/infra-tests#30` added byte-exact hash assertions for all four paths with gateway-stream xfail'd (non-strict) referencing #701 — it will show **XPASS** in the infra suite once this reaches prod, then the marker can be removed.

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -33,6 +33,14 @@ use uuid::Uuid;
 // Timeout for synchronous usage recording before response is returned
 const USAGE_RECORDING_TIMEOUT_SECS: u64 = 5;
 
+// Upper bound on leading SSE control events (keepalive comments, blank
+// lines) consumed while peeking for the first parsed chunk. Real upstreams
+// emit zero before the first data chunk; the cap stops a misbehaving or
+// malicious upstream from stalling response start or growing the stash
+// unbounded. Past the cap we proceed without an Inference-Id header and let
+// the remaining stream (including the buffered control events) flow through.
+const MAX_LEADING_CONTROL_EVENTS: usize = 32;
+
 /// Insert validated E2EE headers into a provider `extra` HashMap.
 fn insert_encryption_headers(
     encryption_headers: &crate::routes::common::EncryptionHeaders,
@@ -1120,7 +1128,11 @@ async fn chat_completions_inner(
                 // upstream keepalive comment) may precede the first data
                 // chunk; consume and stash them so they are still forwarded
                 // to the client in order (they're part of the signed byte
-                // stream — issue #701).
+                // stream — issue #701). Bounded by MAX_LEADING_CONTROL_EVENTS
+                // so a misbehaving upstream that only emits keepalives can't
+                // stall response start or grow this buffer unbounded — past
+                // the cap we proceed without an Inference-Id and let the rest
+                // flow through the byte stream below.
                 let mut leading_control: Vec<
                     Result<inference_providers::SSEEvent, inference_providers::CompletionError>,
                 > = Vec::new();
@@ -1135,6 +1147,9 @@ async fn chat_completions_inner(
                         _ => break None,
                     };
                     if is_control {
+                        if leading_control.len() >= MAX_LEADING_CONTROL_EVENTS {
+                            break None;
+                        }
                         if let Some(ev) = peekable_stream.next().await {
                             leading_control.push(ev);
                         }
@@ -1218,7 +1233,9 @@ async fn chat_completions_inner(
                                     // and are non-byte-verifiable anyway since the TEE
                                     // signs under the canonical model name, not the alias
                                     // the client requested.
-                                    if event.raw_passthrough && !auto_redact_enabled && !alias_served
+                                    if event.raw_passthrough
+                                        && !auto_redact_enabled
+                                        && !alias_served
                                     {
                                         if event.is_done_marker() {
                                             upstream_done
@@ -1268,15 +1285,17 @@ async fn chat_completions_inner(
                                     let alias_warning =
                                         pending_warning.lock().ok().and_then(|mut g| g.take());
                                     let json_data = match alias_warning {
-                                        Some(warning) => serde_json::to_value(&chunk).map(|mut v| {
-                                            if let Some(obj) = v.as_object_mut() {
-                                                obj.insert(
-                                                    "warning".to_string(),
-                                                    serde_json::Value::String(warning),
-                                                );
-                                            }
-                                            v.to_string()
-                                        }),
+                                        Some(warning) => {
+                                            serde_json::to_value(&chunk).map(|mut v| {
+                                                if let Some(obj) = v.as_object_mut() {
+                                                    obj.insert(
+                                                        "warning".to_string(),
+                                                        serde_json::Value::String(warning),
+                                                    );
+                                                }
+                                                v.to_string()
+                                            })
+                                        }
                                         None => serde_json::to_string(&chunk),
                                     }
                                     .unwrap_or_else(|e| {
@@ -1730,8 +1749,11 @@ async fn completions_inner(
                 // header, consuming any leading control events. This route
                 // reshapes chat chunks into text-completion format, so
                 // control lines are never forwarded (no byte passthrough
-                // here by design) and the stash can be discarded.
+                // here by design) and the consumed events can be discarded.
+                // Bounded so a keepalive-only upstream can't stall the
+                // response: past the cap we proceed without an Inference-Id.
                 let mut peekable_stream = Box::pin(stream.peekable());
+                let mut control_skipped = 0usize;
                 let inference_id = loop {
                     let is_control = match peekable_stream.as_mut().peek().await {
                         Some(Ok(event)) => {
@@ -1743,6 +1765,10 @@ async fn completions_inner(
                         _ => break None,
                     };
                     if is_control {
+                        if control_skipped >= MAX_LEADING_CONTROL_EVENTS {
+                            break None;
+                        }
+                        control_skipped += 1;
                         peekable_stream.next().await;
                     }
                 };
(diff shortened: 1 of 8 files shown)


PierreLeGuen (changes requested):
Two issues need to be fixed before this can merge:

1. `crates/inference_providers/tests/integration_tests.rs:657-672` still treats every streamed item as a parsed chat chunk. With this PR the mock/provider stream now also emits the chunk-less `[DONE]` control event, so `test_reasoning_content` panics on the new valid event. This is already failing in GitHub CI (`Test Suite` / `test_reasoning_content`) and I reproduced it locally with `cargo test -p inference_providers --test integration_tests test_reasoning_content -- --nocapture`. Please update the integration test/stream consumers to skip `None` control events where appropriate.

2. `crates/inference_providers/src/vllm/mod.rs:1117` and `crates/inference_providers/src/vllm/mod.rs:1598` still only inspect `peekable.peek().await` once to decide whether the first SSE payload is a retryable in-stream `HttpError`. After this parser change, a leading comment/blank/control line is now surfaced as `Ok(SSEEvent { chunk: None, ... })`, so a stream like `: keepalive\n\ndata: {"error":{"code":503,...}}\n\n` will bypass rotation and return the canonical backend error instead of trying the rotation fallback. The same control-skipping/stashing pattern used in the pool/route peeks should be applied here before deciding that no retryable first error exists.

Checks run locally: `git diff --check origin/main...HEAD` passed; `cargo test -p inference_providers sse_parser -- --nocapture` passed; `cargo test -p services streaming_chat_id_mapping_available_immediately -- --nocapture` passed; `cargo test -p inference_providers --test integration_tests test_reasoning_content -- --nocapture` failed as above.

Evrard-Nil:
Both addressed, thanks:

1. **`test_reasoning_content`** — fixed in 9b9b77b1 (raced your review; CI was green on that push): added a `None => {}` arm for control events.

2. **Rotation peeks masking first-chunk errors** — fixed in 36fa2e84. Both sites (`chat_completion_stream` and `try_chat_completion_stream_rotation`) now go through a shared `VLlmProvider::peek_first_payload_status` helper that consumes leading chunk-less control events before classifying the first real payload, and re-attaches them in order on the success path (they're part of the signed byte stream, so dropping them would break byte-exactness). Covered by three new unit tests, including the exact `: keepalive\n\n` + in-stream 503 scenario you described — it now returns `Some(503)` and replays the control events ahead of the error.

(Local note: the `models.rs` `detect_audio_content_type` doctest failure is pre-existing on main — nextest in CI doesn't run doctests.)

Evrard-Nil:
Thanks — addressed the review feedback and rebased on latest `main` (resolved a conflict with the alias-warning feature: alias-served streaming responses now stay on the re-serialization path so the `warning` injection is preserved, and passthrough is disabled for them since they're non-byte-verifiable under the canonical model name anyway).

On the **unbounded leading-control-event buffering** (three inline comments on the chat route, text route, and provider pool): good catch in principle. Real vLLM/SGLang upstreams emit zero control frames before the first data chunk, but I've capped the peek-and-stash at `MAX_LEADING_CONTROL_EVENTS = 32` at all four peek sites (chat route, text route, provider pool, and the vLLM rotation first-chunk classifier). Past the cap we proceed without the optional `Inference-Id` header / sticky-routing pin rather than buffering indefinitely; the already-buffered control events still flow through to the client so byte-exactness holds. Commit ff8b5485.

On the **`event.chunk` move from a shared reference** (completions/mod.rs `InterceptStream`): no change needed — the code matches by reference (`if let Some(StreamChunk::Chat(ref chat_chunk)) = event.chunk` and `event.chunk.is_none()`), so nothing is moved out of the `&event`. It compiles and the `InterceptStream` unit tests pass.

All green locally after rebase: 335 unit + 541 e2e + 8 integration (mock).

PierreLeGuen (approved):
Reviewed the streaming passthrough changes across the SSE parser, chat/text completion routes, InterceptStream, provider-pool pinning, vLLM first-payload peeking, response-service consumers, mocks, and updated tests. I did not find a substantive correctness issue in the current diff.

Local checks run:
- `cargo test -p inference_providers sse_parser -- --nocapture`
- `cargo test -p inference_providers --test integration_tests test_reasoning_content -- --nocapture`
- `cargo test -p inference_providers peek_first_payload_status -- --nocapture`
- `cargo test -p services streaming_chat_id_mapping_available_immediately -- --nocapture`
- `cargo test -p api completions::tests -- --nocapture`

GitHub checks at review time: Lint and security_audit passed; Test Suite was still pending.

PR #714, opened 2026-06-04
fix(inference_provider_pool): cover both algos when backend_count < ALGOS.len()

### LINKED ISSUE #710: Ed25519 pubkey not registered in pubkey_to_providers when a model is added mid-restart
## Symptom

`POST /v1/chat/completions` against a freshly-added inference_url model returns:

```
HTTP 421
{"error":{"message":"The encryption key is no longer valid. Please refresh your attestation report and retry.","type":"provider_error","param":null,"code":null}}
```

…only when `X-Signing-Algo: ed25519`. The same model + same request with `X-Signing-Algo: ecdsa` returns 200.

The error comes from `services::inference_provider_pool::get_providers_with_fallback`: `pubkey_to_providers.get(pub_key)` returns `None` for the ed25519 key the client sent (correctly fetched from `GET /v1/attestation/report`).

## Repro on staging (2026-06-03)

Newly added model `deepseek-ai/DeepSeek-V4-Flash` via `PATCH /v1/admin/models`. End-to-end E2EE works direct against `dsv4-flash.completions.near.ai` (ecdsa + ed25519, non-streaming + streaming + system-message — all 8/8 pass in `test_direct_e2ee.py`). Via `cloud-stg-api.near.ai`:

| algo | path | result |
|---|---|---|
| ecdsa | `cloud-stg-api → completions.near.ai` | ✅ 200 |
| ed25519 | `cloud-stg-api → completions.near.ai` | ❌ 421 (no provider found) |
| ecdsa | direct `dsv4-flash.completions.near.ai` | ✅ 200 |
| ed25519 | direct `dsv4-flash.completions.near.ai` | ✅ 200 |

Both pubkeys returned by `cloud-stg-api/v1/attestation/report?signing_algo={ecdsa,ed25519}` match the keys returned by the direct endpoint, so the discovery FETCH worked; the cache WRITE for ed25519 didn't.

Persisted across 4+ scheduled refresh cycles (>20 min) and two manual `PATCH /v1/admin/models` re-loads.

## Suspected cause

During the original `load_inference_url_models` call, the backend container was in a crash-loop (TDX-incompatible flags being iterated on). `discover_model` issues rotation calls for `[ecdsa, ed25519]` in sequence. Likely one of:

1. The ed25519 rotation call hit a transient backend error during the crash-loop window, the failure was logged but the per-algo cache update was skipped silently — and subsequent refreshes don't re-attempt to populate the ed25519 entry because they only refresh keys already present.
2. The discover loop is short-circuiting (continuing/breaking) on first algo success.

Code under suspicion: `services::inference_provider_pool::discover_model` and the per-algo cache update in `load_inference_url_models` / its refresh path.

## Repro steps (general)

1. Add a new vllm inference_url model via `PATCH /v1/admin/models` while the backend container is restarting.
2. Wait for the backend to stabilize.
3. `GET /v1/attestation/report?model=X&signing_algo=ed25519` → returns key K
4. `POST /v1/chat/completions` with `X-Model-Pub-Key: K` and `X-Signing-Algo: ed25519` → 421
5. Same with `signing_algo=ecdsa` → 200

## Expected

Both algos should be re-discovered on every refresh tick, not just whatever was cached at the first successful sync. A failed discover should not leave the algo entry empty; subsequent refreshes should populate it.

## Workaround

Bouncing cloud-stg-api forces a fresh discovery and the ed25519 entry populates. Unclear if a `PATCH /v1/admin/models` with `isActive=false` then `isActive=true` would do the same — untested.

## Test artifacts

- infra-tests PR: nearai/infra-tests#31 documents the failures + which tests cover the gap.
- 3 failing cases on staging: `test_stg_e2ee.py::TestStaging{NonStreaming,Streaming}::test_*[ed25519-DeepSeek-V4-Flash]`

Evrard-Nil:
Fixes #710.

## Bug

`discover_model` uses

```rust
let algo = ALGOS[i % ALGOS.len()].to_string();   // ALGOS = ["ecdsa","ed25519"]
```

across `i in 0..backend_count`. For `backend_count == 1`, only `i = 0` is ever evaluated, so only `"ecdsa"` is fetched. The ed25519 signing pubkey never lands in `pubkey_to_providers`, and chat completions with `X-Signing-Algo: ed25519` return 421 `no provider found for model … with public key …` — even though the live `/v1/attestation/report?signing_algo=ed25519` (which hits the backend directly, bypassing the cache) returns the correct ed25519 pubkey.

Models with `backend_count >= 2` were unaffected because indices 0 and 1 modulo 2 happen to cover both algos.

## Empirical cross-check

Across prod today (`completions.near.ai/backends/count?domain=...`):

| Model | `backend_count` | ed25519 routing pre-fix |
|---|---|---|
| `dsv4-flash` | 1 | ❌ 421 |
| `qwen3-6-27b` | 1 | ❌ (not added to cloud-api yet — same bug would apply) |
| `qwen3-6-35b` | 2 | ✅ |
| `qwen3-30b` | 2 | ✅ |
| `gpt-oss-120b` | 2 | ✅ |
| `gemma-4-31b` | 3 | ✅ |
| `glm-5-1` | 4 | ✅ |

`X-Encryption-Version: 2` was ruled out as a cause: that header gates payload v1/v2 codec selection at the backend, not routing. Without it, every multi-backend model returns 400 `Decryption failed`; with it, 200. The 421 from DSV4 is independent and identical regardless of the header.

## Fix

Decouple the algo from the backend index. Total calls = `max(backend_count, ALGOS.len())`:

- `backend_index = i % backend_count` for rotation-SNI routing
- `algo = ALGOS[i % ALGOS.len()]` for algo selection

For `backend_count >= ALGOS.len()` (≥2 today) this is identical to the previous behavior — same number of calls, same algo coverage. For `backend_count == 1` it issues **two** calls to the same backend (one per algo).

The `complete_coverage` check in `apply_pin_update` still keys off `backend_count` (not `call_count`): under the single-backend case, both algo calls hit the same backend so the deduped fingerprint set has size 1, matching `backend_count`. Rapid-eviction semantics preserved.

## What's untouched

- TLS-cert pinning logic (`apply_pin_update`, `FingerprintState`) — unchanged.
- `pubkeys_by_algo` write semantics (first-write-wins per algo) — unchanged; just gets at least one write per algo now.
- Refresh / reuse path at `sync_inference_url_models` — already merges `pubkeys_by_algo` additively. With this fix, the ed25519 entry actually gets harvested and backfilled on the same cycle the bug previously silently dropped it.
- `DiscoveryOutcome::observed_fingerprints` — docstring updated; for `backend_count == 1` it now contains the same fingerprint twice (one entry per call). The deduped set used for pin updates is `verified_this_round`.

## Verification plan

- [x] `cargo build -p services`
- [x] `cargo clippy -p services -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --lib -p services inference_provider_pool` → 48 pass / 0 fail
- [ ] Post-deploy on staging: PATCH a single-backend model (e.g. add DSV4 there), then `POST /v1/chat/completions` with `X-Signing-Algo: ed25519` + correct `X-Model-Pub-Key` → expect 200, not 421.
- [ ] Post-deploy on prod: same check against the existing DSV4 deployment.

Existing tests use `MockProvider` and don't traverse the live `discover_model` HTTP path, so no unit test in this file directly exercises the bug. The behavioral change is small enough — and the empirical evidence narrow enough — that I held off introducing a wiremock-based test for the discovery loop in this PR. Happy to add one in a follow-up if reviewers want it.

## Cost

Discovery is one cycle every 5 minutes per model. For single-backend models, this adds **one extra HTTP call per 5 min**. Negligible.

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -61,10 +61,12 @@ struct DiscoveryOutcome {
     /// they're identical across backends of the same model.
     pubkeys_by_algo: HashMap<String, String>,
     /// Per-call verified TLS fingerprints observed in this pass, in launch
-    /// order (by index — `futures::future::join_all` preserves the input
-    /// order, not completion order). With rotation, each index lands on a
-    /// distinct backend, so under complete coverage every entry is unique
-    /// and `observed_fingerprints.len() == backend_count`.
+    /// order (`futures::future::join_all` preserves input order, not
+    /// completion order). One entry per call, not per backend, so under
+    /// complete coverage `observed_fingerprints.len() == max(backend_count,
+    /// ALGOS.len())`. When `backend_count == 1`, the two algo calls hit
+    /// the same backend and entries repeat — the set of *distinct*
+    /// fingerprints is `verified_this_round` (used for pin updates).
     observed_fingerprints: Vec<String>,
     /// Per-call failure reasons that prevented a fingerprint observation, in
     /// launch order. Each entry is `"{category}: {detail}"` where category
@@ -766,10 +768,18 @@ impl InferenceProviderPool {
         None
     }
 
-    /// Run N attestation-discovery calls against a model URL, each on its own
+    /// Run attestation-discovery calls against a model URL, each on its own
     /// minimal `reqwest::Client` (fresh TCP connection, isolated `FingerprintState`),
-    /// alternating signing algorithms to harvest both ECDSA and Ed25519 signing
-    /// pubkeys in the same pass.
+    /// covering every (backend_index, signing_algo) pair needed to harvest
+    /// both ECDSA and Ed25519 signing pubkeys in the same pass.
+    ///
+    /// The number of calls is `max(backend_count, ALGOS.len())`: one per
+    /// backend (for TLS-cert fingerprint coverage across all serving CVMs)
+    /// and at least one per algo (so both ECDSA and Ed25519 pubkeys are
+    /// fetched even when a model has only a single backend). For
+    /// `backend_count >= ALGOS.len()`, this degenerates to one call per
+    /// backend; for `backend_count == 1` it issues two calls to the same
+    /// backend, one per algo.
     ///
     /// Why a fresh client per call: reqwest with HTTP/2 multiplexes many
     /// concurrent requests onto a single TCP connection, which hashes to a
@@ -780,8 +790,8 @@ impl InferenceProviderPool {
     /// (see model-proxy PR #27). A fresh-TCP probe to that SNI is routed
     /// deterministically to `backends_sorted_by_address[N % healthy_count]`,
     /// bypassing the least-connections LB. We fetch the healthy count from
-    /// `/backends/count`, then fan out one call per backend index. One cycle
-    /// = full coverage.
+    /// `/backends/count`, then fan out calls across backend indices and
+    /// algos. One cycle = full coverage.
     ///
     /// Why an isolated Bootstrap state per call: if discovery calls shared
     /// the caller's `fingerprint_state`, the first call that completed and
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Inspected the single-file diff in `crates/services/src/inference_provider_pool/mod.rs`, including the discovery fan-out change, pin replacement invariant, provider/pubkey mapping consumers, rotation helpers, tests, and CI metadata.

Local checks passed:
- `cargo test -p services --lib inference_provider_pool -- --nocapture` (48 passed)

Remote checks at review time: lint, security audit, and claude-review were passing; GitHub Test Suite was still pending. I did not find substantive correctness or operational issues in this PR.

Issue #697, opened 2026-06-01
Sampling params (top_k, repetition_penalty, min_p, top_a) hard-400 on OpenAI-source passthrough models


Evrard-Nil:
## Summary

`openai/*` models (OpenAI-source passthrough) return **HTTP 400** when the request includes a sampling parameter that OpenAI's API doesn't recognize — `top_k`, `repetition_penalty`, `min_p`, `top_a`. cloud-api forwards the request body verbatim and OpenAI rejects the unknown argument:

```
{"error":{"message":"Unrecognized request argument supplied: top_k", ...}}
```

For OpenRouter listing this is a problem: OR clients send these optional sampling params broadly, and any request that includes one and gets routed to an OpenAI-backed model hard-fails. Anthropic and self-hosted (vLLM/SGLang) paths already handle unknown sampling params gracefully (accept + ignore or apply) — only the OpenAI-source path 400s.

## Evidence (live prod, `openai/gpt-4.1`)

| Param sent | OpenAI-source (`gpt-4.1`) | Anthropic | self-hosted vLLM (Qwen3-30B) |
|---|---|---|---|
| `top_k: 5` | **400** unrecognized | 200 | 200 (applied) |
| `repetition_penalty: 1.8` | **400** | 200 (ignored) | 200 (applied) |
| `min_p: 0.5` | **400** | 200 | 200 (applied) |
| `top_a: 0.5` | **400** | 200 | 200 (accepted/ignored) |

## Repro

```bash
KEY=sk-...
curl -s -w " HTTP %{http_code}\n" https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"openai/gpt-4.1","messages":[{"role":"user","content":"hi"}],"max_tokens":10,"top_k":5}'
# HTTP 400  Unrecognized request argument supplied: top_k
```

## Suggested fix

Strip sampling parameters outside the provider's accepted set before forwarding to **OpenAI-source** providers (the same graceful-ignore behavior the Anthropic path already has). i.e. the OpenAI adapter should only forward params OpenAI's API accepts and silently drop the rest, so an OR client sending `top_k`/`min_p`/`top_a`/`repetition_penalty` gets a 200 with those params ignored rather than a 400.

## Related
[...]

Issue #696, opened 2026-06-01
temperature: non-default values rejected with 400 on anthropic/claude-opus-4-7 passthrough


Evrard-Nil:
## Summary

`anthropic/claude-opus-4-7` returns **HTTP 400** for any **non-default `temperature`** value, even though it advertises `temperature` in `supported_sampling_parameters`. Only `temperature: 1.0` (and omitting it, which defaults to 1.0) succeeds.

```
{"error":{"message":"Provider failed for model 'anthropic/claude-opus-4-7': `temperature` is deprecated for this model.","type":"invalid_request_error",...}}
```

This is an OpenRouter-listing blocker: OR and OpenAI-SDK clients routinely send `temperature: 0`, `0.7`, etc. Those all hard-fail on this model.

## Evidence (live prod)

| Request | HTTP |
|---|---|
| no `temperature` (cloud-api defaults to 1.0) | **200** |
| `temperature: 1.0` | **200** |
| `temperature: 0.7` | **400** `temperature is deprecated for this model` |
| `temperature: 0.0` | **400** (same) |
| `top_p: 0.5` | 200 (top_p is fine) |

`anthropic/claude-opus-4-6` accepts `temperature: 0.5` → 200, so this is specific to opus-4-7 (mirrors Anthropic deprecating `temperature` on certain newer models).

## Repro

```bash
KEY=sk-...
for T in 1.0 0.7 0.0; do
  curl -s -w " HTTP %{http_code}\n" https://cloud-api.near.ai/v1/chat/completions \
    -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
    -d "{\"model\":\"anthropic/claude-opus-4-7\",\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}],\"max_tokens\":10,\"temperature\":$T}" \
    | python3 -c 'import sys,json;d=json.load(sys.stdin);print(d.get("error",{}).get("message","OK")[:70])'
done
```

## Suggested fix

The Anthropic adapter (`crates/inference_providers/src/external/anthropic/`) should not forward a `temperature` that the target model rejects. Options:
1. **Drop `temperature` for models that don't support non-default values** (per-model capability flag), forwarding only `top_p` — cleanest for OR interop.
2. Clamp/ignore `temperature` on these models and log a debug note.
[...]

PR #695, opened 2026-06-01
fix(completions): accept chat `stop` as string or array (OpenAI-spec / OpenRouter)

Evrard-Nil:
## Summary

OpenAI's chat-completions spec allows `stop` to be **either a single string or an array of strings**. `ChatCompletionRequest.stop` was typed `Option<Vec<String>>`, so a bare-string `"stop": "foo"` failed serde deserialization with **HTTP 422**:

```
Failed to deserialize the JSON body into the target type:
stop: invalid type: string "foo", expected a sequence
```

The array form (`"stop": ["foo"]`) worked fine. OpenRouter and the OpenAI SDK both send the **bare-string** form, so this surfaces as spurious 422s — found during OpenRouter-listing verification.

## Fix

The legacy `/v1/completions` (`CompletionRequest`) struct already uses the flexible `StopSequences` untagged enum (`Single(String) | Many(Vec<String>)`). This PR reuses it for the chat struct too:

- `ChatCompletionRequest.stop`: `Option<Vec<String>>` → `Option<StopSequences>`
- Flatten to `Vec<String>` at the domain boundary via `into_vec()` (conversions.rs + completions.rs), so nothing downstream changes — the domain `ChatCompletionParams.stop` stays `Option<Vec<String>>`.
- Derive `Serialize` on `StopSequences` (the request struct is serialized elsewhere). Untagged serialization round-trips correctly: `Single` → string, `Many` → array.

## Verification

```bash
# Before: HTTP 422.  After: HTTP 200.
curl -s -w '%{http_code}\n' https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-30B-A3B-Instruct-2507","messages":[{"role":"user","content":"count 1 to 9"}],"max_tokens":40,"stop":"5"}'
```

Unit tests added (`cargo test -p api --lib stop`, both pass):
- `conversions::tests::test_chat_completion_stop_accepts_string_or_array` — string → `["STOP"]`, array → `["a","b"]`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #668, opened 2026-05-26
response_format: json_object returns markdown-fenced JSON on passthroughs; json_schema ignored on Gemini


Evrard-Nil:
## Summary

`response_format: {"type": "json_object"}` on Anthropic and Gemini passthroughs returns content wrapped in **markdown code fences** (`\`\`\`json\n{...}\n\`\`\``) instead of raw JSON. Every `JSON.parse(response.content)` will throw.

And `response_format: {"type": "json_schema", ...}` is **completely ignored** on Gemini — returns plain text unrelated to the schema.

Self-hosted models honor both correctly (tested in iter 4: gpt-oss-120b returns valid raw JSON).

## Repro

```bash
KEY=sk-...

# Anthropic — returns markdown-fenced JSON
curl -s https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"anthropic/claude-haiku-4-5","messages":[{"role":"user","content":"Give me a JSON: {\"city\":\"Paris\"}"}],"max_tokens":200,"response_format":{"type":"json_object"}}' \
  | jq -r .choices[0].message.content
# ```json
# {
#   "city": "Paris"
# }
# ```

# Gemini-3-pro — same
curl -s ... -d '{"model":"google/gemini-3-pro",...,"response_format":{"type":"json_object"}}' \
  | jq -r .choices[0].message.content
# ```json
# {
#   "city": "Paris"
# }
# ```

# Gemini-3-pro with json_schema — totally broken
curl -s ... -d '{"model":"google/gemini-3-pro","messages":[{"role":"user","content":"Give me a tiny weather record for Paris."}],"max_tokens":300,"response_format":{"type":"json_schema","json_schema":{"name":"weather","strict":true,"schema":{"type":"object","additionalProperties":false,"required":["city","temp_c","conditions"],"properties":{"city":{"type":"string"},"temp_c":{"type":"number"},"conditions":{"type":"string","enum":["sunny","cloudy","rain","snow"]}}}}}}'
# Returns: " Highest: 42.6°C / " — not even JSON, took 36s
# gemini-2.5-flash with same: "Here's a tiny weather record for Paris:" — plain text, ignores schema entirely
```

## Why it matters

`response_format: json_object` is the standard way to enforce raw JSON output in the OpenAI SDK. Users:
1. Send `response_format: {"type":"json_object"}`
2. Get back a string they assume is JSON
[...]

PR #715, opened 2026-06-04
test(e2e): port infra-tests param/health/reasoning coverage to mocked integration tests

Evrard-Nil:
## What

Closes a gap found while auditing infra-tests against cloud-api's integration suite: several **cloud-api behaviours that the live infra-tests exercise had no mocked integration coverage** — they only failed (or xfailed) against prod/staging. This adds them as `MockProvider`-backed e2e tests so they run in CI without live TEE/inference backends.

### New test files (`crates/api/tests/e2e_all/`)

**`openrouter_params.rs`** — ported from infra-tests `test_openrouter_params.py`. The live suite asserts behavioural outcomes against real models; here we assert the equivalent **cloud-api contract** on the self-hosted path: each parameter is **accepted** (HTTP 200, never 400/422) and **forwarded** to the provider (typed `ChatCompletionParams` field, or the `extra` passthrough map for the long tail). These guard the rejection/drop bugs the live suite tracks:
  - `stop` array + bare string (PR #695)
  - non-default `temperature` (#696)
  - sampling extras `top_k`/`top_a`/`repetition_penalty`/`min_p` (#697)
  - `seed`, `logprobs`/`top_logprobs`
  - `tool_choice` `required`/`none` (#619)
  - `response_format: json_schema` (#668)

**`reasoning.rs`** — ported from `test_reasoning.py`. Splits the live test's intent into the two halves cloud-api owns: provider `reasoning_content` is surfaced (non-streaming + streaming), and `chat_template_kwargs` is forwarded intact.

**`health.rs`** — ported from `cloud_api_helpers.check_health`. `GET /v1/health` is public (no auth) and reports `status=ok`. Previously only a unit test existed.

## Why mocked, not 1:1 ported

infra-tests buckets into (A) cloud-api API behaviour, (B) TEE/CVM/crypto/model-proxy/attestation, and (C) staging mirrors. Only **(A)** is meaningfully portable to mocked integration tests — (B) inherently needs live attestation/signing/SNI fan-out and already has the right mock-level coverage (`chat_encryption.rs`, `signature_verification.rs`, `crates/services/tests/attestation_verification_test.rs`). This PR ports the (A) gaps.

The mock deliberately ignores params it can't honour (it won't truncate on `stop` or emit `logprobs`), so the tests assert **acceptance + passthrough** — which is exactly cloud-api's responsibility and exactly where #695/#696/#697-style bugs live.

## Test plan
- `cargo test -p api --test e2e_all -- openrouter_params reasoning health` → **13 passed, 0 failed** (against local Postgres)
- `cargo fmt -p api --check` clean
- `cargo clippy -p api --tests` clean


PierreLeGuen (approved):
Reviewed the added mocked e2e coverage for health, OpenRouter/OpenAI parameter forwarding, and reasoning propagation, plus the surrounding chat request conversion, completion service extraction, mock provider, and health route wiring. No substantive issues found.

Checks: cargo fmt --all --check passed. cargo test -p api --test e2e_all health::test_health_endpoint_public_and_ok compiled the e2e_all binary but could not execute locally because the test bootstrap could not connect to local Postgres (connection refused). GitHub CI currently shows Lint, Test Suite, security_audit, and claude-review passing.

Evrard-Nil:
Thanks for the reviews — addressed in 01438076.

**gemini** (reasoning.rs duplication): extracted a shared `setup()` helper.

**claude + Copilot** (streaming reasoning substring): now parses the `data:` SSE lines and reassembles the reasoning value from the deltas, asserting it equals what the provider emitted.

**Copilot** (`seed`/`logprobs`/`top_logprobs` are typed fields, assertions on `extra` will fail): this is the one case where the bot suggestions don't apply. Although `ChatCompletionParams` *has* those typed fields, the service conversion in `crates/services/src/completions/mod.rs` (`create_chat_completion_stream` / `create_chat_completion`) **hardcodes `seed: None`, `logprobs: None`, `top_logprobs: None`** and forwards the original values through the `extra` passthrough map. So asserting on `params.extra` is what reflects the real request to the provider — the tests pass (13/13) precisely because of this. I've added a module-level note documenting it so it doesn't trip up future readers. (Worth a follow-up question for maintainers: is leaving those typed fields unpopulated intentional, or should the conversion map them? Either way the passthrough behaviour is what's tested here.)

**Copilot** (parametrized `param:` key is literal): `json!` actually evaluates bare-identifier keys as expressions, so each iteration used a distinct key — but I've made it explicit via `insert(param, …)` to remove the ambiguity.

Re-ran: `cargo test -p api --test e2e_all -- openrouter_params reasoning health` → 13 passed, 0 failed; fmt + clippy clean.

Issue #613, opened 2026-05-20
Chat completions: top_logprobs validation gap — accepted without logprobs=true and above 20


Evrard-Nil:
## Summary

`POST /v1/chat/completions` accepts invalid `top_logprobs` values that should be rejected per the OpenAI spec.

| Request | Expected | Actual |
|---|---|---|
| `top_logprobs: 2` without `logprobs: true` | 400 (per spec, requires logprobs=true) | 200, request runs |
| `top_logprobs: 100` (spec range is 0–20) | 400 out-of-range | 200, request runs |

These silently-accepted requests proceed to inference (cost is incurred) and return `logprobs: null`, masking client bugs that would otherwise be caught upfront.

## Repro

```bash
KEY=sk-32c0476395fd40c795725fc101f33304

# Should 400: top_logprobs without logprobs=true
curl -s -w "%{http_code}\n" https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"zai-org/GLM-5.1-FP8","messages":[{"role":"user","content":"hi"}],"max_tokens":10,"top_logprobs":2}'
# → 200

# Should 400: top_logprobs > 20
curl -s -w "%{http_code}\n" https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"zai-org/GLM-5.1-FP8","messages":[{"role":"user","content":"hi"}],"max_tokens":10,"logprobs":true,"top_logprobs":100}'
# → 200
```

## Suggested fix

Reject at the API boundary:
- If `top_logprobs` is set, require `logprobs: true` → otherwise 400 `"top_logprobs requires logprobs to be true"`.
- If `top_logprobs > 20` or `< 0` → 400 `"top_logprobs must be in the range 0..=20"`.

Saves inference cost on malformed requests and gives clients a real error to act on.

---
Found during routine cloud-api smoke test.

Evrard-Nil:
Same validation-gap family — adding one more case found in routine smoke testing:

**`stop` array accepts more than 4 entries** (OpenAI spec caps at 4):

```bash
curl -s https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
[...]

PR #716, opened 2026-06-04
fix(completions): forward frequency_penalty/presence_penalty to self-hosted backends (#622)

### LINKED ISSUE #622: frequency_penalty / presence_penalty silently dropped on all self-hosted models
## Summary

`frequency_penalty` and `presence_penalty` parameters have **no observable effect** on any self-hosted model. With `temperature: 0` and a highly-repetitive instruction, output is byte-identical with and without the penalty at the OpenAI maximum value (2.0). The most likely cause is that cloud-api drops these fields before forwarding upstream.

## Evidence

Same prompt `"Repeat the word banana ten times separated by spaces."`, `temperature: 0`, `max_tokens: 80`. Penalty 0 vs penalty 2.0:

| Model | Engine | fp=0 | fp=2.0 |
|---|---|---|---|
| `Qwen/Qwen3-30B-A3B-Instruct-2507` | vLLM | "banana banana banana banana banana banana banana banana banana banana" | **identical** |
| `openai/gpt-oss-120b` | vLLM | "banana × 10" | **identical** |
| `zai-org/GLM-5.1-FP8` | SGLang | "banana × 10" | **identical** |
| `google/gemma-4-31B-it` | SGLang | "banana × 10" | **identical** |

Penalty 2.0 should heavily suppress repeated tokens — instead nothing changes across engines. Same when testing a soft prompt ("Write a 30-word story about a cat") — penalty 0 and penalty 2.0 produce the same story word-for-word.

## Repro

```bash
for fp in 0 2.0; do
    -d "{\"model\":\"Qwen/Qwen3-30B-A3B-Instruct-2507\",\"messages\":[{\"role\":\"user\",\"content\":\"Repeat the word banana ten times separated by spaces.\"}],\"max_tokens\":80,\"temperature\":0,\"frequency_penalty\":$fp}" \
    | jq -r .choices[0].message.content
done
# both lines identical
```

## Suggested fix

Trace the parameter through the request-forwarding layer; likely missing in the Rust → upstream JSON mapper. Both vLLM and SGLang accept `frequency_penalty` and `presence_penalty` on their `/v1/chat/completions` surface, so if cloud-api forwarded them they'd take effect.

## Related (same general theme)

- Out-of-range values (e.g. `frequency_penalty: 3.0` or `-3.0`) are silently accepted with HTTP 200. Spec range is [-2.0, 2.0]. Same validation-gap family as #613.
- Wrong-type values (`frequency_penalty: "oops"`) return HTTP 422 with raw plain-text serde error (`Failed to deserialize the JSON body into the target type...`) — same envelope-bypass as #609.

---

Evrard-Nil:
## Summary

`frequency_penalty` and `presence_penalty` had no observable effect on any self-hosted (vLLM/SGLang) model — output was byte-identical at penalty 0 vs 2.0.

**Root cause:** Both are typed fields on `ChatCompletionRequest`, so serde `#[flatten] extra` never captured them. The service layer (`create_chat_completion[_stream]`) hardcodes the typed `ChatCompletionParams.frequency_penalty`/`presence_penalty` slots to `None`. With nothing in `extra` and the typed slots nulled, both penalties were dropped before reaching the backend.

**Fix:** In `convert_chat_request_to_service`, forward the typed penalty fields into the `extra` passthrough map — matching how `seed`/`logprobs`/`top_logprobs` and the text-completion path already work. The vLLM adapter spreads `extra` into the upstream JSON body, so both penalties now reach the self-hosted backend.

Scope kept tight: only penalty forwarding. Out-of-range validation is a separate issue (same family as #613) and intentionally not addressed here.

## Testing
- `cargo build --workspace` — passes
- `cargo test -p inference_providers --lib --tests` — 180 + 8 pass (the one doctest failure on `detect_audio_content_type` pre-exists on main, unrelated)
- `cargo test --test e2e_all -- openrouter` — 10/10 pass, incl. new `test_penalties_accepted_and_forwarded` asserting both penalties land in `params.extra`

Fixes #622

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -288,6 +288,26 @@ fn convert_chat_request_to_service(
     body_hash: RequestBodyHash,
     request_id: Uuid,
 ) -> ServiceCompletionRequest {
+    // `presence_penalty` / `frequency_penalty` are typed fields on
+    // `ChatCompletionRequest`, so `#[serde(flatten)] extra` never captures them.
+    // The service layer hardcodes the typed `ChatCompletionParams` penalty slots
+    // to `None`, so forward them through `extra` (matching seed/logprobs and the
+    // text-completion path) — otherwise they are silently dropped before reaching
+    // the self-hosted backend (nearai/cloud-api #622).
+    let mut extra = request.extra.clone();
+    if let Some(presence_penalty) = request.presence_penalty {
+        extra.insert(
+            "presence_penalty".to_string(),
+            serde_json::json!(presence_penalty),
+        );
+    }
+    if let Some(frequency_penalty) = request.frequency_penalty {
+        extra.insert(
+            "frequency_penalty".to_string(),
+            serde_json::json!(frequency_penalty),
+        );
+    }
+
     ServiceCompletionRequest {
         request_id,
         model: request.model.clone(),
@@ -325,7 +345,7 @@ fn convert_chat_request_to_service(
         store: None,
         body_hash: body_hash.hash.clone(),
         response_id: None, // Direct chat completions API calls don't have a response_id
-        extra: request.extra.clone(),
+        extra,
     }
 }
 
--- a/crates/api/tests/e2e_all/openrouter_params.rs
+++ b/crates/api/tests/e2e_all/openrouter_params.rs
@@ -422,3 +422,55 @@ async fn test_json_schema_response_format_accepted_and_forwarded() {
         "response_format.type not preserved on passthrough"
     );
 }
+
+// ── frequency_penalty / presence_penalty (nearai/cloud-api #622) ─────────────
+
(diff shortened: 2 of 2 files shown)


PierreLeGuen (approved):
Reviewed the chat completion conversion, request models, service ChatCompletionParams construction, vLLM/OpenAI-compatible serialization, Anthropic extra filtering, and the new e2e regression. The change matches the existing passthrough pattern used for seed/logprobs and the text-completion path, and I did not find a substantive issue.

Checks: `cargo fmt --check` passed. `cargo test -p api --test e2e_all openrouter_params::test_penalties_accepted_and_forwarded` compiled but could not run locally because the test DB bootstrap failed to connect to Postgres (`Connection refused`). GitHub CI shows Lint, Test Suite, and security_audit passing.

PR #717, opened 2026-06-04
fix(openai): strip non-OpenAI sampling params before forwarding (#697)

### LINKED ISSUE #697: Sampling params (top_k, repetition_penalty, min_p, top_a) hard-400 on OpenAI-source passthrough models

Evrard-Nil:
## Summary

`openai/*` (OpenAI-source passthrough) models returned **HTTP 400** when a request carried a sampling param OpenAI's API doesn't recognize — `top_k`, `repetition_penalty`, `min_p`, `top_a`. cloud-api forwarded the body verbatim and OpenAI rejected the unknown argument (`Unrecognized request argument supplied: top_k`). OpenRouter clients send these optional knobs broadly, so any such request routed to an OpenAI-backed model hard-failed.

## Fix

In the OpenAI-compatible adapter, strip those passthrough-only sampling extras (they land in `ChatCompletionParams.extra`) before forwarding — but **only when the upstream is OpenAI-source** (`api.openai.com` / `*.openai.azure.com`), reusing the same scoping as the existing `strip_reasoning_effort_if_unsupported` mitigation.

- Other openai-compatible providers (Together, Groq, Fireworks, OpenRouter) tend to accept/apply these knobs → left untouched.
- vLLM/SGLang/Anthropic use different backends → unaffected.
- Result: 200 with the unsupported param silently ignored (the same graceful-ignore semantics the Anthropic path already has) instead of 400.

## Tests

- 5 new unit tests in `openai_compatible.rs`: OpenAI + Azure strip all four knobs; non-OpenAI providers preserve them; supported extras (`seed`/`logprobs`/`response_format`) survive; no-op when none present.
- `cargo build --workspace` ✅; `cargo test -p inference_providers --lib` ✅ (185 passed). The pre-existing `detect_audio_content_type` doctest failure also reproduces on `main` and is unrelated.
- e2e `cargo test --test e2e_all -- openrouter` ✅ (9 passed), incl. `test_sampling_extras_accepted_and_forwarded` (vLLM-path contract unchanged: these still ride through in `extra` for self-hosted models).

Fixes #697

### DIFF
--- a/crates/inference_providers/src/external/openai_compatible.rs
+++ b/crates/inference_providers/src/external/openai_compatible.rs
@@ -90,7 +90,7 @@ fn strip_reasoning_effort_if_unsupported(
     if params.tools.as_ref().is_none_or(|t| t.is_empty()) {
         return;
     }
-    if !base_url.contains("api.openai.com") && !base_url.contains(".openai.azure.com") {
+    if !is_openai_source(base_url) {
         return;
     }
     if params.extra.remove("reasoning_effort").is_some() {
--- a/crates/inference_providers/src/external/openai_compatible.rs
+++ b/crates/inference_providers/src/external/openai_compatible.rs
@@ -108,7 +108,17 @@ fn strip_reasoning_effort_if_unsupported(
 /// because they tend to accept (or apply) the extra sampling knobs below, and
 /// stripping them there would silently change behaviour.
 fn is_openai_source(base_url: &str) -> bool {
-    base_url.contains("api.openai.com") || base_url.contains(".openai.azure.com")
+    // Match on the parsed URL *host* (lower-cased), not a substring of the whole
+    // URL. A substring check would both miss mixed-case hosts (`API.OPENAI.COM`)
+    // and misclassify look-alikes such as `api.openai.com.evil.example` as
+    // OpenAI-source, silently mutating requests bound for the wrong upstream.
+    match url::Url::parse(base_url)
+        .ok()
+        .and_then(|u| u.host_str().map(str::to_ascii_lowercase))
+    {
+        Some(host) => host == "api.openai.com" || host.ends_with(".openai.azure.com"),
+        None => false,
+    }
 }
 
 /// Sampling knobs that OpenAI's `/v1/chat/completions` does **not** recognise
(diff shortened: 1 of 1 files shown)


Evrard-Nil:
Addressed the automated review feedback in the two follow-up commits:
- `is_openai_source` now parses the URL host and matches case-insensitively, so look-alikes like `api.openai.com.evil.example` no longer misclassify and mixed-case hosts work (added tests for both). Also reused it in `strip_reasoning_effort_if_unsupported` for consistency.
- Collapsed the per-key `warn!` into a single `debug!` line — these params are sent broadly by OpenRouter clients and prod runs at info level, so per-key warn would flood logs.

PierreLeGuen (approved):
Reviewed the single-file change in crates/inference_providers/src/external/openai_compatible.rs plus the surrounding external-provider default injection path, request model flattening, and OpenRouter passthrough tests. The strip happens after provider defaults are injected and before both streaming and non-streaming OpenAI-compatible forwards, and it preserves non-OpenAI-compatible providers.

Local checks: cargo fmt --check passed; cargo test -p inference_providers openai_compatible::tests::test_strip_sampling_params -- --nocapture passed (5 tests). I also attempted cargo test -p api --test e2e_all openrouter_params -- --nocapture, but it failed before exercising the tests because the local admin Postgres bootstrap connection was refused. GitHub checks are green (Lint, Test Suite, security_audit, claude-review).

Issue #619, opened 2026-05-20
tool_choice='none' ignored on vLLM-served models (Qwen3.5-122B, gpt-oss-120b)


Evrard-Nil:
## Summary

`tool_choice: "none"` is supposed to suppress all tool calls and force a text response. On vLLM-served models, this is ignored — the model still emits a `tool_calls` array. SGLang-served models honor it correctly.

| Model | Engine | tool_choice=none | Result |
|---|---|---|---|
| `Qwen/Qwen3.5-122B-A10B` | vLLM | weather prompt | **calls `get_weather` 3/3** ❌ |
| `openai/gpt-oss-120b` | vLLM | weather prompt | **calls `get_weather` 2/2** ❌ |
| `zai-org/GLM-5.1-FP8` | SGLang | weather prompt | text reply, no tool call ✓ |

## Why this matters

`tool_choice: "none"` is the standard way for an agent to say "I gave you tools earlier, but on *this* turn I want you to summarize/answer without invoking anything." With this broken on vLLM models, agents can't safely disable tools mid-conversation.

## Repro

```bash
KEY=sk-32c0476395fd40c795725fc101f33304
curl -s https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model":"Qwen/Qwen3.5-122B-A10B",
    "messages":[{"role":"user","content":"What is the weather in Paris?"}],
    "tools":[{"type":"function","function":{"name":"get_weather","description":"Get weather.","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}],
    "tool_choice":"none",
    "max_tokens":500
  }' | jq '.choices[0] | {finish_reason, tool_calls: .message.tool_calls}'
```

Returns:
```json
{"finish_reason":"tool_calls","tool_calls":[{"function":{"arguments":"{\"city\":\"Paris\"}","name":"get_weather"}, ...}]}
```

## Suggested fix

Cleanest place to enforce this is at the cloud-api boundary: **when `tool_choice == "none"`, strip the `tools` array from the upstream request entirely** before forwarding to vLLM. With no tools available, the model can't call any. The OpenAI-spec semantic ("model is forbidden from calling any tool") is preserved end-to-end. This works regardless of whether the underlying engine honors `tool_choice`.

Separate small finding while testing tool_choice:
[...]

PR #152, opened 2025-11-06
fix: tool call streaming

think-in-universe:
Fixes #150

PR #718, opened 2026-06-04
fix(tools): honor tool_choice=none by stripping tools before forwarding (#619)

Evrard-Nil:
## What

Honor `tool_choice: "none"` end-to-end: when a request sets `tool_choice: "none"`, cloud-api now strips the `tools` array (typed field **and** any `tools` left in the `extra` passthrough map) before forwarding upstream. Backends that ignore `tool_choice` (notably vLLM-served Qwen / gpt-oss) therefore have nothing to call, so the OpenAI semantic ("model is forbidden from calling any tool this turn") holds regardless of backend behavior. `tool_choice: "none"` itself is still forwarded (harmless, preserves intent for backends that honor it).

## Scope change after review

This PR originally also rewrote `finish_reason` (`stop`→`tool_calls`) on forced-tool calls. @PierreLeGuen correctly flagged that doing so in cloud-api breaks **byte-exact TEE response-hash verification** — the chat signature is the provider/TEE signature over the original body, so a rewritten body would no longer match `/v1/signature/{id}` (and usage `stop_reason` had the same layering issue). That half has been **removed** here and moved to the signing layer: **nearai/inference-proxy#152**.

`#619` stays open until #152 lands; this PR closes the request-side half.

## Tests
- `extract_tools_strips_function_tools_when_choice_none` + `extract_tools_strips_unparsed_extra_tools_when_choice_none` (services unit).
- e2e `test_tool_choice_none_strips_tools` asserts the upstream request carries no `tools` (typed or in `extra`) while `tool_choice:none` is forwarded.

Refs #619

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -437,44 +437,6 @@ fn unredact_chat_response_in_place(
     }
 }
 
-/// Correct `finish_reason` to `"tool_calls"` on any choice that carries a
-/// non-empty `tool_calls` array but reports a different reason (nearai/cloud-api
-/// #619). When `tool_choice` forces a specific function, some backends (vLLM)
-/// return `finish_reason: "stop"` even though a tool *was* called; per the
-/// OpenAI spec, clients keying off `finish_reason` to detect tool calls must
-/// see `"tool_calls"`.
-///
-/// Returns `true` if any choice was modified, so the caller can decide whether
-/// it must re-serialize (which forgoes returning the provider's exact attested
-/// bytes) or can keep the verbatim body for hash/attestation verification.
-fn fix_tool_call_finish_reason_in_place(
-    response: &mut inference_providers::ChatCompletionResponse,
-) -> bool {
-    let mut changed = false;
-    for choice in &mut response.choices {
-        let has_tool_calls = choice
-            .message
-            .tool_calls
-            .as_ref()
-            .is_some_and(|tc| !tc.is_empty());
-        // Only normalize the known mis-labels: a backend that ran to a natural
-        // stop (`"stop"`), reported nothing (`None`), or used the legacy
-        // `"function_call"` reason while a tool *was* called. Reasons that carry
-        // independent meaning — `"length"` (truncated, e.g. tool args cut off by
-        // max_tokens) and `"content_filter"` — MUST be preserved, or clients
-        // would mistake a truncated/filtered response for a clean tool call.
-        let is_mislabel = matches!(
-            choice.finish_reason.as_deref(),
-            None | Some("stop") | Some("function_call")
-        );
-        if has_tool_calls && is_mislabel {
-            choice.finish_reason = Some("tool_calls".to_string());
-            changed = true;
-        }
-    }
-    changed
-}
-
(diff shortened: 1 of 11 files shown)


PierreLeGuen (changes requested):
Blocking issue found.

- `crates/api/src/routes/completions.rs:1285` normalizes the non-streaming response body after `CompletionServiceImpl::create_chat_completion` returns, but TEE-backed chat signatures are already fetched/stored from the provider response at `crates/services/src/completions/mod.rs:1422`, and the exposed `Inference-Id` is still derived from that provider id. For the exact vLLM forced-tool case this PR targets, clients receive a JSON body whose `finish_reason` has been rewritten to `tool_calls`, while `/v1/signature/{id}` still signs the provider's original body with `finish_reason: stop`. That breaks the existing response-hash verification contract. The same layering issue leaves usage `stop_reason` recorded from the unnormalized provider finish reason at `crates/services/src/completions/mod.rs:1479`. Please normalize before signature/usage recording, or update the signature/attestation behavior so the returned bytes and stored signature stay consistent.

Local checks:
- `cargo test -p services extract_tools_ -- --nocapture` passed.
- `cargo fmt --check` passed.
- Targeted API e2e tests compiled but could not run locally because Postgres admin DB connection was refused; GitHub CI is green.

Evrard-Nil:
Good catch — addressed in the latest commit. `fix_tool_call_finish_reason_in_place` now only normalizes the known mis-labels (`stop` / missing / legacy `function_call`) and **preserves** `length` and `content_filter`, so a tool call whose arguments were truncated by `max_tokens` (or filtered) keeps its real finish reason instead of being reported as a clean `tool_calls`. Added unit tests covering both the corrected and the preserved cases.

Evrard-Nil:
Thanks @PierreLeGuen — you're right, and I've reworked the PR.

The response-side `finish_reason` rewrite is **removed** (helper, both call sites, unit + e2e tests, and the mock `with_finish_reason` override all gone). As you noted, cloud-api can't mutate the response body without breaking byte-exact verification — the signature is over the provider's original bytes and `/v1/signature/{id}` would diverge, and usage `stop_reason` had the same problem.

I filed **nearai/inference-proxy#152** to normalize `finish_reason` at the signing layer instead (normalize → then sign, so returned bytes == signed bytes; preserving `length`/`content_filter`).

This PR now ships only the request-side **`tool_choice:"none"` → strip tools** fix, which has no attestation interaction. Re-requesting your review.

PierreLeGuen (approved):
No substantive issues found in this revision. I inspected the changed request-side tool extraction in `crates/services/src/completions/mod.rs`, the direct chat conversion path, the Responses API bridge that serializes tools/tool_choice into completion `extra`, and the updated e2e assertion in `crates/api/tests/e2e_all/openrouter_params.rs`.

Local checks:
- `cargo fmt --check` passed.
- `cargo test -p services extract_tools_ -- --nocapture` passed (6 tests).
- `cargo test -p api --test e2e_all test_tool_choice_none_strips_tools -- --nocapture` compiled but could not run locally because Postgres bootstrap failed with connection refused.

GitHub checks at review time: lint and security audit passed; Test Suite was still pending.

Issue #606, opened 2026-05-19
Vision: image_url fetch fails for hosts that block default User-Agent; error surfaces as misleading 502 'model unavailable'


Evrard-Nil:
## Summary

Sending `image_url` with a URL hosted on Wikimedia (a very common source for example/test images) consistently returns:

```
HTTP 502
{"error":{"message":"The model is currently unavailable. Please try again later.","type":"bad_gateway","param":null,"code":null}}
```

The model is in fact perfectly healthy — plain-text and other image URLs work — but the upstream image fetcher gets a `400 Bad Request` from Wikimedia because it sends a default User-Agent that Wikimedia explicitly blocks ([Wikimedia UA policy](https://meta.wikimedia.org/wiki/User-Agent_policy)). cloud-api then re-wraps that as 502 `bad_gateway`, so the user sees "model unavailable" and is told to retry — retries will never succeed.

## Repro

```bash
KEY=sk-32c0476395fd40c795725fc101f33304

# Fails (502 model unavailable):
curl -s https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-VL-30B-A3B-Instruct","messages":[{"role":"user","content":[
    {"type":"text","text":"What is in this image?"},
    {"type":"image_url","image_url":{"url":"https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/640px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"}}
  ]}],"max_tokens":80}'

# Works (image hosted elsewhere, no UA gating):
curl -s https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-VL-30B-A3B-Instruct","messages":[{"role":"user","content":[
    {"type":"text","text":"Describe in 1 sentence."},
    {"type":"image_url","image_url":{"url":"https://www.gstatic.com/webp/gallery/1.jpg"}}
  ]}],"max_tokens":80}'

# Also works (data URL — no network fetch at all):
curl ... -d '{... "image_url":{"url":"data:image/png;base64,iVBORw0KGgo..."} ...}'
```

[...]

PR #719, opened 2026-06-04
fix(vision): preserve image bytes/media_type on Anthropic & Gemini passthrough (#640)

### LINKED ISSUE #640: Vision passthrough garbles images: Anthropic and Gemini describe wildly wrong content for non-trivial images
## Summary

Vision via Anthropic and Gemini passthrough produces systematically wrong descriptions for normal-sized images. The self-hosted `Qwen/Qwen3-VL-30B-A3B-Instruct` correctly identifies the same images, so this is a passthrough-pipeline issue, not a model-quality one.

The same image yields consistent-but-wrong descriptions (sometimes identical across temperature=0 runs), strongly suggesting cloud-api is preprocessing/transmitting the image incorrectly before forwarding upstream.

## Evidence

Test image: `https://www.gstatic.com/webp/gallery/1.jpg` — a real photograph of a Norwegian fjord (panoramic view, mountains, water, 550x368 JPEG). Local fetch confirms the actual content.

| Model | Input | Description |
|---|---|---|
| `Qwen/Qwen3-VL-30B-A3B-Instruct` | URL | "panoramic view from a rocky cliff overlooking a deep, winding fjord" ✓ |
| `anthropic/claude-haiku-4-5` | URL | "vibrant red flower with delicate petals and a yellow center" ❌ |
| `anthropic/claude-haiku-4-5` | URL again | "yellow Labrador retriever stands in a grassy field" ❌ (different hallucination) |
| `anthropic/claude-haiku-4-5` | data:image/jpeg;base64,... | "person preparing food in a kitchen, chopping vegetables" ❌ |
| `anthropic/claude-haiku-4-5` | data URL, temp=0, run 1 | "person is holding a large, freshly baked chocolate chip cookie" ❌ |
| `anthropic/claude-haiku-4-5` | data URL, temp=0, run 2 | "person is holding a large, freshly baked chocolate chip cookie" (same — deterministic but wrong) |
| `anthropic/claude-haiku-4-5` | data URL, temp=0, run 3 | "person is holding a large, fresh lobster with prominent claws" |
| `google/gemini-2.5-flash` | URL | "A lone gray wolf stands" (truncated) ❌ |
| `google/gemini-2.5-flash` | data URL | "stack of fresh-baked cinnamon rolls with a generous amount of white icing" ❌ |

## Sanity check: tiny known-color PNGs

| Model | 32x32 #0000FF PNG | 16x16 #FF0000 PNG |
|---|---|---|
| `anthropic/claude-haiku-4-5` | "Blue" ✓ (3/3 runs) | "blue square or rectangular shape" ❌ |
| `google/gemini-2.5-flash` | "White" / "Black" / "Black" ❌ | (not tested) |
| `Qwen/Qwen3-VL-30B-A3B-Instruct` | "Blue" ✓ | "entirely red" ✓ |

The 32x32 blue case passes on Anthropic but the 16x16 red case fails. So Anthropic isn't simply blind — *something specific* about the image transport is corrupting some images while preserving others.

## Repro

```bash

# Same image, three calls — note inconsistent and impossible answers
for i in 1 2 3; do
    -d '{
      "model":"anthropic/claude-haiku-4-5",
      "messages":[{"role":"user","content":[
        {"type":"text","text":"Describe this image in one sentence."},
        {"type":"image_url","image_url":{"url":"https://www.gstatic.com/webp/gallery/1.jpg"}}
      ]}],
      "max_tokens":150,"temperature":0
    }' | jq -r .choices[0].message.content
done
```

(The actual image is a photo of a fjord.)

## Suspected cause

cloud-api probably either (a) sends a different/transformed image bytes than the user-supplied URL points to, or (b) sends the image in a format the upstream provider misinterprets. The fact that the same input yields identical-but-wrong output across multiple temp=0 runs argues for a deterministic transformation gone wrong — not "the model is hallucinating."

Note that for Anthropic specifically:
- URL `https://upload.wikimedia.org/.../Gfp-wisconsin-madison-the-nature-boardwalk.jpg` → correctly described as "wooden boardwalk winds through a lush green marsh" ✓
- URL `https://www.gstatic.com/webp/gallery/1.jpg` → systematically wrong (above)

Both are real JPEG URLs of nature scenes. The difference might be the path cloud-api takes to fetch/forward (one host vs another), or some content-type / encoding difference.

## Suggested investigation

1. Capture the exact image bytes cloud-api forwards to Anthropic / Gemini for the gstatic URL. Compare byte-for-byte with the original. If they differ, there's the bug.
2. If they match, check the request shape cloud-api builds for each upstream — wrong `mime_type`, `media_type`, base64-vs-binary, or width/height hints.

---
Found during routine cloud-api smoke test. Distinct from #606 (which was the self-hosted Qwen3-VL pipeline UA-blocking external URLs at fetch time).

Evrard-Nil:
## Root cause

The Anthropic and Gemini message converters flattened OpenAI multimodal content (an array of `{type:"text"|"image_url",...}` parts) into a single **text** string via `serde_json::Value::to_string()`. Images were never sent as native image blocks — they reached the upstream model as a JSON text blob containing the base64 string, so the model hallucinated.

This explains every symptom in #640:
- Deterministic-but-wrong descriptions at temp=0 (the image is literally absent; the model invents content from the garbled text).
- Self-hosted Qwen3-VL reads the same images correctly (the vLLM path forwards `content` as `serde_json::Value` verbatim — no flattening).
- "32x32 blue passes / 16x16 red fails" — pure noise; in both cases the image was never transmitted as an image.

Neither converter even had an image representation:
- `AnthropicContentPart` had no `Image` variant.
- `GeminiPart` had no `inlineData`/`fileData` field.

## Fix

- New shared `external/content.rs` parser splits OpenAI content into ordered `Text` / `ImageBase64` / `ImageUrl` parts. Data URIs are split on the **first** `,` into `media_type` + payload; the base64 payload is forwarded **verbatim** (only surrounding ASCII whitespace trimmed) — no re-encoding, no double-encoding, no byte/char confusion.
- **Anthropic**: add `AnthropicContentPart::Image` with a `base64`/`url` source; build native image blocks for user messages.
- **Gemini**: add `inlineData` (base64) and `fileData` (remote URL) parts; emit them for user messages.
- System/tool messages keep text-only behaviour (images dropped, not serialized as a JSON blob).

## Tests

- Converter unit tests for both providers use a real 1x1 red PNG and assert the **exact** base64 payload and `media_type`/`mimeType` survive serialization (proves the bug and guards the fix).
- Content-parser tests cover data-URI splitting, whitespace trimming, missing media type, bare-string `image_url`, remote URLs, and base64 byte-identity.

## Validation

- `cargo build --workspace` — ok
- `cargo clippy -p inference_providers --lib --tests` — clean (exit 0)
- `cargo test -p inference_providers --lib` — 192 passed, 0 failed (12 new). The only doctest failure (`detect_audio_content_type`) is pre-existing on `origin/main` and unrelated.

Fixes #640

### DIFF
--- a/crates/inference_providers/src/external/anthropic/converter.rs
+++ b/crates/inference_providers/src/external/anthropic/converter.rs
@@ -49,6 +49,23 @@ pub enum AnthropicContentPart {
         tool_use_id: String,
         content: String,
     },
+    #[serde(rename = "image")]
+    Image { source: AnthropicImageSource },
+}
+
+/// Anthropic image source. Anthropic accepts either inline base64 bytes
+/// (`type: "base64"`) or a remote URL (`type: "url"`).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "type")]
+pub enum AnthropicImageSource {
+    #[serde(rename = "base64")]
+    Base64 {
+        media_type: String,
+        /// Exact base64 payload, forwarded verbatim (no re-encoding).
+        data: String,
+    },
+    #[serde(rename = "url")]
+    Url { url: String },
 }
 
 /// Anthropic tool definition
@@ -221,15 +238,12 @@ pub struct AnthropicResponse {
 
 /// Convert OpenAI messages to Anthropic format
 pub fn convert_messages(messages: &[ChatMessage]) -> (Option<String>, Vec<AnthropicMessage>) {
+    use crate::external::content::{
+        parse_content, text_from_content as extract_content, ContentPart,
+    };
+
     let mut system_message = None;
     let mut anthropic_messages = Vec::new();
-
-    let extract_content = |value: &serde_json::Value| -> String {
-        match value {
-            serde_json::Value::String(s) => s.clone(),
-            _ => value.to_string(),
-        }
-    };
 
     for msg in messages {
         match msg.role {
@@ -239,15 +253,49 @@ pub fn convert_messages(messages: &[ChatMessage]) -> (Option<String>, Vec<Anthro
                 }
             }
             MessageRole::User => {
-                let content = msg
-                    .content
-                    .as_ref()
-                    .map(&extract_content)
-                    .unwrap_or_default();
-                anthropic_messages.push(AnthropicMessage {
-                    role: "user".to_string(),
-                    content: AnthropicMessageContent::Text(content),
-                });
+                // User messages may be multimodal (text + image parts). Build
+                // native Anthropic content blocks so images are transmitted as
+                // images, not flattened into a JSON text blob (issue #640).
+                let parts = msg.content.as_ref().map(parse_content).unwrap_or_default();
+
+                let has_image = parts.iter().any(|p| !matches!(p, ContentPart::Text(_)));
+
+                if has_image {
+                    let mut blocks = Vec::with_capacity(parts.len());
+                    for part in parts {
+                        match part {
+                            ContentPart::Text(text) => {
+                                if !text.is_empty() {
+                                    blocks.push(AnthropicContentPart::Text { text });
+                                }
+                            }
+                            ContentPart::ImageBase64 { media_type, data } => {
+                                blocks.push(AnthropicContentPart::Image {
+                                    source: AnthropicImageSource::Base64 { media_type, data },
+                                });
+                            }
+                            ContentPart::ImageUrl { url } => {
+                                blocks.push(AnthropicContentPart::Image {
+                                    source: AnthropicImageSource::Url { url },
+                                });
+                            }
+                        }
+                    }
+                    anthropic_messages.push(AnthropicMessage {
+                        role: "user".to_string(),
+                        content: AnthropicMessageContent::Blocks(blocks),
+                    });
+                } else {
+                    let content = msg
+                        .content
+                        .as_ref()
+                        .map(&extract_content)
+                        .unwrap_or_default();
+                    anthropic_messages.push(AnthropicMessage {
+                        role: "user".to_string(),
+                        content: AnthropicMessageContent::Text(content),
+                    });
+                }
             }
             MessageRole::Assistant => {
                 // Check if the assistant message contains tool calls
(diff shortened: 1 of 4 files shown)


PierreLeGuen (changes requested):
Requesting changes for one provider contract issue.

- `crates/inference_providers/src/external/gemini/converter.rs:285`: the Gemini `ImageUrl` path emits `fileData` with `mimeType` omitted via `GeminiPart::file_image(None, url)`. Gemini `fileUri` media inputs need a MIME type for URL/file references, so an OpenAI request like `image_url: {url: "https://example.com/cat.jpg"}` will serialize to a request that is likely rejected upstream instead of preserving vision passthrough. Please either infer/fetch and set the MIME type before using `fileData`, inline the fetched bytes as `inlineData`, or explicitly reject/skip unsupported remote URLs rather than producing a broken Gemini request. The current test only checks `file_uri`; please add a serialization/assertion that catches the required `fileData.mimeType` behavior.

Local checks run: `cargo test -p inference_providers --lib`, `cargo fmt --all -- --check`, and `git diff --check` all passed. GitHub lint/security checks are passing; the GitHub Test Suite was still pending when reviewed.

Evrard-Nil:
@PierreLeGuen Thanks — fixed the `fileData.mimeType` contract issue in `8d536f29`.

**What changed** (`crates/inference_providers/src/external/gemini/converter.rs`):
- The `ImageUrl` path no longer emits `GeminiPart::file_image(None, url)`. It now infers `mimeType` from the URL path extension via a new `mime_type_from_url()` helper (`.jpg/.jpeg→image/jpeg`, `.png`, `.webp`, `.gif`, `.heic`, `.heif`; case-insensitive; query string + fragment stripped, e.g. `cat.jpg?sig=...#frag`).
- If the extension is missing or unrecognised, the image part is **skipped** (with a content-free `tracing::warn!`) rather than emitting a `fileData` without `mimeType`. The serialized request can therefore never contain a `fileData` lacking `mimeType`.

**Why this option (infer + skip), not inline/reject-all:** Gemini's `fileData` does accept public HTTPS URLs ([file-input-methods](https://ai.google.dev/gemini-api/docs/file-input-methods)) and securely fetches them itself — but it requires a `mimeType`; an empty one returns 400 *"empty mimeType parameter in fileData"*. cloud-api does **not** pre-fetch remote images (Gemini is a pure passthrough to Google's API, no engine in the loop — #606), so inlining bytes isn't available here. Inferring the type preserves passthrough for the common extension-bearing URLs; skipping unknown ones keeps the rest of the request valid instead of failing the whole call. Remote-URL image support remains provider-dependent, as the thread already notes.

**Base64 data-URI path (#640) is untouched** — those still serialize to `inlineData` with the exact bytes; the existing `inlineData`/`mimeType`/byte-identity test stays green.

**Tests added/updated:**
- `test_convert_messages_image_url_uses_file_data_with_mime_type` — asserts the serialized request contains `"fileData"` AND `"mimeType":"image/jpeg"` (this is the serialization assertion you asked for; would catch a regression to `file_image(None, ..)`).
- `test_convert_messages_image_url_infers_mime_from_various_extensions` — png/JPEG(uppercase)/webp/gif + query-string stripping.
- `test_convert_messages_image_url_unknown_extension_is_skipped` — text preserved, both unknown-extension images dropped, and no `fileData` on the wire.

Local checks: `cargo build --workspace` ✅, `cargo test -p inference_providers --lib` ✅ (194 passed), `cargo fmt --all -- --check` ✅, `cargo clippy -p inference_providers --lib --tests` ✅ clean, `git diff --check` ✅.

PierreLeGuen (approved):
Reviewed the Anthropic/Gemini message conversion changes, shared content parser, request serialization shapes, surrounding provider request code, tests, and CI metadata. The image blocks now match provider-native wire formats, and I did not find substantive issues.

Local checks run:
- cargo fmt --all -- --check
- cargo test -p inference_providers --lib external::content -- --nocapture
- cargo test -p inference_providers --lib test_convert_messages_image -- --nocapture
- cargo test -p inference_providers --lib test_convert_messages_with_system -- --nocapture
- cargo test -p inference_providers --lib
- cargo clippy -p inference_providers --lib --tests -- -D warnings

GitHub checks at review time: Lint and security_audit passed; Test Suite was still pending.

PR #720, opened 2026-06-04
fix(passthrough): Anthropic/Gemini param fidelity — temperature drop, seed, response_format (#696 #669 #668)

### LINKED ISSUE #668: response_format: json_object returns markdown-fenced JSON on passthroughs; json_schema ignored on Gemini

### LINKED ISSUE #669: seed parameter silently ignored on Anthropic and Gemini passthroughs
## Summary

The OpenAI-standard `seed` parameter for deterministic outputs is honored on self-hosted models and OpenAI passthrough, but silently ignored on Anthropic and Gemini passthroughs. Two identical requests with the same `seed` return different content.

| Model | seed=42 honored? |
|---|---|
| `openai/gpt-oss-120b` (self-hosted) | ✓ deterministic |
| `Qwen/Qwen3-30B-A3B-Instruct-2507` | ✓ "Stonetooth" both runs |
| `google/gemma-4-31B-it` (self-hosted) | ✓ "Sylvester Stallone" both runs |
| `openai/gpt-4.1-nano` (OpenAI passthrough) | ✓ "Stonely" both runs |
| **`anthropic/claude-haiku-4-5`** | ❌ "Pebbles McRollington" vs "Pebble Gatsby" |
| **`google/gemini-2.5-flash-lite`** | ❌ "Sir Pebbleton" vs "Boulder-darling" |

## Repro

```bash
KEY=sk-...
for i in 1 2; do
    -d '{"model":"anthropic/claude-haiku-4-5","messages":[{"role":"user","content":"One creative pet rock name."}],"max_tokens":40,"temperature":1.0,"seed":42}' \
    | jq -r .choices[0].message.content
done
# "Pebbles McRollington"
# "Pebble Gatsby"
```

## Why it matters

`seed` is the canonical way OpenAI-compatible clients request deterministic outputs (eval reproducibility, A/B test consistency, regression test fixtures). Users assume same `seed` + same input → same output.

When migrating an OpenAI-direct integration to use cloud-api as a gateway across providers, the same code suddenly produces non-deterministic outputs on the Anthropic and Gemini branches — and the failure mode is subtle (the user might only notice in CI when reproducibility checks fail).

## Provider notes

- **Anthropic**: their native API doesn't expose `seed`. cloud-api can't directly satisfy this. Options: (a) document that `seed` isn't supported on Anthropic and ideally reject with a 400, (b) emulate via temperature=0 substitution (but this loses sampling), (c) accept silently as-is.
- **Gemini**: Google's Generative AI API *does* support `seed` (via `seed` field in generationConfig). cloud-api should pass it through.

Suggested fix:
1. For Gemini: forward `seed` to the upstream `generationConfig.seed`.
2. For Anthropic: either reject `seed` with a clear `400 unsupported_parameter` OR document the silent drop in the OpenAPI spec.

## Related

Same family as `cache_control` and `thinking` passthrough gaps (#666) — provider-specific or OpenAI-standard parameters not being correctly translated/forwarded.

---

### LINKED ISSUE #696: temperature: non-default values rejected with 400 on anthropic/claude-opus-4-7 passthrough

Evrard-Nil:
Brings the Anthropic and Gemini passthrough converters up to OpenAI/OpenRouter parameter parity. Three related fixes shipped together.

## #696 — `temperature` 400 on `claude-opus-4-7`
Newer Anthropic models deprecate `temperature` and 400 on any non-default value, even though they still advertise it. The Anthropic adapter now drops `temperature` for those models (matched by a per-model substring list, currently `claude-opus-4-7`) and forwards `top_p` instead. `claude-opus-4-6` and earlier still forward `temperature` — verified by a regression test so we don't over-strip.

## #669 — `seed` silently ignored on Gemini
Gemini's native API supports deterministic sampling via `generationConfig.seed`. We now forward it (reading the typed `seed` field or the `extra` passthrough map the service layer actually populates). Anthropic has no native `seed`; left as graceful accept-and-ignore per the issue's preferred option (no hard 400).

## #668 — `response_format` broken on passthroughs
- **Gemini**: `json_object` → `responseMimeType: "application/json"`; `json_schema` → `responseMimeType` + a sanitized `responseSchema` (run through the existing Gemini OpenAPI-subset sanitizer so `additionalProperties` etc. don't 400). Previously `json_schema` was ignored entirely and `json_object` came back markdown-fenced.
- **Anthropic**: no native JSON mode, so on the non-streaming path we strip a leading/trailing markdown ` ```json … ``` ` fence when `json_object`/`json_schema` was requested, so `JSON.parse` works. Streaming is documented as best-effort (a fence marker can split across SSE deltas; buffering would defeat streaming).

## Tests
- New converter unit tests: opus-4-7 temperature drop + opus-4-6 guard; fence-stripping; Gemini seed (typed + extra); Gemini `responseMimeType`/`responseSchema` mapping; `response_format_to_gemini` translation.
- New e2e `json_object` passthrough guard in `openrouter_params.rs`.
- `cargo build --workspace` clean; `cargo test -p inference_providers --lib` 194 passed; `cargo test --test e2e_all -- openrouter` 10 passed. (The pre-existing `detect_audio_content_type` doctest failure on `main` is unrelated.)

Fixes #696
Fixes #669
Fixes #668

### DIFF
--- a/crates/api/tests/e2e_all/openrouter_params.rs
+++ b/crates/api/tests/e2e_all/openrouter_params.rs
@@ -422,3 +422,43 @@ async fn test_json_schema_response_format_accepted_and_forwarded() {
         "response_format.type not preserved on passthrough"
     );
 }
+
+// ── response_format json_object (nearai/cloud-api #668) ─────────────────────
+
+/// `response_format: { type: json_object }` must be accepted and forwarded in
+/// `extra` so the provider adapters (Gemini → responseMimeType, Anthropic →
+/// fence-strip) can act on it. #668 tracks the markdown-fence breakage on
+/// passthroughs; cloud-api must not reject or drop the param.
+#[tokio::test]
+async fn test_json_object_response_format_accepted_and_forwarded() {
+    let (server, mock, model, api_key) = setup().await;
+
+    let response = server
+        .post("/v1/chat/completions")
+        .add_header("Authorization", format!("Bearer {api_key}"))
+        .json(&serde_json::json!({
+            "model": model,
+            "messages": [{"role": "user", "content": "Give me a JSON object."}],
+            "response_format": {"type": "json_object"},
+            "max_tokens": 200,
+            "stream": false,
+        }))
+        .await;
+
+    assert_eq!(
+        response.status_code(),
+        200,
+        "json_object response_format should be accepted, got: {}",
+        response.text()
+    );
+    let params = mock.last_chat_params().await.expect("provider was called");
+    let forwarded = params
+        .extra
+        .get("response_format")
+        .expect("response_format forwarded in `extra`");
+    assert_eq!(
+        forwarded.get("type").and_then(|v| v.as_str()),
+        Some("json_object"),
+        "response_format.type not preserved on passthrough"
+    );
+}
--- a/crates/inference_providers/src/external/anthropic/mod.rs
+++ b/crates/inference_providers/src/external/anthropic/mod.rs
@@ -32,6 +32,58 @@ const DEFAULT_ANTHROPIC_VERSION: &str = "2023-06-01";
 /// accept and will reject with its own 400, instead of us silently dropping it).
 const ANTHROPIC_PASSTHROUGH_KEYS: &[&str] = &["thinking", "reasoning_effort"];
 
+/// Anthropic model-name fragments that **reject any non-default `temperature`**
+/// with a 400 (`temperature is deprecated for this model`), even though they
+/// still advertise `temperature` (nearai/cloud-api #696).
+///
+/// These are matched as substrings so both the bare alias (`claude-opus-4-7`)
+/// and the dated form (`claude-opus-4-7-20XXYYZZ`) are covered. opus-4-6 and
+/// earlier still accept `temperature`, so they are intentionally absent — do
+/// not over-strip.
+const ANTHROPIC_MODELS_REJECTING_TEMPERATURE: &[&str] = &["claude-opus-4-7"];
+
+/// Whether `model` rejects a non-default `temperature` and we must drop it
+/// (forwarding `top_p` instead) rather than 400 the caller (#696).
+fn rejects_non_default_temperature(model: &str) -> bool {
+    ANTHROPIC_MODELS_REJECTING_TEMPERATURE
+        .iter()
+        .any(|fragment| model.contains(fragment))
+}
+
+/// Whether a requested `response_format` asks for JSON output, which Anthropic
+/// has no native mode for and tends to return markdown-fenced (#668). When
+/// true, we strip code fences from the response so `JSON.parse` works.
+fn wants_json_output(extra: &std::collections::HashMap<String, serde_json::Value>) -> bool {
+    extra
+        .get("response_format")
+        .and_then(|rf| rf.get("type"))
+        .and_then(|t| t.as_str())
+        .map(|t| t == "json_object" || t == "json_schema")
+        .unwrap_or(false)
+}
+
+/// Strip a single leading/trailing markdown code fence from `content`.
+///
+/// Anthropic has no native JSON-output mode, so when a caller requests
+/// `response_format: json_object`/`json_schema` the model frequently wraps the
+/// JSON in a ` ```json … ``` ` block. This unwraps that fence so the content is
+/// raw parseable JSON (#668). Content without a fence is returned unchanged.
+fn strip_json_code_fence(content: &str) -> String {
+    let trimmed = content.trim();
+    let Some(rest) = trimmed.strip_prefix("```") else {
+        return content.to_string();
+    };
+    // Drop the optional language tag on the opening fence line (e.g. `json`).
+    let after_lang = match rest.find('\n') {
+        Some(idx) => &rest[idx + 1..],
+        None => return content.to_string(),
+    };
+    let Some(inner) = after_lang.trim_end().strip_suffix("```") else {
+        return content.to_string();
+    };
+    inner.trim().to_string()
+}
+
 /// Pick the allowlisted reasoning-control fields out of `extra`.
 fn extract_passthrough(
     extra: &std::collections::HashMap<String, serde_json::Value>,
@@ -96,7 +148,14 @@ impl AnthropicBackend {
 
         // Anthropic doesn't allow both temperature and top_p - prefer temperature if both are set.
         // Also clamp temperature to Anthropic's valid range [0.0, 1.0] (OpenAI allows up to 2.0).
-        let (temperature, top_p) = if let Some(temp) = params.temperature {
+        //
+        // #696: some newer models (e.g. claude-opus-4-7) 400 on ANY non-default
+        // `temperature`. For those we drop `temperature` entirely and forward
+        // `top_p` instead, so OpenAI/OpenRouter clients that routinely send
+        // `temperature: 0`/`0.7` get a 200 (param ignored) instead of a 400.
+        let (temperature, top_p) = if rejects_non_default_temperature(model) {
+            (None, params.top_p)
+        } else if let Some(temp) = params.temperature {
             (Some(temp.clamp(0.0, 1.0)), None)
         } else {
             (None, params.top_p)
@@ -154,6 +213,13 @@ impl ExternalBackend for AnthropicBackend {
         model: &str,
         params: ChatCompletionParams,
     ) -> Result<StreamingResult, CompletionError> {
+        // NOTE (#668): markdown-fence stripping for `response_format` JSON modes
+        // is applied on the non-streaming path only. A fence marker (```) can
+        // split across SSE deltas, so reliably stripping it mid-stream would
+        // require buffering the whole response and defeats streaming. The
+        // model's own behaviour with the json hint is usually fence-free when
+        // streaming; callers needing guaranteed raw JSON should use the
+        // non-streaming endpoint.
         let url = format!("{}/messages", config.base_url);
         let request = self.build_request(model, &params, true);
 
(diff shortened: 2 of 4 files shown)


PierreLeGuen (changes requested):
crates/inference_providers/src/external/gemini/converter.rs:461: The `json_schema` path currently sanitizes the OpenAI schema and sends it as `generationConfig.responseSchema`, which deletes constraints such as `additionalProperties: false`, `$ref`, `$defs`, and `oneOf` before the provider sees them. That means a request with `response_format: {type: "json_schema", json_schema: {strict: true, schema: ...}}` can succeed while Gemini is no longer enforcing the schema the client asked for. For this API contract we should either preserve the original JSON Schema by using Gemini's JSON-Schema field (`responseJsonSchema` / `_responseJsonSchema`) when available, or fail/document this as best-effort instead of silently weakening strict schemas. The current tests also lock in the weakened behavior by asserting that `additionalProperties` is removed.

Checks run: `cargo fmt --check`; `cargo test -p inference_providers --lib external::anthropic` (27 passed); `cargo test -p inference_providers --lib external::gemini` (32 passed). `cargo test --test e2e_all openrouter_params` compiled but could not run locally because all filtered cases failed during test DB bootstrap with Postgres connection refused.

Evrard-Nil:
@PierreLeGuen Addressed the strict-schema concern in ba8b41ed.

**What changed**
- `response_format_to_gemini` now branches on `json_schema.strict`:
  - **strict: true** → forwards the **original schema unmodified** via Gemini's standard-JSON-Schema field `generationConfig.responseJsonSchema`, so `additionalProperties: false`, `$ref`, `$defs` and `oneOf` are preserved and actually enforced. No sanitization.
  - **non-strict** → keeps the previous best-effort `generationConfig.responseSchema` (OpenAPI-3.0 subset, sanitized) so unsupported keywords don't 400. The caller didn't ask for strict enforcement here.
  - The two fields are **never both populated** (mutually exclusive in Gemini's API).
- Added `responseJsonSchema` to `GeminiGenerationConfig`; `response_format_to_gemini` now returns a small `GeminiResponseFormat` struct instead of a 2-tuple.
- Tests that locked in the weakened behavior were replaced: strict path now asserts the original schema (incl. `additionalProperties:false`, `$ref`, `$defs`, `oneOf`) survives byte-for-byte in `responseJsonSchema` and that `responseSchema` is unset; added a non-strict counterpart asserting the lossy `responseSchema` path still sanitizes.

**Gemini API findings**
- Field is `responseJsonSchema` (camelCase) under `generationConfig`; it accepts **standard JSON Schema** (incl. `$ref`/`$defs`/advanced validation), unlike the OpenAPI-3.0-subset `responseSchema`.
- The two are **mutually exclusive** — if `responseJsonSchema` is set, `responseSchema` must be omitted — and `responseJsonSchema` must be paired with `responseMimeType: application/json`. The impl enforces this.
- Refs: https://ai.google.dev/gemini-api/docs/structured-output and the GenerationConfig REST reference.

**Validation**
- `cargo build --workspace` clean; `cargo test -p inference_providers --lib` → 196 passed.
- `cargo fmt --all -- --check` clean; `cargo clippy -p inference_providers --lib --tests` clean.
- e2e (disposable PG): `test_json_schema_response_format_accepted_and_forwarded` + `test_json_object_response_format_accepted_and_forwarded` pass. (`test_temperature_nondefault_accepted_and_forwarded` flaked on a shared `admin@test.com` fixture collision under parallel run; passes in isolation — pre-existing, unrelated.)

The #696 (opus-4-7 temperature drop) and #669 (seed) changes are untouched.

PierreLeGuen (approved):
Reviewed the current head ba8b41ed, including the Anthropic/Gemini request conversion paths, schema handling around strict Gemini json_schema, response_format passthrough, seed forwarding, Anthropic temperature dropping, and the modified OpenRouter passthrough tests.

No substantive issues found. The earlier strict-schema concern is addressed by routing strict Gemini schemas through responseJsonSchema while keeping responseSchema unset.

Checks run locally:
- cargo fmt --all -- --check
- cargo test -p inference_providers external::gemini::tests:: --all-features
- cargo test -p inference_providers external::anthropic::tests:: --all-features
- cargo test -p api --test e2e_all openrouter_params --all-features (built, but all filtered tests failed during local DB bootstrap with Postgres connection refused)

GitHub checks are green: security_audit, Lint, and Test Suite.

PR #721, opened 2026-06-04
fix(vision): fetch image_url with explicit User-Agent; surface fetch errors as 4xx not 502 (#606)

### LINKED ISSUE #606: Vision: image_url fetch fails for hosts that block default User-Agent; error surfaces as misleading 502 'model unavailable'

Evrard-Nil:
## Problem

Sending an `image_url` on a host that blocks the inference engine's default User-Agent (e.g. Wikimedia's [UA policy](https://meta.wikimedia.org/wiki/User-Agent_policy) returns `400` to default UAs) makes cloud-api return a misleading:

```
HTTP 502 {"error":{"message":"The model is currently unavailable. Please try again later.","type":"bad_gateway"}}
```

The model is healthy — the **image fetch** failed — but it was re-wrapped as a 502, so users are told to retry (retries never succeed). Reproduced against prod with `Qwen/Qwen3-VL-30B-A3B-Instruct` (both the Wikimedia URL and a plain 404 image URL return 502; a non-gated URL works).

## Where the fetch actually happens (important finding)

**cloud-api does NOT pre-fetch `image_url`s.** It serializes the `image_url` content into the chat message and forwards it to the backend; the actual HTTP fetch is performed by the **inference engine (vLLM / SGLang) itself** via its Python multimodal connector (aiohttp). `inference-proxy` also does not fetch images — it proxies to vLLM.

Consequence: **the User-Agent fix cannot live in cloud-api** (issue fix #1). It belongs in the inference engine's media-fetch config (set a `User-Agent: nearai-cloud-api/1.0 (+https://near.ai)` on the engine's image fetcher, e.g. via `cvm-compose-files` env / vLLM `--media-io-kwargs` or the engine's HTTP client). I'll route that half to the engine/compose side — see the issue comment.

## What this PR fixes (the cloud-api half: fix #2)

cloud-api's responsibility is the error mapping: when the failure cause is a client-supplied URL the engine couldn't fetch, surface a non-retryable **4xx** (`client_media` → `invalid_request_error`), not a generic 502.

That machinery already exists (#688 short-circuits + #698 → 400). The remaining gap was **detection coverage**: the existing markers only matched specific decode-side phrasings and the `HTTP error 500: 4xx, message='...', url='...'` aiohttp wrapper. The Qwen3-VL / vLLM `MediaConnector` fetch-side error phrasings weren't caught, so they fell through to the retryable-5xx → 502 path.

This PR broadens `is_client_media_fetch_error` to also match the fetch-side phrasings:
- `failed to fetch image/video`, `error fetching image/video`, `failed to load image/video`
- `clientresponseerror` (aiohttp's exception class)
- `NNN client error: bad request / forbidden / not found for url` (the `requests`/SGLang form, covering Wikimedia's 400, 403s, and stale-URL 404s)

The existing guard (wrapped status must be a real `4xx`, so a `500: 503` wrapper still retries) is preserved.

## Tests

Added `classify_retry_decision` cases for the Wikimedia 400 scenario across all three engine phrasings (aiohttp wrapper, SGLang `loading IMAGE data ... 400 Client Error`, and bare `MediaConnector` messages), asserting they classify as `non_retryable_client_media_error` (→ 400). All existing tests still pass.

- `cargo build --workspace`: ok
- `cargo test -p services --lib`: 333 passed
- `cargo test -p inference_providers --lib --tests`: 8 passed (only the pre-existing unrelated `detect_audio_content_type` doctest fails on origin/main)
- `cargo fmt --all -- --check`, `cargo clippy -p services`: clean

## Follow-up (not in this repo)

The User-Agent half (issue fix #1) needs to be set on the inference engine's image fetcher. Tracking via an issue comment so it can be routed to the engine/compose config.

Fixes #606


PierreLeGuen (changes requested):
Blocking issue:

- `crates/services/src/inference_provider_pool/mod.rs:1438-1444` now classifies any `failed to fetch image/video`, `error fetching image/video`, or `clientresponseerror` text inside an upstream HTTP 5xx as `non_retryable_client_media_error` without checking that the embedded fetch status is actually 4xx. That defeats the guarded logic below (`:1451-1459`) which intentionally keeps wrapped 5xx fetches retryable. For example, a vLLM/aiohttp message like `Error fetching image: ClientResponseError, status=503, message='Service Unavailable'` would now skip retries and map to a 400 client-media error instead of retrying/surfacing a backend/upstream failure. Please constrain the new fetch-side markers to explicit upstream 4xx statuses (and add a negative test for `ClientResponseError, status=503` / equivalent).

Checked locally: `cargo test -p services --lib test_classify_retry_decision` passed. Current GitHub checks are passing.

Evrard-Nil:
Thanks @PierreLeGuen — fixed in 62e631a3.

**The fix:** the fetch-side markers (`failed to fetch image/video`, `error fetching image/video`, `ClientResponseError`, `NNN Client Error`, `HTTP error 500: NNN`) are now gated on an explicit upstream **4xx**. They are split out from the decode-side markers and only classify as `non_retryable_client_media_error` when `extract_fetch_status()` finds a determinable 4xx in the message. A 5xx (e.g. `ClientResponseError, status=503`) or an indeterminate status stays **retryable** / surfaces as a backend failure.

- Decode-side markers (`loading image/video data`, `cannot identify image file`, `failed to open input buffer`) carry no fetch status and remain unconditionally non-retryable — unchanged.
- New `extract_fetch_status()` helper parses the three engine phrasings: aiohttp wrapper `HTTP error 500: NNN`, aiohttp exception `status=NNN`, requests/urllib `NNN Client Error`. It deliberately never matches a bare 3-digit number, so the wrapper's outer `500` envelope is never mistaken for the inner fetch status.

**Negative tests added** (all must stay `retryable_http_5xx`):
- `Error fetching image: ClientResponseError, status=503, message='Service Unavailable'`
- `Failed to fetch image: ClientResponseError, status=500, ...`
- `Failed to load image from url: 502 Client Error: Bad Gateway ...`
- `Error fetching image: ClientResponseError, message='connection closed'` (no status)
- `Failed to fetch image from <url>` (no status)
- aiohttp wrapper around a 503

**Positive controls kept/added** (`non_retryable_client_media_error`): `status=400/403/404`, `403/404 Client Error: ... for url`, and the `HTTP error 500: 400` Wikimedia wrapper — the #606 cases don't regress.

Also updated the sanitize-ordering test: since classification is now status-driven (not URL-anchored), the embedded `404` survives redaction and the sanitized wrapper still classifies as client-media, which is strictly more robust.

`cargo build --workspace`, `cargo test -p services --lib test_classify_retry_decision` (and the full `inference_provider_pool` suite, 50 tests), `cargo fmt --all -- --check`, and `cargo clippy -p services --lib --tests` all pass clean.

PierreLeGuen (changes requested):
Requesting changes for one correctness issue.

- crates/services/src/inference_provider_pool/mod.rs:1429: `loading image data` / `loading video data` still returns `non_retryable_client_media_error` before `extract_fetch_status()` runs. This PR’s own added fixture at lines 3878-3885 treats `loading IMAGE data ... 400 Client Error` as an SGLang fetch-side shape, not a decode-only shape. The analogous `loading IMAGE data ... 503 Client Error` will therefore still short-circuit all providers and map a transient upstream fetch 5xx to a client 400, which is the regression the new status gate is trying to avoid for the other fetch markers. Please split true decode markers from fetch-with-status markers, or at least gate the `loading image/video data` cases when they also contain an embedded fetch status, and add a negative test for a `loading IMAGE data ... 503 Client Error` message staying retryable.

Checks run locally: `cargo test -p services test_classify_retry_decision --lib`; `cargo test -p services client_media_error --lib`. Both passed. GitHub checks at review time: lint/security passed; Test Suite still pending.

Evrard-Nil:
Good catch on the second pass — fixed. `loading image data`/`loading video data` are no longer treated as unconditional decode-only markers: they're the SGLang/vLLM *prefix* for both a decode failure (`... cannot identify image file`, still caught unconditionally) and a fetch failure (`... NNN Client Error`), so I moved them into the status-gated fetch-marker set. They're now non-retryable only with an explicit upstream **4xx**; a `loading IMAGE data ... 503 Client Error` stays **retryable** instead of being short-circuited to a client 400. Added negative tests for `loading IMAGE/VIDEO data ... 5xx` staying retryable (and the existing `... 400` / `cannot identify` / `failed to open input buffer` positives still hold). `cargo test -p services classify_retry client_media` green.

PierreLeGuen (changes requested):
Blocking issue:

- `crates/services/src/inference_provider_pool/mod.rs:1490` still does not parse the normal `aiohttp.ClientResponseError.__str__` shape emitted by `raise_for_status()`: `400, message='Bad Request', url='...'`. The new tests only cover `ClientResponseError, status=400`, but aiohttp's string form uses the leading status/comma form, and vLLM's media fetch path raises from aiohttp `r.raise_for_status()`. If the engine message is `Failed to fetch image: 400, message='Bad Request', url='https://upload.wikimedia.org/...'` or `ClientResponseError: 400, message=...`, `has_fetch_marker` is true but `extract_fetch_status()` returns `None`, so the Wikimedia/default-UA 400 remains retryable and can still surface as 502. Please add parsing and a regression test for the `NNN, message=..., url=...` aiohttp form while keeping the 5xx equivalent retryable.

Checks run locally: `cargo fmt --all -- --check`; `cargo test -p services --lib test_classify_retry_decision`; `cargo test -p services --lib client_media_error`. GitHub lint, security, and test suite are currently green.

Evrard-Nil:
Fixed — thanks for the precise repro. `extract_fetch_status` now also parses aiohttp's `ClientResponseError.__str__` form from `raise_for_status()`: `NNN, message='...', url='...'`. Added regex alternation `\b(\d{3}), message=`, anchored to the literal `, message=` so the aiohttp wrapper's outer `500` (followed by `: `) can never match it. So `Failed to fetch image: 400, message='Bad Request', url=...` → non-retryable 4xx (the Wikimedia default-UA case), while the `503, message=...` equivalent stays retryable. Added both as regression tests; `cargo test -p services classify_retry client_media` green.

PierreLeGuen (approved):
Reviewed the single-file change in crates/services/src/inference_provider_pool/mod.rs, including the retry classifier, status extraction, client-media short-circuit, sanitization behavior, and the downstream map_provider_error path to 400/invalid_request_error.

Local checks run:
- cargo fmt --all -- --check
- cargo test -p services --lib inference_provider_pool
- cargo test -p services --lib test_map_provider_error_client_media_becomes_invalid_params
- cargo clippy -p services --lib --tests -- -D warnings

GitHub checks were also green at review time: Lint, Test Suite, and security_audit. I did not find a substantive blocking issue.

PR #723, opened 2026-06-04
Add model deprecation notification workflow

PierreLeGuen:
## Summary
- add planned model deprecation preview/confirm admin endpoints
- persist deprecation dates while keeping models active and expose them through /v1/models
- target active org owners/admins from 30-day model usage, send Resend notifications, and audit per-recipient outcomes

## Validation
- cargo check -p services -p database -p api
- cargo test --lib
- cargo test -p services email::tests

## Notes
- Full cargo test still requires the e2e admin database; locally the e2e bootstrap failed with connection refused at crates/api/tests/common/db_setup.rs:57.


Evrard-Nil (approved):
Solid backend for the model-deprecation workflow. Reviewed locally on `bd8e158`.

**Verified locally:**
- `cargo test -p services --lib` → **335/335 pass** (21 in email module incl. `render_model_deprecation_email_includes_expected_content_and_escapes_html` + `resend_sender_builds_model_deprecation_payload`).
- `cargo build -p services` clean.
- CI: claude-review + security_audit + Lint + Test Suite all SUCCESS.

**Design is right:**

1. **Two-endpoint preview/confirm split** — preview is side-effect-free and returns counts, confirm does the work. Right shape for an admin UI that needs to show blast-radius before pulling the trigger.

2. **Idempotency story is end-to-end:**
   - DB `UNIQUE (model_id, successor_model_name, deprecation_date, recipient_user_id, organization_id)` on `model_deprecation_email_deliveries`.
   - Pre-flight `list_sent_model_deprecation_delivery_keys` returns `(user_id, org_id)` pairs already in `Sent` status for the (model, successor, date) tuple → those are skipped on retry.
   - `ON CONFLICT DO UPDATE SET ... updated_at = NOW()` lets a Failed row be retried into Sent without violating the unique constraint.
   - Combined: re-running confirm after a partial failure only re-sends the failed/missing recipients. ✓

3. **Recipient scoping is correct.** 30-day usage window via `organization_usage_log`; only active owners/admins of active orgs (`om.role IN ('owner', 'admin')`, `o.is_active`, `u.is_active`); de-duplicated by lowercase email both within the run (`email_results` map) and across runs (`already_sent_emails`). The cross-org dedup is subtle but right — a single human shouldn't get five copies because they happen to own five orgs.

4. **Alias resolution.** The recipient query unions `models.model_name` and active `model_aliases.alias_name` → if the model was used under an old alias during the 30-day window, those orgs are still notified. Important for migrations.

5. **`parse_deprecation_date` is strict.** Accepts `YYYY-MM-DD` (defaults to 13:00 UTC per OpenRouter spec) or `YYYY-MM-DDTHH:00:00Z` (whole-hour UTC). Rejects: non-UTC offsets, non-zero minute/second/nanosecond, anything else. Returns `None` → 400 with a clear error message. No silent truncation.

6. **Audit table semantics.**
   - `model_id` / `recipient_user_id` / `organization_id` are FKs with `ON DELETE CASCADE`.
   - `initiated_by_user_id` is FK with `ON DELETE SET NULL` — historical audit survives admin deletion. Good split.
   - `status IN ('sent', 'failed', 'skipped')` check constraint matches the Rust `ModelDeprecationEmailStatus` variants.

7. **Privacy.** All new logging keys off `model_id` / `successor_model_id`. No recipient emails or org names in logs. Matches CLAUDE.md.

8. **Email rendering** is HTML-escaped via `escape_html` on every interpolation site (verified by `...includes_expected_content_and_escapes_html` test). Honest body copy: "No traffic has been rerouted automatically. Existing calls continue to use the current model until it is removed in a later step."

9. **`models_service.invalidate_models_cache()`** after the upsert so `/v1/model/list` reflects the new `deprecation_date` immediately (no 5-min TTL wait). Same pattern as the existing model-edit paths.

**Worth flagging (non-blocking):**

1. **No direct unit test for `preview_model_deprecation` / `confirm_model_deprecation`.** The email rendering + Resend payload are tested, but the complex dedup/skip-tracking logic in `confirm_model_deprecation` (cross-org email dedup, `already_sent_emails` derivation, the `counted_emails` set) has no coverage at the service layer. The PR body notes `e2e_test` requires a real admin DB that wasn't available locally; would be cheap to add a mock-repo + mock-EmailSender unit test that exercises the three statuses + the dedup branches. **Strongest follow-up I can think of.**

2. **`drop(model); drop(successor);` in `preview_model_deprecation`** is a code smell — destructure with `let (_, _, recipients) = ...` instead. Same effect, cleaner.

3. **`_deprecation_date`** is passed to `validate_and_load_model_deprecation` but unused (prefixed with underscore). Either use it (e.g., reject planned-deprecation-in-the-past) or drop it from the signature.

4. **No transactional boundary** between `upsert_model_pricing` (sets the date) and the email send loop. A panic in the middle leaves the model deprecated with partial notifications. The idempotency story above means a retry will catch up the unsent recipients — but worth documenting in the operator runbook that a re-confirm is the recovery path.

5. **`already_sent_emails` cross-row dedup** is correct but reads as "Skipped: already sent for this deprecation" even when the *specific* (user_id, org_id) row wasn't in the prior `Sent` set — it's the *email* that was. A clearer `email_last_error` value like "Already sent to this recipient via a different org membership in a prior run" would help auditing.

6. **Migration uses `uuid_generate_v4()`** which requires the `uuid-ossp` extension; `V0001`+ presumably already enable it, but worth a sanity check that no fresh-install path skips it.

LGTM as-is. The above are post-merge polish items, not gates.

PR #724, opened 2026-06-04
fix(anthropic): drop top_p too for opus-4-7 — #696 follow-up (regression)

Evrard-Nil:
## Regression

#720 (#696) made `anthropic/claude-opus-4-7` **400 on every request**, including requests with no sampling params. Verified live on staging:

| Request | prod (pre-#720) | staging (#720) |
|---|---|---|
| no sampling params | **200** ✅ | **400** `top_p is deprecated` ❌ |
| `temperature: 0.7` | 400 `temperature` | 400 `top_p` |
| `top_p: 0.5` | **200** ✅ | **400** `top_p` ❌ |

## Root cause

opus-4-7 rejects **both** non-default `temperature` *and* any `top_p` (```top_p is deprecated for this model```). #696 dropped `temperature` and forwarded `params.top_p` instead — but `top_p` **defaults to `Some(1.0)`** at deserialization (`api/src/models.rs` `default_top_p`), so the gateway now sends `top_p: 1.0` on every opus-4-7 request and 400s all of them. The old code never hit this because it always sent `temperature` (also defaulted to 1.0, which opus-4-7 accepts) and dropped `top_p`.

## Fix

For temperature-rejecting models (opus-4-7), drop **both** `temperature` and `top_p` and let the model use its own defaults → request succeeds (params ignored), matching the OpenRouter-interop goal of #696. Unit test updated to cover the defaulted `top_p=1.0` no-params case.

Follow-up to #720 / #696. Verified on `cloud-stg-api` after merge will flip opus-4-7 back to 200.

### DIFF
--- a/crates/inference_providers/src/external/anthropic/mod.rs
+++ b/crates/inference_providers/src/external/anthropic/mod.rs
@@ -42,8 +42,9 @@ const ANTHROPIC_PASSTHROUGH_KEYS: &[&str] = &["thinking", "reasoning_effort"];
 /// not over-strip.
 const ANTHROPIC_MODELS_REJECTING_TEMPERATURE: &[&str] = &["claude-opus-4-7"];
 
-/// Whether `model` rejects a non-default `temperature` and we must drop it
-/// (forwarding `top_p` instead) rather than 400 the caller (#696).
+/// Whether `model` rejects a non-default `temperature` (and also `top_p`), so we
+/// must drop BOTH sampling knobs rather than 400 the caller (#696). opus-4-7
+/// returns `temperature is deprecated` / `top_p is deprecated` for either.
 fn rejects_non_default_temperature(model: &str) -> bool {
     ANTHROPIC_MODELS_REJECTING_TEMPERATURE
         .iter()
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Reviewed the Anthropic adapter diff and surrounding request/defaulting/serialization paths. The change correctly omits both `temperature` and `top_p` for `claude-opus-4-7` variants, including the API default `top_p = Some(1.0)` case, while preserving normal sampling behavior for other Anthropic models.

Local checks run:
- `cargo test -p inference_providers external::anthropic::tests::` (20 passed)

GitHub check rollup at review time: lint, security_audit, and claude-review were passing; the full Test Suite was still in progress, with no failures observed.

PierreLeGuen:
What about we drop the default of top_p = 1 on every request in cloud API @Evrard-Nil ?

Evrard-Nil:
@PierreLeGuen good call — done in #726: dropped the `temperature`/`top_p` = Some(1.0) defaults so cloud-api forwards only what the caller sets (and deleted the two default fns). Build + all unit tests stay green; the response repo/route already use `unwrap_or(1.0)` for storage so DB/usage records are unchanged. One nuance flagged in that PR: clients that sent no temperature previously got an implicit 1.0, now they get the model's own default (no-op for 1.0-identity backends, but a real cross-provider change).

Kept this PR (#724) separate and minimal as the regression fix: #726 fixes the no-params/injected-default case generally, while #724 still gracefully drops an *explicitly-sent* temperature/top_p for opus-4-7 (OR clients do send those). They touch different files (models.rs vs anthropic/mod.rs) — no conflict.

PR #727, opened 2026-06-05
fix(patroni): tolerate string "unknown" lag from stopped cluster members

Evrard-Nil:
## Problem

Patroni's `/cluster` REST endpoint reports `lag` as an integer for streaming members, but emits the **string** `"unknown"` for members it cannot measure — `stopped`, `crashed`, or `start failed` replicas. Our `ClusterMember.lag: Option<i64>` rejects that string, so a **single** unhealthy member fails the **entire** `/cluster` parse:

```
Failed to parse cluster info: error decoding response body
  (invalid type: string "unknown", expected i64)
```

This fired in production today (2026-06-05) — 151 occurrences on `cloud-api` and ~96 on `chat-api`/`chat-api-worker` — triggered by a stopped replica (`postgres-prod-vzeis375`, `"lag": "unknown"`) appearing in the member list. It cleared the instant the replica was removed.

## Impact

- **Running service:** the background refresh logs `Failed to refresh cluster state` and `ClusterState` freezes at its last-known-good value — the service is blind to any real failover during the window.
- **Restart/deploy:** `Database::new` calls `discovery.update_cluster_state().await?`, so the parse error aborts startup with `No leader found in cluster during initialization`. A pod restarting while a stopped replica is in the list **cannot boot the DB layer at all**, despite a perfectly healthy leader.

## Fix

Custom `deserialize_lag` that accepts an integer or coerces any non-integer value (`"unknown"`, `null`, absent) to `None`. The stopped member already sorts last in `get_replicas_by_lag` (`unwrap_or(i64::MAX)`) and is excluded from selection by its non-`running`/`streaming` state, so behavior is unchanged for healthy members — the cluster just parses instead of erroring.

## Tests

- `test_member_with_string_lag_unknown` — single stopped member with `"lag": "unknown"` parses, `lag == None`.
- `test_cluster_with_stopped_member_string_lag` — the **exact** production payload that previously failed now parses all 3 members; streaming member keeps `lag == Some(0)`.

`cargo test -p database patroni` → 4 passed.

## Note

`chat-api` has a byte-identical `patroni_discovery.rs` and the same bug — companion PR opened there.

### DIFF
--- a/crates/database/src/patroni_discovery.rs
+++ b/crates/database/src/patroni_discovery.rs
@@ -1,11 +1,35 @@
 use anyhow::{anyhow, Result};
 use reqwest::Client;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{Mutex, RwLock};
 use tokio::time;
 use tracing::{debug, error, info, warn};
+
+/// Patroni reports `lag` as an integer for streaming members, but emits the
+/// string `"unknown"` for members it cannot measure (e.g. `stopped`, `crashed`,
+/// or `start failed` replicas). A bare `Option<i64>` rejects that string and
+/// fails the entire `/cluster` parse, which would freeze topology discovery (or
+/// abort startup) over a single unhealthy member. Coerce any non-integer value
+/// to `None` so the rest of the cluster still parses.
+fn deserialize_lag<'de, D>(deserializer: D) -> Result<Option<i64>, D::Error>
+where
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Inspected the Patroni lag deserialization change, the cluster state filtering, and read-replica selection paths. The new parser preserves integer lag values while mapping `"unknown"`/null/missing lag to `None`, and stopped members still do not enter the usable replica set.

Local checks run:
- `cargo test -p database patroni` (4 passed)
- `cargo fmt --check`

Remote CI at review time: lint/security passed; full Test Suite was still pending.

PR #728, opened 2026-06-05
feat(metrics): per-model billed-usage metrics

Evrard-Nil:
## What

Adds a per-model **billed-usage** metric family, emitted at the single authoritative billing point — `UsageServiceImpl::record_usage()`. This is where *every* inference type flows through after cost is computed, so coverage is uniform — unlike the existing `cloud_api.tokens.*` metrics which only fire on the chat-completions streaming/non-streaming path.

## New metrics

All OTLP counters, dimensioned by `model` + `inference_type` + `environment`:

| Metric | Meaning |
|---|---|
| `cloud_api.billed.requests` | billed requests (one per recorded usage row) |
| `cloud_api.billed.input_tokens` | billed input tokens |
| `cloud_api.billed.output_tokens` | billed output tokens |
| `cloud_api.billed.cache_read_tokens` | billed cache-read tokens |
| `cloud_api.billed.input_cost_usd` | input cost (nano-dollars) |
| `cloud_api.billed.output_cost_usd` | output cost (nano-dollars) |

## Design notes

- **Coverage:** recorded in the central `record_usage()` billing path, so chat, image, audio, rerank, embedding and privacy-classify are all captured (the existing `cloud_api.tokens.*` only cover chat completions).
- **No double counting:** gated by the existing `log.was_inserted` idempotency guard, so retried requests with the same `inference_id` don't inflate the counters.
- **`inference_type` tag:** added to the new metrics *and* to the existing `cloud_api.cost.usd` total. This is additive/backward-compatible in Datadog (queries summing over all `inference_type` values are unaffected) and lets you slice billed usage by operation type. `total == input + output` is derivable from the breakdown.
- **`cloud_api.cost.usd`** keeps its existing semantics (still guarded by `total_cost > 0`) so current dashboards are unaffected.
- **Cardinality:** the tag fan-out is `model × inference_type × environment` — bounded and low, consistent with the `consts.rs` "low-cardinality tags only" rule (no org/workspace/api_key).

## Files changed

- `crates/services/src/metrics/consts.rs` — 6 metric-name constants + `TAG_INFERENCE_TYPE`
- `crates/services/src/metrics/mod.rs` — counter descriptions
- `crates/services/src/usage/mod.rs` — emit block

## Testing

- `cargo check -p services` — clean
- `cargo fmt -p services` — clean
- `cargo test -p services usage:: --lib` — 8 passed

### DIFF
--- a/crates/services/src/metrics/consts.rs
+++ b/crates/services/src/metrics/consts.rs
@@ -22,6 +22,17 @@ pub const METRIC_REQUEST_ERRORS: &str = "cloud_api.request.errors";
 // Cost metrics
 pub const METRIC_COST_USD: &str = "cloud_api.cost.usd";
 
+// Billed-usage metrics (emitted once per newly-recorded usage row, covering every
+// inference type, dimensioned by model + inference_type). These are the authoritative
+// "what we billed" counters; tokens.input/output above are throughput-oriented and only
+// cover the chat-completions path.
+pub const METRIC_BILLED_REQUESTS: &str = "cloud_api.billed.requests";
+pub const METRIC_BILLED_INPUT_TOKENS: &str = "cloud_api.billed.input_tokens";
+pub const METRIC_BILLED_OUTPUT_TOKENS: &str = "cloud_api.billed.output_tokens";
+pub const METRIC_BILLED_CACHE_READ_TOKENS: &str = "cloud_api.billed.cache_read_tokens";
+pub const METRIC_BILLED_INPUT_COST_USD: &str = "cloud_api.billed.input_cost_usd";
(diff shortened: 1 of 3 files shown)


PierreLeGuen (approved):
Reviewed the diff and surrounding usage, repository, metrics, tests, and workflow code. I did not find substantive correctness issues: the new billed counters are emitted from the central `record_usage` path and are gated on `log.was_inserted`, which matches the committed usage row/balance update path and skips idempotent duplicates.

Local checks run:
- `cargo fmt --check`
- `cargo test -p services --lib usage::tests`
- `cargo clippy -p services --all-targets --all-features -- -D warnings`

GitHub checks were still in progress at review time; no failures were reported.

PR #729, opened 2026-06-05
feat(billing): bill interrupted Anthropic streams for input tokens

Evrard-Nil:
## Problem

Part of nearai/infra#98 — *"Implement billing for interrupted streams"*. We see many:

> `Stream interrupted before usage stats received (client disconnect or provider error)` (`stream_error: false`)

When a stream is interrupted (client disconnect, or provider error before the final usage chunk), `InterceptStream` has no `last_usage_stats` and the request goes **unbilled** even though tokens were consumed.

## Root cause (per-provider)

Billing happens in `InterceptStream` (`crates/services/src/completions/mod.rs`), which captures usage from any chunk carrying a `usage` object. Whether partial usage is *available* at interruption depends on the provider:

| Provider | Incremental usage? | Billed on interruption before this PR? |
|---|---|---|
| Internal vLLM/sglang | `continuous_usage_stats:true` → every chunk | ✅ already |
| Gemini | per-chunk `usageMetadata` | ✅ already |
| **Anthropic** | input at `message_start`, output only in final `message_delta`; input never attached to a chunk | ❌ |
| OpenAI / openai_compatible | usage only in final chunk (OpenAI rejects `continuous_usage_stats`) | ❌ (no mid-stream data available — out of scope) |

## Fix

Attach the input tokens Anthropic reports at `message_start` to the first (role) chunk, via a new `ChunkContext::role_chunk_with_usage` helper. `InterceptStream` then captures it as `last_usage_stats`, so an interrupted Anthropic stream is billed for the prompt tokens Anthropic charged us for. Completion tokens stay `0` until the final chunk (Anthropic gives no mid-stream output count — we bill only what's real). On a clean stream the final chunk's full usage overwrites this, so billing is unchanged.

This mirrors what Gemini already does (usage on every chunk).

## Scope / known limitation

Real partial usage only — **no estimation**. OpenAI/openai_compatible interrupted streams still can't be billed because the provider sends no mid-stream usage; those keep the existing WARN and bill \$0. Documented intentionally.

The inference-proxy side (billing interrupted internal-model streams from its own usage reporter) is in nearai/inference-proxy#155.

## Tests

- `test_message_start_chunk_carries_input_usage` — new regression test.
- Full `inference_providers` lib suite green (224 tests).

### DIFF
--- a/crates/inference_providers/src/chunk_builder.rs
+++ b/crates/inference_providers/src/chunk_builder.rs
@@ -50,6 +50,21 @@ impl ChunkContext {
 
     /// Create a chunk with assistant role (typically first chunk)
     pub fn role_chunk(&self) -> ChatCompletionChunk {
+        self.role_chunk_with_usage(None)
+    }
+
+    /// Create a chunk with assistant role, optionally carrying early usage stats.
+    ///
+    /// Some providers (notably Anthropic) only emit complete token usage in the
+    /// final stream event, but expose the prompt/input token count at the very
+    /// start (Anthropic's `message_start`). Attaching that partial usage to the
+    /// first chunk lets the billing layer (`InterceptStream`) capture it as
+    /// `last_usage_stats`, so an interrupted stream (client disconnect or
+    /// provider error before the final chunk) is still billed for the input
+    /// tokens the provider already charged us for. On a clean stream the final
+    /// chunk's complete usage overwrites this partial value, so billing is
+    /// unaffected.
+    pub fn role_chunk_with_usage(&self, usage: Option<TokenUsage>) -> ChatCompletionChunk {
         self.build(
             ChatDelta {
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):
Inspected the Anthropic stream parser/chunk builder diff and the downstream `InterceptStream` billing path. The early `message_start` usage is overwritten by the final usage chunk on clean streams and records prompt-only usage on interrupted streams, with no accumulation/double-billing path found.

Checks:
- `gh pr checks 729` passes: Lint, Test Suite, security_audit, claude-review
- `cargo test -p inference_providers --lib --tests` passes: 224 unit tests, 8 integration tests, 1 ignored
- Note: full `cargo test -p inference_providers` also runs an unrelated pre-existing doctest in untouched `crates/inference_providers/src/models.rs` that fails on a private import.

Issue #169, opened 2025-11-11
Fix conversation items listing order


PierreLeGuen:
Currently listing items from a conversation don't follow OpenAI spec.

PR #734, opened 2026-06-08
fix(billing): accept per-type usage on /v1/internal/usage (#169 prereq)

Evrard-Nil:
## What

Extends the internal usage wire union so inference-proxy can report direct-`sk-` usage with the **correct inference type** instead of mislabeling everything as `chat_completion`.

Adds four input-token-billed, output-less variants to `RecordUsageApiRequest` (`crates/services/src/usage/ports.rs`):

- `embedding`, `rerank`, `score`, `privacy_classify`

Each bills `input_tokens × input_cost_per_token` (no output cost, no cache pricing) via the existing per-type cost arm in `record_usage`. Also folds `Score` into that cost arm — it previously fell through to the chat-token formula (equivalent today since score carries 0 output tokens, but a latent mislabel/mispricing hazard).

## Why

This is the **prerequisite half** of nearai/infra#169. Today inference-proxy's `/v1/internal/usage` reporter only knows two shapes, so:

- **privacy-classify** (direct `sk-`) is **unbilled** — its response has no top-level `usage` object, so it falls through to the chat extractor and reports nothing.
- **embeddings / rerank / score** are billed but **mislabeled** as `chat_completion`, corrupting per-type analytics on `organization_usage_log.inference_type`.

cloud-api must accept the new type tags **before** the proxy starts sending them, otherwise the new-type reports would be rejected and usage dropped. This PR is that step.

## Safety

**Purely additive.** Nothing emits the new types until the inference-proxy side ships (tracked in nearai/infra#169), so existing traffic and billing are unchanged. The internal ack (`RecordUsageResponse`) still echoes the chat-completion shape; the **stored** `inference_type` carries the correct label.

## Test plan

- `cargo test --test e2e_all usage_recording` — adds `test_record_input_only_usage_types` (records embedding/rerank/score/privacy_classify, asserts `input_tokens × input_rate`, no output cost) and `test_record_input_only_usage_rejects_zero_tokens`. ✅ 6/6 pass locally (incl. chat/image baselines).
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo fmt --all --check` clean.

## Follow-ups

- inference-proxy side of #169 (per-endpoint `UsageType` + correct extraction + SAFE-LOUD `catch_all`) — depends on this being deployed.
- Part of the broader billing refactor (cloud-api#602 / infra#98 residual / infra#169).

### DIFF
--- a/crates/api/tests/e2e_all/usage_recording.rs
+++ b/crates/api/tests/e2e_all/usage_recording.rs
@@ -868,3 +868,74 @@ async fn test_internal_usage_returns_401_on_wrong_token() {
 
     assert_eq!(response.status_code(), 401);
 }
+
+/// `/v1/internal/usage` accepts the input-token-billed kinds (embedding,
+/// rerank, score, privacy_classify) that inference-proxy now reports for
+/// direct `sk-` requests. Each bills `input_tokens × input_rate` with no
+/// output cost. See nearai/infra#169. (The internal ack echoes the
+/// chat-completion shape; the *stored* `inference_type` carries the label.)
+#[tokio::test]
+async fn test_record_input_only_usage_types() {
+    let server = enable_internal_usage_server().await;
+    setup_qwen_model(&server).await;
+    let id = provision_identity(&server).await;
+
+    for (idx, ty) in ["embedding", "rerank", "score", "privacy_classify"]
+        .iter()
+        .enumerate()
+    {
+        let response = post_internal_usage(
+            &server,
+            &id,
+            serde_json::json!({
+                "type": ty,
+                "model": "Qwen/Qwen3-30B-A3B-Instruct-2507",
+                "input_tokens": 100,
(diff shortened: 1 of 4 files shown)


PierreLeGuen (approved):
Reviewed the `/v1/internal/usage` wire variants, usage-service validation/pricing paths, database usage adapter/idempotency flow, and the added e2e coverage. The new input-only variants map to the existing inference type strings and use the intended input-token-only billing branch; I did not find a substantive issue.

Local checks:
- `cargo fmt --all -- --check` passed.
- `cargo test -p services usage::` passed (8/8).
- `cargo test --test e2e_all usage_recording` compiled but could not run locally because PostgreSQL was not available; all selected tests failed during DB bootstrap with connection refused at `crates/api/tests/common/db_setup.rs:57`.

GitHub checks at review time: Lint, security_audit, and claude-review passed; Test Suite was still pending.

PR #736, opened 2026-06-08
fix(billing): bill the x-auto-redact classify pass (#602)

### LINKED ISSUE #602: Billing for x-auto-redact
The new `x-auto-redact` header (redacts PII from prompts to 3p providers, see PR #599) needs a billing model — probably a small per-call uplift or a separate line item on usage.

Source: May 13 other-todos list.

Evrard-Nil:
## What

Bills the privacy-filter **classify pass** that `x-auto-redact` runs before a chat completion. Closes nearai/cloud-api#602.

- `detect::parse_response` now sums the classify response's `data[].usage.input_tokens` (previously discarded).
- The count is threaded out through `redact_messages` → `maybe_redact` to the chat handler.
- The handler records a `PrivacyClassify` usage row on `openai/privacy-filter` (`input_tokens × input_rate`, no output cost), mirroring the existing explicit `/v1/privacy/classify` recording (fresh v4 id, one record per request).

## Why

The classify pass is a **real inference call** to the PII model (GPU cost), and the explicit `/v1/privacy/classify` endpoint already bills it — but the inline auto-redact path threw the usage away, so auto-redact requests got the classify for free.

## Design notes

- **Billed on `requested`, not `enabled`.** A request that opts in but contains no PII (`auto_redact_enabled == false`) still ran a classify pass and is charged.
- **Fail-closed preserved.** Billing fires only on the detector's `Ok` branch; a timed-out/unavailable detector already returns 503 and is never billed, never mutates messages.
- **Best-effort recording.** A model-lookup or record failure is logged and swallowed so it never fails the user's completion (which bills separately via `InterceptStream`). The classify already cost GPU time, so it's billed even if the downstream completion later fails.
- **No idempotency key needed** beyond the fresh v4 id: each request runs exactly one classify (a real, distinct billable inference), matching how the explicit privacy endpoints record.

## Test plan

- `cargo test -p services auto_redact::detect` — token-sum parsing (incl. multi-fragment + out-of-range-index cases). ✅
- `cargo test --test e2e_all auto_redact` — **21/21 pass**, including a new `auto_redact_bills_classify_pass` that drives a real `x-auto-redact` chat and asserts (via DB) exactly one `privacy_classify` usage row with the mock's 10 input tokens. ✅
- `cargo clippy -p services -p api --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean. (Verified in an isolated target dir.)

## Scope

Independent of the other billing PRs; part of the broader billing refactor (infra#169 / infra#98 residual). No migration, no wire change.

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -503,12 +503,36 @@ async fn bill_auto_redact_classify(
     api_key: &crate::middleware::auth::AuthenticatedApiKey,
     classify_tokens: i64,
 ) {
-    // Clamp to the i32 the usage row stores: an implausibly large count is
-    // capped (billed) rather than dropped, and 0/negative is nothing to bill.
-    let input_tokens = classify_tokens.clamp(0, i32::MAX as i64) as i32;
-    if input_tokens == 0 {
-        return;
+    if classify_tokens <= 0 {
+        return; // nothing to bill
     }
+    // Cap an anomalous count at the same `MAX_REASONABLE_TOKENS` the explicit
+    // `/v1/privacy/{classify,redact}` paths use, emitting the same anomaly
+    // metric — so x-auto-redact (which shares the privacy-filter classify
+    // call) can't massively overcharge on a buggy/malicious provider response
+    // instead of just clamping to i32::MAX. See nearai/cloud-api#602 review.
+    const MAX_REASONABLE_TOKENS: i32 = 1_000_000;
+    let input_tokens = if classify_tokens > MAX_REASONABLE_TOKENS as i64 {
+        tracing::error!(
+            token_count = classify_tokens,
+            max_expected = MAX_REASONABLE_TOKENS,
+            model = auto_redact::DEFAULT_PII_MODEL,
+            "auto_redact: provider returned unreasonable classify token count - capping"
+        );
+        let model_tag = format!("model:{}", auto_redact::DEFAULT_PII_MODEL);
+        let reason_tag = format!(
+            "reason:{}",
+            services::metrics::consts::REASON_TOKEN_OVERFLOW
+        );
+        app_state.metrics_service.record_count(
+            services::metrics::consts::METRIC_PROVIDER_TOKEN_ANOMALIES,
+            1,
+            &[model_tag.as_str(), reason_tag.as_str()],
+        );
+        MAX_REASONABLE_TOKENS
+    } else {
+        classify_tokens as i32 // safe: 0 < classify_tokens <= 1_000_000
+    };
 
     let model = match app_state
         .models_service
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -417,11 +417,15 @@ fn convert_chat_request_to_service(
 /// detector + rewrite messages in place. Returns the placeholder map for
 /// downstream un-redact; an empty map is also returned when auto-redact is
 /// off.
+/// The tuple is `(requested, map, classify_input_tokens)`. `classify_input_tokens`
+/// is the number of tokens the privacy-filter billed for the classify pass,
+/// for the caller to charge (nearai/cloud-api#602); it is 0 when auto-redact
+/// is off or there was nothing to classify.
 async fn maybe_redact(
     headers: &header::HeaderMap,
     service_request: &mut ServiceCompletionRequest,
     pool: &services::inference_provider_pool::InferenceProviderPool,
-) -> Result<(bool, RedactionMap), AutoRedactError> {
+) -> Result<(bool, RedactionMap, i64), AutoRedactError> {
     let header_values: Vec<&str> = headers
         .get_all(auto_redact::AUTO_REDACT_HEADER)
         .iter()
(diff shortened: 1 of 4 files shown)


PierreLeGuen (changes requested):
Requesting changes.

- `crates/api/src/routes/completions.rs:1184`: `bill_auto_redact_classify(...).await` runs inline after the detector succeeds but before the user's completion is dispatched. That helper does a model lookup and `usage_service.record_usage(...).await` with no timeout. Since this billing is explicitly best-effort, a slow/exhausted billing DB or model lookup can now stall every `x-auto-redact` completion even though the classify pass already succeeded; normal chat usage recording is not on this pre-dispatch path. Please bound this with a timeout and/or async fallback like the existing usage-recording helpers so billing failures are logged but cannot block the completion request.

Checks run locally: `cargo test -p services auto_redact` passed; `cargo fmt --check` passed; `git diff --check origin/main...HEAD` passed. I also attempted `cargo test -p api --test e2e_all auto_redact_bills_classify_pass -- --nocapture`; it compiled but could not run because the local Postgres test DB connection was refused. GitHub lint/security/claude checks are passing; Test Suite was still pending when reviewed.

Evrard-Nil:
Thanks @PierreLeGuen — good catch, addressed in ebb7ebef.

The classify billing no longer touches the pre-dispatch hot path: it's now a **time-bounded (10s) spawned task**, so a slow/exhausted billing DB or model lookup is logged but can never stall an `x-auto-redact` completion. The classify already succeeded by then, and the chat itself still bills separately via `InterceptStream`.

Also hardened token-overflow handling from the bot review: `detect::parse_response` sums with `saturating_add`, and `bill_auto_redact_classify` clamps to `i32::MAX` rather than dropping the bill.

The e2e test now polls (bounded) for the async usage row. Local: `cargo test --test e2e_all auto_redact` → 21/21 pass; clippy + fmt clean. (The DB-refused issue you hit is the local test Postgres — `DATABASE_PORT=5432` against the `cloud-api-test-pg` container.)

PierreLeGuen (changes requested):
I found two issues that should be fixed before merge:

1. `crates/api/src/routes/completions.rs:508` clamps auto-redact classify usage to `i32::MAX`, while the existing `/v1/privacy/classify` and `/v1/privacy/redact` paths cap anomalous provider token counts at `MAX_REASONABLE_TOKENS = 1_000_000` and emit anomaly metrics (`crates/api/src/routes/completions.rs:4793-4817`, `crates/api/src/routes/completions.rs:5306-5329`). This means the same privacy-filter classify call can bill up to 2,147,483,647 tokens only when invoked through `x-auto-redact`, which can massively overcharge on a buggy or malicious provider response. Please reuse the same cap/metric handling for auto-redact billing.

2. `crates/services/src/auto_redact/detect.rs:86-110` now makes `usage.input_tokens` part of the main detector response deserialization. A malformed billing-only field, such as the wrong type or an out-of-range JSON number, now turns a valid span response into `AutoRedactError::Internal` and fails the completion, whereas this metadata should be best-effort billing and the explicit privacy endpoints tolerate billing parse problems without failing the upstream response. Please parse usage tolerantly, for example from `serde_json::Value` or a custom deserializer that treats invalid token counts as 0/clamped, so span application is not coupled to usage metadata validity.

Checks:
- `cargo test -p services auto_redact` passed.
- `cargo test -p api --test e2e_all auto_redact` compiled but failed locally because the test Postgres admin connection was refused.
- GitHub checks currently pass: Lint, Test Suite, security_audit.

Evrard-Nil:
Both addressed in 814a8261, thanks:

1. **Cap consistency** — `bill_auto_redact_classify` now caps at the same `MAX_REASONABLE_TOKENS = 1_000_000` as the explicit `/v1/privacy/{classify,redact}` paths and emits the same `METRIC_PROVIDER_TOKEN_ANOMALIES` / `reason:token_overflow` metric, instead of clamping to `i32::MAX`. So x-auto-redact can no longer overcharge on a buggy/malicious provider count. (The 1M cap is currently duplicated ~4× across the privacy handlers; folding it into one helper is noted for the billing refactor in infra#170 rather than expanding this PR.)
2. **Tolerant usage parse** — `DetectItem.usage` is now an `Option<serde_json::Value>` (not a typed struct), and `input_tokens` is extracted leniently (invalid type / out-of-range / absent → 0). A malformed billing-only field can no longer fail span deserialization or the completion. Added `parse_tolerates_malformed_usage`.

Local: `cargo test -p services auto_redact::detect` (5/5, incl. the new tolerance test) + `cargo test --test e2e_all auto_redact` (21/21) + clippy/fmt clean (isolated target; the `DATABASE_PORT=5432` note is the local `cloud-api-test-pg` container if you want to re-run the e2e).

PierreLeGuen (approved):
Reviewed the current diff locally, including the auto-redact token plumbing, tolerant detector usage parsing, the spawned/time-bounded classify billing task, the usage-service record shape, and the added E2E coverage. The earlier blocking issues on hot-path billing and anomalous/malformed token handling are addressed.

Checks run locally:
- `cargo test -p services auto_redact::detect::tests -- --nocapture` passed (5/5).
- `git diff --check origin/main...HEAD` passed.
- `cargo test -p api auto_redact_bills_classify_pass --test e2e_all -- --nocapture` compiled, then failed to run because the local test Postgres admin connection was refused.

GitHub checks are green: Lint, Test Suite, and security_audit.

Issue #735, opened 2026-06-08
Epic: first-class SNI-rotation + pluggable attested 3p providers (FleetRouter + ProviderRegistry + VerificationEngine)


Evrard-Nil:
Tracking epic for the cloud-api inference-stack refactor. Goal: (1) make SNI-rotation first-class via **stable address-derived backend handles** (fixes the GLM-5.1 `"does not match any attested fingerprint"` signature-fetch bug + the pin-eviction hazard), (2) a pluggable teep-style **attested-provider** abstraction behind a single registry, (3) a per-model **attestation policy** (NEAR + attested-3p fallback / attested-3p-only / non-attested) where fallback can never cross an attestation boundary.

### Phases (each independently shippable; flag-gated; GLM-5.1 as canary)
- [ ] **P0** model-proxy: stable-handle SNI `-b<hash>` + `select_backend_by_handle` + `/backends/list` — nearai/model-proxy#38 *(additive; deps: none)*
- [ ] **P1** cloud-api: extract `FleetRouter` from `VLlmProvider` (still `u64`, flag-gated, characterization tests first) *(deps: none)*
- [ ] **P2** cloud-api: `ProviderRegistry` + capability handles; move the 5 NEAR-only no-ops off `InferenceProvider`; delete `ExternalProvider` attestation stubs *(deps: P1)*
- [ ] **P3** cloud-api: `FleetRouter` `u64`→`BackendHandle`; per-handle `FingerprintState`; handle-based discovery (preserve pubkey harvest). **Fixes the bug.** *(deps: P0 deployed to gpu11+gpu30, P1, P2)*
- [ ] **P4** cloud-api: `VerificationEngine` factor pipeline + per-step traits + `RawAttestation` + `Blocked()`; port NEAR verbatim (signing_algo fix, shared codec, golden vectors) *(deps: P2)*
- [ ] **P5** cloud-api: DB migration — drop `chk_external_provider_no_attestation`; add `attestation_policy` + `providers`/`model_providers(fallback_order)` *(deps: none; schema-only)*
- [ ] **P6** cloud-api: loader-merge NEAR+3p under one model_id; `ProviderTier` vs `RequiredPolicy` filter in `get_providers_with_fallback` (fail-closed, within-tier RR) *(deps: P2, P5)*
[...]

PR #737, opened 2026-06-08
feat(db): per-model attestation_policy migration (V0056)

Evrard-Nil:
## What

**P5** of the inference-stack refactor (epic #735). Schema-only and rollout-safe: this migration introduces policy storage without changing current runtime behavior.

- Keeps `chk_external_provider_no_attestation` in place (no premature enablement of attested external providers before policy-aware runtime/write paths land).
- Adds `models.attestation_policy VARCHAR(32)` as **nullable** with **no default**.
- Adds `chk_attestation_policy_valid` with `attestation_policy IS NULL OR attestation_policy IN (...)` where allowed values are:
  `near_only | near_or_attested_3p | attested_3p_only | non_attested`.
- Backfills existing `models` rows:
  - `provider_type='vllm'` → `near_only`
  - `provider_type='external'` → `non_attested`
- Adds nullable `model_history.attestation_policy` audit mirror and backfills it where `provider_type` is known.

## Policy semantics (for the later routing phase)
- `near_only` — only NEAR AI TEE backends (today's vLLM behavior)
- `near_or_attested_3p` — prefer NEAR, may fall back to an **attested** 3rd party
- `attested_3p_only` — only attested 3rd parties
- `non_attested` — any provider, incl. plaintext 3rd parties

Routing/enforcement and insert-time population of this column land in a later phase; this migration intentionally allows `NULL` as “not yet classified”.

## Migration number
Next free is **V0056** (the dir jumps V0053 → V0055; V0054 never existed). Refinery `load_sql_migrations` picks up the new file automatically — no code change.

## Verification
Applied the **full chain V0001..V0056** on a clean Postgres 16 — all apply cleanly. Asserted: `models.attestation_policy` is nullable with no default, `chk_attestation_policy_valid` present, `chk_external_provider_no_attestation` still present, and `model_history.attestation_policy` nullable.

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>

### DIFF
--- a/crates/database/src/migrations/sql/V0056__add_attestation_policy.sql
+++ b/crates/database/src/migrations/sql/V0056__add_attestation_policy.sql
@@ -1,6 +1,6 @@
 -- Per-model attestation policy: what a model REQUIRES of the providers that
--- may serve it. Supersedes the binary `attestation_supported` flag for routing
--- (wired up in a later phase); this migration only adds the data model.
+-- may serve it. This migration only adds the data model as groundwork — no code
+-- reads or writes the column yet, and behavior is unchanged.
 --
 -- Policy values:
 --   near_only           - only NEAR AI TEE backends (current behavior for vLLM models)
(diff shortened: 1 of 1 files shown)


PierreLeGuen (changes requested):
I found two blockers in this migration-only change.

- `crates/database/src/migrations/sql/V0056__add_attestation_policy.sql:17`: dropping `chk_external_provider_no_attestation` now allows admins to persist `provider_type=external` with `attestation_supported=true`, but the current runtime still treats `model.attestation_supported` as authoritative for E2EE rejection and signature storage (`crates/services/src/completions/mod.rs:1235`, `:1288`, and the non-streaming path too). That creates an inconsistent external-provider state before the new policy/routing code exists. CI is already catching this: `external_providers::test_external_model_cannot_have_attestation_enabled` now returns 200 instead of being rejected. Please either keep the old invariant until the code that consumes `attestation_policy` lands, or update the write/runtime paths and tests in this PR so attested 3P providers are represented safely.

- `crates/database/src/migrations/sql/V0056__add_attestation_policy.sql:21`: the new column defaults every future insert to `non_attested`. Existing vLLM rows are backfilled to `near_only` at lines 25-26, but current model inserts do not write `attestation_policy` (`crates/database/src/repositories/model.rs:520-572`), so any vLLM model created after this migration and before the later routing phase will be stored as `non_attested`. When routing starts reading this column, those models can be misclassified and allowed to use less-attested providers. The migration needs a rollout-safe default/trigger/app write path, or the backfill needs to be repeated when the consuming code lands.

Checks: inspected the diff, migration chain context, model write paths, completion attestation use, and GitHub Actions metadata. GitHub Actions `Test Suite` is failing on the external-provider attestation test. I also ran the targeted test locally; it compiled, then failed to connect to local Postgres (`Connection refused`), so local runtime reproduction was blocked by environment.

Evrard-Nil:
Thanks @PierreLeGuen — both blockers were right; the migration is now strictly additive with no behavior change:

**Blocker 1 (premature constraint drop).** Reverted — `chk_external_provider_no_attestation` is **kept**. You're right that dropping it while the runtime still treats `attestation_supported` as authoritative (`completions/mod.rs:1235/1288`) creates inconsistent attested-external state, and it turned `test_external_model_cannot_have_attestation_enabled` red. The drop + attested-3p enablement now move to the policy-aware routing phase, which updates the write/runtime paths **and** that test in the same PR. CI should be green again (constraint still rejects external+attestation).

**Blocker 2 (misleading default).** The column is now **NULLABLE with no default**. Existing rows are still backfilled (`vllm→near_only`, `external→non_attested`), but a row created in the window before the routing code lands is left **NULL** ("not yet classified") instead of being misclassified as `non_attested`. The routing phase sets it on insert, re-backfills NULLs, then tightens to `NOT NULL`. The `CHECK` now allows `NULL OR <valid>`.

Also addressed the bots: backfilled the `model_history` mirror where `provider_type` is known (@gemini-code-assist); re: @Copilot's bloat note — without a column default every row legitimately needs a value, and `models` is tiny, so the full one-time UPDATE is fine.

Re-verified the full V0001..V0056 chain on clean Postgres 16. Re-requesting review.

PierreLeGuen (approved):
Reviewed the V0056 migration, surrounding model/model_history migrations, repository row mappings, and current CI metadata. I did not find substantive issues.

Checks run locally:
- Applied all SQL migrations through V0056 on disposable Postgres 15 with psql.
- Simulated pre-V0056 vLLM/external rows and verified backfill maps to near_only/non_attested.
- Ran `cargo test -p database migrations --lib` (compiled successfully; no matching tests ran).

PR #739, opened 2026-06-08
Preserve hidden-state fields in chat completions (supersedes #679)

Evrard-Nil:
## Summary

Enables **hidden-state passthrough** for mechanistic-interpretability research — a client sets `return_hidden_states` (optionally `layers`) on a chat-completions request and receives the backend's per-layer activations back through cloud-api.

Supersedes #679 by @satojandro, carried forward with credit (`Co-authored-by`). Thanks for the report and the original patch — this reworks the mechanism so it actually reaches the client (see below) and addresses the review feedback on that PR.

## Why a new PR instead of patching #679

#679 added the catch-all at the **top level** of `ChatCompletionResponse`/`ChatCompletionChunk`. But sglang attaches hidden states **per-choice**, so a top-level catch-all can't capture them and they were still dropped on re-serialized paths. Verified against sglang `entrypoints/openai/protocol.py`:

| | `hidden_states` location |
|---|---|
| Non-streaming | `choices[].hidden_states` (sibling of `message`) |
| Streaming | `choices[].delta.hidden_states` |

## Changes

- **Core fix:** `#[serde(flatten)] extra` on **`ChatCompletionResponseChoice`** — preserves non-streaming per-choice `hidden_states` on the re-serialized (auto-redact) path. The default non-streaming path already returns the provider's raw bytes, and streaming is already covered by the existing `ChatDelta.extra`; this closes the one remaining gap.
- Top-level `extra` on `ChatCompletionResponse`/`ChatCompletionChunk` to preserve other unknown top-level provider fields (e.g. sglang `sglext`).
- **Request side:** no typed fields. `return_hidden_states`/`layers` already pass through verbatim via the existing `ChatCompletionParams.extra`, which is serialized straight to vLLM/sglang. This avoids dead plumbing **and** the silent-drop-on-mistype behavior the reviewers flagged on #679 (a mis-typed value now passes through rather than being dropped). Locked with a test.
- Regression tests for both sglang response shapes + request passthrough; fixed a stale `detect_audio_content_type` doctest import.

## Caveat

This is a **no-op until an sglang backend is launched with `--enable-return-hidden-states`** (cvm-conf / infra side). cloud-api is the proxy-layer half and is now ready.

## Testing

- `cargo check --workspace --all-targets`
- `cargo test`: `inference_providers --lib` (227), `services --lib` (342), `api --lib` (131), doctests — all green.

---

### ⚠️ Depends on an inference-proxy change (not sufficient alone)

inference-proxy sits between cloud-api and sglang and does its own deserialize/reassemble:
- **Streaming** already works through it (`SseTransformer` is `serde_json::Value`-based and preserves `delta.hidden_states`); this PR's `ChatDelta`-level coverage already existed.
- **Non-streaming** is dropped *inside inference-proxy*: it forces `stream:true` upstream and reassembles via `StreamAssembler`, which rebuilds `message` from only `role/content/reasoning_content/tool_calls` (no catch-all). So this PR's `ChatCompletionResponseChoice.extra` only takes effect once inference-proxy preserves `hidden_states` in that reassembly.
- Also note: forcing upstream streaming means only **last-token** hidden states are produced; full per-token activations need a true non-streaming upstream call.

This PR is the cloud-api half (correct + harmless on its own); the inference-proxy reassembler is the gating change for non-streaming.

### DIFF
--- a/crates/api/src/conversions.rs
+++ b/crates/api/src/conversions.rs
@@ -937,4 +937,37 @@ mod tests {
         let domain: ChatCompletionParams = many.into();
         assert_eq!(domain.stop, Some(vec!["a".to_string(), "b".to_string()]));
     }
+
+    /// The mechanistic-interpretability workflow sends `return_hidden_states`
+    /// / `layers` on the chat-completions request. We deliberately do NOT
+    /// model these as typed fields: they ride the flattened `extra` catch-all
+    /// on both `ChatCompletionRequest` and `ChatCompletionParams`, so they
+    /// survive the request -> params conversion and serialize back out to the
+    /// backend verbatim (vLLM/sglang reads them directly). This is the request
+    /// half of hidden-state passthrough; the response half is covered by the
+    /// per-choice / delta catch-alls in `inference_providers::models`.
+    #[test]
+    fn return_hidden_states_request_fields_pass_through_to_backend() {
+        let req: ChatCompletionRequest = serde_json::from_str(
+            r#"{"model":"sglang-model","messages":[],"return_hidden_states":true,"layers":[0,5,11]}"#,
+        )
+        .unwrap();
+
+        let params: ChatCompletionParams = req.into();
+
+        // Land in the catch-all that is serialized straight to the backend.
+        assert_eq!(
+            params.extra.get("return_hidden_states"),
+            Some(&serde_json::json!(true))
+        );
+        assert_eq!(
+            params.extra.get("layers"),
+            Some(&serde_json::json!([0, 5, 11]))
+        );
+
+        // And survive serialization of the outgoing backend request body.
+        let wire = serde_json::to_value(&params).unwrap();
+        assert_eq!(wire["return_hidden_states"], serde_json::json!(true));
+        assert_eq!(wire["layers"], serde_json::json!([0, 5, 11]));
+    }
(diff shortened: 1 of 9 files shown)


PierreLeGuen (approved):
Reviewed the diff and surrounding chat completion request/response paths, including vLLM response parsing, streaming chunk reserialization, auto-redact reserialization, and external provider constructors. The new flattened catch-alls preserve hidden-state and other unknown provider fields on the reserialized paths without changing the raw passthrough path.

Local checks run:
- `cargo test -p inference_providers hidden_states --lib`
- `cargo test -p inference_providers response_preserves_top_level_unknown_fields_round_trip --lib`
- `cargo test -p api return_hidden_states_request_fields_pass_through_to_backend --lib`

GitHub checks at review time: lint, claude-review, and security_audit passing; Test Suite still pending.

Evrard-Nil:
Companion inference-proxy PR opened: nearai/inference-proxy#163 — teaches the SSE reassembler to preserve `delta.hidden_states` (+ unknown fields) at choice level. Both are needed for the **non-streaming** path; this cloud-api PR handles cloud-api's own re-serialize (auto-redact) path. Streaming already works end-to-end via the existing `ChatDelta.extra`.

PR #741, opened 2026-06-08
P1: group providers by attestation (attested::nearai / non_attested) + extract FleetRouter

Evrard-Nil:
## Summary

Two coordinated, behavior-preserving changes to the `inference_providers` crate, both groundwork for the inference-stack refactor (epic #735):

1. **Regroup the crate by attestation.** Providers are now organized along the axis that actually matters for trust — *is the backend attested?* — instead of by engine/implementation name. `vllm/` → `attested/nearai/`, `external/` → `non_attested/external/`, with a new `attested/` parent ready for attested third parties (e.g. Chutes) as siblings of `nearai`.
2. **Extract the NEAR-AI routing state into a `Fleet`.** All model-proxy routing machinery moves out of the ~3.6k-line provider module into a cohesive `Fleet` (`attested/nearai/fleet.rs`), leaving `nearai::Provider` a thin `InferenceProvider` adapter over a single `Arc<Fleet>`.

**Runtime behavior is unchanged.** The only externally visible change is the renamed module/type paths of this crate — a deliberate, breaking API migration (see *API compatibility* below). This is **P1** of #735; its sole purpose is to create a clean seam for **P3** (the GLM-5.1 signature-fetch fix) and the **P2** capability registry.

## Why rename (`vllm` was a misnomer)

`vllm` named an *engine*, but the type was never about vLLM specifically — these are **NEAR AI's own TEE-hosted, attested backends served via model-proxy**, and today the engine behind them is frequently **SGLang, not vLLM**. The name actively misled readers about what the code routes to.

The new taxonomy keys on the property the inference stack branches on — attestation:

```
inference_providers/src/
├── attested/
│   └── nearai/        (was vllm/)   NEAR's TEE fleet  → nearai::{Provider, Config, Fleet}
│       ├── mod.rs                   thin InferenceProvider adapter + trait impl
│       ├── fleet.rs                 Fleet: all model-proxy routing state/logic
│       └── prefix_router.rs
└── non_attested/
    └── external/      (was external/)  plaintext 3p (OpenAI/Anthropic/Gemini) — type names unchanged
```

This is exactly the grouping P2's `ProviderTier { Near, Attested3p, NonAttested }` + capability registry build on, and it gives attested-3p providers an obvious home (`attested/<provider>/`) without another reshuffle later.

### Renamed public items
| Before | After |
|---|---|
| `inference_providers::VLlmProvider` | `inference_providers::nearai::Provider` |
| `inference_providers::VLlmConfig` | `inference_providers::nearai::Config` |
| `inference_providers::vllm::*` | `inference_providers::attested::nearai::*` (re-exported as `nearai`) |
| `inference_providers::external::*` | `inference_providers::non_attested::external::*` (still re-exported at crate root) |

## API compatibility

This **is a breaking change to this crate's public surface**, and it is intentional — not papered over with shims. Rationale:

- `inference_providers` is an **unpublished, internal workspace crate** (`version = 0.0.0`). Its *only* consumers are `crates/api` and `crates/services`, both `path` dependencies in this same workspace. There are no external/downstream consumers to break.
- **All in-workspace call sites are updated in this PR**, and `cargo build --workspace` is green. The migration is complete, not partial.
- **No compatibility shims** (`pub use … as VLlmProvider`, etc.) were added on purpose: a `vllm` re-export would re-introduce the exact misleading name this change exists to remove, and would leave a second name for the same type for future readers to trip over. For an internal crate with all callers in-tree, a clean rename is strictly better than a deprecated alias.

(Earlier review noted the description claimed "public surface unchanged" — that was wrong and is corrected here: the runtime behavior is unchanged; the public *names* deliberately are not.)

## Why extract `Fleet`

model-proxy load-balances per-TCP at L4 with no application session affinity. To send a completion **and its later signature fetch** to the *same* backend, the provider accumulated a lot of routing machinery — prefix-bucket sticky H2 clients, a rotation-SNI fallback addressed by a positional `u64` index (`<canonical>-i<N>`), and signature/bucket pin maps — all interleaved with the I/O methods, attestation, TLS-SPKI pinning, and E2EE handling in one file.

That **positional rotation index is the root cause of the production GLM-5.1 `"…does not match any attested fingerprint"` signature-fetch failures**: `-i<N>` resolves to `sort_by_address(healthy)[N % len]`, recomputed live, so the index a completion was served on can point at a *different* backend by the time the signature is fetched — whose cert isn't in the pinned set. The fix (**P3**) is to address backends by a **stable handle** instead of a positional index, but that is unsafe to attempt inside the current monolith. P1 isolates the routing state + logic into `Fleet` so P3 can swap `u64 → BackendHandle` in one place without touching the completion path.

## What changed

- **New `attested::nearai::fleet::Fleet`** owns everything model-proxy-specific: the prefix router + per-bucket sticky clients; rotation-SNI addressing (`rotation_parts` / `rotation_count` / `rotation_url`); the four pin maps (`pending`/`signature` × `bucket`/`rotation`); the healthy-backend count; and the per-provider `config`, HTTP clients, TLS roots, `FingerprintState`, backend verifier, and verification semaphore.
- **All routing / attestation / I/O methods moved onto `Fleet`**: prefix routing, pin/unpin, rotation fallback (`try_chat_completion[_stream]_rotation`), `get_or_verify_bucket_client` / `pre_warm` / `clear_bucket`, header + rotation-client building, E2EE + tracing header prep, and the entire `InferenceProvider` impl (chat / text / image / audio / score / rerank / embeddings / privacy / models / attestation / `get_signature`).
- **`nearai::Provider` is reduced to `{ fleet: Arc<Fleet> }`**: constructors build a `Fleet` and wrap it; the pool-facing methods and the `InferenceProvider` impl delegate to `self.fleet`.
- **Caller updates**: `services::inference_provider_pool` and the integration tests now use `nearai::{Provider, Config}`; `non_attested::external` and crate-root re-exports keep their names.

## Design decisions

- **Single mechanical extraction — no `FLEET_ROUTER` dual-path flag.** A behavior-preserving *move* can't change behavior; duplicating ~1k lines of hot-path code behind a runtime toggle adds drift risk for no gain. The flag belongs with P3, where there's an actual behavioral difference (handle vs index) to gate.
- **Inherent helpers named distinctly from trait methods.** `Fleet`'s `InferenceProvider` impl forwards to inherent helpers; those helpers are named `pin_chat` / `unpin_chat` / `store_backend_count` (not the trait names) so the forwarding is unambiguous and a removed/renamed helper is a *compile error*, not silent self-recursion. This is the same class of bug as the `get_attestation_report` recursion caught in review (now fixed + regression-tested).
- **`fleet: Arc<Fleet>`** so `pre_warm` can keep spawning background bucket-verification tasks (`self: Arc<Self>`).
- **Inner `Mutex` instead of `Arc<Mutex>`** for the relocated maps — they were never cloned independently and are now owned by the shared `Arc<Fleet>`, removing a redundant refcount layer.

## Expected impact

- **Runtime behavior: none.** Same routing, retries, attestation, E2EE, and signature logic.
- **Performance: negligible.** One extra `Arc` deref on the provider→fleet hop; one fewer `Arc<Mutex>` layer on the pin maps. No new hot-path allocations.
- **Risk: low.** No DB / wire / config / migration changes. The change is confined to one internal crate's structure; all callers are in-tree and updated.
- **Unblocks:** P3's `u64 → BackendHandle` swap (the GLM-5.1 fix) now sits behind a single `Fleet` seam; P2's capability registry builds on the `attested::nearai` grouping + thin-wrapper shape; attested 3p providers get a home under `attested/`.

## Testing

Kept green and validated end-to-end at every sub-step:

- **Characterization tests** added first, pinning pre-extraction behavior: the sticky/signature routing state machine (incl. the empty-`chat_id` asymmetry) and `get_signature`'s fetch + 404-retry precedence over a `TcpListener` mock backend. Plus a regression test that drives `get_attestation_report` through `nearai::Provider` to catch forwarding mistakes.
- `cargo test -p inference_providers --lib` → **234 pass**; `cargo fmt --all -- --check`; `cargo clippy`; `cargo build --workspace` (proves every caller compiles against the renamed API).
- **Full local `cargo test --test e2e_all` → 546 passed** (real Postgres, server end-to-end). The inference path runs through `MockProvider` in e2e, so e2e covers workspace/integration while the unit tests cover `Fleet`'s own logic.
- The timing-sensitive `test_timeout_does_not_trigger_bucket_clear_retry` is now `#[serial]` (it asserts wall-clock elapsed and flaked under parallel-pool CPU contention).

## Rollout

No flag, no migration, no config change — a self-contained refactor + internal rename. Reviewable commit-by-commit: regroup modules → characterization tests → small behavior-preserving relocations (state → addressing → buckets → fields → helpers → trait impl) → delegation hardening.

## Follow-up

- **P3** swaps the positional rotation index for a stable, address-derived `BackendHandle` inside `Fleet` (depends on model-proxy#38, already merged, being live on both proxy peers).
- **P2** (ProviderRegistry + capability handles + `ProviderTier`) builds on the `attested::nearai` grouping and thin-wrapper shape.


PierreLeGuen (changes requested):
Blocking issue found during local review.

- `crates/inference_providers/src/vllm/mod.rs:2167`: `VLlmProvider`'s `InferenceProvider::get_attestation_report` delegator calls `self.get_attestation_report(...)`, which resolves back to the same trait method on `VLlmProvider` because there is no inherent provider method left after the move. Any attestation report request through the normal provider wrapper will recurse forever / overflow instead of reaching the moved `FleetRouter` implementation at line 1283. This should delegate to `self.fleet.get_attestation_report(...).await`, and it needs a small test that calls the method through `VLlmProvider` so future extraction steps catch forwarding mistakes.

Checks run locally: `cargo test -p inference_providers` (passed), `git diff --check origin/main...HEAD` (passed). GitHub checks were green when reviewed.

Evrard-Nil:
Great catch @PierreLeGuen (and the claude-review bot) — fixed in the latest push.

`VLlmProvider::get_attestation_report` now delegates to `self.fleet.get_attestation_report(...).await` like every sibling. Root cause: it was the only **multiline** `self.fleet` receiver in the delegating impl (long signature), so the `s/self\.fleet$/self/` wrap-form flip I used during the 4c move clipped its `.fleet` and it resolved back to itself → unconditional recursion → stack overflow on `/v1/attestation/report` through a real `VLlmProvider`. The green run missed it because e2e exercises inference via `MockProvider` and `unconditional_recursion` is only warn-level.

Added the regression test you suggested: `get_attestation_report_delegates_to_fleet_without_recursing` drives the call through `VLlmProvider` and asserts it returns (a transport error, no backend) rather than overflowing — so any future forwarding mistake in the wrapper is caught. I also re-audited all 16 delegators; this was the only one affected (the rest are single-line `self.fleet.X`). 234 unit + fmt + clippy + `services` build green; re-running the full local e2e now. Re-requesting review.

PierreLeGuen (approved):
Reviewed current head 2698e0d after the delegation fix. I inspected the FleetRouter extraction in crates/inference_providers/src/vllm/fleet.rs and crates/inference_providers/src/vllm/mod.rs, including the VLlmProvider adapter, signature/attestation forwarding, bucket/rotation pin state, constructor state transfer, and service call sites that use Arc<dyn InferenceProvider>. No blocking issues found.\n\nChecks run locally: cargo fmt --check, git diff --check origin/main...HEAD, cargo test -p inference_providers (234 unit tests, 8 integration tests, doc tests passed; existing ignored tests unchanged). GitHub checks on the PR head are also passing: Lint, Test Suite, security_audit.

Evrard-Nil:
**Update — two changes since the last review:**

**1. Fixed the `get_attestation_report` recursion** (your earlier blocker): now delegates to `self.fleet.get_attestation_report(...)`, plus a regression test (`get_attestation_report_delegates_to_fleet_without_recursing`) that drives it through `VLlmProvider`/`nearai::Provider` and asserts it returns rather than overflowing.

**2. Scope expanded per maintainer request — attestation taxonomy + rename.** `vllm` was a misnomer (it's NEAR AI's own TEE-hosted attested fleet, often SGLang not vLLM). The crate is now grouped by the real axis — **attestation** — to prepare for attested 3p providers (e.g. Chutes):

```
attested/nearai/   (was vllm/)   -> nearai::Provider / nearai::Fleet / nearai::Config
non_attested/external/ (was external/) -> OpenAI/Anthropic/Gemini (type names unchanged)
```

`nearai::*` (bare names under the `attested::nearai` module, re-exported via `pub use attested::nearai`); `non_attested::external::*` stays re-exported at the crate root so non-attested callers are unchanged. Pure rename/move, no behavior change.

> Heads-up: this grew the PR from the reviewed mechanical extraction into extraction + rename (folded in at the maintainer's request rather than a separate PR), so it's worth a fresh pass. Git tracks the moves as renames, so the diff stays readable.

**All green:** 234 unit + fmt + clippy + `cargo build` (workspace) + integration-test compile + **local `e2e_all` → 546 passed** at each step. Re-requesting review.

PierreLeGuen (changes requested):
Requesting changes for an API compatibility regression.

- `crates/inference_providers/src/lib.rs:58` now exposes only the new `attested`/`non_attested` layout plus `nearai`, and the old public `vllm`/`external` modules and `VLlmProvider`/`VLlmConfig` re-exports are gone. Any caller using the previous public API (`inference_providers::{VLlmProvider, VLlmConfig}`, `inference_providers::vllm::...`, or `inference_providers::external::...`) no longer compiles. Since the PR is described as refactor-only/no behavior change with unchanged public constructors, please keep compatibility shims/re-exports or make this an explicit breaking API migration.

Checks: `cargo fmt --all -- --check` passed; `cargo check -p services` passed; `cargo test -p inference_providers --lib` failed locally twice on `attested::nearai::tests::test_timeout_does_not_trigger_bucket_clear_retry` at `crates/inference_providers/src/attested/nearai/mod.rs:3223` (the isolated test passed, the same full command passed once on `origin/main`, and GitHub's Test Suite is green).

Evrard-Nil:
@PierreLeGuen thanks — addressed all three points. Head is now `HEAD` after two commits (delegation hardening + a stale-doc-path fix).

**1. API compatibility (your blocker).** Taking the "explicit breaking API migration" option rather than shims, and I've rewritten the PR description to say so honestly (the old "public surface unchanged" line was wrong — corrected). Reasoning:

- `inference_providers` is an unpublished internal workspace crate (`version = 0.0.0`); its only consumers are `crates/api` and `crates/services`, both `path` deps in this workspace. No external/downstream consumers.
- All in-workspace call sites are updated in this PR — I grepped the whole tree for `VLlmProvider` / `VLlmConfig` / `inference_providers::vllm` / `inference_providers::external` and the only remaining hit was a stale doc-comment path, now fixed. `cargo build --workspace` is green, so every caller compiles against the renamed API.
- No `pub use … as VLlmProvider` shims on purpose: re-exporting the `vllm` name would reintroduce the exact misnomer this change removes and leave two names for one type. For an internal crate with all callers in-tree, a clean rename beats a deprecated alias.

**2. Copilot's delegation footgun (lines 1273/1277/1281).** Fixed the same way the `get_attestation_report` recursion was: the inherent helpers on `Fleet` are now named distinctly from the trait methods — `pin_chat` / `unpin_chat` / `store_backend_count` — so the `impl InferenceProvider for Fleet` forwarders can't accidentally resolve to themselves. A removed/renamed helper is now a compile error instead of silent self-recursion. (The `get_attestation_report` delegator is fixed + covered by a regression test that drives it through `nearai::Provider`.)

**3. Flaky `test_timeout_does_not_trigger_bucket_clear_retry`.** Marked `#[serial]` — it asserts wall-clock elapsed (< 1700ms single timeout vs ~2s erroneous retry), so it shouldn't run under the parallel pool's CPU contention. That matches what you saw (failed twice under the full `--lib` run, passed isolated and on CI).

Local re-run on the new head: `cargo build --workspace`, `cargo test -p inference_providers --lib` (234 pass), `cargo fmt --all -- --check`, `cargo clippy`, and full `cargo test --test e2e_all` (546 pass). Re-requesting review.

PierreLeGuen (commented):
No blocking issues found in the provider grouping / Fleet extraction.

Inspected the diff around `inference_providers` module exports, `nearai::Provider` -> `Fleet` delegation, signature pin/unpin routing state, services provider-pool construction, and stale `vllm`/`external` import references. GitHub CI currently reports Lint, Test Suite, and security_audit passing.

Local checks run:
- `git diff --check origin/main...HEAD` passed
- `cargo check -p inference_providers` passed
- `cargo check -p services` passed
- `cargo test -p services --lib` passed: 343 passed, 1 ignored
- `cargo test -p inference_providers --lib` had one timing-sensitive failure in `attested::nearai::tests::test_timeout_does_not_trigger_bucket_clear_retry` at 1.713s vs the 1.700s bound; rerunning that exact test passed in 1.06s. Since CI/nextest passed and the exact test passed alone, I’m treating this as a non-blocking flake, but it is worth tightening because `Makefile` still has a plain `cargo test --lib --bins` path and `#[serial]` does not isolate the test from non-serial harness load.

Evrard-Nil:
@PierreLeGuen good catch on `#[serial]` — you're right that it only serializes against other `#[serial]` tests and does nothing about the non-serial async load that actually skews the wall-clock, and the `Makefile`'s plain `cargo test` path wouldn't honor it anyway.

Fixed properly in `d2d73409` by making the test **deterministic** instead of tightening the bound: the test already had a timing-independent assertion that captures the exact regression — `accept_count == 1`. A bucket-clear retry on timeout opens a *second* backend connection, so one TCP accept proves no retry fired. I dropped the redundant `elapsed < 1700ms` assertion (and the now-unneeded `#[serial]`) and rely on the connection count. No coverage lost — the behavioral check is strictly more precise than the latency proxy, and it can't flake under harness load.

Verified by running the full `cargo test -p inference_providers --lib` (the flaky condition) 3× back-to-back: 234 pass each, no flake. fmt + clippy clean.

PierreLeGuen (approved):
Reviewed the Fleet extraction and provider module reshuffle locally, focusing on Provider -> Fleet trait delegation, signature/attestation routing state, public re-exports/path updates, and services call sites. I did not find substantive issues.

Local checks run:
- cargo fmt --check
- cargo test -p inference_providers (234 unit tests, 8 integration tests, doc tests)
- cargo test -p services (343 unit tests + attestation verification tests)

GitHub checks are also green: Lint, Test Suite, security_audit.

PR #742, opened 2026-06-08
fix(build): refresh stale openssl pins to restore reproducible build

Evrard-Nil:
## Problem

`Verify Reproducible Build` has been failing on **every push since ~2026-05-25** (60+ consecutive red runs). Cross-machine builds (ubuntu-latest/22.04/24.04) all match and so does the *first* sequential build — only the **second sequential build** (same runner, after `docker system prune`) diverges.

## Root cause

Diffing the two debug OCI tars the workflow uploads on mismatch shows the images are byte-identical except for **two lines** in `/etc/apt/preferences.d/pinned-packages`:

```
- Pin: version 3.0.18-1~deb12u2     (libssl3 + openssl)
+ Pin: version 3.0.20-1~deb12u1
```

The committed runtime pins for `libssl3`/`openssl` (`3.0.18-1~deb12u2`) **no longer exist in the bookworm mirror** — Debian shipped the `3.0.20-1~deb12u1` security update and dropped 3.0.18. With `Pin-Priority: 1001` on a missing version, apt installs `3.0.20` anyway. So:

1. Build #1 bakes the stale `3.0.18` pin text into the image, but `build-image.sh`'s post-build step extracts the *actually-installed* `3.0.20` and overwrites `pinned-packages-runtime.txt`.
2. Build #2 reuses that working dir, bakes `3.0.20` → different digest.

The pins no longer round-trip, which is exactly what the sequential idempotency check is designed to catch. (Direct proof the version is unavailable: build #1 was *given* the 3.0.18 pin yet still installed 3.0.20.)

## Fix

Bump the two runtime pins to the version apt actually installs today (`3.0.20-1~deb12u1`), restoring the build's fixed point. Evidence-scoped: the builder-stage pins did **not** drift in the diff, so they're left untouched.

## Follow-up (not in this PR)

This will recur on every Debian security bump that removes the pinned version. Durable options worth discussing: pin apt to a `snapshot.debian.org` timestamped index so pinned versions never disappear, or stop baking `preferences.d/pinned-packages` into the final layer. Tracking separately.


PierreLeGuen (approved):
Inspected the one-file diff in pinned-packages-runtime.txt plus the surrounding Dockerfile, build-image.sh, and verify-reproducible-build workflow. The updated libssl3/openssl runtime pins match the versions installed by the runtime bootstrap path, and a lightweight local reproduction of the runtime APT sequence round-tripped exactly back to the committed pinned-packages-runtime.txt.\n\nChecks: git diff --check passed; local runtime pin round-trip passed; GitHub PR checks Lint, Test Suite, security_audit, and claude-review are passing.

PR #744, opened 2026-06-08
fix(responses): propagate upstream errors

lloydmak99:
## Summary
- preserve completion/provider errors through the Responses API stream task
- include typed error/status metadata on response.failed events
- return non-streaming /v1/responses failures as HTTP errors instead of 200 status=failed fallbacks
- add e2e coverage for mocked 429 and missing-model non-streaming Responses API failures

Fixes #511

## Validation
- cargo fmt --all -- --check
- cargo check -p services -p api
- cargo test -p api --test e2e_all --no-run

Runtime e2e execution requires local Postgres at localhost:5432; this environment does not have Postgres listening and Docker is not installed.


Evrard-Nil (approved):
🤖 Reviewed locally at `0950cac` (cloned, built, ran the masking unit tests, traced the error paths). **Safe to approve — the key security invariant is preserved.** Approving.

The #1 thing I checked: does propagating upstream errors risk leaking **401/403** to clients (they signal *our* backend-creds bug, must be masked as 5xx)? **No.** This PR consumes the already-masked `services::completions::CompletionError` produced at the single masking site `map_provider_error` (`crates/services/src/completions/mod.rs:687`): 401/403 → `ProviderError{status:500, "model currently unavailable"}` (raw message + IPs dropped), 429 → `RateLimitExceeded`, 5xx → 502, vLLM-404 → 400. The new `completion_http_status_code` (`responses/errors.rs`) only ever sees that masked status — it can't observe a raw 401/403. Verified:
- **Non-streaming** (`responses.rs:521-527`) routes through `map_domain_error_to_status` — the *same* helper chat completions uses → 401/403 client-facing = **500**, 429 = **429**, missing-model = **400**.
- **Streaming** `response.failed` (`service.rs:247-259`) derives status/body from the same masked error. Both paths mutually consistent.
- Unit test `test_map_provider_error_401_masks_auth_details` passes (asserts 500, no key/IP leak); new e2e `test_responses_provider_error_429_propagated_non_streaming` asserts 429. Info-leak: auth/5xx use the fixed generic message; only genuine client-side 4xx (400/422/vLLM-404) forward the provider message, identical to existing chat-completions behavior — not new.

Local: `cargo fmt --check` clean, `cargo check -p services -p api` clean, `cargo clippy` clean, `cargo test -p api --test e2e_all --no-run` compiles, `map_provider_error` units 18/18 pass. (Full e2e needs local Postgres — couldn't run, but the masking is unit-covered and the path provably can't unmask.)

NITs (none blocking):
1. `responses/errors.rs` `completion_http_status_code`/`completion_response_error` duplicate the 502/503/504 mapping already in `conversions.rs:219-224` + `common.rs:18` — 3 copies that could drift; consider delegating to one.
2. **Pre-existing (not this PR):** a bare **407** from a vLLM backend isn't in the `(401|403)` mask arm and would pass through — the documented invariant names 401/403/**407**. Shared verbatim with chat-completions, not worsened here, and 407 isn't realistically emitted by our backends — worth a separate follow-up to add 407 (and 511) to the mask.
3. Cheap add: an explicit `test_responses_provider_error_401_masked_non_streaming` asserting client gets 500 (not 401) would lock the invariant at the API boundary (the 429 test is a ready template). Coverage is adequate today via the services unit test, so optional.

Nice fix — closes the 200-status=failed-fallback gap for non-streaming and threads typed error metadata cleanly. LGTM.

PierreLeGuen (changes requested):
Requesting changes for one correctness gap.

- `crates/services/src/responses/service.rs:1160` / `crates/services/src/responses/service.rs:1221` / `crates/services/src/responses/service.rs:1309`: `process_completion_stream` now captures provider stream failures as `stream_error_cause`, but `process_response_stream` still treats any agent-loop `Err` with partial output or usage as `ResponseStatus::Completed` and falls through to `emit_completed`. That means a provider stream that emits some text and then returns 429/5xx is surfaced as a successful completed Responses response, and the non-streaming route will return 200 with partial output instead of propagating the upstream failure. The added tests cover failures before any output; please distinguish upstream provider errors from client-disconnect partial-save behavior and add coverage for partial output followed by an upstream error.

Local checks run: `cargo fmt --all -- --check`; `cargo check -p services -p api`; `cargo test -p api --test e2e_all --no-run`.

Evrard-Nil (changes requested):
🤖 Withdrawing my earlier approval — @PierreLeGuen is right, and I under-weighted this. Concurring with his CHANGES_REQUESTED.

My review had actually spotted the same mid-stream path (`process_response_stream` / the `service.rs:723-731` in-stream error branch that sets `stream_error=true` and breaks without populating the typed `failed_error`/`status_code`), but I mis-scoped it as "pre-existing / out of scope for #511." That was wrong: a provider stream that emits partial output and **then** returns 429/5xx being reported as `ResponseStatus::Completed` → `emit_completed` (200 with partial output, non-streaming) directly contradicts this PR's own goal of *propagating upstream errors*. A whole class of upstream error is still swallowed as success. So it belongs in this PR, not a follow-up.

The nuance Pierre names is the right framing: **distinguish an upstream provider error from a client-disconnect partial-save.**
- partial output → client disconnect: legitimately save partial / complete.
- partial output → upstream 429/5xx (`stream_error_cause` set at `service.rs:1160/1221/1309`): must surface as `response.failed` (streaming) and a non-200 HTTP error (non-streaming), carrying the same masked status the initial-failure path now produces.

And +1 to his test request: add coverage for **partial output followed by an upstream error** (the current new tests only exercise failure *before* any output, which is why this slipped through — mine included).

The initial-failure propagation + 401/403→5xx masking I verified earlier still hold and are good; this is the remaining gap before I can re-approve. Happy to re-review promptly once the mid-stream upstream-error path is threaded through and the partial-then-error test is added.

lloydmak99:
Addressed the requested mid-stream error gap in latest commit `1bee7af`.

Changes:
- Provider stream errors with partial output now mark the response as `failed`, update the stored response status to failed, and return `response.failed` instead of falling through to `response.completed`.
- Client-disconnect / local emitter failures without a typed provider cause still keep the existing partial-save behavior.
- Added mock support for partial output followed by a provider stream error.
- Added regression coverage: `test_responses_partial_output_then_stream_error_429_non_streaming`.

Local validation:
- `cargo check -p inference_providers -p services -p api`
- `cargo clippy -p inference_providers -p services -p api --all-targets -- -D warnings`
- `cargo test -p api --test e2e_all --no-run`

Evrard-Nil (approved):
🤖 Re-reviewed the force-pushed fix at `1bee7af` (cloned, built, ran `provider_errors` against local Postgres — 11 passed). **The gap is fully fixed — clearing my changes-request / approving.**

The mid-stream partial-output-then-upstream-error case is now handled correctly, with a clean typed discriminant:
- `process_completion_stream`'s `Err(e)` arm (`service.rs:724-740`) — items are `Result<SSEEvent, inference_providers::CompletionError>`, so an `Err` here is always an **upstream** error → mapped via `map_provider_error` → `ResponseError::Completion` carried out as `stream_error_cause`.
- A failed SSE write to the client (`service.rs:656-689`) sets `stream_error=true` but leaves `stream_error_cause=None` → **client disconnect**.
- The branch that tells them apart (`service.rs:1444-1450`): upstream → `Completion(_)`; disconnect → `StreamInterrupted`. Streaming emits `response.failed` (`service.rs:233-271`) instead of `emit_completed`; non-streaming returns the event's non-2xx status (`responses.rs:525-531`) instead of 200. Partial is persisted, then the error is returned — not silently completed.

Non-regressions I specifically checked:
- **401/403→5xx masking preserved** on BOTH the initial-failure and the new mid-stream paths (both go through `map_provider_error` → masked 500; 429→429, vLLM-404→400). No raw upstream auth status can reach the client mid-stream.
- **Client disconnect still saves partial as Completed** (`StreamInterrupted`→`Completed`, `service.rs:1164/1292`) — no false failures on normal cancels.
- The exact coverage requested is present: `test_responses_partial_output_then_stream_error_429_non_streaming` streams a partial then 429 and asserts **HTTP 429, not 200/completed** (plus stream-created-429 and missing-model-400 cases). All pass locally.

`cargo fmt --check` / `cargo check` / `cargo clippy -p services -p api` all clean.

Optional NITs (non-blocking): the new error tests are all non-streaming — a streaming-mode assertion (collect SSE, expect a `response.failed` frame after partial deltas) would close the loop; and a responses-path 401→500 test would lock the masking at the API boundary (logic is shared with chat-completions, so it's structurally guaranteed today). Nice fix — thanks for the quick turnaround. (Note @PierreLeGuen's review is also on the prior head; deferring to him to clear his.)

PierreLeGuen (approved):
Reviewed the Responses API error propagation changes, including the non-streaming event collection path, typed response.failed metadata, completion/provider error mapping, mock stream-error scaffolding, and the new e2e coverage. Also verified the branch merges cleanly into current origin/main.

Checks run locally:
- cargo fmt --all -- --check
- cargo check -p services -p api
- cargo test -p api --test e2e_all --no-run
- merged current origin/main with the PR locally and reran cargo check -p services -p api

I did not find substantive blocking issues.

PR #745, opened 2026-06-09
PR1: add ProviderTier { Near, Attested3p, NonAttested } (Chutes attested-provider scaffolding)

Evrard-Nil:
## Summary

**PR1 of the Chutes third-party-attested-provider integration.** Adds a single trust-tier enum, `ProviderTier`, to the `inference_providers` crate. Pure scaffolding — a `pub enum` + `is_attested()` + one unit test, **no behavior change**. Later PRs use it to select a per-provider measurement policy and report-data verifier at pool-construction time.

## Background

PR #741 reorganized `inference_providers` along the **attestation axis** — `attested::nearai` (NEAR's own TEE fleet) and `non_attested::external` (plaintext 3p) — with an `attested/` parent that already anticipates third-party attested providers (its module doc literally says *"Future: attested 3p (e.g. Chutes)"*). This enum names that axis as a type so the verifier and pool can branch on it.

The branch is **trust**, not engine or operator: *can we cryptographically verify what served the response?*

## The tiers

| Variant | Maps to | Meaning |
|---|---|---|
| `Near` | `attested::nearai` | NEAR AI's own TEE fleet. Full chain: per-request TDX quote + GPU evidence + TLS-SPKI pin + per-response signature, rooted in NEAR's own KMS / compose-manager / gateway. |
| `Attested3p` | `attested::<provider>` (e.g. Chutes) | A third party with a *verifiable, response-bound* TEE attestation. Earns the "attested" badge **only** when the per-response bindings NEAR's verifier requires (fresh nonce, signing key, TLS fingerprint, authenticated measurement) are present; otherwise it degrades to `NonAttested`. |
| `NonAttested` | `non_attested::external` | Plaintext third parties (OpenAI / Anthropic / Gemini / OpenRouter). No verified badge. |

`is_attested()` returns true for `Near` and `Attested3p` — the gate for the "verified" badge.

## Why an enum now (decision + example)

The Chutes design (manually validated against Chutes' real `/evidence` endpoint) requires the verifier to apply **per-provider** policy: a Chutes TDX measurement must be *mathematically incapable* of validating a NEAR backend, and an `Attested3p` provider must be held to stricter checks (e.g. TCB-up-to-date enforced, no padded-address `report_data` fallback) than NEAR's own fleet inherits by default. Selecting that policy needs a tier discriminator chosen at pool construction — never derived from attacker-influenced report fields. This enum is that discriminator.

Example (later PR, illustrative):
```rust
let (policy, report_data_verifier) = match tier {
[...]

PR #746, opened 2026-06-09
PR2: per-provider fail-closed MeasurementPolicy (Chutes attested-provider)

Evrard-Nil:
## Summary

**PR2 of the Chutes attested-provider integration** (stacked on #745's `ProviderTier`). Introduces a per-provider, **fail-closed** `MeasurementPolicy` and threads it through `AttestationVerifier`. NEAR's own-fleet verification path is **byte-for-byte unchanged**; the new behavior only applies to attested third-party providers (none wired yet — that's PR7).

No provider uses the `Attested3p` policy at runtime in this PR; this is the verifier groundwork that PR7 (Chutes) selects per-tier at pool construction.

## The problem it fixes

Today the OS-image-hash gate is a single global env (`ALLOWED_IMAGE_HASHES`) that is **fail-open**: when empty it *skips* the check entirely (`verification.rs` old `:224`/`:232`), and one `Arc<AttestationVerifier>` is shared across every provider. That is acceptable for NEAR's own fleet, but for a third-party attested provider it has two holes:

1. **Fail-open on misconfiguration** — an empty allowlist would silently accept *arbitrary* software measurements.
2. **Cross-provider bleed** — a measurement allowlisted for one provider would also satisfy another's backend.

## What changed

- **New `crates/services/src/attestation/measurement.rs`** — `MeasurementPolicy { tier, allowed_os_image_hashes, require_tcb_up_to_date }` with **private fields** (construction only via safe constructors):
  - `near(..)` / `near_from_env()` — reproduce the *exact* previous semantics (empty allowlist = skip). Allowlist entries are normalized (trim, strip `0x`, lowercase).
  - `attested3p(..)` — `Attested3p` tier; `require_tcb_up_to_date` defaults **true** (stricter than NEAR). Allowlist entries are normalized.
  - **`assert_enforceable()`** — the fail-closed guard: an `Attested3p` policy with an empty allowlist is **unconditionally rejected**, never skipped. No-op for `Near` (keeps skip). The guard no longer depends on any flag a caller can flip.
  - `enforces_image_hash()` — true iff the allowlist is non-empty.
  - `allows_image_hash(hash)` — normalizes the lookup hash before checking, so `0xDEADBEEF` and `deadbeef` are equivalent.
- **`AttestationVerifier`** now holds a `MeasurementPolicy` instead of the bare `allowed_image_hashes` + `require_tcb_up_to_date` fields.
  - `new(allowed, pccs, require_tcb)` and `from_env()` keep their signatures and build a **`Near`** policy → existing callers/tests untouched, NEAR behavior identical.
  - New `with_policy(policy, pccs)` constructor for per-provider verifiers (PR7).
  - `verify_attestation_report` now calls `self.policy.assert_enforceable()?` as **step 0** (before any quote work), and reads the TCB flag + image-hash allowlist from the policy.
  - NEAR error string preserved verbatim: `"ALLOWED_IMAGE_HASHES configured but no os-image-hash in event log"` — log-based monitoring is unaffected.

## Decision: keep NEAR fail-open-on-empty, make 3p fail-closed

We deliberately do **not** flip NEAR to fail-closed in this PR — that would be a behavior change to the production fleet (some envs run without `ALLOWED_IMAGE_HASHES`) and is out of scope. The fail-closed semantics are gated strictly on `ProviderTier::Attested3p`, and are **unconditional** — there is no flag a caller can set to bypass them:

```rust
pub fn assert_enforceable(&self) -> Result<(), AttestationVerificationError> {
    if self.tier == ProviderTier::Attested3p && self.allowed_os_image_hashes.is_empty() {
        return Err(AttestationVerificationError::ImageHashMismatch(
            "attested third-party provider has no os-image-hash allowlist configured \
             (fail-closed: empty allowlist would accept arbitrary software)".into(),
        ));
    }
    Ok(())
}
```

Because fields are private and construction is only via safe constructors, it is **impossible** to assemble an `Attested3p` policy in a fail-open state. Because the tier is chosen at construction from `ProviderTier` (never from a report field), a Chutes policy can never validate a NEAR backend or vice-versa.

## Tests

- `measurement::tests` (5): empty-Near enforceable+skips, non-empty-Near enforces, **empty-Attested3p rejected** (unconditional), non-empty-Attested3p enforced+TCB-required, **allowlist entries normalized** (`0xDEADBEEF` matches `deadbeef`).
- `attestation_verification_test` (2, deterministic, no network):
  - `attested3p_empty_allowlist_fails_closed_before_verification` — proves the guard fires *first* (an empty report yields the fail-closed error, not `MissingField("intel_quote")`).
  - `near_empty_allowlist_does_not_fail_closed` — proves `Near` still reaches real verification (historical behavior preserved).
- **Existing GLM-5 / Qwen3.5 fixture tests pass unchanged** (4/4) → NEAR path intact.
- Full `services` lib suite 348 pass; `cargo build --workspace` green; fmt + clippy clean; local `cargo test --test e2e_all` green.

## Impact

- **Runtime behavior:** none today (no `Attested3p` provider wired). NEAR identical.
- **Risk:** low — additive type + a refactor that preserves the `new`/`from_env` API; the only new control-flow (the fail-closed guard) is unreachable for `Near`.
- **Unblocks:** PR3 (`ReportDataVerifier` strict) and PR7 (Chutes `with_policy(chutes_policy(), ..)`).

🤖 Generated with <a href="https://claude.com/claude-code">Claude Code</a>

### DIFF
--- a/crates/services/src/attestation/measurement.rs
+++ b/crates/services/src/attestation/measurement.rs
@@ -9,9 +9,11 @@
 //!
 //! [`MeasurementPolicy`] makes the policy explicit and **per-provider**, selected
 //! by [`ProviderTier`] at construction time — never derived from attacker-
-//! influenced report fields. NEAR's behavior is reproduced exactly by
-//! [`MeasurementPolicy::near_from_env`] / [`MeasurementPolicy::near`], so the
-//! existing verification path is byte-for-byte unchanged.
+//! influenced report fields. Its fields are private: a policy can only be built
+//! through the safe constructors below, so it is impossible to construct an
+//! attested-tier policy that silently skips its measurement check. NEAR's
+//! behavior is reproduced exactly by [`MeasurementPolicy::near_from_env`] /
+//! [`MeasurementPolicy::near`].
 
 use std::collections::HashSet;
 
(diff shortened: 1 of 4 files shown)


PierreLeGuen (changes requested):
Requesting changes for one security-sensitive policy invariant issue.

Finding:
- `crates/services/src/attestation/measurement.rs:104`: `assert_enforceable()` only fails closed for `Attested3p` when `require_dstack_event_log` is true, but that flag is public and is not actually used by `AttestationVerifier` to change verification behavior in this PR. A caller can construct `MeasurementPolicy { tier: ProviderTier::Attested3p, allowed_os_image_hashes: HashSet::new(), require_dstack_event_log: false, ... }`, pass it to `with_policy()`, and the verifier will still require/replay an event log but will skip the OS image hash allowlist entirely. That contradicts the PR’s core invariant that Attested3p + empty allowlist is rejected fail-closed, and it creates a footgun for the later provider wiring. Until a real non-event-log fallback exists and enforces its own measurement pins, please make the fields private / route construction through safe constructors, or reject empty Attested3p allowlists unconditionally.

Checked locally:
- `cargo test -p services attestation::measurement -- --nocapture` passed (4 unit tests)
- `cargo test -p services --test attestation_verification_test attested3p_empty_allowlist_fails_closed_before_verification -- --nocapture` passed
- `cargo test -p services --test attestation_verification_test near_empty_allowlist_does_not_fail_closed -- --nocapture` passed
- `cargo check -p services` passed
- `git diff --check origin/main...HEAD` passed

GitHub checks were also green when refreshed (Lint, Test Suite, security_audit, claude-review).

Evrard-Nil:
Thanks all — addressed in `311d7510`.

**@PierreLeGuen (blocking) + @gemini-code-assist — the `require_dstack_event_log` footgun.** You're right: that flag gated the fail-closed check but had no other effect in this PR, so `{ tier: Attested3p, allowed_os_image_hashes: empty, require_dstack_event_log: false }` would pass `assert_enforceable()` while the OS-image-hash check was skipped, and the register-pin path it implied doesn't exist yet. Fixed by **removing the field entirely from this PR** — the guard is now **unconditional** for `Attested3p` + empty allowlist:

```rust
pub fn assert_enforceable(&self) -> Result<(), AttestationVerificationError> {
    if self.tier == ProviderTier::Attested3p && self.allowed_os_image_hashes.is_empty() {
        return Err(/* fail-closed */);
    }
    Ok(())
}
```

The flag will return in the later PR that adds the real, **measurement-enforcing** register-pin fallback (where it actually changes verification behavior and is enforced), not as a no-op knob that can disable safety.

**@PierreLeGuen — safe construction.** `MeasurementPolicy` fields are now **private**; a policy can only be built via `near()` / `near_from_env()` / `attested3p()`, with accessors (`tier()`, `require_tcb_up_to_date()`, `enforces_image_hash()`, `allows_image_hash()`). It's no longer possible to assemble an attested-tier policy in a fail-open state.

**@Copilot — allowlist normalization.** All constructors now normalize entries (trim, strip `0x`, lowercase); `allows_image_hash()` normalizes the looked-up hash too. Added a test (`allowlist_entries_are_normalized`) covering `0xDEADBEEF` matching a bare lowercase hash.

**@Copilot — error string.** Restored the original `"ALLOWED_IMAGE_HASHES configured but no os-image-hash in event log"` text so NEAR log-based monitoring is unaffected. The "byte-for-byte unchanged for NEAR" claim now holds (the only constructor-level behavior change is case-insensitive allowlist matching, which is strictly more lenient).

Local: 5 policy unit tests + 2 fail-closed integration tests, GLM-5/Qwen fixtures unchanged, services lib 348 pass, `e2e_all` 546 pass, fmt + clippy clean. Re-requesting review.

PierreLeGuen (approved):
Reviewed the current head (311d751) locally, including the full diff in `crates/services/src/attestation/{measurement,verification}.rs`, the attestation tests, existing verifier call sites in the provider pool, PR discussion, and GitHub checks. The prior fail-open construction issue is fixed: `MeasurementPolicy` fields are private, the unused event-log flag is removed, and `Attested3p` + empty allowlist now fails closed unconditionally before quote parsing.

Local checks run:
- `cargo test -p services measurement::tests --lib`
- `cargo test -p services --test attestation_verification_test allowlist`

GitHub checks were green when refreshed: Lint, Test Suite, security_audit.

Issue #43, opened 2025-10-15
Bug: fail to query `/model/{model_name}` with public model name


think-in-universe:
After #7 is implemented, now we can configure the public model names. When querying with [`/models`](https://nearai-platform-api-dev-59501de18ae5.herokuapp.com/docs#/Models/models) endpoint, we have the below model list with public model names (except `nearai/gpt-oss-120b` which has another issue #42).

```json
{
  "object": "list",
  "data": [
    {
      "id": "deepseek/deepseek-chat-v3-0324",
      "object": "model",
      "created": 0,
      "owned_by": "system"
    },
    {
      "id": "deepseek/deepseek-chat-v3.1",
      "object": "model",
      "created": 0,
      "owned_by": "system"
    },
    {
      "id": "nearai/gpt-oss-120b",
      "object": "model",
      "created": 0,
      "owned_by": "system"
    },
    {
      "id": "qwen/qwen3-30b-a3b-instruct-2507",
      "object": "model",
      "created": 0,
      "owned_by": "system"
    }
  ]
}
```

But if we query the `[/model/{model_name}`](https://nearai-platform-api-dev-59501de18ae5.herokuapp.com/docs#/Models/get_model_by_name) endpoint with public model name `deepseek/deepseek-chat-v3.1`, the below error response is received.

```json
{
  "error": {
    "message": "Model 'deepseek/deepseek-chat-v3.1' not found",
    "type": "model_not_found",
    "param": null,
    "code": null
  }
}
```

This breaks the model page in the frontend. e.g. the [deepseek v3.1 model details page](https://nearai-cloud-dev-git-feat-api-migration-near-agent.vercel.app/models/deepseek/deepseek-chat-v3.1)

<img width="425" height="269" alt="Image" src="https://github.com/user-attachments/assets/77f1e818-9474-4a51-a71a-26e0f92541ed" />

PR #747, opened 2026-06-09
feat(admin): list organization members endpoint

Evrard-Nil:
## What

Adds `GET /v1/admin/organizations/{org_id}/members` — an admin-authenticated endpoint that lists the members of **any** organization with full user details (email, last login, active status), consistent with `/v1/admin/users`.

## Why

The cloud-api admin UI needs to show an org's members from a per-org overview page. The existing member-facing `GET /v1/organizations/{org_id}/members` endpoint:
- requires the caller to be a **member** of the org (403 otherwise) — platform admins aren't members of arbitrary orgs, and
- returns only **public** user fields (no email/role/last-login).

So it can't back an admin view. This new endpoint reuses the admin auth middleware (`AdminAppState` + `AdminUser`) and does not require membership.

## Changes

- **ports**: `AdminOrganizationMemberInfo`; `AdminRepository::{list_organization_members, count_organization_members, organization_exists}`; `AdminService::list_organization_members`.
- **repo** (`admin_composite`): `organization_members ⋈ users` join (active users only), count, and existence check.
- **route**: `200` list / `404` unknown org / `401` unauthenticated. The raw DB role string is mapped to `MemberRole` with a safe `Member` fallback so a malformed row can never be misrepresented as owner/admin.
- **models**: `ListAdminOrganizationMembersResponse`; registered in OpenAPI.
- **e2e**: owner-listed-with-details, org-not-found (404), unauthenticated (401).

## Testing

`cargo build`, `cargo clippy`, `cargo fmt` clean. The 3 new e2e tests pass against a local Postgres:

```
test admin_organization_members::test_admin_list_organization_members_unauthorized ... ok
test admin_organization_members::test_admin_list_organization_members_includes_owner ... ok
test admin_organization_members::test_admin_list_organization_members_org_not_found ... ok
```

## Related

Frontend consumer: nearai/nearai-cloud-admin-ui#43 (org overview + members page). That PR is gated on this endpoint being deployed.

## Update — also adds a single-org getter

During review (and to support the UI overview in nearai-cloud-admin-ui#43), this PR also adds **`GET /v1/admin/organizations/{org_id}`** — a single active org by id (404 on missing/inactive), reusing the org SELECT via a shared row mapper. Covered by `test_admin_get_organization_ok_and_not_found` (200 / 404 / 404-deactivated).


PierreLeGuen (changes requested):
Requesting changes for one correctness issue plus the current CI-blocking lint failure.

- `crates/database/src/repositories/admin_composite.rs:1054-1056` and `:1098-1100`: the admin organization-member list/count joins `users` but filters `u.is_active = true`. This makes deactivated users silently disappear from both `members` and `total`, even though users are soft-deleted in place (`users.is_active = false`) and the response exposes `user.is_active`; `/v1/admin/users` also includes inactive users by default unless an `is_active` filter is supplied. Please either include inactive members in this admin endpoint or add an explicit active-status filter, and add coverage for an inactive member.
- `crates/api/tests/e2e_all/admin_organization_members.rs:34`: `cargo fmt --all -- --check` fails on the multi-line `assert!`. This is already failing the PR's Lint check; run rustfmt before merging.

Local checks run: `cargo check -p services -p database -p api` passed. `cargo fmt --all -- --check` failed as noted. I did not run the full e2e suite locally.

lloydmak99:
## Code Review

**Verdict: Approve with one question**

Straightforward admin endpoint, consistent with the pattern used by `list_users` and `list_organizations`. Route, handler, service, and repo layers all updated in step. E2E tests cover the three essential cases.

### Question

**`is_active = true` filter on the SQL query** (`admin_composite.rs`)
The query joins on `u.is_active = true`, so a deactivated user who is still an org member won't appear in the response — but `total` would need to match that filtered count, not the full member count. Verify that the count query (not visible in this diff) also applies `AND u.is_active = true`; if it counts all members and the list only returns active ones, pagination arithmetic will be off for orgs with deactivated members.

Also worth deciding: should an admin listing org members see *all* members including deactivated ones? The `is_active` field is returned in the response, so the UI could filter client-side rather than dropping rows server-side.

### Nits
- `member_role_from_db_str` fallback to `Member` on unknown role values is the right safe default.
- `Extension(_admin_user)` correctly enforces auth without needing the user value.

lloydmak99 (approved):
Reviewed — approve. Clean admin endpoint consistent with existing patterns. Question on is_active=true filter: verify the count query also applies the same filter so pagination totals stay consistent. See comment for details.

Evrard-Nil:
Thanks @PierreLeGuen @lloydmak99 — addressed in 7a97f526:

- **Inactive members now included.** Dropped `u.is_active = true` from both the list **and** count queries, so soft-deleted members appear (matching `/v1/admin/users`) and `total` stays consistent with the rows. The response already exposes `user.is_active` for the UI to badge.
- **Coverage added:** new e2e `test_admin_list_organization_members_includes_inactive` — creates an org (active owner) + a soft-deleted member, asserts `total == 2` and that the inactive member is listed with `is_active == false`.
- **Lint:** rustfmt was already fixed in b5c4b5f6; Lint is green.
- **@lloydmak99's count-consistency question:** confirmed — list and count apply identical filters (now: none beyond `organization_id`).
- **gemini:** existence check now only runs when `total == 0`, saving a round-trip in the common case.
- **Copilot:** `organization_exists` trait signature reflowed onto one line.

All 4 e2e tests pass locally against Postgres.

PierreLeGuen (changes requested):
The endpoint itself is solidly built — admin middleware covers the new route, the list and count queries use matching filters, full-detail exposure matches the existing `/v1/admin/users` precedent, the role-string fallback fails toward least privilege, and there are e2e tests for the happy path, inactive members, 404, and 401. Two issues need fixing before merge.

**1. Inconsistent handling of soft-deleted organizations (blocking).**
`organization_exists` (`crates/database/src/repositories/admin_composite.rs:1110`) filters `is_active = true`, but `list_organization_members` (:1056) and `count_organization_members` (:1102) join only `organization_members`/`users` with no constraint on the org's activity — and org deletion is a soft delete that keeps member rows (`crates/database/src/repositories/organization.rs:328`). Since the existence check only runs when `total == 0` (`crates/services/src/admin/mod.rs:670`), a deactivated org **with** members returns 200 with the full member list, while a deactivated org **without** members returns 404. The same org id flips between "exists" and "not found" based on member count, and the result disagrees with `/v1/admin/organizations` (:982), which hides inactive orgs. Fix: pick one behavior — either add `JOIN organizations o ON o.id = m.organization_id AND o.is_active = true` (or a `WHERE EXISTS` guard) to both list and count so deactivated orgs always 404, or drop the `is_active` filter from `organization_exists` so admins can always inspect a deactivated org's members. Either way, add e2e coverage for the deactivated-org case, and while there, a test for an active org with zero members (the path that exercises `organization_exists`).

**2. Unrelated ops changelog swept into the PR (important).**
Commit 7a97f52 adds a 105-line `CHANGELOG-prod-to-staging.md` — a prod→staging deploy verification doc covering PRs #708–#728 with internal operational details (container image digests, staging hostnames like `cloud-stg-api.near.ai`, live incident notes, internal test-infra results). It has nothing to do with this feature and pollutes the PR's scope. Drop it from this PR; if the team wants deploy changelogs versioned, do that in a dedicated PR/location.

**Minor.**
- `crates/database/src/repositories/admin_composite.rs:1056` — `ORDER BY m.joined_at DESC` has no unique tiebreaker; members created in the same transaction (e.g., bulk invites) share a timestamp, so paginated pages can repeat or skip rows. Use `ORDER BY m.joined_at DESC, m.id`.

**Checks run locally:** `cargo fmt --all -- --check`, `cargo check -p api -p services -p database` (including `--tests`), and `cargo clippy -p api -p services -p database -- -D warnings` — all clean. The new e2e tests were not run locally (no PostgreSQL available); CI's Test Suite, Lint, and security audit are green on the PR head.

Evrard-Nil:
Thanks @PierreLeGuen — second round addressed in 165e4413:

**1. Soft-deleted-org inconsistency (blocking).** Went with "deactivated orgs always 404" to match `/v1/admin/organizations` (which hides inactive orgs). Both the list and count queries now `JOIN organizations o ON o.id = m.organization_id AND o.is_active = true`, so a deactivated org returns an empty set → `total == 0` → existence check → 404, regardless of surviving member rows. Coverage added:
- `test_admin_list_organization_members_deactivated_org` → 404
- `test_admin_list_organization_members_empty_org` → 200/empty (the path that actually exercises `organization_exists`)

**2. Unrelated ops changelog.** Dropped `CHANGELOG-prod-to-staging.md` from the branch — it was swept in by a `git add -A`. Gone from the PR.

**3. (minor) ORDER BY tiebreaker.** Now `ORDER BY m.joined_at DESC, m.id`.

Also, to support the UI overview's request to fetch one org directly (raised on #43), I added `GET /v1/admin/organizations/{org_id}` here (single org by id, 404 on missing/inactive, shared row mapper with the list). Covered by `test_admin_get_organization_ok_and_not_found` (200 / 404 / 404-deactivated).

All 7 e2e tests pass locally; `cargo fmt`/`clippy -D warnings` clean.

PierreLeGuen (commented):
This adds `GET /v1/admin/organizations/{org_id}/members` plus a single-org getter `GET /v1/admin/organizations/{org_id}`. The implementation is solid: admin auth is enforced by the existing `admin_middleware` layer, pagination is validated with a stable sort tiebreaker (`joined_at DESC, m.id`), soft-deleted orgs consistently 404 across list/count/existence checks (with tests for empty-org, deactivated-org, and inactive-member edge cases), the single-org SQL reuses a shared row mapper, and logging emits IDs only. Exposing member emails/last-login to platform admins matches the existing `/v1/admin/users` surface. No blocking issues — a few non-blocking observations:

- `crates/api/src/routes/admin.rs:217` — the `utoipa::path` for `list_organization_members` omits a `400` response, but `validate_limit_offset` returns 400 for `limit <= 0`, `limit > 1000`, or negative offset. Other admin endpoints (e.g. lines 162, 894) document 400; add `(status = 400, description = "Invalid request", body = ErrorResponse)` so the OpenAPI contract matches behavior.
- `crates/api/src/routes/admin.rs:204` — `member_role_from_db_str` silently coerces unknown role strings to `Member`, whereas `role_str_to_domain_role` in `crates/database/src/repositories/organization.rs:78` errors on invalid roles. The DB `CHECK (role IN ('owner','admin','member'))` constraint makes this unreachable today, but if the constraint is ever relaxed an admin would silently see wrong data. Consider logging a warning (role string + member_id only) on the fallback arm.
- `crates/api/src/routes/admin.rs:156` and `:257` — `error!(...)` fires on every 404 for unknown/deactivated orgs, so probing nonexistent org IDs floods error-level logs. Match `OrganizationNotFound` first and log it at `debug!`/`warn!`, reserving `error!` for the `InternalError` arm.
- Process note: the PR title/body describe only the members endpoint, but the diff also ships the single-org getter. It's well-tested and consistent, but the description should mention it so the new admin surface isn't missed by reviewers or changelog tooling.

Also verified: the `get_organization as get_admin_organization` import alias avoids the collision with the non-admin handler and the route registration uses the aliased name correctly; the diff contains only the 9 feature-relevant files.

Checks run locally: `cargo fmt --all -- --check` clean; `cargo clippy` across `api`/`services`/`database` (all targets) clean with no warnings; the 7 new e2e tests (`cargo test --test e2e_all admin_organization_members`) all pass against a fresh Postgres container with CI-matching env. GitHub CI (Lint, Test Suite, security_audit) is green on the PR head. The full e2e suite and vLLM integration tests were not run locally; CI's Test Suite covers them.

Evrard-Nil:
Thanks @PierreLeGuen — all four non-blocking items addressed in f8aaab2f:

1. **400 in the OpenAPI contract** — added `(status = 400, ...)` to `list_organization_members` (it calls `validate_limit_offset`). `get_organization` takes no limit/offset, so no 400 there.
2. **`member_role_from_db_str` fallback** — now `warn!`s with the role string + `member_id` (ids only) instead of silently coercing.
3. **404 log spam** — both handlers now match `OrganizationNotFound` first and log at `debug!`, `Unauthorized` at `warn!`, and reserve `error!` for `InternalError`. Probing unknown org ids no longer floods error logs.
4. **Process note** — updated the PR description to call out the single-org getter (`GET /v1/admin/organizations/{org_id}`).

7/7 e2e pass; `fmt`/`clippy -D warnings` clean.

lloydmak99 (approved):
Reviewed updated diff — my is_active question is fully answered:

- Inactive *users* are now included (is_active filter removed from JOIN); is_active field is returned so the UI can render the status badge.
- Inactive *organizations* are excluded via JOIN (not WHERE), so both list and count apply the same filter — pagination totals stay consistent.
- organization_exists is called only when total == 0, avoiding a round-trip in the common case. The logic correctly distinguishes deactivated-org (404) from active-org-with-no-members (200/empty).
- row_to_admin_org_info helper eliminates the duplicated struct mapping between list_organizations and get_organization.
- New tests cover: inactive member included, empty-org 200, deactivated-org 404, get_org 200/404/deactivated, unauthorized. Comprehensive.

No new issues.

PierreLeGuen (commented):
The PR is in good shape: both new admin endpoints are registered inside the admin middleware layer (`crates/api/src/lib.rs:1693`) so auth is enforced consistently, the `ORDER BY m.joined_at DESC, m.id` pagination tiebreaker is backed by the UUID PK from the V0001 migration, the single-org SQL is field-for-field identical to `list_all_organizations` via the shared `row_to_admin_org_info` mapper, and the `total == 0 → organization_exists` check correctly distinguishes an empty org (200) from a missing/deactivated one (404). New log statements emit IDs only, and exposing member emails/last-login is intentional and admin-gated, consistent with `/v1/admin/users`.

- `crates/api/src/routes/admin.rs:1881` — `member_role_from_db_str` duplicates role semantics that already exist as `services::organization::ports::MemberRole` plus the conversions in `crates/api/src/conversions.rs:378`. The repo port returns a raw `String` role; parsing in the repository (where the CHECK constraint lives) and reusing the existing conversion would avoid a second source of truth. The fallback's `warn!` also logs the raw unexpected DB value — harmless given the CHECK constraint, but an ID-only message would be safer under the project's logging rules.
- `crates/api/tests/e2e_all/admin_organization_members.rs` — no test exercises `limit`/`offset` on the members endpoint, even though the code comments call out the pagination tiebreaker specifically. A two-member org paged with `limit=1` would lock in that behavior cheaply.
- `crates/api/src/routes/admin.rs:1795` — the `get_organization` handler re-implements the `AdminOrganizationInfo → AdminOrganizationResponse` mapping (spend-limit scale 9, usage block) that `list_organizations` already contains; a shared helper would keep the two from drifting.
- `crates/services/src/admin/mod.rs:924` — the list/count/exists queries run on separate connections with no shared snapshot, so `total` can briefly disagree with the returned rows under concurrent membership changes. This mirrors the existing `list_organizations` pattern and is fine for an admin UI; if exact consistency ever matters, a single query with `COUNT(*) OVER ()` or one transaction would fix it.

Local checks: `cargo fmt --all -- --check` clean; `cargo check -p api -p services -p database --tests` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test --test e2e_all admin_organization_members` against a temporary Postgres container (CI-matching env, `DEV=true`) — all 7 new tests pass; regression runs of `admin_analytics` (24 passed) and `general` (53 passed), which cover the refactored org-list mapping. CI on the PR head (Lint, Test Suite, security_audit) is green.

Evrard-Nil:
Rebased onto `main` (resolved a conflict with the merged pricing-changes work, #758) and addressed your non-blocking items in a1ef5ffd:

- **Shared response mapper** — extracted `admin_org_info_to_response`, now used by both `list_organizations` and `get_organization`, so the spend-limit scale / usage block can't drift.
- **ID-only logging** — the `member_role_from_db_str` fallback now logs by `member_id` only (no raw DB role value).
- **Pagination test** — added `test_admin_list_organization_members_pagination`: a two-member org paged with `limit=1` returns each member exactly once across pages, locking in the `joined_at DESC, m.id` tiebreaker.
- **Separate connections / snapshot** — acknowledged; left as-is since it mirrors the existing `list_organizations` pattern and is fine for an admin UI (noted in your review).

The `member_role_from_db_str` → repo-side parsing suggestion: I kept the mapping at the API boundary for now (consistent with how the handler converts the other fields) rather than threading a typed role through `AdminOrganizationMemberInfo`; happy to do that refactor in a follow-up if you'd prefer.

⚠️ **Local test caveat:** my Docker test-Postgres host is wedged at 100% disk, so I could only confirm the rebased branch **compiles + is fmt/clippy-clean** this round — the e2e assertions ran green twice before the rebase, and CI runs the full suite on its own Postgres. Watching CI now.

PierreLeGuen (approved):
This PR adds `GET /v1/admin/organizations/{org_id}/members` and `GET /v1/admin/organizations/{org_id}`, and the implementation is careful and consistent with existing admin patterns.

**What I verified:**

- **Auth**: Both handlers require `Extension<AdminUser>` and are registered under `build_admin_routes`, matching every other admin endpoint; the 401 e2e test confirms enforcement.
- **Privacy**: All log statements emit IDs only (the `warn!` in `member_role_from_db_str` and the debug-level 404 logging are deliberately ID-only). Member emails appear only in the response body, gated behind admin auth, consistent with `/v1/admin/users`.
- **Correctness**: Pagination orders by `joined_at DESC, m.id` (UUID PK tiebreaker) and is locked in by a no-repeat test. List, count, and existence queries all apply the same `JOIN ... AND o.is_active = true` filter, so totals can't drift from rows and soft-deleted orgs consistently 404 (tested). The role-string fallback to `Member` is unreachable today given the DB CHECK constraint and fails toward least privilege. The single-org query shares `row_to_admin_org_info` with the list query, preventing mapper drift.
- **Contracts**: New paths and schemas are registered in OpenAPI with no duplicates; limit/offset reuse the existing `validate_limit_offset` (1–1000, non-negative).

Two non-blocking notes, neither requiring changes:

- `crates/services/src/admin/mod.rs:1035` — the list and count queries run via `tokio::join!` on separate pool connections without a shared snapshot, so `total` can briefly disagree with the returned page under concurrent membership changes. This matches every other paginated admin endpoint; only worth revisiting if the team ever moves to single-query `COUNT(*) OVER ()` pagination.
- `crates/api/src/routes/admin.rs:2192`, `:2283` — both handlers match `AdminError::Unauthorized`, which the underlying service methods can never return (they only produce `OrganizationNotFound`/`InternalError`). Dead but harmless arms copied from the established handler pattern; fine to keep for uniformity.

**Checks run:**
- `cargo check -p api -p services -p database --tests` — clean
- `cargo clippy` (`-D warnings`, both targeted packages and workspace/all-targets) — clean
- Targeted e2e (`cargo test --test e2e_all admin_organization_members` against local PostgreSQL with `DEV=true`) — all 8 new tests pass (owner included, inactive members included, empty org 200, deactivated org 404, pagination no-repeat, org-not-found 404, single-org getter 200/404, unauthorized 401)
- GitHub CI: Lint, Test Suite, and security_audit passed

lloydmak99 (approved):
Approve — polish commits look correct. Shared org mapper reduces duplication, new paging tests cover offset arithmetic, and the ID-only warn avoids logging user data on role fallback. No new issues from my previous review.

PR #748, opened 2026-06-09
PR3: strict report_data binding for all attested tiers, NEAR included (remove padded-address fallback)

Evrard-Nil:
## Summary

**PR3 of the Chutes attested-provider integration** (follows #745 `ProviderTier`, #746 `MeasurementPolicy`). Replaces the inline `report_data` check with a pluggable `ReportDataVerifier` and applies a **single strict binding rule to every attested provider — NEAR's own fleet included**: `report_data[0:32]` MUST equal `SHA256(signing_address ‖ tls_cert_fingerprint)` (a TLS fingerprint is **mandatory**) and `report_data[32:64]` MUST equal the per-request nonce. The legacy zero-padded-`signing_address` fallback is **removed**.

> **This is a deliberate NEAR hardening, not a no-op refactor.** Per review feedback: the code will be audited and we want NEAR held to the same bar as third parties — no weaker path for our own fleet. See *Validation* for why it's safe and what must be checked in staging before prod.

## Background — why the fallback is removed

A TDX quote's 64-byte `report_data` is sealed inside the Intel-signed quote and binds it to *this* request and *this* backend:

- `report_data[32:64]` = caller's per-request nonce (**freshness**).
- `report_data[0:32]` = `SHA256(signing_address ‖ tls_cert_fingerprint)` — binds the signing key **and** the live TLS endpoint into the quote.

The verifier historically also accepted a fallback when no fingerprint was present: `report_data[0:32] = signing_address` zero-padded. That fallback **drops the TLS co-binding** — an attacker who can influence the JSON `signing_address` could satisfy `report_data[0:32]` against a quote not bound to the TLS endpoint we're actually talking to (a `breaks_trust=critical` finding in the design's adversarial review). There is no reason NEAR should retain a weaker path than a third party, so the fallback is removed for everyone.

## What changed

New `crates/services/src/attestation/report_data.rs`:
- **`trait ReportDataVerifier`** — `verify(report_data, signing_address, tls_cert_fingerprint, nonce)`.
- **`StrictBoundReportDataVerifier`** — the single verifier used for all attested tiers: requires `tls_cert_fingerprint` (rejects when absent), re-checks the nonce, verifies the `SHA256(addr‖fp)` binding.

`AttestationVerifier` holds an `Arc<dyn ReportDataVerifier>` (the seam is kept so a provider with a *different* `report_data` layout can plug its own verifier in a later PR), set to `StrictBoundReportDataVerifier` for every tier. `new()` / `from_env()` (NEAR) and `with_policy()` (3p, PR7) all get strict. The unused `signing_algo` argument is dropped.

## Validation — why this is safe for NEAR

NEAR backends already bind the TLS fingerprint into `report_data`; the padded-address fallback was never exercised in practice. Evidence:

- The existing **live fixture tests now run under the strict verifier** and pass: `test_verify_glm5_attestation` and `test_verify_qwen35_attestation` fetch *real* attestation reports from production NEAR backends and verify them strictly (they also assert `tls_cert_fingerprint.is_some()`). So real NEAR reports carry the binding.
- `report_data::tests` (4): accepts the fingerprint binding; **rejects missing fingerprint**; **rejects stale-nonce replay**; rejects a wrong fingerprint binding.
- PR2 fail-closed tests unchanged; full `services` lib green; `cargo build --workspace` green; local `cargo test --test e2e_all` green; fmt + clippy clean.

**Staging gate before prod promotion:** this changes NEAR verification, so the full infra-tests staging suite must be run across **all** NEAR models after deploy. Any NEAR model whose report does not bind a fingerprint would now be (correctly) rejected and drop from rotation — that would surface as that model failing attestation/discovery in staging, and must be fixed on the backend before promotion. The 2 fixture models are already proven; the rest are validated by the staging suite.

## Impact

- **Behavior:** NEAR `report_data` verification is now strict (no fallback). Intended to be a no-op for the live fleet (fingerprint always present), but it *is* a behavior change — hence the staging gate above. No `Attested3p` provider is wired yet.
- **Risk:** medium-low and self-announcing — a NEAR model lacking a fingerprint fails loudly in staging rather than silently downgrading trust. Mitigated by fixtures (2 models proven) + full staging validation.
- **Unblocks:** PR7 (Chutes) reuses the same strict binding.

### DIFF
--- a/crates/services/src/attestation/mod.rs
+++ b/crates/services/src/attestation/mod.rs
@@ -1,9 +1,11 @@
 pub mod measurement;
 pub mod models;
+pub mod report_data;
 pub mod verification;
 use dstack_sdk::dstack_client;
 pub use measurement::MeasurementPolicy;
 pub use models::{AttestationError, ChatSignature, SignatureLookupResult};
+pub use report_data::{NearReportDataVerifier, ReportDataVerifier, StrictBoundReportDataVerifier};
 use std::sync::Arc;
 pub use verification::{AttestationVerificationError, AttestationVerifier, VerifiedAttestation};
 
--- a/crates/services/src/attestation/mod.rs
+++ b/crates/services/src/attestation/mod.rs
@@ -5,7 +5,7 @@ pub mod verification;
 use dstack_sdk::dstack_client;
 pub use measurement::MeasurementPolicy;
 pub use models::{AttestationError, ChatSignature, SignatureLookupResult};
-pub use report_data::{NearReportDataVerifier, ReportDataVerifier, StrictBoundReportDataVerifier};
+pub use report_data::{ReportDataVerifier, StrictBoundReportDataVerifier};
 use std::sync::Arc;
 pub use verification::{AttestationVerificationError, AttestationVerifier, VerifiedAttestation};
 
--- a/crates/services/src/attestation/report_data.rs
+++ b/crates/services/src/attestation/report_data.rs
@@ -8,14 +8,17 @@
 //!   when a TLS fingerprint is present, binding the signing key **and** the live
 //!   TLS endpoint into the quote.
 //!
-//! NEAR's own fleet has historically also accepted a *fallback* form for the
-//! first 32 bytes when no fingerprint is present: `signing_address` zero-padded
-//! to 32 bytes. That fallback drops the TLS co-binding, so for an **attested
-//! third party** (whose nodes we do not operate) it is unsafe — an attacker who
-//! can set `signing_address` could satisfy the check without binding the
-//! connection. [`StrictBoundReportDataVerifier`] therefore requires the
-//! fingerprint binding and forbids the padded-address fallback, while
-//! [`NearReportDataVerifier`] preserves today's behavior byte-for-byte.
+//! The verifier has historically also accepted a *fallback* for the first 32
+//! bytes when no fingerprint was present: `signing_address` zero-padded to 32
+//! bytes. That fallback drops the TLS co-binding — an attacker who can set the
+//! JSON `signing_address` could satisfy the check against a quote that isn't
+//! bound to the TLS endpoint we're actually talking to. [`StrictBoundReportDataVerifier`]
+//! removes that fallback and requires the fingerprint binding for **every**
+//! attested provider, NEAR's own fleet included: a uniform, no-exceptions bar.
+//! NEAR backends already bind the fingerprint (the fallback was never exercised
+//! in practice), so this is a hardening, but it must be validated across the
+//! whole NEAR fleet in staging before promotion — any model that does not bind
+//! a fingerprint would now be (correctly) rejected.
 
 use sha2::{Digest as Sha2Digest, Sha256};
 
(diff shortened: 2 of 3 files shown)


PierreLeGuen (approved):
Reviewed the report-data verifier split, tier-based selection from MeasurementPolicy, the moved NEAR fallback behavior, strict Attested3p fingerprint binding, and related provider-pool call paths. I did not find substantive correctness or security regressions in this patch.

Local checks run:
- cargo fmt --check
- cargo test -p services --lib (354 passed, 1 ignored)

GitHub checks were still pending when reviewed, not failing.

lloydmak99:
## Code Review

**Verdict: Approve**

Clean, well-motivated extraction. The security reasoning in the module-level doc comment is clear and the trait boundary guarantees the verifier is selected at construction (from `policy.tier()`), never from a report field — that's the right invariant.

### Minor observations

**`_signing_algo` removal is intentional and correct** — the parameter was already unused (prefixed `_`) in the old `verify_report_data`; removing it tidies the call site without behaviour change.

**`bytes32` test helper panics on wrong length** (`report_data.rs:200`)
`a.copy_from_slice(&v)` panics if `hex::decode` returns ≠ 32 bytes. Fine for test-internal constants (they're hardcoded 64-hex-char strings), but worth a comment so a future test author doesn't reach for `bytes32` with an arbitrary hex string.

**`NearReportDataVerifier` padded-address branch copies at most `addr_bytes.len()` bytes** — correct, handles the case where the signing address is shorter than 32 bytes without panicking.

### Test coverage
All four meaningful cases are covered (Near+fp, Near no-fp, Strict+fp, Strict no-fp) plus stale-nonce and wrong-fingerprint cross-checks. Good.

Evrard-Nil:
**Scope change (pushed `24d9119c`):** per review, the strict `report_data` verifier now applies to **NEAR's own fleet too**, not just third parties — we want NEAR held to the same bar for the audit, with no weaker path for our own backends.

- Removed `NearReportDataVerifier` + the padded-`signing_address` fallback entirely. `StrictBoundReportDataVerifier` is the single verifier for all attested tiers: `tls_cert_fingerprint` is **mandatory**, `report_data[0:32] = SHA256(addr ‖ fp)`, `report_data[32:64] = nonce`.
- The `Arc<dyn ReportDataVerifier>` seam stays so PR7 can plug a different-layout verifier if a provider needs one.

**This is a deliberate NEAR behavior change** (no longer "byte-for-byte unchanged"). Why it's safe + how it's gated:
- The live fixture tests `test_verify_glm5_attestation` / `test_verify_qwen35_attestation` now run **under the strict verifier** against real production NEAR backends and pass → real NEAR reports carry the fingerprint binding; the fallback was never exercised.
- **Before prod promotion**, the full infra-tests staging suite must pass across **all** NEAR models. Any model not binding a fingerprint now fails loudly in staging (correct) rather than silently downgrading trust — that's the intended safety property.

Local: report_data unit tests (incl. rejects-missing-fp, rejects-stale-nonce-replay), PR2 fail-closed tests, services lib 352, `e2e_all` 546, fmt + clippy all green.

lloydmak99:
## Code Review — updated diff

Reviewed the updated diff (all tiers now strict, `NearReportDataVerifier` removed). Logic is clean and the behavior change is intentional and well-documented in the module comment.

**One stale doc comment contradicts the implementation** (`crates/services/src/attestation/verification.rs`, `report_data_verifier` field):

```rust
/// Per-tier `report_data` binding verifier. NEAR keeps the historical
/// behavior (TLS-fingerprint binding with a padded-address fallback); an
/// attested third party gets the strict verifier (fingerprint binding
/// required, no fallback). Selected from `policy.tier()` at construction.
report_data_verifier: Arc<dyn ReportDataVerifier>,
```

This doc was written for the earlier version of the PR where NEAR and 3p used different verifiers. The implementation now uses `StrictBoundReportDataVerifier` for all tiers — the padded-address fallback is gone for NEAR too. The comment says the opposite of what the code does, which will mislead a future reader trying to understand the trust model.

Suggested replacement:
```rust
/// `report_data` binding verifier — `StrictBoundReportDataVerifier` for all
/// attested providers, NEAR included. The `Arc<dyn>` seam lets a future
/// provider with a different report_data layout plug its own verifier.
report_data_verifier: Arc<dyn ReportDataVerifier>,
```

Everything else looks correct. The `claude[bot]` note about the live-cert comparison (fingerprint in report vs. actual TLS peer cert) is the remaining open security item for a future PR.

PR #749, opened 2026-06-09
PR4: per-tier GPU-evidence requirement policy (attested 3p must present GPU evidence)

Evrard-Nil:
## Summary

**PR4 of the Chutes attested-provider integration** (follows #745 `ProviderTier`, #746 `MeasurementPolicy`, #748 strict `ReportDataVerifier`). Makes **GPU-evidence-required** a per-tier policy. NEAR keeps its historical best-effort behavior; an attested third party that advertises confidential GPU compute must actually present verifiable, nonce-bound GPU evidence. NEAR behavior is unchanged.

This completes the per-tier verifier policy story — TCB floor (#746), report_data binding (#748), and now GPU evidence — before the Chutes provider code (PR5–PR7).

## Background

`verify_gpu_evidence` is best-effort today: a report with no `nvidia_payload` returns `Ok(None)` ("acceptable for non-GPU CVMs"). When GPU evidence *is* present, the caller-nonce binding (`payload.nonce == request_nonce`) and NVIDIA NRAS verification are already mandatory.

For NEAR's own fleet (mix of GPU and non-GPU CVMs) the absent-is-OK behavior is correct. But a third party like Chutes advertises NVIDIA confidential compute (Hopper) — if it simply omitted the GPU evidence, the current code would pass it with `gpu_verdict = None`, silently dropping the GPU half of the attestation. An attested 3p must prove what it claims.

## What changed

- `MeasurementPolicy` gains a private `require_gpu_evidence: bool` + `require_gpu_evidence()` accessor:
  - `near()` / `near_from_env()` → `false` (historical best-effort).
  - `attested3p()` → `true`.
- `verify_gpu_evidence`: the absent-`nvidia_payload` branch now consults the policy — `Ok(None)` when not required (NEAR), **`Err(GpuVerificationFailed)`** when required (attested 3p). The mandatory nonce binding + NRAS verification on present evidence are unchanged.

```rust
_ => {
    if self.policy.require_gpu_evidence() {
        return Err(GpuVerificationFailed("GPU evidence required by policy but absent"));
    }
    return Ok(None); // NEAR non-GPU CVMs
}
```

## Tests

- `measurement::tests::gpu_evidence_required_only_for_attested3p` — `near`/`near_from_env` → false, `attested3p` → true.
- `verification::tests::gpu_evidence_absent_ok_for_near_but_rejected_for_attested3p` — deterministic (no network; the absent branch returns before any NRAS call): NEAR verifier → `Ok(None)`, attested-3p verifier → `Err`.
- Existing GLM-5/Qwen fixture tests (GPU evidence present) verify unchanged; full `services` lib green; `cargo build --workspace` green; `cargo test --test e2e_all` green; fmt + clippy clean.

## Impact

- **Behavior:** none for NEAR (`require_gpu_evidence = false`, identical to today). No `Attested3p` provider is wired yet.
- **Risk:** low — additive policy field + one guarded branch; NEAR path unchanged.
- **Unblocks:** PR7 (Chutes) — `attested3p()` policy now requires the GPU evidence Chutes advertises.

### DIFF
--- a/crates/services/src/attestation/measurement.rs
+++ b/crates/services/src/attestation/measurement.rs
@@ -43,6 +43,11 @@ pub struct MeasurementPolicy {
     allowed_os_image_hashes: HashSet<String>,
     /// Reject attestations whose TDX TCB status is not `UpToDate`.
     require_tcb_up_to_date: bool,
+    /// Require GPU (NVIDIA NRAS) evidence to be present. When false, a report
+    /// with no `nvidia_payload` verifies with `gpu_verdict = None` (best-effort,
+    /// for NEAR's non-GPU CVMs). When true, absent GPU evidence is rejected — an
+    /// attested third party that claims confidential GPU compute must prove it.
+    require_gpu_evidence: bool,
 }
 
 impl MeasurementPolicy {
@@ -58,6 +63,9 @@ impl MeasurementPolicy {
                 .map(|s| normalize_hash(s))
                 .collect(),
             require_tcb_up_to_date,
+            // NEAR's own fleet keeps the historical best-effort GPU behavior
+            // (non-GPU CVMs verify with gpu_verdict = None).
+            require_gpu_evidence: false,
         }
     }
 
@@ -73,6 +81,9 @@ impl MeasurementPolicy {
                 .map(|s| normalize_hash(s))
                 .collect(),
             require_tcb_up_to_date: true,
+            // A third party advertising confidential GPU compute must present
+            // verifiable, nonce-bound GPU evidence — absent is rejected.
+            require_gpu_evidence: true,
         }
     }
 
@@ -104,6 +115,7 @@ impl MeasurementPolicy {
             tier: ProviderTier::Near,
             allowed_os_image_hashes,
             require_tcb_up_to_date,
+            require_gpu_evidence: false,
         }
     }
 
@@ -133,6 +145,12 @@ impl MeasurementPolicy {
     /// Whether attestations with a non-`UpToDate` TCB status must be rejected.
     pub fn require_tcb_up_to_date(&self) -> bool {
         self.require_tcb_up_to_date
+    }
+
+    /// Whether GPU evidence must be present (absent GPU evidence is rejected
+    /// rather than verifying with `gpu_verdict = None`).
+    pub fn require_gpu_evidence(&self) -> bool {
+        self.require_gpu_evidence
     }
 
     /// Whether the OS-image-hash allowlist is enforced for this policy.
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):
Reviewed the diff and surrounding attestation policy/verifier code. The new `require_gpu_evidence` policy is wired through `MeasurementPolicy::attested3p`, and the missing-`nvidia_payload` branch fails closed for attested 3P while preserving NEAR best-effort behavior.

Local checks passed: `cargo fmt --check`; `cargo test -p services attestation::measurement::tests --no-fail-fast`; `cargo test -p services gpu_evidence --no-fail-fast`; `cargo test -p services attestation::verification::tests --no-fail-fast`.

GitHub checks observed: lint, security_audit, and claude-review passed; Test Suite was still pending at review time. I found no substantive issues.

Evrard-Nil:
Good catch @gemini-code-assist (security-high) / @Copilot — fixed in `7d5b4083`.

The gap: `get("nvidia_payload").and_then(|v| v.as_str())` returns `None` when the field is **present but not a string** (object/number/bool), so a malformed payload was treated as "absent" and could silently skip GPU verification on the best-effort (NEAR) path — a type-confusion bypass.

Now the three cases are explicit, and present-but-malformed is never mistaken for absent:
```rust
let nvidia_payload_str = match attestation_report.get("nvidia_payload") {
    None | Some(Value::Null) => None,                       // absent
    Some(Value::String(s)) if s.is_empty() => None,         // absent
    Some(Value::String(s)) => Some(s.as_str()),             // verify this
    Some(_) => return Err(GpuVerificationFailed(            // MALFORMED -> always reject
        "nvidia_payload is present but is not a string (malformed report); refusing to skip GPU verification")),
};
```
- **absent / null / empty** → policy decides (`Ok(None)` for NEAR, `Err` for an attested 3p that requires GPU evidence).
- **non-empty string** → verified as before.
- **present but not a string** → `Err` for **every** tier — never silently skipped.

Test extended (`gpu_evidence_absent_ok_for_near_but_rejected_for_attested3p`): a non-string `nvidia_payload` now errors for both NEAR and attested-3p verifiers. services lib 354 + fixtures green; fmt + clippy clean. Re-requesting review.

PierreLeGuen (approved):
Reviewed the attestation policy and verifier changes in `crates/services/src/attestation/measurement.rs` and `crates/services/src/attestation/verification.rs`, including the policy constructors, absent/malformed `nvidia_payload` handling, and surrounding attestation flow.

Checks run locally:
- `cargo fmt --check`
- `cargo test -p services attestation`
- `cargo clippy -p services --all-targets -- -D warnings`

Remote checks were also green when refreshed: Lint, Test Suite, and security_audit. I did not find any blocking or substantive issues.

lloydmak99:
## Code Review

**Verdict: Approve**

Clean, well-scoped change. The explicit 3-way match (`absent/null/empty → None`, `non-empty string → verify`, `other type → error`) is the right fix for the type-confusion bypass, and the policy-driven reject-vs-best-effort split is implemented correctly.

### One gap not in prior comments

**Empty-string `nvidia_payload` untested under `require_gpu_evidence = true`** (`crates/services/src/attestation/verification.rs`)

The test `gpu_evidence_absent_ok_for_near_but_rejected_for_attested3p` covers the absent-key case and the non-string malformed case, but not `nvidia_payload: ""` (present, empty string). Per the new match arm, an empty string maps to `None`, then the policy check fires — so an attested 3p sending an empty payload should be rejected. Worth adding a test case:

```rust
let mut empty_payload = serde_json::Map::new();
empty_payload.insert("nvidia_payload".to_string(), json!(""));
// NEAR: best-effort → Ok(None)
assert!(matches!(near.verify_gpu_evidence(&empty_payload, "00").await, Ok(None)));
// attested 3p: require_gpu_evidence → rejected
assert!(a3p.verify_gpu_evidence(&empty_payload, "00").await.is_err());
```

Not a correctness bug (the code handles it correctly), but the coverage gap means a future refactor could accidentally merge the empty-string and non-empty-string branches without a test catching it.

### Nits already noted
- `claude[bot]` caught the stale `// best-effort` comment at verification.rs:253 — that one's the remaining cleanup item.

PR #750, opened 2026-06-09
Fix Responses in_progress status serialization

### LINKED ISSUE #621: Responses API streaming: status payload field is 'inprogress' (no underscore), should be 'in_progress'
## Summary

In the `/v1/responses` streaming API, the `status` field inside payload bodies during an in-progress response is `"inprogress"` (one word). The OpenAI spec uses `"in_progress"` (with underscore). Clients matching against the standard string never see "in progress" state.

Inconsistency: the **event name** `response.in_progress` does use the underscore correctly — only the **status field inside the JSON payload** drops it.

## Repro

```bash
curl -sN https://cloud-api.near.ai/v1/responses \
  -d '{"model":"openai/gpt-oss-120b","input":"Count to 3","max_output_tokens":80,"stream":true}' \
  | grep -oE '"status":"[a-z_]+"' | sort -u
# "status":"completed"
# "status":"inprogress"   ← should be "in_progress"
```

The `completed` value is correct; only `inprogress` is wrong. Non-streaming responses (`status: "completed"` at top level) are fine — this is streaming-payload-only.

## Suggested fix

Whatever serializes the streaming payload's `status` enum needs to use the kebab/underscore form. A `serde(rename = "in_progress")` on the `InProgress` variant (Rust) or equivalent.

Per OpenAI Responses API spec, valid status values are: `queued`, `in_progress`, `completed`, `failed`, `incomplete`, `cancelled`.

## While poking at /v1/responses (informational, not a bug to fix here)

Every response object echoes back a default `tools: [{"type":"web_search","search_context_size":"medium","user_location":{"type":"approximate","country":"US"}}]` whether or not the request included tools. Clients who serialize the response object verbatim and replay it as the next request will inherit web search. Worth confirming this is intentional.

---

lloydmak99:
## Problem

Responses streaming already emits the correct `response.in_progress` event name, but the embedded response object status serialized `ResponseStatus::InProgress` as `inprogress` because the enum used `serde(rename_all = "lowercase")`.

That does not match the OpenAI Responses API spelling (`in_progress`) or the internal database spelling, and it can break clients that validate the streamed status payload.

Fixes #621

## Change

- Serialize `ResponseStatus::InProgress` as `in_progress` in both API-facing and service Responses models.
- Serialize `ResponseItemStatus::InProgress` as `in_progress` as well, so output item status fields stay consistent with response status fields.
- Keep accepting the legacy `inprogress` spelling during deserialization with `serde(alias = "inprogress")`.
- Add API and service model regression tests for both the corrected serialization and legacy deserialization behavior.

## What Else Is Needed

- No database migration or backfill is needed. Database response statuses already use `in_progress`, and the new aliases keep legacy serialized payloads readable.
- After merge, deploy `cloud-api` so `/v1/responses` streaming clients receive the corrected status spelling.
- Deployment note: this PR starts emitting `in_progress` from new instances. During a rolling deploy, old instances from before this PR do not have the legacy-read alias for the new spelling, so a brief mixed-version window could affect long-lived pending `FunctionCall` or `McpApprovalRequest` response items. A fast/all-at-once deploy avoids the mixed-version window; a two-phase deploy with an accept-only alias first would eliminate it entirely.

## Tests

- `cargo test -p services response_status_`
- `cargo test -p api response_status_`
- `cargo test --lib --bins`


PierreLeGuen (approved):
Reviewed the enum serialization change in crates/api/src/models.rs and crates/services/src/responses/models.rs, plus the surrounding response item persistence and streaming paths. The snake_case change fixes the in_progress wire spelling while preserving the legacy inprogress read path, and I did not find a substantive regression.

Local checks run:
- cargo fmt --all -- --check
- cargo test -p services response_status_
- cargo test -p api response_status_

GitHub checks are also passing: Lint, Test Suite, security_audit, and claude-review.

Evrard-Nil (approved):
🤖 Reviewed locally at HEAD (cloned, built, ran the new tests — green). **Correct, careful fix — approving.**

- Switches `rename_all = "lowercase"` → `"snake_case"` on `ResponseStatus` + `ResponseItemStatus` in **both** `crates/api/src/models.rs` and `crates/services/src/responses/models.rs`, so `InProgress` now serializes as `in_progress` (matching OpenAI's Responses spelling and the DB), fixing the `inprogress` mismatch (#621).
- **No collateral wire changes:** `InProgress` is the *only* multi-word variant in either enum (the rest — Completed/Failed/Cancelled/Queued — are single words where lowercase ≡ snake_case), so the `rename_all` switch changes nothing except the intended `inprogress`→`in_progress`.
- **Backward-compatible reads:** `#[serde(alias = "inprogress")]` on `InProgress` means any legacy-persisted `inprogress` value still deserializes correctly — good attention to existing data / in-flight payloads.
- Tests added in both files cover both directions (serialize → `in_progress`; deserialize legacy `inprogress` → `InProgress`); the serialization test passes locally and the crate builds clean.

LGTM.

PR #752, opened 2026-06-09
PR5: attested::chutes::Provider data-path skeleton (OpenAI-compatible, hard-off)

Evrard-Nil:
## Summary

**PR5 of the Chutes attested-provider integration** — the first Chutes *provider* code (PR1–PR4 were verifier groundwork). Adds an `attested::chutes::Provider` **data-path skeleton**: an OpenAI-compatible inference client over `https://llm.chutes.ai/v1`, implementing `InferenceProvider`. It is **not wired into the provider pool** and cannot serve traffic — a hard-off skeleton. The attestation path that earns it the `Attested3p` badge is added in a later PR.

## Design (decisions + rationale)

- **Reuse the proven OpenAI-compatible data path.** Chutes' inference API is OpenAI-compatible, identical in shape to NEAR's external providers. So `chutes::Provider` wraps an inner `ExternalProvider` (configured `OpenAiCompatible { base_url: https://llm.chutes.ai/v1 }`) and **delegates** every data-path method (chat, streaming, text, image, audio, score, rerank, embeddings, privacy, models). No duplicated HTTP/SSE code; the data path is already tested + in production.
- **Why a distinct type under `attested::chutes` (not just an external provider):** Chutes is *attested* — a later PR bolts a `ChutesBackendVerifier` onto it that fetches `/evidence`, transforms the TDX quote + GPU evidence into NEAR's attestation format, and runs the shared verifier under the strict `Attested3p` policy (`MeasurementPolicy::attested3p` + `StrictBoundReportDataVerifier` from #746/#748/#749). An external provider can never be attested; this one will be.
- **Attestation methods are explicitly stubbed, not delegated.** `get_attestation_report` / `get_signature` return an explicit *"not yet wired (behind a hard-off gate)"* error rather than the inner external provider's *"attestation unsupported"* message — so the skeleton can never be mistaken for a verified provider, and the difference from `external` (which is permanently non-attested) is clear in the code. They become real in the verifier-wiring PR.
- **Hard-off = no construction path.** Nothing in `inference_provider_pool` builds a `chutes::Provider` in this PR, so it is unreachable at runtime. The pool wiring + an explicit enable flag land with the verifier PR; enabling Chutes in production stays gated until the open questions about Chutes' evidence format (exact nonce→`report_data` derivation, per-response signing, signed golden-measurement manifest) are confirmed.

## What changed

- New `crates/inference_providers/src/attested/chutes/mod.rs`:
  - `chutes::Config { base_url, api_key, model_name, timeout_seconds }` + `Config::new(..)` defaulting `base_url` to `https://llm.chutes.ai/v1`.
  - `chutes::Provider { data_path: ExternalProvider }` implementing `InferenceProvider` (delegate data path; stub attestation/signature).
- `attested/mod.rs`: `pub mod chutes;`.

## Tests

- `chutes::tests`: construction keeps the model id + defaults to the Chutes base URL; `get_attestation_report` / `get_signature` return the "not yet wired" error (the skeleton never appears verified).
- The data path itself is the already-tested `ExternalProvider` / `OpenAiCompatibleBackend`; the live Chutes data path + real attestation get an integration test in the verifier-wiring PR.
- `inference_providers` lib green; `cargo build --workspace` green; `cargo test --test e2e_all` green (Chutes is unwired, so e2e is unaffected — confirms no break); fmt + clippy clean.

## Impact

- **Runtime behavior:** none — `chutes::Provider` is not constructed anywhere.
- **Risk:** very low — additive module, data path reuses proven code, no pool/config/DB/wire changes.
- **Unblocks:** PR6 (evidence models + `/evidence`→NEAR-format transform) and PR7 (`ChutesBackendVerifier` + pool wiring under the `Attested3p` policy + enable flag).

### DIFF
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -84,9 +84,22 @@ impl Config {
         }
     }
 
+    /// Override the inference base URL (defaults to [`DEFAULT_BASE_URL`]). Used
+    /// by integration tests / mock servers and by the verifier-wiring PR, which
+    /// sources the URL from config rather than hardcoding it.
+    pub fn with_base_url(mut self, base_url: impl Into<String>) -> Self {
+        self.base_url = base_url.into();
+        self
+    }
+
     /// The OpenAI-compatible inference base URL.
     pub fn base_url(&self) -> &str {
         &self.base_url
+    }
+
+    /// The per-request timeout in seconds (always > 0).
+    pub fn timeout_seconds(&self) -> i64 {
+        self.timeout_seconds
     }
 }
 
@@ -286,12 +299,41 @@ mod tests {
 
     #[test]
     fn config_non_positive_timeout_falls_back_to_default() {
-        // Guards the `timeout_seconds as u64` cast in the HTTP backend.
-        let redacted = format!("{:?}", Config::new("k".into(), "m".into(), -5));
-        assert!(redacted.contains("timeout_seconds: 300"));
-        // And the api_key must not leak through Debug.
-        assert!(redacted.contains("[redacted]"));
-        assert!(!redacted.contains("\"k\""));
+        // Guards the `timeout_seconds as u64` cast in the HTTP backend. Assert on
+        // the accessor, not Debug output, so formatting changes don't break it.
+        assert_eq!(
+            Config::new("k".into(), "m".into(), -5).timeout_seconds(),
+            DEFAULT_TIMEOUT_SECONDS
+        );
+        assert_eq!(
+            Config::new("k".into(), "m".into(), 0).timeout_seconds(),
+            DEFAULT_TIMEOUT_SECONDS
+        );
+        assert_eq!(
+            Config::new("k".into(), "m".into(), 42).timeout_seconds(),
+            42
+        );
+    }
+
+    #[test]
+    fn config_debug_redacts_api_key() {
+        // Use a realistic-length, distinctive secret so the leak check is meaningful.
+        let secret = "cpk_super_secret_value_0123456789abcdef";
+        let redacted = format!("{:?}", Config::new(secret.into(), "m".into(), 60));
+        assert!(redacted.contains("[redacted]"), "key must be redacted");
+        assert!(
+            !redacted.contains(secret),
+            "api_key must never appear in Debug output, got: {redacted}"
+        );
+    }
+
+    #[test]
+    fn with_base_url_overrides_default() {
+        let c = Config::new("k".into(), "m".into(), 60).with_base_url("http://127.0.0.1:9999/v1");
+        assert_eq!(c.base_url(), "http://127.0.0.1:9999/v1");
+        // And the override survives into the provider.
+        let p = Provider::new(c);
+        assert_eq!(p.model_name(), "m");
     }
 
     #[tokio::test]
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -1,0 +1,267 @@
+//! Chutes — an attested third-party inference provider (`ProviderTier::Attested3p`).
+//!
+//! Chutes (chutes.ai) serves models from Intel TDX + NVIDIA confidential-compute
+//! TEEs. Its **data path** is OpenAI-compatible (`https://llm.chutes.ai/v1`), so
+//! this provider reuses the existing OpenAI-compatible backend for chat / image /
+//! audio / etc., exactly like NEAR's external providers.
+//!
+//! Its **attestation path** is what makes it `Attested3p` rather than
+//! `NonAttested`: a later PR wires a `ChutesBackendVerifier` that fetches Chutes'
+//! `/evidence`, transforms the TDX quote + GPU evidence into NEAR's attestation
+//! format, and runs the shared verifier under the strict `Attested3p` policy
+//! (`MeasurementPolicy::attested3p` + `StrictBoundReportDataVerifier`).
+//!
+//! # Status — skeleton, behind a hard-off gate
+//!
+//! This is the **data-path skeleton** (PR5). It is **not constructed anywhere**:
+//! nothing in the provider pool builds a `chutes::Provider` yet, so it cannot
+//! serve traffic. The pool wiring + an explicit enable flag land with the
+//! attestation verifier (PR7); turning Chutes on in production is gated behind
+//! that flag until the open questions about Chutes' evidence format are
+//! confirmed. Until then `get_attestation_report` / `get_signature` return an
+//! explicit "not yet wired" error so this provider can never be mistaken for a
+//! verified one.
+
+use async_trait::async_trait;
+use std::sync::Arc;
+
+use crate::non_attested::external::{ExternalProvider, ExternalProviderConfig, ProviderConfig};
+use crate::{
+    AttestationError, AudioTranscriptionError, AudioTranscriptionParams,
+    AudioTranscriptionResponse, ChatCompletionParams, ChatCompletionResponseWithBytes,
+    ChatSignature, CompletionError, CompletionParams, EmbeddingError, ImageEditError,
+    ImageEditParams, ImageEditResponseWithBytes, ImageGenerationError, ImageGenerationParams,
+    ImageGenerationResponseWithBytes, InferenceProvider, ListModelsError, ModelsResponse,
+    PrivacyClassifyError, RerankError, RerankParams, RerankResponse, ScoreError, ScoreParams,
+    ScoreResponse, StreamingResult,
+};
+
+/// Default Chutes OpenAI-compatible inference base URL.
+pub const DEFAULT_BASE_URL: &str = "https://llm.chutes.ai/v1";
+
+/// Configuration for a Chutes attested provider.
+///
+/// The attestation-specific fields (chute id, `/evidence` endpoint, golden
+/// measurements) are added by the PR that wires the verifier; this skeleton only
+/// needs what the OpenAI-compatible data path uses.
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// OpenAI-compatible inference base URL (e.g. `https://llm.chutes.ai/v1`).
+    pub base_url: String,
+    /// Chutes API key (`cpk_...`). A secret — sourced from env/secret store,
+    /// never hard-coded or logged.
+    pub api_key: String,
+    /// The model id as served by Chutes (e.g. `zai-org/GLM-5.1-TEE`).
+    pub model_name: String,
+    /// Per-request timeout, seconds.
+    pub timeout_seconds: i64,
+}
+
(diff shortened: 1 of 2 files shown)


lloydmak99:
## Code Review

**Verdict: Approve** (two minor items below, `claude[bot]`'s `Debug`/api_key finding is the blocking one to address first)

### New findings not in prior comments

**`pub mod chutes` allows external crates to construct the provider** (`crates/inference_providers/src/attested/mod.rs`)

The hard-off gate relies on "nothing constructs `chutes::Provider`." With `pub mod chutes`, any crate that depends on `inference_providers` can call `chutes::Provider::new(...)` directly and bypass that intent. Using `pub(crate)` gives a compile-time guarantee that no code outside `inference_providers` can construct the provider until PR7 deliberately widens the visibility. Easy change while the module is still new.

**`AttestationError::FetchError` is the wrong variant for "not yet wired"** (`attested/chutes/mod.rs:231`)

`FetchError` implies a transient network failure — callers that inspect error variants for retry/backoff logic would retry a permanently hard-coded stub indefinitely. `AttestationError::Unsupported` (or similar "not available" variant) would be the semantically correct choice. Low risk while the provider is never constructed, but worth fixing before PR7 registers it.

PierreLeGuen (changes requested):
Found one blocking security issue.

- `crates/inference_providers/src/attested/chutes/mod.rs:47-53`: `Config` derives `Debug` while carrying the Chutes `api_key`, so any future init/error/debug log of this config will print the full provider credential. This is especially risky because this skeleton is intended to be wired into env/pool config next, and the adjacent `ExternalProviderConfig` already uses a custom `Debug` impl that redacts the same field. Please remove the `Debug` derive or add a redacting `Debug` implementation before introducing the secret-bearing config.

I inspected the diff, provider-pool reachability, and the delegated `ExternalProvider` data path. Local checks passed: `cargo test -p inference_providers chutes --lib` and `cargo fmt --check`. GitHub check rollup was green at review time.

lloydmak99 (approved):
Reviewed — approve (address claude[bot]'s Debug/api_key leak before merging). Two additional items not in prior comments: (1) pub mod chutes should be pub(crate) to compile-time enforce the hard-off gate — external crates can currently construct Provider directly; (2) AttestationError::FetchError is the wrong variant for 'not yet wired' — implies retryable network failure, callers inspecting error variants for retry logic would retry a permanently-hard-coded stub.

Evrard-Nil:
Thanks all — addressed in `de2b1e95`.

**@PierreLeGuen (blocking) / @Copilot / claude[bot] — `Config` `Debug` leaking `api_key`.** Fixed: `Config` fields are now **private** (only constructor is `Config::new`) with a **custom `Debug` that redacts the key**:
```rust
impl std::fmt::Debug for Config {
    fn fmt(&self, f) -> ... {
        f.debug_struct("Config")
            .field("base_url", &self.base_url)
            .field("api_key", &"[redacted]")
            .field("model_name", &self.model_name)
            .field("timeout_seconds", &self.timeout_seconds)
            .finish()
    }
}
```
No `{:?}` can expose the secret now, and the private fields also close item 2.

**@Copilot — `timeout_seconds: i64` → `as u64` underflow.** `Config::new` now replaces a non-positive timeout with `DEFAULT_TIMEOUT_SECONDS` (300), and private fields prevent a struct-literal bypass. New test `config_non_positive_timeout_falls_back_to_default` covers it (and asserts the key is redacted, not leaked).

**@Copilot — brittle `{:?}` test assertions.** The attestation/signature tests now match the concrete variant (`AttestationError::FetchError(msg)` / `CompletionError::CompletionError(msg)`) and assert on the message.

**@lloydmak99 — `pub(crate) mod chutes` for compile-time hard-off.** Good instinct, but `pub(crate)` would make the skeleton (used only in `#[cfg(test)]`) trip the `dead_code` lint, since it's not reachable from the crate's public API — it'd need an `#[allow(dead_code)]`, which is its own smell and would have to be removed in PR7 anyway. The hard-off is already enforced structurally: **nothing constructs a `chutes::Provider`** (no pool wiring), so it's unreachable at runtime regardless of visibility. PR7 adds the wiring + an explicit enable flag. Happy to switch to `pub(crate)+allow` if you'd prefer the stronger compile-time signal in the interim.

Local: inference_providers lib 239 + 4 chutes tests; fmt + clippy clean. Re-requesting review.

Evrard-Nil:
⚠️ Heads-up: `copilot-swe-agent[bot]` pushed `cdd144f2 "Apply remaining changes"` on top of my fix commit, which **accidentally committed 260 build-artifact files** (256 `.o` + 3 `.bin` + a lock) under a stray `false/inference_providers-.../` directory — the agent built with `--target-dir false`, and `.gitignore` only covers `target/`, so they got staged. It contained **zero source changes**.

I've force-pushed the branch back to my clean commit `de2b1e95` (all the review fixes from the earlier comment are intact); the junk commit is dropped. CI is re-running on `de2b1e95`.

PierreLeGuen (commented):
This PR is well-scoped and safe to land: the 315-line `attested::chutes::Provider` skeleton is additive, and the hard-off claim holds — `chutes::Provider`/`chutes::Config` are referenced only by the `pub mod chutes;` declaration in `attested/mod.rs` and are never constructed by `inference_provider_pool` or anywhere else in the workspace, so runtime impact is zero.

Verified details:

- **Delegation is complete.** Every `InferenceProvider` method is either delegated to the inner `ExternalProvider` or deliberately stubbed. The three default-impl hooks not overridden here (`pin_chat_connection`, `unpin_chat_connection`, `set_backend_count`) are also not overridden by `ExternalProvider`, so behavior matches the wrapped provider exactly.
- **Stub errors are consistent and unambiguous.** `AttestationError::FetchError` / `CompletionError::CompletionError` match the variants `ExternalProvider` uses for its attestation-unsupported path (`crates/inference_providers/src/non_attested/external/mod.rs:329-356`), with a distinct "not yet wired" message so the skeleton can never appear verified. Tests pin the variants, not just message substrings.
- **Privacy.** `Config` uses a manual `Debug` impl that redacts the API key, fields are private (so the timeout guard can't be bypassed), and a test asserts the key doesn't leak through `Debug`.
- **Doc claims check out.** `DEFAULT_TIMEOUT_SECONDS = 300` matches the external-provider fallback (`crates/config/src/types.rs:1027-1030`), and the `timeout_seconds as u64` wrap hazard it guards is real (e.g. `crates/inference_providers/src/non_attested/external/openai_compatible.rs:210`). `ProviderConfig::OpenAiCompatible.model_name: None` correctly means the supplied model id is used verbatim.

Non-blocking observations for the follow-up wiring PR:

- `crates/inference_providers/src/attested/chutes/mod.rs:74` — `Config::new` always hardcodes `base_url = DEFAULT_BASE_URL`, yet the field is documented as configurable and exposed via `base_url()`. The verifier-wiring PR's integration tests (and any staging/mock-server use) will need an override; consider a `with_base_url(...)` builder now, or drop the accessor until needed. Relatedly, `api_key`/`model_name` aren't validated non-empty — fine for a skeleton, but worth adding when this becomes constructible from config.
- `crates/inference_providers/src/attested/chutes/mod.rs:307` — `config_non_positive_timeout_falls_back_to_default` asserts by substring-matching `Debug` output (`"timeout_seconds: 300"`), which breaks on any formatting change, and the secret-leak assertion uses a 1-char key. A `timeout_seconds()` accessor plus a realistic-length key would make both assertions robust.

Checks run locally:
- `cargo test -p inference_providers --lib` — 239 passed, 0 failed (includes the 4 new chutes tests)
- `cargo clippy -p inference_providers --all-targets --all-features -- -D warnings` — clean
- `cargo check --workspace` — passed (no downstream breakage from the new module)
- grep audit confirming the module is unreachable at runtime
- E2E suite skipped (requires PostgreSQL; change is unwired so exposure is nil). CI: lint and security audit passed; test suite was still in progress at review time but the full lib suite passed locally.

Evrard-Nil:
Thanks for the thorough re-review @PierreLeGuen — addressed both follow-up observations in `a76b7faf` (and the branch is on the latest main via the update merge):

1. **`base_url` configurability.** Added a `Config::with_base_url(..)` builder so the URL is genuinely overridable (the verifier-wiring PR will source it from config, and integration/mock-server tests need it) — the `base_url()` accessor is now meaningful rather than dead. `Config::new` still defaults to `DEFAULT_BASE_URL`.
2. **Brittle test assertions.** Added a `timeout_seconds()` accessor; `config_non_positive_timeout_falls_back_to_default` now asserts on it (`-5`/`0` → 300, `42` → 42) instead of substring-matching `Debug`. Split the secret-leak check into `config_debug_redacts_api_key` using a realistic-length key (`cpk_super_secret_value_…`) and asserting the full secret never appears in `Debug`. Added `with_base_url_overrides_default`.

(Non-empty validation of `api_key`/`model_name`: agreed it belongs when this becomes constructible from config — I'll add it in the verifier-wiring PR where the values come from env/DB.)

inference_providers lib 241 + 6 chutes tests; fmt + clippy clean. Re-requesting review.

PierreLeGuen (approved):
This PR cleanly adds the `attested::chutes::Provider` skeleton (`crates/inference_providers/src/attested/chutes/mod.rs` plus the `pub mod chutes;` declaration) with no runtime behavior change. No substantive issues found.

What I verified:

- **Genuinely hard-off**: nothing outside the module's own tests constructs `chutes::Provider` — no pool wiring, config, route, or DB changes. A workspace-wide check confirms downstream crates are unaffected.
- **Delegation is sound**: all `InferenceProvider` data-path methods delegate to the inner `ExternalProvider`. The trait defaults that aren't overridden (`pin_chat_connection`, `unpin_chat_connection`, `set_backend_count`) are no-ops in `ExternalProvider` as well, so nothing is silently bypassed.
- **Model-name semantics**: `model_name: None` in the `OpenAiCompatible` variant correctly causes the chutes `Config.model_name` to be used as the effective upstream model id (`external/mod.rs:241`), matching the documented intent.
- **Secret handling**: `Config` fields are private, `Config::new` is the only constructor, and the custom `Debug` impl redacts `api_key` (with a test asserting the secret never appears in `Debug` output). No logging is added, so privacy rules are satisfied.
- **Attestation fails closed**: `get_attestation_report` / `get_signature` return explicit errors with tests matching the concrete error variants — the skeleton cannot appear verified.
- The non-positive-timeout fallback to 300s in `Config::new` matches the config-layer default (`crates/config/src/types.rs:1030`), is documented, and is covered by a test.

Checks run locally: `cargo fmt --all -- --check`, `cargo clippy -p inference_providers --all-targets -- -D warnings`, `cargo clippy --workspace --all-targets`, `cargo build --workspace` / `cargo check --workspace`, `cargo test -p inference_providers --lib` (241 tests passed, including 6/6 chutes tests), and `git diff --check` — all clean. GitHub CI (Test Suite, security_audit) was still in progress at review time; the DB-backed/e2e suite was not run locally since this PR is confined to the `inference_providers` crate skeleton with no runtime/route/DB changes.

PR #753, opened 2026-06-09
Standardize attestation error envelopes

### LINKED ISSUE #657: /v1/signature/{chat_id} returns non-standard error envelope: {"error": "string"} instead of {"error": {message, type, ...}}
## Summary

`GET /v1/signature/{chat_id}` uses a different error response shape than every other endpoint. Clients catching errors uniformly via `response.error.message` break here.

## Repro

```bash
curl -s -w "HTTP %{http_code}\n" \
  "https://cloud-api.near.ai/v1/signature/chatcmpl-fake-id?model=openai/gpt-oss-120b&signing_algo=ecdsa" \
  -H "Authorization: Bearer $KEY"
```

```json
{"error":"Signature not found: chatcmpl-fake-id:ecdsa"}
HTTP 404
```

vs. the standard envelope returned by every other endpoint:

```json
// /v1/files/{bad-id}
{"error":{"message":"Invalid file ID format: file-bad","type":"invalid_request_error","param":null,"code":null}}

// /v1/responses/{fake-id}
{"error":{"message":"Get response not yet implemented","type":"not_implemented","param":null,"code":null}}

// /v1/chat/completions (no auth)
{"error":{"message":"Missing authorization header","type":"missing_auth_header","param":null,"code":null}}
```

The signature endpoint emits `error` as a string; everywhere else `error` is an object with `message`/`type`/`param`/`code`.

## Why it matters

Clients normalizing errors typically do something like:
```python
err = resp.json().get("error")
msg = err.get("message") if isinstance(err, dict) else err
```
…or break on the first crash. The signature endpoint forces every consumer to add a special case.

## Suggested fix

Return the standard envelope here too:
```json
{
  "error": {
    "message": "Signature not found: chatcmpl-fake-id:ecdsa",
    "type": "not_found_error",
    "param": null,
    "code": null
  }
}
```

## Bonus minor: `signing_algo` not validated

```bash
curl -s "https://cloud-api.near.ai/v1/signature/x?signing_algo=rsa" -H "..."
# {"error":"Signature not found: x:rsa"}
```

The endpoint accepts any string for `signing_algo` (e.g. `rsa`, `foo`) and only fails at lookup time. A nicer surface would 400 on unknown algos upfront.

## What works

Happy path is healthy:
```json
{
  "text":"openai/gpt-oss-120b:<hash>:<hash>",
  "signature":"0x...",
  "signing_address":"0x56d070df1c6be444b007839ef9cf67cec7c12b8b",
  "signing_algo":"ecdsa"
}
```

Distinct from the closed signature-API issues (#10, #18, #19, #308) which were about hash matching, db errors, and the mock-address bug — none of those were about the response envelope shape.

---

lloydmak99:
## Problem

Fixes #657.

`GET /v1/signature/{chat_id}` returned a non-standard error shape:

```json
{ "error": "Signature not found: ..." }
```

Other API errors use the shared envelope with `error.message`, `error.type`, `error.param`, and `error.code`, so clients that normalize errors through `response.error.message` needed a signature-endpoint special case.

The issue thread also called out the same local string-only envelope on `GET /v1/attestation/report`, plus invalid nonce/signing parameter errors being reported as HTTP 503 instead of client-side 400s.

## Change

- Replaced the attestation route-local `ErrorResponse { error: String }` with the shared `crate::models::ErrorResponse` envelope.
- Mapped `/v1/signature/{chat_id}` errors into standard envelopes:
  - missing signatures return `404` with `type: not_found_error`
  - invalid parameters return `400` with `type: invalid_request_error`
  - internal/provider/repository failures return `500` with `type: internal_server_error`
- Added route-level `signing_algo` validation for `/v1/signature/{chat_id}` so unsupported algorithms return `400` before repository lookup.
- Mapped `/v1/attestation/report` invalid client parameters to `400`, including nonce and signing algorithm errors, while preserving `503` for provider unavailability.
- Updated OpenAPI schema registration to use the shared error model.
- Added focused unit tests for the signature 404 envelope, invalid attestation nonce status/param mapping, invalid `signing_algo`, and supported algorithms.

## What Else Is Needed

- CI and reviewer validation.
- Release/deploy notes should call out the intentional error-body compatibility change for clients reading attestation `error` values as strings; they should read `error.message` after this change.
- Broader error-shape/status audits for unrelated endpoints should stay in their dedicated issues; this PR is scoped to the attestation routes raised in #657 and its thread.

## Tests

- `cargo test -p api --lib attestation`
- `cargo test --lib --bins`


PierreLeGuen (approved):
Reviewed the attestation error-envelope changes against the linked issue, the route diff, the shared ErrorResponse model, service AttestationError sources, OpenAPI schema registration, docs references, and relevant tests/CI metadata.

No substantive issues found. Local checks run: cargo test -p api routes::attestation::tests and cargo fmt --all -- --check. GitHub reports Security audit, Lint, and Test Suite passing.

Evrard-Nil (approved):
🤖 Reviewed independently at `f464a01` (built, clippy, unit tests — green). **Safe to approve.**

- Standardizes both attestation endpoints onto the shared OpenAI-compatible `ErrorResponse { error: { message, type, param, code } }` (drops the route-local `{error: String}`) — now consistent with the rest of cloud-api.
- **Upstream-auth-masking invariant SAFE:** the only backend-wrapping variant, `ProviderError` (attestation report path, `services/attestation/mod.rs:764`), still maps to **503** (`attestation.rs:392-397`) — a backend 401/403/407 during attestation never surfaces as client 4xx. The signature endpoint can only yield `SignatureNotFound`/`RepositoryError`, so no backend error reaches it. Error message body is byte-identical to pre-PR (`e.to_string()`) — no new info-leak surface.
- The `503→400` change for invalid client params (bad nonce / unsupported `signing_algo`, via the new pre-lookup `validate_signing_algo`) is a correctness improvement (stops blaming our backend for client mistakes).
- The string→object error-body shape change is intentional/documented (#657) and aligns attestation with every other endpoint; no e2e asserts on the old shape; 4 new unit tests pass; CI green.
- NIT only: `ClientError` is a dead variant (never constructed) — harmless.

LGTM.

PR #754, opened 2026-06-10
fix(mcp): accept initialized notifications

hanakannzashi:
## Summary
- Accept MCP `notifications/initialized` as a JSON-RPC notification without requiring an `id`.
- Return `202 Accepted` with an empty body for the initialized notification while preserving existing JSON-RPC error behavior for other missing-id requests.
- Add e2e coverage for the initialized notification path.

## Validation
- `cargo test -p api --test e2e_all --no-run`
- `cargo clippy -p api --test e2e_all -- -D warnings`

Note: running the new e2e test directly previously reached local Postgres setup and failed with `password authentication failed for user "postgres"`; compile/no-run and clippy pass.

### DIFF
--- a/crates/api/src/routes/mcp_server.rs
+++ b/crates/api/src/routes/mcp_server.rs
@@ -5,7 +5,7 @@ use crate::models::ErrorResponse;
 use axum::{
     extract::{Extension, Json, State},
     http::StatusCode,
-    response::Json as ResponseJson,
+    response::{IntoResponse, Json as ResponseJson, Response},
 };
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
@@ -20,6 +20,7 @@ const MCP_PROTOCOL_VERSION: &str = "2024-11-05";
 const MCP_SERVER_VERSION: &str = "1.0.0";
 const WEB_SEARCH_TOOL_NAME: &str = "web_search";
 const WEB_SEARCH_TOOL_DESCRIPTION: &str = "Search the web and return structured search results.";
+const MCP_INITIALIZED_NOTIFICATION: &str = "notifications/initialized";
 // JSON-RPC reserves -32000..-32099 for server-defined errors. We use a small
 // stable subset here to translate HTTP-layer auth/usage/rate-limit failures
 // into MCP-framed responses without changing the underlying REST middleware.
@@ -136,20 +137,17 @@ struct McpWebSearchArgs {
     operators: Option<bool>,
 }
 
-fn ok_response(id: Option<Value>, result: Value) -> ResponseJson<McpResponse> {
+fn ok_response(id: Option<Value>, result: Value) -> Response {
     ResponseJson(McpResponse {
         jsonrpc: "2.0",
         id: response_id(id),
         result: Some(result),
         error: None,
     })
+    .into_response()
 }
 
-fn error_response(
-    id: Option<Value>,
-    code: i64,
-    message: impl Into<String>,
-) -> ResponseJson<McpResponse> {
+fn error_response(id: Option<Value>, code: i64, message: impl Into<String>) -> Response {
     ResponseJson(McpResponse {
         jsonrpc: "2.0",
         id: response_id(id),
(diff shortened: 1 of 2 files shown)


think-in-universe (approved):

Evrard-Nil (approved):

PierreLeGuen (changes requested):
The fix itself is correct and well-tested: a POST to `/mcp` with `{"jsonrpc": "2.0", "method": "notifications/initialized"}` and no `id` now gets `202 Accepted` with an empty body instead of a `-32600` error, which is what the MCP Streamable HTTP transport requires. Widening the helper return types from `ResponseJson<McpResponse>` to `Response` is a clean, behavior-preserving way to allow the bodiless 202.

**Important — the fix is scoped to a single notification method** (`crates/api/src/routes/mcp_server.rs:246-251`). JSON-RPC 2.0 defines *any* request without an `id` as a notification and forbids replying to it, and the MCP Streamable HTTP transport requires `202 Accepted` with no body for any posted notification. Real clients send more than the handshake: the official TypeScript/Python SDKs emit `notifications/cancelled` when a tool call is aborted, plus `notifications/progress` and `notifications/roots/list_changed`. With this PR, those still receive a `200` with a JSON-RPC error object (`id: null`) — the exact failure mode being fixed here, just on the next notification a client sends. The generalization is essentially one line: keep the `jsonrpc == "2.0"` guard and match `request.method.starts_with("notifications/")` (or treat any valid-`jsonrpc` message without an `id` as a notification per spec), then add an e2e case for `notifications/cancelled`.

**Minor** — `crates/api/tests/e2e_all/mcp_server.rs:42-62`: `test_mcp_missing_id_returns_jsonrpc_error` enshrines replying `-32600` to an id-less `initialize`, which is itself a notification under JSON-RPC 2.0. If the guard is generalized as above, this test's expected behavior should be revisited rather than kept as a contract.

Local checks: `cargo fmt --all -- --check` passed; `cargo clippy -p api --lib -- -D warnings` and `cargo clippy -p api --test e2e_all -- -D warnings` passed; the e2e target compiles (`--no-run`). I also ran the full MCP server e2e suite against a temporary `postgres:16` container with CI-equivalent env — 8/8 passed, including the new `test_mcp_initialized_notification_is_accepted_without_id`. GitHub CI (Lint, Test Suite, security audit) is green on the PR head.

PierreLeGuen (approved):
This fixes a real protocol bug: MCP clients send `notifications/initialized` (and other `notifications/*` messages) without an `id`, and the handler previously rejected every id-less message with a JSON-RPC `-32600` error. Responding to a notification violates JSON-RPC 2.0, and the MCP Streamable HTTP transport spec requires a bodyless **202 Accepted** when a POST contains only notifications — which is exactly what this implements.

Verified:

- **Spec conformance** — `is_mcp_notification` requires `jsonrpc == "2.0"` plus a `notifications/` method prefix, covering all MCP-defined notifications (`initialized`, `cancelled`, `progress`, `roots/list_changed`), and the head commit generalizes correctly from `notifications/initialized` alone to the full prefix. The 202-with-empty-body response matches the transport spec.
- **Edge cases** — a malformed id-less message (wrong/missing `jsonrpc`, or non-`notifications/` method) still gets the pre-existing `-32600` with `id: null`; a `notifications/*` method sent *with* an id falls through to "Method not found", which is harmless; `"id": null` deserializes to `None` via `Option<Value>` and is correctly treated as a notification candidate.
- **Auth/limits** — the auth middleware still short-circuits unauthenticated requests with 401 before the handler, so notifications require a valid API key. They bypass rate-limit/usage checks, but those only ever applied to `tools/call` and the notification path does no work, so there's no new exposure.
- **Refactor** — switching the helpers from `ResponseJson<McpResponse>` to `Response` is the minimal way to mix a bodyless 202 with JSON responses; no behavior change on existing paths.
- **Tests** — the renamed `test_mcp_non_notification_missing_id_returns_jsonrpc_error` preserves coverage of the error path, and the new `test_mcp_notifications_are_accepted_without_id` covers all four notification methods, asserting both status and empty body.

One non-blocking, pre-existing note: at `crates/api/src/routes/mcp_server.rs:249-258`, an id-less message whose method lacks the `notifications/` prefix still receives a JSON body, while strict JSON-RPC 2.0 says any id-less request must get no response. All MCP notifications use the prefix, the behavior predates this PR, and the renamed test deliberately pins it — no change needed.

**Checks:** `cargo fmt --all -- --check` passed; `cargo clippy -p api --lib --tests -- -D warnings` and `cargo clippy -p api --test e2e_all -- -D warnings` passed; `cargo test -p api --test e2e_all mcp_server --no-run` compiled cleanly. The new e2e test couldn't run locally (no Postgres available), but CI runs `cargo nextest run --test e2e_all` against a real Postgres service and Lint, Test Suite, and security_audit are all green on the PR head.

PR #755, opened 2026-06-10
PR6: Chutes /evidence models + NEAR-format transform (data layer for the verifier)

Evrard-Nil:
## Summary

**PR6 of the Chutes attested-provider integration** (follows #745/#746/#748/#749 verifier groundwork + #752 data-path skeleton). Adds the **Chutes `/evidence` serde models** and a **pure, fail-closed transform** from a Chutes instance's evidence to NEAR's attestation-report shape — the data layer the Chutes verifier (next PR) will consume. Still behind the hard-off gate; nothing is wired into the pool.

## What changed

- `attested/chutes/evidence.rs` — serde models for the real `GET /chutes/{chute_id}/evidence?nonce=` response (verified against the live endpoint): `EvidenceResponse { evidence: [InstanceEvidence { quote, gpu_evidence: [GpuEvidence{certificate,evidence,arch}], instance_id, certificate }], failed_instance_ids }` + an `instance(id)` lookup.
- `attested/chutes/attestation.rs` — `to_near_report(instance, request_nonce)` + helpers (`intel_quote_hex`, `tls_cert_fingerprint`, `nvidia_payload`) + a `TransformError`. Maps the **format-identical** fields the shared verifier can use:

  | NEAR field | Source |
  |---|---|
  | `intel_quote` (hex) | base64 `quote` → raw → hex |
  | `tls_cert_fingerprint` | `SHA256(SPKI(certificate))` = `report_data[32:64]` |
  | `nvidia_payload` | `{arch, evidence_list:[{certificate,evidence}]}` |
  | `request_nonce`, `instance_id` | passthrough |
- Added `base64` dep; reuses the existing `spki_verifier::compute_spki_fingerprint_from_der`.

## Key design point (grounded in the real evidence + Chutes' docs)

Chutes' `report_data` layout is **not** NEAR's, so "transform → reuse NEAR's verifier unchanged" doesn't hold — and this transform is deliberately honest about it. Per Chutes' docs + our own decoding of live quotes:

- `report_data[0:32] = SHA256(nonce ‖ instance_e2e_pubkey)` — the ML-KEM-768 `e2e_pubkey` comes from `GET /e2e/instances/{chute}` (**not** in `/evidence`).
- `report_data[32:64] = SHA256(SPKI(certificate))` (confirmed empirically).
- Per-response integrity is **ChaCha20-Poly1305 AEAD** over the ML-KEM E2EE channel keyed by that attested key — not a separate signature.
- No dstack-style RTMR3 event log → measurement is **register-pin** (MRTD/RTMR), not replay.

So the transform **does not fabricate** `signing_address` or `event_log`, and leaves the GPU nonce unset. The Chutes-specific verifier (next PR) — slotting into the `ReportDataVerifier` seam from #748 — fetches the `e2e_pubkey` and checks `report_data[0:32] = SHA256(nonce ‖ e2e_pubkey)` + `[32:64] = tls_cert_fingerprint`, runs DCAP on `intel_quote` and NRAS on `nvidia_payload`, and register-pins the measurement. (A few precise details — exact preimage encoding, instance correlation between `/e2e/instances` and `/evidence`, signed golden measurements — are open questions out to Chutes; flag-on stays blocked on them.)

## Tests

`evidence::tests` (3): parses the real shape, instance lookup, `failed_instance_ids` optional. `attestation::tests` (7): base64-quote→hex; **real-cert SPKI fingerprint** (a captured Chutes instance cert fixture → 64-hex); `nvidia_payload` assembly (no fabricated nonce); `to_near_report` maps all shared fields and does **not** invent `signing_address`/`event_log`; fail-closed on bad/empty base64 and absent GPU evidence. inference_providers lib + `cargo build --workspace` + `e2e_all` green; fmt + clippy clean.

## Impact

- **Runtime:** none — pure functions, not constructed/wired anywhere.
- **Risk:** low — additive module + `base64` dep; fail-closed.
- **Unblocks:** PR7 (`ChutesBackendVerifier`: fetch `/e2e/instances` + `/evidence`, transform, verify the Chutes-shaped bindings via a `ChutesReportDataVerifier`, pool wiring + enable flag).


PierreLeGuen (commented):
This is a clean, additive data layer: serde models for Chutes' `/evidence` response plus a pure transform to NEAR's attestation-report shape, with nothing wired into the runtime. I cross-checked the transform against the consumer of the NEAR report shape in `crates/services/src/attestation/verification.rs` — the string-typed `nvidia_payload` convention matches, and the deliberate omission of the GPU nonce means a Chutes report accidentally fed to NEAR's default verifier fails closed at the nonce check. The transform validates before assembling, has no panic paths, and the test coverage (including the real captured cert fixture and all three fail-closed error paths) is solid. No blocking issues; a few minor observations below.

**Findings (all minor, non-blocking):**

- `crates/inference_providers/src/attested/chutes/mod.rs:12` — the module docs still say Chutes will use `StrictBoundReportDataVerifier`, but the new transform's docs correctly note that Chutes' `report_data` layout differs from NEAR's and needs a provider-specific `ReportDataVerifier`. Worth fixing now so the upcoming verifier PR isn't misled in a security-sensitive path.
- `crates/inference_providers/src/attested/chutes/attestation.rs:88-101` — `nvidia_payload` takes `gpu_evidence[0].arch` and applies it to the whole payload without checking the remaining entries agree. A mixed-arch instance would produce a mislabeled NRAS payload; NRAS would presumably reject it, but a transform documented as fail-closed should validate homogeneity itself (e.g. a `TransformError::MixedGpuArch` when any entry's arch differs from the first).
- `crates/inference_providers/src/attested/chutes/evidence.rs:38-44` — `#[serde(default)]` on `gpu_evidence` and `failed_instance_ids` covers a *missing* key but not an explicit JSON `null`, which would fail deserialization of the entire `EvidenceResponse`. Python-backed APIs commonly emit `null` for empty lists, and `failed_instance_ids` is purely informational, so a `null` there would needlessly make all evidence for the chute unparseable (availability-only impact, since the path fails closed). Consider a null-tolerant `deserialize_with`, or a test/doc note confirming the live endpoint omits rather than nulls these fields.
- `crates/inference_providers/src/attested/chutes/testdata/instance_cert.b64:1` — the committed fixture is a real Chutes instance cert whose SANs embed a node hostname (`h200-gd-245`) and public IP. It's a public self-signed attestation cert with no key material, already served by the public `/evidence` endpoint, so this is fine — just confirming it was committed deliberately.

The earlier suggestion to trim whitespace from base64 input isn't needed: the live endpoint returns clean base64, and explicit failure is the safer behavior in an attestation pipeline.

**Checks run locally:** `cargo fmt --all -- --check` clean; `cargo clippy -p inference_providers --all-targets` no warnings; `cargo check -p inference_providers` clean; `cargo test -p inference_providers --lib` — 251 passed (16 chutes-specific, including the 10 new evidence/attestation tests); manual cross-check of the transform output against `verification.rs` and `spki_verifier::compute_spki_fingerprint_from_der`; decoded the cert fixture with openssl. Not run: full-workspace/e2e suite (PostgreSQL-dependent). CI Lint and security_audit passed; Test Suite was still in progress at review time.

Evrard-Nil:
Thanks all — addressed in `b2e064f4`:

- **@gemini-code-assist (base64 trimming):** `decode_b64` now `.trim()`s before decoding, covering the quote, instance cert, and per-GPU cert/evidence. Added a test that a quote with surrounding whitespace/newline still decodes.
- **@Copilot (`nvidia_payload` arch):** no longer silently uses GPU[0]'s `arch`. It now fails closed on an empty arch (`EmptyGpuArch`) or mixed archs across GPUs (`InconsistentGpuArch { first, other }`), so the payload's single `arch` can't mislabel the evidence list. Tests added for both.
- **@Copilot (doc inconsistency):** fixed `chutes/mod.rs` — it no longer claims `StrictBoundReportDataVerifier`. Chutes' `report_data` layout differs (`SHA256(nonce ‖ e2e_pubkey)`), so it needs a **Chutes-specific** report-data verifier, consistent with `attestation.rs`.
- **@Copilot (`to_near_report` is partial):** clarified the doc — it intentionally returns a **subset** of NEAR fields. `signing_address`/`event_log` are N/A for Chutes (no separate signer; register-pin, not an event log), and `report_data` isn't a map key at all (it lives inside the decoded `intel_quote`). The Chutes verifier (next PR) consumes this partial map + adds the Chutes-shaped binding checks.

inference_providers lib 254 + 19 chutes tests; fmt + clippy clean. Re-requesting review.

FYI on the broader effort: I confirmed from Chutes' docs that `report_data[0:32] = SHA256(nonce ‖ ML-KEM-768 e2e_pubkey)` (key from `GET /e2e/instances/{id}`) and that per-response integrity is ChaCha20-Poly1305 AEAD over the E2EE channel. A few precise details (exact preimage encoding, `/e2e/instances`↔`/evidence` correlation, signed golden measurements) are out to Chutes and gate flag-on (PR8), not this data layer.

PierreLeGuen (commented):
This PR is in good shape: the code is additive with no runtime wiring, fail-closed throughout, well-documented, and the field conventions match the existing verifier (`intel_quote` hex, `tls_cert_fingerprint` as SPKI SHA-256 matching `compute_spki_hash`, `nvidia_payload` as a JSON string as `verify_gpu_evidence` expects). No payload content is logged. A few non-blocking observations:

- **Trim-then-emit mismatch in `nvidia_payload()`** — `crates/inference_providers/src/attested/chutes/attestation.rs:110-115`: each GPU `certificate`/`evidence` is validated via `decode_b64`, which trims surrounding whitespace, but the payload then emits the **untrimmed** originals (`json!({ "certificate": g.certificate, ... })`). A whitespace-padded value passes local validation yet ships padded base64 to NRAS, so the validation no longer attests the exact bytes emitted. Suggest emitting the trimmed strings (`g.certificate.trim()` / `g.evidence.trim()`), as is already done for `arch`.
- **Committed real-instance cert fixture** — `crates/inference_providers/src/attested/chutes/testdata/instance_cert.b64:1`: the fixture is a valid self-signed cert (public-key material only, no secrets), but its SANs embed a third party's node identifiers (hostname `h200-gd-245`, private IP `192.168.100.2`, public IP `103.175.197.20`). This is already publicly served by Chutes' `/evidence` endpoint so risk is low, but consider whether a synthetic self-signed cert with the same structure would exercise the SPKI path equally well before permanently committing another operator's identifiers.
- **Doc note for PR7** — `attestation.rs:96-101` vs `crates/services/src/attestation/verification.rs:458-469`: the transform intentionally omits `nonce` from `nvidia_payload`, but the shared `verify_gpu_evidence` hard-fails when `payload.nonce != request_nonce` (absent ⇒ `""`). Safely fail-closed today, but the upcoming Chutes verifier can't reuse `verify_gpu_evidence` as-is — it must inject the Chutes-derived GPU nonce or use a parallel GPU path. Worth stating in the module docs so PR7 doesn't assume drop-in reuse.

The module comment at `mod.rs:13-14` correctly notes that Chutes' `report_data` layout differs from NEAR's and needs a Chutes-specific report-data verifier, and `nvidia_payload()` correctly enforces consistent arch across GPU entries.

Local checks: `cargo test -p inference_providers --lib --bins` (254 passed, including all new chutes evidence/attestation tests), `cargo check -p inference_providers`, `cargo clippy -p inference_providers --all-targets -- -D warnings` (clean), `cargo fmt --check` (clean), `git diff --check` (clean), and the committed cert fixture decoded with openssl (valid self-signed DER X.509, CN=attestation-service, no secrets). Full-workspace/e2e suite skipped (PostgreSQL-dependent; not relevant to this pure-code change). CI at review time: Lint and security_audit green, Test Suite still in progress.

PierreLeGuen (commented):
This is a clean, additive data layer: serde models for Chutes' `/evidence` response plus a pure transform into the NEAR attestation-report shape. Nothing is wired into the provider pool, the hard-off gate is untouched, and the module docs are commendably explicit about which bindings the future Chutes-specific verifier must still check. I confirmed the field contracts against the shared verifier (`crates/services/src/attestation/verification.rs`): `intel_quote` as hex and `nvidia_payload` as a JSON-encoded string match what `verify_attestation`/`verify_gpu_evidence` consume, and feeding this output to the existing verifier fails closed (missing `signing_address`/`event_log`, and an absent GPU `nonce` never matches the request nonce) — consistent with the PR's stated flag-off contract until the PR7 verifier lands.

Two minor findings, neither blocking:

- `crates/inference_providers/src/attested/chutes/attestation.rs:110-112` — trim-vs-emit mismatch in `nvidia_payload()`. Each GPU's `certificate`/`evidence` is validated via `decode_b64`, which trims surrounding whitespace, but the **untrimmed** originals are pushed into `evidence_list` (while `arch` on line 114 is correctly emitted trimmed). If Chutes ever returns whitespace-padded base64 — the exact case the trim tolerance exists for — local validation passes but NRAS receives different bytes than were validated, producing a failure far from the root cause. Emit `g.certificate.trim()` / `g.evidence.trim()` so validation and emission operate on the same bytes.
- `crates/inference_providers/src/attested/chutes/testdata/instance_cert.b64:1` — the committed fixture is a real captured instance certificate whose SANs embed a Chutes node hostname (`h200-gd-245`), private IPs, and a public IP, now permanently in repo history. No key material is included and the data is presumably retrievable from the public `/evidence` endpoint, so risk is low — but the test only asserts "valid X.509 → 64-hex SPKI fingerprint", which a synthetic self-signed DER cert exercises equally well. Prefer swapping in a generated fixture, or confirm Chutes treats this as public.

Checks run locally: `cargo test -p inference_providers --lib` (254 passed, including the 19 new chutes tests), `cargo clippy -p inference_providers --all-targets -- -D warnings` (clean), `cargo fmt --all -- --check` (clean), `cargo check -p inference_providers`, `git diff --check`, and an `openssl x509` decode of the cert fixture. Full-workspace build and e2e suite were skipped (require PostgreSQL); CI on the PR head shows Lint, Test Suite, and security_audit green.

PierreLeGuen (commented):
This PR is additive and unwired at runtime — serde models for Chutes' `/evidence` response plus a pure transform into the NEAR attestation-report subset — so there's no behavioral change to the data plane. I cross-checked the transform output against the consuming verifier in `crates/services/src/attestation/verification.rs` and the shapes line up: `intel_quote` as a hex string (the consumer strips an optional `0x` prefix), `nvidia_payload` as a stringified JSON object, and the report as a `serde_json::Map`. Not fabricating `signing_address`/`event_log`/GPU nonce is the right call: the verifier fails closed on a missing payload nonce, so this partial report can't accidentally pass NEAR's default verification path. The transform has no panic paths and is fail-closed throughout. A few non-blocking observations:

- **Trim-vs-emit mismatch in `nvidia_payload()`** — `crates/inference_providers/src/attested/chutes/attestation.rs:109-118`. `decode_b64` trims surrounding whitespace before validating each GPU `certificate`/`evidence`, but the *original untrimmed* strings are inserted into the NRAS `evidence_list` (note `arch` is correctly emitted trimmed). If Chutes ever returns whitespace-padded base64, local validation passes on bytes that differ from what NRAS receives, which may then reject it. Fix: emit `g.certificate.trim()` / `g.evidence.trim()` (or re-encode the decoded bytes) so validation and emission cover the same bytes.

- **Real-instance cert committed as a fixture** — `crates/inference_providers/src/attested/chutes/testdata/instance_cert.b64:1`. The cert's SANs embed a live third-party node's details: hostname `h200-gd-245`, internal IP `192.168.100.2`, public IP `103.175.197.20`. It contains only a public key and is publicly served by Chutes' `/evidence` endpoint, so risk is low — but it permanently records another operator's network details in git history, and the test only asserts "valid X.509 → SHA256(SPKI) hex", which a synthetic self-signed cert would exercise equally. Consider swapping in a synthetic fixture, or confirm Chutes is fine with this being public.

- **Optional hardening in `intel_quote_hex()`** — `attestation.rs:72-74` accepts any non-empty base64, so a 1-byte "quote" transforms successfully. Acceptable for a data layer since downstream DCAP verification fails closed, but a cheap minimum-length/header sanity check (the TDX quote header is fixed-size) would surface malformed evidence as a clearer `TransformError` instead of a downstream DCAP parse error.

Checks run locally: `cargo check -p inference_providers`, `cargo clippy -p inference_providers --all-targets -- -D warnings`, and `cargo fmt --check` are all clean; `cargo test -p inference_providers --lib` passes 254 tests including the new chutes evidence/attestation tests. I also decoded the committed cert fixture with openssl (basis for the fixture note) and verified the output contract against the verifier in `crates/services/src/attestation/verification.rs`. The full e2e suite was skipped (requires PostgreSQL; CI's `Test Suite`, `Lint`, and `security_audit` checks all passed on this head).

Evrard-Nil:
Thanks @PierreLeGuen — both addressed in `f50ebb18`:

1. **trim-vs-emit mismatch in `nvidia_payload()`.** Fixed: each GPU's `certificate`/`evidence` is now both *validated and emitted* as the trimmed bytes, so what NRAS receives is exactly what we validated even if Chutes pads the base64:
   ```rust
   let certificate = g.certificate.trim();
   let evidence = g.evidence.trim();
   decode_b64("gpu_evidence.certificate", certificate)?;
   decode_b64("gpu_evidence.evidence", evidence)?;
   evidence_list.push(json!({ "certificate": certificate, "evidence": evidence }));
   ```

2. **real cert fixture.** Good catch — and it also trips our standing "no raw IPs/hostnames in repo" rule (the SANs embed `h200-gd-245` + a public IP). Replaced it with a **synthetic** self-signed cert (`testdata/synthetic_cert.b64`, `CN=chutes-test-fixture`, no SANs/IPs) and removed `instance_cert.b64`. The unit test only asserts "valid X.509 → 64-hex SPKI fingerprint", which the synthetic cert exercises identically; the *real* Chutes cert gets exercised by the live integration test in the verifier PR.

inference_providers lib 254 + 19 chutes tests; fmt + clippy clean. Re-requesting review.

PierreLeGuen (commented):
This PR is small and purely additive: serde models for Chutes' `/evidence` response plus a pure, fail-closed transform into NEAR's attestation-report shape, behind the existing hard-off gate with nothing wired into the provider pool. I cross-checked the transform's output contract against the consuming verifier in `crates/services/src/attestation/verification.rs` — field names line up, `nvidia_payload` is emitted as a JSON string as the verifier expects, `intel_quote` hex matches, and even if a partial report were misrouted into the shared verifier today it fails closed (missing `signing_address` fails `StrictBoundReportDataVerifier`; the deliberately-unset GPU nonce fails the nonce check at `verification.rs:458`). The committed cert fixture is genuinely synthetic (self-signed `CN=chutes-test-fixture`, no real hostnames/IPs, valid to 2036), and no logging of payload content was added.

No blocking issues. A few non-blocking findings:

- **Pin the SPKI fingerprint in the test** — `crates/inference_providers/src/attested/chutes/attestation.rs:169`: the test asserts only that the fingerprint is 64 hex characters, and nothing in the codebase pins a known-good fingerprint to a known cert (`spki_verifier.rs` only covers the invalid-DER path). A regression in `compute_spki_fingerprint_from_der` — e.g. hashing the full cert DER instead of the SPKI — would pass all tests and only surface later as a runtime `report_data[32:64]` mismatch, far from the root cause. Suggest asserting the fixture's exact value: `assert_eq!(fp, "e7c25815d0d940fea893d56984e131788afa6e931920093c9c2896fb04dea0da")`.
- **Stale test name** — same test: `tls_fingerprint_from_real_cert_is_64_hex` (and the "real cert" expect message at `attestation.rs:176`) predates the swap to a synthetic fixture. Rename to something like `tls_fingerprint_from_synthetic_cert_matches_expected` so a future reader doesn't assume a real Chutes cert is committed.
- **Explicit `null` fails deserialization** — `crates/inference_providers/src/attested/chutes/evidence.rs:48-49`: `#[serde(default)]` on `gpu_evidence` and `failed_instance_ids` only covers absent keys; an explicit JSON `null` (common from Python-backed APIs) would make the whole `EvidenceResponse` unparseable. Fail-closed is right for security-relevant fields, but `failed_instance_ids` is purely informational, so this is stricter than needed. Either add a null-tolerant `deserialize_with`, or document that the live endpoint is confirmed to omit rather than null these fields. Relatedly, `decode_b64` (`attestation.rs:60`) requires canonical padded standard base64 — fine as a fail-closed interop choice, just worth knowing if Chutes' encoding ever varies.

Local checks run: `cargo test -p inference_providers --lib` (254 passed, including all 19 new `attested::chutes` tests), `cargo fmt --all --check` clean, `cargo clippy -p inference_providers --all-targets -- -D warnings` clean, `cargo check --workspace` clean. Fixture sanity-checked by decoding `testdata/synthetic_cert.b64` with openssl. GitHub CI was still queued/in-progress at review time; e2e tests were skipped (require PostgreSQL, not exercised by this unwired code).

Evrard-Nil:
Thanks @PierreLeGuen — all three addressed in `df48f873`:

1. **Pinned the SPKI fingerprint.** The test now asserts the exact value (`e7c25815…dea0da`) — I independently recomputed it via `openssl x509 -pubkey | openssl pkey -pubin -outform DER | sha256` and it matches yours. This catches a `compute_spki_fingerprint_from_der` regression (e.g. hashing the whole cert DER) that the length-only check missed.
2. **Renamed** `tls_fingerprint_from_real_cert_is_64_hex` → `tls_fingerprint_from_synthetic_cert_matches_expected` (and the `expect` message) — no more "real cert" misnomer after the synthetic swap.
3. **Explicit `null` tolerated.** Added a `null_default` `deserialize_with` on `gpu_evidence` and `failed_instance_ids`, so an explicit JSON `null` (common from Python-backed APIs) deserializes to empty rather than failing the whole `EvidenceResponse` — important since `failed_instance_ids` is purely informational. New test `explicit_null_lists_tolerated`.

On the two earlier non-blocking notes I'll carry forward rather than expand this PR's scope: the **GPU-nonce / can't-reuse-`verify_gpu_evidence`-as-is** point is exactly what PR7's Chutes verifier handles (it injects the Chutes-derived GPU nonce / uses a parallel GPU path — noted for that PR), and the optional **`intel_quote` min-length/header sanity check** I can add in PR7 alongside the DCAP call. Happy to fold either in here instead if you'd prefer.

inference_providers lib 255 + 20 chutes tests; fmt + clippy clean. Re-requesting review.

PierreLeGuen (commented):
This PR adds two pure, additive modules under `attested/chutes/` — serde models for the Chutes `/evidence` response and a fail-closed transform into NEAR's attestation-report map shape — plus a synthetic test fixture. Nothing is wired into runtime paths; the provider stays behind the existing hard-off gate.

The output shape matches the existing consumer contract precisely: `intel_quote` is emitted as hex (`crates/services/src/attestation/verification.rs:151-160` strips an optional `0x` and hex-decodes), and `nvidia_payload` as a JSON string (`verification.rs:420-455` requires string-or-absent). The deliberate omissions (`signing_address`, `event_log`, GPU nonce) are documented rather than fabricated — the right fail-closed posture for an attestation data layer. The pinned SPKI fingerprint in the test was independently verified against openssl (SHA256 of the fixture's SPKI DER matches exactly), and the fixture cert is confirmed synthetic (CN=`chutes-test-fixture`), so no real node material lands in repo history. No logging is added, and `base64 0.22.1` already existed in the lockfile, so no new duplicate dependency.

Two non-blocking observations, both fail-closed:

1. **Null-tolerance asymmetry** — `crates/inference_providers/src/attested/chutes/evidence.rs:46`: the top-level `evidence: Vec<InstanceEvidence>` is not `null`-tolerant, while sibling list fields `failed_instance_ids` and `gpu_evidence` are (with the stated rationale that the Python backend commonly emits `null` for empty lists). An `"evidence": null` response — plausible when a chute has zero live instances — would fail the whole parse with an opaque serde error instead of reaching the clean "instance not found" path. Suggest applying the same `#[serde(default, deserialize_with = "null_default")]` to `evidence`, or commenting why strictness is intended there.

2. **Partial report vs. the shared GPU verifier** — `crates/inference_providers/src/attested/chutes/attestation.rs:96`: the emitted `nvidia_payload` deliberately omits `nonce`, but the existing `verify_gpu` path (`verification.rs:458-470`) hard-fails when the payload nonce (absent → `""`) mismatches the request nonce. This is documented and nothing consumes the transform yet, but it means the follow-up verifier PR cannot route this report through the current shared GPU-verification call unchanged — flagging it so that PR explicitly supplies/checks the Chutes-derived nonce.

**Checks run locally:** `cargo fmt --check` (clean), `cargo clippy -p inference_providers --all-targets -- -D warnings` (clean), `cargo check --workspace` (clean), `cargo test -p inference_providers --lib` (255 passed, including all 20 chutes evidence/attestation tests), `git diff --check` (clean), plus the independent openssl SPKI fingerprint verification of the test fixture. The e2e suite was skipped (requires PostgreSQL; the PR touches no API/database surface). GitHub CI was still queued/in-progress at review time.

Evrard-Nil:
Thanks @PierreLeGuen — addressed in `24ee34b4`:

1. **Null-tolerance asymmetry on `evidence`.** Applied the same `#[serde(default, deserialize_with = "null_default")]` to the top-level `evidence` field so it matches its siblings. An `"evidence": null` response (zero live instances) now parses to an empty list and reaches the clean "instance not found" path instead of an opaque serde error. Added `explicit_null_evidence_tolerated` to lock the behavior. Comment on the field states the rationale.

2. **Partial report vs. the shared GPU verifier (omitted `nonce`).** Agreed, and intentional — leaving it as a no-change forward-flag. The transform's doc comment already calls out that the follow-up Chutes verifier (PR7) must supply/check the **Chutes-derived** GPU nonce rather than routing this report through `verify_gpu` unchanged (which hard-fails on the absent-nonce mismatch at `verification.rs:458-470`). I'll wire exactly that in PR7.

Local checks: `cargo fmt --check` clean, `cargo clippy -p inference_providers --all-targets -- -D warnings` clean, `cargo test -p inference_providers --lib chutes` → 21 passed.

PierreLeGuen (approved):
This PR is a purely additive data layer: serde models for Chutes' `GET /chutes/{id}/evidence` response (`evidence.rs`) plus a pure, fail-closed transform (`attestation.rs`) producing the subset of NEAR attestation-report fields that are format-identical. Nothing is wired at runtime; the Chutes provider stays behind its hard-off gate.

I cross-checked the transform against its downstream consumers in `crates/services/src/attestation/verification.rs` and everything aligns:

- `intel_quote` is consumed as hex (`verification.rs:151-160`, optional `0x` prefix stripped); the base64→hex conversion matches.
- `nvidia_payload` is consumed as a JSON-encoded string (`verification.rs:420-451`); `nvidia_payload(instance)?.to_string()` matches, and the `{arch, evidence_list}` shape matches what's POSTed to NRAS.
- The deliberately omitted fields fail closed: if a Chutes-transformed report were ever fed to NEAR's current shared verifier, the absent GPU nonce mismatches `request_nonce` (`verification.rs:458-468`) and the missing `signing_address`/`event_log` produce `MissingField` errors. The module docs are explicit about which checks belong to the upcoming Chutes-specific verifier.
- `tls_cert_fingerprint` reuses `compute_spki_fingerprint_from_der` (`spki_verifier.rs:88-94`); the pinned fingerprint in the test fixture was independently re-derived with `openssl` and matches. The cert fixture is synthetic (no real hostnames/IPs).
- `null_default` tolerance on the three list fields is correct and tested; required scalars stay strict. No logging added, so no privacy concerns. The `base64 = "0.22"` dependency follows the crate's existing style.

Two non-blocking notes for the follow-up verifier PR:

- `crates/inference_providers/src/attested/chutes/attestation.rs:92` — `decode_b64` trims only surrounding whitespace, so PEM-style line-wrapped base64 (internal newlines, common for certificate fields) would fail to decode. It fails closed and the live endpoint reportedly returns single-line strings, so this is informational; consider stripping internal whitespace or documenting the single-line assumption.
- `crates/inference_providers/src/attested/chutes/evidence.rs:444` — `EvidenceResponse::instance()` returns evidence for an id even if it also appears in `failed_instance_ids`. No security impact (the verifier re-checks all cryptographic bindings), but the next PR should decide explicitly how to treat an instance listed in both.

Local checks: `cargo test -p inference_providers --lib` (256 passed, including the 21 new chutes evidence/attestation tests), `cargo check -p inference_providers`, `cargo clippy -p inference_providers --all-targets -- -D warnings` (clean), `cargo fmt --check` (clean), `git diff --check` (clean). Full e2e suite skipped (requires PostgreSQL; the change is additive pure functions fully covered by the unit runs). CI at review time: Lint and security audit passing; Test Suite still in progress.

Evrard-Nil:
Thanks for the approval @PierreLeGuen 🙏 — both follow-up notes are tracked for the verifier PR (PR7), leaving this approved+green PR unchurned:

1. **`decode_b64` internal whitespace (PEM-style wrapping).** Currently fails closed and the live `/evidence` strings are single-line, so no behavior change needed here. PR7 will either strip internal whitespace defensively or assert the single-line assumption explicitly when it starts consuming these against real quotes.
2. **`instance()` vs `failed_instance_ids`.** PR7 (the `ChutesBackendVerifier`) will make this explicit — an id present in `failed_instance_ids` should be treated as unusable rather than verified, independent of the cryptographic re-checks. I'll add a dedicated test for the both-lists case.

Holding off on merge for the user.

PR #758, opened 2026-06-10
PR7: Chutes as a fully attested inference provider (E2EE, hard-off)

Evrard-Nil:
## Chutes as a fully attested inference provider (`ProviderTier::Attested3p`)

Adds [Chutes](https://chutes.ai) as the first **third-party attested** inference provider. Every Chutes chat request runs a complete trust chain and is served over a verified end-to-end-encrypted channel — **behind a hard-off flag (`ENABLE_CHUTES`, default off)**, so this is zero behavior change until explicitly enabled.

> The companion verifier-hardening PRs (#745 `ProviderTier`, #746 fail-closed `MeasurementPolicy`, #748 strict `ReportDataVerifier` incl. NEAR, #749 per-tier GPU policy) and the data-layer PR (#755) already merged. This PR adds the Chutes-specific verification + transport + provider + pool wiring.

### Why this isn't "transform Chutes' attestation → reuse the NEAR verifier"

The load-bearing anti-substitution values (`report_data`, measurements) are sealed inside the Intel-signed TDX quote — a JSON transform can't fabricate them. And Chutes' bindings genuinely differ from NEAR's, so it needs its **own** verifier, not `StrictBoundReportDataVerifier`. All bindings below were confirmed **empirically against the live GLM-5.1-TEE fleet** (2026-06-10), not assumed.

### The trust chain (what `ChutesBackendVerifier::verify_instance` checks)

1. **TDX quote** — Intel DCAP (`dcap-qvl`), TCB floor, debug bit clear. *Shared verbatim with NEAR* via the extracted `verify_tdx_quote` (NEAR behavior byte-for-byte unchanged — all 31 services attestation tests pass).
2. **`report_data[0:32] == SHA256(nonce ‖ e2e_pubkey)`** — freshness + E2EE-key binding. The inputs are the **string** forms (the 64-hex query nonce + the base64 ML-KEM-768 key), UTF-8-concatenated — matching Chutes' reference verifier `hashlib.sha256((nonce + e2e_pubkey).encode())`. *Gotcha:* decoding either to raw bytes does **not** match. Confirmed 5/5 live.
   - Example: `report_data[0:32] = 344a44eb… ⇒ SHA256("344a44eb…" ‖ "<base64 e2e_pubkey>")`.
3. **`report_data[32:64] == SHA256(SPKI(instance cert))`** — binds the instance's attested key (Chutes' own verifier omits this; we add it). Confirmed 6/6 live.
4. **Measurement** — register-pin MRTD + RTMR0-2 against a **vetted snapshot** of Chutes' published golden values (`GET https://api.chutes.ai/servers/tee/measurements`). Live GLM-5.1-TEE matches `8xh200 v1.3.0` byte-for-byte. RTMR3 is a runtime register (published as 0) and intentionally **not** pinned. Chutes ships **no** RTMR3 event log, so this replaces NEAR's event-log replay + image-hash path.
   - ⚠️ The endpoint is **unsigned** plain JSON over TLS, so we pin a snapshot in code rather than fetch-at-verify-time. The narrowed ask to Chutes is to *sign* it.
5. **GPU** — NVIDIA NRAS, with the **Chutes-derived** nonce (`SHA256(nonce ‖ e2e_pubkey)`, the same value sealed in `report_data[0:32]` and bound into the GPU SPDM evidence), not the raw caller nonce. Reuses the shared `verify_gpu_evidence`.

### Why E2EE transport (not plain TLS, not verified-direct-TLS)

`llm.chutes.ai`/`api.chutes.ai` terminate TLS at a **GoDaddy-CA gateway cert**, not the attested key (verified empirically). The per-instance attested cert is internal-only — its public IP exposes only plaintext `:8080` (uvicorn), with no reachable TLS port presenting it. So verified-direct-TLS (our `verified-proxy` pattern) is **not** available for GLM-5.1-TEE. The verified channel must be Chutes' **ML-KEM-768 + ChaCha20-Poly1305 E2EE**, which rides over the gateway: a request encapsulated to a *specific* attested instance's `e2e_pubkey` can only be decrypted by that instance, so the response is bound to verified software even through load balancing.

**Protocol matched byte-for-byte to Chutes' own clients** — the `pqcrypto` Python reference and, decisively, Chutes' own RustCrypto reference (`github.com/chutesai/e2ee-test`, `ml-kem 0.3.2`), so interop is proven by construction:
- KEM = **ML-KEM-768 FIPS 203** (RustCrypto `ml-kem 0.3.2`, pinned; *not* round-3 Kyber).
- KDF = `HKDF-SHA256(salt = mlkem_ct[:16], info = e2e-{req,resp,stream}-v1)`.
- AEAD = ChaCha20-Poly1305 (12-byte nonce, empty AAD).
- Request blob = `mlkem_ct[1088] ‖ nonce[12] ‖ ct ‖ tag[16]`, plaintext = `gzip(json(openai_request + {e2e_response_pk}))`; non-stream response symmetric (gzipped); streaming = one `e2e_init` + per-chunk frames (not gzipped).
- Endpoints (two-host): model→`chute_id` via `llm.chutes.ai/v1/models`; E2EE via `api.chutes.ai` `/e2e/instances` + `/e2e/invoke` with `X-Instance-Id` + a single-use `X-E2E-Nonce` pinning a specific attested instance.

### Architecture

- `inference_providers::attested::chutes`: `report_data`, `measurements`, `e2ee` (+ `e2ee_stream`), `client`, `verifier_port`, `Provider`.
- The verifier (`ChutesBackendVerifier`) lives in `services` (needs DCAP + NRAS); the `Provider` reaches it through the **`ChutesInstanceVerifier` port** (dependency inversion, since `services → inference_providers` is one-way). The pool injects the concrete verifier.
- Fail-closed throughout: empty measurement allow-list, any binding mismatch, an unvetted measurement, or absent GPU evidence is an **error** — never an unverified fallback. Non-chat modalities + `get_signature` return explicit "unsupported" (Chutes integrity is the AEAD channel, no separate signature).

### Config / enabling

`ENABLE_CHUTES=true` + `CHUTES_API_KEY[_FILE]` + `CHUTES_MODELS=<comma-separated>` (+ optional `PCCS_URL`). Default off.

### Testing

- 56 `inference_providers::attested::chutes` unit tests (report_data, measurement pin, E2EE full client→instance→client round-trips for non-stream + stream, tamper/wrong-session fail-closed, client serde + headers, streaming control paths, provider).
- services attestation tests green incl. the fail-closed ordering; **NEAR's path byte-for-byte unchanged.**
- The one thing local tests can't cover is interop with Chutes' compiled `aegis` server lib — proven by one **live E2EE round-trip on staging** before any prod flag-on (the RustCrypto reference makes this low-risk).

### Open question to Chutes (does not block this PR; gates the prod "attested" badge)

Sign `/servers/tee/measurements` (cosign / in-toto / transparency log) so the golden measurements are cryptographically anchored rather than trust-the-API-over-TLS. We ship today by pinning a vetted snapshot.


Evrard-Nil:
Thanks @gemini-code-assist — all three addressed in `c256d07a`:

1. **(high) `0x`-prefixed custom nonce in `get_attestation_report`.** Fixed: the caller's nonce is now normalized (strip an optional `0x`, validate 32-byte hex, lowercase) and the bare-hex form is used — so a `0x`-prefixed nonce is *honored* (and bound consistently) rather than silently falling through to a random one. We send the normalized bare-hex to Chutes and bind against the same string.
2. **(medium) per-line `Vec` alloc in the SSE loop.** Now takes the line owned (`from_utf8_lossy(&buf[..pos]).into_owned()`) before `drain`, dropping the intermediate `Vec` (one allocation per line, no borrow held across the buffer mutation).
3. **(medium) double `trim()` in `measurements::norm`.** Trims once.

`cargo test -p inference_providers --lib chutes` → 56 passed; fmt + clippy clean.

PierreLeGuen (changes requested):
The core of this PR is solid: the E2EE protocol, report_data binding, and measurement pinning are fail-closed, well-documented, and well-tested (including tamper cases), and the `verify_tdx_quote` extraction in the NEAR path is a faithful behavior-preserving refactor. However, there are two integration-layer issues that break the feature when the flag is actually turned on, so I'd like to see those fixed before merge.

**Blocking / important**

1. **Env-registered Chutes providers are wiped by the pool refresh.** `crates/api/src/lib.rs:678-699` registers Chutes models from `CHUTES_MODELS` via `pool.register_provider`, but the periodic refresh (`crates/services/src/inference_provider_pool/mod.rs:3280, 3360-3402`) builds `valid_model_names` solely from the DB-backed `fetch_inference_url_models()`/`fetch_external_models()` and then calls `remove_stale_providers`. Config-only Chutes models are in neither source, so with `ENABLE_CHUTES=true` the provider survives only until the first refresh tick (default 300s; the refresh loop starts at `lib.rs:669`, before registration) and then silently disappears — or gets overwritten if a companion external DB row exists. Fix: track statically-registered (pinned) models in the pool and exclude them from stale removal (or seed `valid_model_names` with them), and add a pool test asserting an env-registered provider survives a refresh cycle.

2. **Streamed Chutes completions will likely record zero usage (billing loss).** `request_body` (`crates/inference_providers/src/attested/chutes/mod.rs:244`) pins only `model` and `stream` — unlike every other provider, which forces `stream_options.include_usage = true` (`non_attested/external/openai_compatible.rs:187`, `attested/nearai/mod.rs:1396`) — and the E2EE SSE adapter (`e2ee_stream.rs:98-103`) deliberately drops Chutes' outer usage-only events. If the inner stream's final chunk carries no usage (the OpenAI-compatible default), `last_usage_stats` stays `None` and `record_usage_and_metrics` returns without recording anything (`crates/services/src/completions/mod.rs:194-201`): streamed tokens are neither billed nor counted against org limits. Fix: inject `stream_options: {"include_usage": true}` when `stream=true`, and/or translate the dropped outer usage event into a usage-bearing parsed chunk; add a test asserting usage survives the E2EE stream path.

3. **Per-request attestation overhead and startup cost** (operational; should gate flag-on if not merge). Every chat request runs `GET /v1/models` (re-resolving the static model→`chute_id` map, `client.rs:144`), `GET /e2e/instances`, `GET /evidence`, an Intel PCCS DCAP collateral fetch, and an NVIDIA NRAS round trip before the invoke (`attested/chutes/mod.rs:185-246`) — multi-second TTFT plus per-request availability/rate-limit dependencies on PCCS and NRAS. Relatedly, `register_provider` (`inference_provider_pool/mod.rs:578-582`) runs signing-key attestation discovery — two full discover→evidence→DCAP→NRAS rounds per model — at startup, for signing keys Chutes intentionally never provides. At minimum cache `chute_id` per model and skip signing-key discovery for Chutes; decide on a short-TTL per-instance verification cache before production flag-on.

**Minor**

- `crates/inference_providers/src/attested/chutes/mod.rs:223-235`: instance selection always takes the first instance and `nonces[0]`, with no check that the token is still unused/valid by invoke time — concurrent requests can collide (TOCTOU), and `nonce_expires_in`/`nonce_expires_at` are parsed but never checked despite the potentially slow DCAP+NRAS chain in between. Suggest one re-discover-and-retry on nonce rejection.
- `crates/inference_providers/src/attested/chutes/report_data.rs:96-98`: the nonce format gate strips an optional `0x` prefix, but `freshness_digest` hashes the verbatim (prefixed) string, so a prefixed nonce passes validation yet binds differently from its bare form. Harmless today (self-generated nonces are bare hex) — reject the prefix outright to remove the trap.
- `crates/config/src/types.rs:1050-1055`: a `CHUTES_API_KEY_FILE` read failure is swallowed with `.ok()`, and an empty `CHUTES_API_KEY` passes as `Some("")`. For a security-sensitive opt-in, log the read error (path only) or fail startup when `ENABLE_CHUTES=true` and no key can be sourced; filter empty keys.
- `crates/inference_providers/src/attested/chutes/attestation.rs:140-157`: the `request_nonce` field in the `to_near_report` output is never compared by the Chutes verifier (which substitutes the derived `SHA256(boot_nonce ‖ e2e_pubkey)` at `services/src/attestation/chutes.rs:144`), but NEAR's `verify_attestation_report` does compare it (`verification.rs:482`). Routing a Chutes partial report through the NEAR verifier would fail confusingly — document that the output must not go through NEAR's path, or drop the field. Similarly, the `e2e_pubkey` echoed in the self-describing report (`mod.rs:472`) is the verbatim string from `/e2e/instances` while the E2EE channel uses the decoded bytes; any mismatch fails closed at encryption, but worth documenting which form the report carries.

**Checks run**

- `cargo check -p api -p inference_providers -p services` — passed.
- `cargo test -p inference_providers --lib attested::chutes` — 56/56 passed.
- `cargo test -p services --lib attestation` — 34 passed, 1 ignored (pre-existing); includes the new Chutes fail-closed ordering tests and the NEAR `verification::tests` regression suite (no failures from the refactor).
- `cargo clippy -p inference_providers -p services -p config -p api --no-deps` — clean.
- `git diff --check origin/main...HEAD` — clean.
- GitHub CI on the PR head: Lint, Test Suite, and security_audit all green.
- Not run: DB-backed e2e suite (no PostgreSQL in this environment) and live Chutes interop (needs staging credentials; the PR notes interop with Chutes' compiled server lib is only provable via a live round-trip).

Evrard-Nil:
Thanks for the thorough review @PierreLeGuen — all addressed in `f273ab08`.

**Blocking**
1. **Refresh wiping env-registered providers.** The pool now tracks **pinned** models (registered out-of-band, not from DB discovery) and excludes them from `remove_stale_providers`. New `register_pinned_provider` is used by the api hook; it also **skips the signing-key attestation discovery** (which Chutes never satisfies — that removes the two startup discover→evidence→DCAP→NRAS round trips you flagged in #3). Added `pinned_provider_survives_refresh` asserting an env-registered provider survives a refresh tick whose valid set lacks it (and that a non-pinned stale model is still removed).
2. **Streamed usage / billing loss.** `request_body` now injects `stream_options.include_usage=true` when `stream=true`, so the final inner chunk carries usage and flows through `record_usage_and_metrics` (matching the other providers). Test asserts it's set for streaming and absent for non-streaming.

**Operational (#3)** — at-minimum items done now: `chute_id` is cached per provider (`tokio::OnceCell`, no `/v1/models` per request), and signing-key discovery is skipped at registration (above). The **short-TTL per-instance verification cache** I've left as a tracked follow-up **before prod flag-on** (the flag is off) — agreed it shouldn't gate merge but must precede enabling.

**Minor**
- `report_data` verifier now **rejects a `0x`-prefixed nonce** outright (it's hashed verbatim, so the prefix would validate yet bind differently) — closes the trap; `get_attestation_report` already normalizes. Test added.
- config: empty `CHUTES_API_KEY` is filtered to `None` (no `Some("")`); a `CHUTES_API_KEY_FILE` read error is logged (path only). With `ENABLE_CHUTES=true` and no key sourced, the api hook already warns and registers nothing.
- Documented that `to_near_report`'s output must **not** be routed through NEAR's `verify_attestation_report` (the `request_nonce` field; Chutes seals `SHA256(nonce‖e2e_pubkey)` at `[32:64]`'s sibling). The live Chutes verifier consumes the individual helpers, not the whole map.
- **TOCTOU / nonce retry** (re-discover + retry on a 403 nonce rejection, + honoring `nonce_expires_*`): tracked as a follow-up alongside the verification cache, before flag-on.

Local: 56 chutes + `pinned_provider_survives_refresh` + services attestation tests pass; fmt + clippy clean across all four crates. Re-requesting your review.

lloydmak99 (approved):
Reviewed the full diff — the E2EE transport (ML-KEM-768 + HKDF-SHA256 + ChaCha20-Poly1305), the Chutes verifier, and the pool/config wiring. Trust chain is sound, the hard-off gate is correctly wired, and the prior round of fixes (nonce normalization, SSE allocation, refresh pinning, streaming billing) all look correct.

**One nit not in prior comments** — `gunzip` in `e2ee.rs` has no decompressed-size bound:

```rust
fn gunzip(comp: &[u8]) -> Result<Vec<u8>, E2eeError> {
    let mut dec = flate2::read::GzDecoder::new(comp);
    let mut out = Vec::new();
    dec.read_to_end(&mut out)  // unbounded
        ...
```

This runs after the AEAD open, so the content is from an attested instance — not a random adversary. But a legitimate backend sending a large response (or a degenerate compression ratio for repetitive token output) could still expand significantly. Adding a cap (e.g. 64 MB via `Read::take`) would give defense-in-depth with no real cost. Non-blocking given the hard-off flag and attestation requirement, but worth a follow-up before this goes to staging-on with real traffic.

Otherwise clean. ✅

PierreLeGuen (changes requested):
This is a carefully built feature overall: the verification chain is fail-closed at every stage, the E2EE protocol has solid round-trip and tamper tests, secrets are redacted, and the whole thing sits behind a hard-off flag. A few issues need fixing before merge, though — most notably a one-sided pinning protection in the provider pool and stream-truncation handling.

**Blocking / important**

- `crates/services/src/inference_provider_pool/mod.rs:3232` (with `register_pinned_provider` at `mod.rs:608`): the pinned-provider protection only guards against stale-*removal*. The periodic refresh does `mappings.model_to_providers.insert(model_name, providers)` for every DB-discovered model without consulting `pinned_models`, so a DB-registered model with the same id (e.g. someone registers `zai-org/GLM-5.1-TEE` through the normal flow) silently replaces the attested, per-request-verified E2EE provider — and since the name stays in `pinned_models`, the substitute is then immune to stale-removal too. That's exactly the substitution the rest of the PR works to prevent. Skip (or loudly reject) pinned model names in the refresh insertion loop and in `register_providers` (`mod.rs:651`), and add a test mirroring `pinned_provider_survives_refresh` for the overwrite direction.

- `crates/api/src/lib.rs:688` / `crates/services/src/completions/mod.rs:1191`: `CHUTES_MODELS` registers providers in the pool, but completions still reject any model lacking an active DB catalog row, and the PR adds no migration/seed/admin step for the configured models. As written, `ENABLE_CHUTES` + key + models can yield a registered-but-unusable model. Add the catalog migration/admin docs, or startup validation for every `CHUTES_MODELS` entry (including intended provider/attestation metadata).

- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:120`: the E2EE SSE adapter treats EOF as normal even when no `data: [DONE]` was seen or `buf` still holds a partial line. The route layer then appends its own `[DONE]`, turning a truncated encrypted stream into an apparently successful completion. Track whether a terminal `[DONE]` was observed and fail the stream on EOF without it.

- `crates/inference_providers/src/attested/chutes/mod.rs:241` (and `verify_and_prepare`): nonce/instance selection is fragile under concurrency. The provider always picks the *first* instance with a non-empty `e2e_pubkey`/nonce list and always uses `inst.nonces[0]`, with multi-second DCAP/NRAS round trips between discovery and `/e2e/invoke`. Two concurrent requests can consume the same single-use token and one fails spuriously; a single failing instance takes down all requests even when healthy instances exist; and there's no emptiness guard on `nonces` at the point of use. Pick a random nonce, iterate over candidate instances on verification/nonce failure (or one retry that re-discovers), and add a cheap `nonces.is_empty()` guard before indexing. If Chutes mints tokens per `/e2e/instances` call, confirm and document that instead.

**Before flag-on (non-blocking now, given the hard-off flag)**

- `crates/services/src/attestation/chutes.rs:3243-3282` / `verification.rs`: every chat request performs discover → evidence fetch → a *fresh* `dcap_qvl::CollateralClient` + PCCS collateral fetch → an NVIDIA NRAS POST before the first inference byte — 4+ external round trips on the hot path, and a PCCS/NRAS outage takes the model fully down. Per-request freshness (the nonce binding) doesn't require re-fetching collateral: cache collateral with a TTL and consider a short-TTL verified-instance cache. Address or explicitly load-test before enabling in production.
- `crates/inference_providers/src/attested/chutes/mod.rs:256-264` / `e2ee_stream.rs:103-104`: streaming usage relies on the inner stream's final chunk carrying usage (outer usage-only events are dropped), but Chutes' API gives no contractual guarantee it will. Validate billing/usage end-to-end against the live staging endpoint as part of the flag-on checklist.

**Minor**

- `crates/services/src/attestation/chutes.rs:144-155`: GPU nonce injection via `as_object_mut()` silently skips if `nvidia_payload` isn't a JSON object. Make it fatal (`ok_or(...)`) rather than proceeding without the binding.
- `0x`-prefix policy is inconsistent: `get_attestation_report` (`chutes/mod.rs:428-463`) normalizes/strips it while `ChutesReportDataVerifier::verify` (`report_data.rs:96-98`) rejects it. Reconcile to one policy (rejecting outright, per the `report_data.rs` doc comment, seems right) before a refactor breaks the contract silently.
- `crates/api/src/lib.rs:426`: `PCCS_URL` is read via `std::env::var` directly instead of through the config crate, and diverges from `AttestationVerifier::from_env()`'s handling. Consolidate to one source of truth.
- `crates/config/src/types.rs:1050-1061`: `CHUTES_API_KEY_FILE` read failure goes to `eprintln!` instead of `tracing::warn!` — opaque in containerized deployments for a security-sensitive opt-in.
- `env.example` doesn't mention the new operator-facing config (`ENABLE_CHUTES`, `CHUTES_API_KEY[_FILE]`, `CHUTES_MODELS`, and the `PCCS_URL` interaction). Add them, defaulted off.
- `e2ee_stream.rs` (document only): content frames are independently AEAD-sealed under one stream key with random nonces and no sequence numbers, so an on-path gateway can drop/reorder/replay individual frames undetected. Inherent to Chutes' published protocol — note it in the module's security notes alongside the unsigned-measurements ask.

**Checks run locally**: `cargo check` and `cargo clippy --no-deps` across `inference_providers`, `services`, `config`, `api` — clean; `cargo test -p inference_providers --lib chutes` — 56 passed; `cargo test -p services --lib attestation` — 34 passed, 1 ignored (includes the new fail-closed ordering tests); `inference_provider_pool::tests::pinned_provider_survives_refresh` — passed. GitHub CI on the PR head: Lint, Test Suite, and security_audit all green. Skipped: DB-backed e2e suite (needs PostgreSQL) and live Chutes/PCCS/NRAS interop (network-dependent; the PR itself gates flag-on behind a staging round-trip).

Evrard-Nil:
Thanks again @PierreLeGuen — round-2 addressed in `944ac620`.

**Blocking**
1. **One-sided pinning.** Fixed all three DB-discovery insert paths (`load_external_providers`, `load_inference_url_models`, `register_providers`) to **skip pinned model names** — a DB model with the same id can no longer clobber the attested per-request-verified provider, and `register_pinned_provider` already skips signing-key discovery. Added `pinned_provider_not_overwritten_by_discovery` (alongside `pinned_provider_survives_refresh`).
2. **Stream truncation.** The E2EE SSE adapter now **fails on EOF without a terminal `[DONE]`** (truncated/interrupted stream) instead of ending cleanly and letting the route mint a success. + `stream_without_done_is_error`.
3. **Nonce/instance concurrency.** `verify_and_prepare` now **iterates over candidate instances** (one bad/unverifiable instance no longer fails all requests — still never an unverified fallback), picks a **pseudo-random nonce token** (not `[0]`) to cut single-use collisions, and the candidate filter guards `nonces` non-emptiness. The deeper *invoke-time* nonce-403 re-discover/retry I've left as a tracked follow-up (below).

**Minor**
- **0x policy reconciled** to one rule: `get_attestation_report` now rejects a `0x`-prefixed nonce (mints fresh), matching `ChutesReportDataVerifier`.
- **GPU nonce injection is fatal** if `nvidia_payload` isn't a JSON object (`MalformedGpuPayload`) — no more silently-unbound GPU evidence.
- **e2ee_stream security note**: documented that frames are independently AEAD-sealed with no sequence numbers, so on-path drop/reorder/replay isn't cryptographically detected (the `[DONE]` check only catches truncation) — same class as the unsigned-measurements caveat.
- **env.example**: added `ENABLE_CHUTES` / `CHUTES_API_KEY[_FILE]` / `CHUTES_MODELS` / `PCCS_URL` (defaulted off) including the catalog requirement.

**On #2 (catalog row)** — you're right that a pool-registered model still needs an active catalog row for `/v1/chat/completions` to route to it. I've documented the operator step in `env.example` for now; how would you prefer to provision it — (a) a migration/seed shipped in this PR, (b) startup validation that warns per `CHUTES_MODELS` entry, or (c) leave it an out-of-band admin step? Happy to add (a) or (b) here.

**Before flag-on (tracked, flag is off):** TTL collateral cache + short-TTL verified-instance cache (the per-request DCAP+NRAS hot-path cost); invoke-time nonce-403 retry; `config` `eprintln!`→`tracing`; `PCCS_URL` single source of truth; live staging billing/usage validation.

57 chutes + pinned/overwrite + services attestation tests green; fmt + clippy clean across all four crates. Re-requesting your review.

PierreLeGuen (changes requested):
This is a strong PR overall: the verification chain (DCAP quote → report_data bindings → pinned measurements → NRAS GPU) is fail-closed at every stage with tests locking the ordering, secrets are redacted, the pool pinning covers all three production insertion paths plus stale-removal, and the feature is verifiably hard-off by default. The `verify_tdx_quote` extraction is a faithful, behavior-preserving refactor of the NEAR path. That said, a few issues — one in the streaming completeness story, two in the attestation contract — should be resolved before this ships as "fully attested."

**Blocking**

- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:80-86` (and the truncation check at ~160-168): a plaintext outer `data: [DONE]` from the untrusted gateway is accepted as a clean stream terminus — even before `e2e_init` or any decrypted frame. The EOF-without-`[DONE]` check exists precisely so a truncated stream can't look like a successful completion, but the gateway (the on-path adversary this E2EE layer is built to distrust) can drop the remaining encrypted frames and inject `data: [DONE]`, yielding an undetectably truncated yet "successful" response. The code already parses an authenticated inner (decrypted) `[DONE]`; if Chutes emits the terminator inside the encrypted channel (verifiable in the planned staging round-trip), accept only that and treat a bare outer `[DONE]` as truncation. If the protocol only terminates with the plaintext outer `[DONE]`, amend the module's security note (which currently claims the check "catches outright truncation") to state the terminator is forgeable, add it to the tracked asks to Chutes alongside missing frame sequence numbers, and reconsider exposing streaming as attested until then.

**Important**

- `crates/services/src/attestation/chutes.rs:137`: verification pins MRTD/RTMR0-2 but skips RTMR3/runtime app measurement. The existing attested-third-party policy hinges on an RTMR3-verified OS/app measurement, so this can certify the boot chain without authenticating the actual runtime inference stack. Verify an authenticated runtime/app measurement, or don't surface Chutes as fully attested until Chutes provides one.
- `crates/inference_providers/src/attested/chutes/mod.rs:545`: `get_signature` always returns unsupported, while the existing attested path fetches and stores provider signatures for every attestation-supported chat. This breaks the `/v1/signature/{chat_id}` contract and adds failure noise on attested Chutes completions. Add a distinct AEAD-integrity capability path, or keep Chutes out of the signature-backed attestation flag.

**Non-blocking**

- `crates/inference_providers/src/attested/chutes/e2ee.rs:87` (`gunzip`): no decompressed-size bound (`read_to_end` into an unbounded `Vec`). A gzip bomb requires a malicious-but-attested instance, so risk is low, but `dec.take(LIMIT)` with an over-limit error is cheap defense-in-depth — this was also raised in an earlier review round and not yet applied.
- `crates/inference_providers/src/attested/chutes/client.rs:122`: the client-wide `timeout` covers the entire response including streamed body reads, so long generations via `invoke_stream` can be cut off. Prefer `connect_timeout` on the builder with explicit per-request control.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`verify_and_prepare`): every request does `/e2e/instances` + `/evidence` + DCAP collateral + an NRAS POST before encapsulating, adding TTFT latency and making PCCS/NRAS hard availability dependencies. Before production flag-on, consider a short-TTL cache of verified `(instance_id, e2e_pubkey, measurement)` — the encapsulation to the attested key, not the per-request quote, is what binds the response.
- `crates/services/src/attestation/chutes.rs:210-218`: golden measurements are pinned in source, so any Chutes VM image update breaks attestation until a redeploy. A config-driven override (e.g., a measurements file path) would let operators rotate pins without a code release; at minimum document the update procedure.

**Checks run locally**: `cargo fmt --all -- --check` and `cargo clippy -p inference_providers -p services --lib` clean; `cargo check` on `inference_providers`, `services`, and `api` passed; `cargo test -p inference_providers --lib attested::chutes` 57 passed (E2EE round-trips, tamper/wrong-session fail-closed, report_data bindings, measurement pinning, stream control paths); `cargo test -p services --lib attestation` 34 passed, 1 ignored (pre-existing network-gated); pinned-pool tests and `cargo test -p config --lib` (23) passed. Not run: the full e2e suite (requires PostgreSQL) and live Chutes interop (network; the PR gates flag-on behind a staging round-trip). GitHub CI Test Suite was still in progress at review time; Lint and security_audit were green.

Evrard-Nil:
Thanks @PierreLeGuen — round-3.

**Blocking — fixed in `16d00344`:**
- **Forgeable stream terminator.** Good catch. The adapter now accepts **only an authenticated inner `[DONE]`** (decrypted from an `e2e` frame) as a clean terminus; a bare *plaintext outer* `[DONE]` from the gateway is ignored, so a dropped-frames + injected-`[DONE]` attack surfaces as truncation rather than a fake success. Module security note rewritten accordingly, and the open item — *confirm on the staging round-trip whether Chutes emits the terminator inside the encrypted channel; if it only sends the outer plaintext `[DONE]`, the terminator is forgeable* — is added to the tracked Chutes asks (with the missing frame sequence numbers + unsigned measurements). Tests updated.

**Non-blocking — also done:**
- `gunzip` now bounds decompression (64 MiB via `take` + over-limit error) — gzip-bomb defense in depth.
- client now uses `connect_timeout` on the builder + **per-request** timeouts on the non-streaming calls only, so long streamed generations via `invoke_stream` aren't cut off by a client-wide total timeout.

**The two "important" items — I'd like your steer (they shape the attestation *claim/scope*):**
- **RTMR3 / runtime-app measurement.** Chutes publishes a golden **RTMR3 = 0** (boot value); the *runtime* RTMR3 (the extended app/IMA layer) is **not published anywhere**, so there is no external value to pin — we genuinely cannot authenticate the running inference stack today, only the boot chain (MRTD/RTMR0-2) + TDX + GPU + freshness + E2EE-key binding. Options: (a) ship with the boot-chain attested, document the runtime-app gap, **qualify the "fully attested" wording**, and add a signed-runtime-RTMR3 ask to Chutes; or (b) don't surface Chutes as attested until Chutes provides a runtime measurement. I lean (a) — it's a real, strong attestation with an honest, documented limitation — but it's a product call.
- **`get_signature` vs the signature-backed flow.** Chutes' integrity is the AEAD channel, not a per-response signature, so it shouldn't be in the signature-fetch path. I'll add a provider **capability** (`supports_chat_signatures()` → false for Chutes) so the completion flow skips the fetch (no noise / contract break), unless you'd rather gate it off the attestation flag differently.

**Tracked before prod flag-on:** short-TTL verified-instance + collateral cache; invoke-time nonce-403 retry; config `eprintln`→`tracing`; `PCCS_URL` single source; config-driven golden-measurement override; live staging billing/usage + terminator verification.

57 chutes tests green; clippy clean. Pushing the capability fix once you confirm the claim direction.

Evrard-Nil:
Update — both "important" items now resolved in `44ae2ed8` (thanks to a correction from @Evrard-Nil on the measurements endpoint).

**#2 (RTMR3 / runtime app measurement) — RESOLVED.** I was wrong that the runtime RTMR3 was unpinnable: `/servers/tee/measurements` returns **two** sets per config — `boot_rtmrs` (RTMR3 = 0) **and `runtime_rtmrs`** whose RTMR3 is the runtime-extended value. Confirmed live: **6/6 GLM-5.1-TEE quotes match `runtime_rtmrs.RTMR3`** for `8xh200 v1.3.0` byte-for-byte. So we now register-pin the **full chain** — MRTD + RTMR0-2 (boot: firmware/kernel/cmdline) **and the runtime RTMR3 (running app/IMA layer)**. A genuine quote running modified software at *either* layer is rejected (`rejects_runtime_rtmr3_mismatch`). This authenticates the actual inference stack, so the "fully attested" claim holds.

**#3 (`get_signature` vs the signature-backed flow) — fixed.** Added an `InferenceProvider::supports_chat_signatures()` capability (default `true`; Chutes `false`). `store_chat_signature_from_provider` now early-returns for AEAD-integrity providers, so the attestation flow skips the signature fetch entirely instead of calling `get_signature` and recording a verification-failure metric on every attested Chutes completion.

**#1 (streaming terminator) + the non-blocking defenses** (gunzip bound, connect_timeout) were in the previous push (`16d00344`).

**Still tracked before prod flag-on** (flag is off): short-TTL verified-instance/collateral cache (per-request DCAP+NRAS hot path); invoke-time nonce-403 retry; config `eprintln`→`tracing`; `PCCS_URL` single source; config-driven golden-measurement override; live staging verification (billing/usage, the inner-vs-outer `[DONE]` terminator, and the unsigned-measurements + frame-ordering asks to Chutes).

All green: 57 chutes + 34 services attestation tests; clippy clean across the 4 crates. Re-requesting your review.

PierreLeGuen (changes requested):
This is a well-engineered PR: fail-closed throughout, hard-off by default, the `verify_tdx_quote` extraction on the NEAR path is behavior-preserving and covered by existing tests, pool pinning prevents DB discovery from substituting an unverified provider (verified across all insertion paths, including `load_external_providers` and the refresh loop, whose first tick correctly fires after the interval), and the protocol gotchas (string-form hash preimages, forgeable outer `[DONE]`) are caught and regression-tested. That said, a few issues should be resolved — or explicitly gated — before this merges as-is, since two of them go to whether streaming can honestly be called attested.

**Important**

- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:16-34` — the Chutes stream protocol has no authenticated sequence numbers, frame count, or transcript hash, yet the provider exposes streaming as an attested data path. An on-path gateway can drop, reorder, or replay AEAD-valid `e2e` frames and still forward an authenticated inner `[DONE]`, yielding a successful but altered/truncated response. Relatedly, if Chutes in practice emits only the plaintext outer `[DONE]` (which this code deliberately ignores), every streaming request will fail as truncation. Either ship non-streaming only, or make the staging verification plus the Chutes-side asks (sequence numbers, guaranteed inner terminator) explicit release criteria before streaming is exposed as attested.
- `crates/inference_providers/src/attested/chutes/client.rs:131` — the streaming client sets only `connect_timeout`; `invoke_stream` has no request or idle/read timeout, so a gateway that stalls after headers holds the request (and concurrency slots) indefinitely. Add a bounded no-progress timeout on body reads that doesn't cap actively streaming long generations.
- `crates/services/src/attestation/chutes.rs:122` + `verify_and_prepare` in `crates/inference_providers/src/attested/chutes/mod.rs` — every chat request performs discovery, evidence fetch, a DCAP collateral fetch (fresh `CollateralClient` each time), and an NVIDIA NRAS POST, serially per candidate. This adds multi-second TTFT and makes PCCS/NRAS hard per-request availability dependencies — an NRAS hiccup takes down all Chutes inference. Before flag-on, add a short-TTL verified-instance cache keyed by `(instance_id, e2e_pubkey)` and/or collateral caching.

**Minor**

- `e2ee_stream.rs:169` — the line buffer grows unboundedly if no newline arrives, and `invoke_nonstream` (`client.rs:217`) reads the response body uncapped (`MAX_DECOMPRESSED` only bounds post-decrypt gunzip). Cap both, consistent with the existing gzip-bomb guard.
- `client.rs:60-66` — `nonce_expires_in`/`nonce_expires_at` are parsed but never used; the single-use invoke token is consumed only after the multi-second DCAP+NRAS verification. Check expiry before invoke and retry once with fresh discovery on token rejection.
- `mod.rs:204` — the model→`chute_id` mapping is memoized forever in a `OnceCell`; a Chutes redeploy under a new chute id breaks the provider until restart. Invalidate on consistent 404s. (`get_attestation_report` also bypasses this cache — harmless inconsistency, worth unifying.)
- `mod.rs:393` vs `report_data.rs` — freshness binding hashes the verbatim `e2e_pubkey` while E2EE encapsulation uses `e2e_pubkey.trim()`. Canonicalize once at the discovery boundary.
- `crates/api/src/lib.rs:679` — `PCCS_URL` is read directly from `std::env` in route wiring; move it into `ExternalProvidersConfig` like every other setting.

None of these are correctness bugs in the dark-launched code, but the streaming-integrity and timeout items bear directly on the PR's attestation claim and are cheap to address (or formally gate) now.

**Checks run locally:** `cargo fmt --check` (clean), `cargo clippy -p inference_providers -p services -p config -p api` (clean), `cargo check --workspace` (clean), `cargo test -p inference_providers attested::chutes` (57 passed), `cargo test -p services --lib attestation` (34 passed, 1 ignored), services attestation integration fixtures (2 passed), pool pinning tests (2 passed). GitHub CI (Lint, Test Suite, security_audit) green. Not run: DB-backed e2e suite (requires PostgreSQL) and live Chutes/PCCS/NRAS interop, which the PR itself gates behind the staging round-trip.

Evrard-Nil:
Thanks @PierreLeGuen — round-4 in `bec8c2a6`.

**Important**
- **Streaming-as-attested → formally gated.** Streaming is now **off by default** behind `CHUTES_ENABLE_STREAMING`. Non-streaming is the honest attested default; `chat_completion_stream` refuses (before any network) unless opted in, with a message that frame ordering isn't cryptographically authenticated. Opt-in is reserved for staging verification + once Chutes adds frame sequence numbers / a guaranteed inner terminator — so streaming is never silently exposed as attested with the integrity gap you flagged. (+ `streaming_gated_off_by_default` test.)
- **`invoke_stream` no-progress timeout.** The client now sets a **read timeout** (inactivity between body reads): a gateway that stalls after headers no longer holds the request/concurrency slot indefinitely, and it does **not** cap an actively streaming long generation (the clock resets per chunk).
- **Per-request DCAP/NRAS hot path.** Agreed it shouldn't gate merge (flag off); tracked as a release criterion before flag-on — a short-TTL verified-instance cache keyed by `(instance_id, e2e_pubkey)` + collateral caching.

**Minor**
- **Bounded reads:** `invoke_nonstream` caps the response body (Content-Length + post-read backstop); the SSE line buffer is capped (no unbounded growth without a newline) — consistent with the gunzip guard.
- **`e2e_pubkey` canonicalized once** (trim) and used for *both* the attestation binding and the encapsulation, so they can't diverge.
- **`PCCS_URL`** (and `CHUTES_ENABLE_STREAMING`) moved into `ExternalProvidersConfig` — one source of truth, no ad-hoc `std::env` reads in route wiring. `env.example` documents them.

**Deferred (tracked, flag-off):** nonce-expiry check + one 403 retry-with-rediscovery; `chute_id` `OnceCell` 404-invalidation; the verified-instance/collateral cache.

58 chutes + 34 services attestation tests green; fmt + clippy clean across the 4 crates. Re-requesting your review.

PierreLeGuen (changes requested):
This is a carefully built PR: fail-closed ordering is unit-tested, the `verify_tdx_quote` extraction is behavior-preserving, secrets are redacted, streaming is honestly gated off by default with the rationale documented, and pool pinning prevents DB discovery from substituting the attested provider. One important issue needs fixing before merge, plus a handful of non-blocking items.

**Important — internal fields leak into the Chutes request body** (`crates/inference_providers/src/attested/chutes/mod.rs:315`): `request_body()` serializes `ChatCompletionParams` directly, and since `extra` is flattened, the JSON sent to Chutes includes internal tracing fields (`x_request_id`, `x_org_id`, `x_workspace_id`) and any client E2EE headers. This leaks internal identifiers to a third party, and because client-facing E2EE headers aren't implemented as response encryption for Chutes, a request carrying partial encryption headers can be treated by the API as E2EE-active while returning a normal Chutes response. Strip the tracing keys before serialization, explicitly reject (or fully implement) client E2EE headers on the Chutes path, and add a unit test asserting these keys never appear in the encrypted request plaintext.

Non-blocking findings:

- `crates/services/src/inference_provider_pool/mod.rs:3163` (eviction sites at 2971/3125/3145): the evicted-models removal path in `load_inference_url_models` deletes `model_to_providers` entries without consulting `pinned_models`, unlike the three guarded insert sites and `remove_stale_providers`. If a DB inference-url model shares a name with a pinned Chutes model and its URL is evicted (e.g. Blocked fingerprint), the pinned mapping is removed and the insert guard then prevents restoration — the model serves nothing until restart. Filter `evicted_models` against the pinned set, mirroring the other four sites.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`chute_id_cache`): the model→`chute_id` mapping is memoized in a `OnceCell` forever. If Chutes redeploys the model under a new `chute_id`, every subsequent request hits `/e2e/instances/{stale-id}` and fails until restart. Invalidate the cache on a 404/"no instances" response so the provider self-heals.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`get_attestation_report`): passes `&inst.e2e_pubkey` to `attest_instance` untrimmed, while the data path canonicalizes with `.trim()` in `verify_and_prepare`. A whitespace-padded key from Chutes would verify on one path and fail on the other — use the trimmed form in both.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`pick_nonce`): the `getrandom::fill` fallback returns `nonces[0]` on RNG failure, so under that condition all concurrent requests to an instance pick the same single-use token — the opposite of the "reduces collisions" intent. A wrapping atomic counter (`fetch_add % nonces.len()`) is infallible and correct under contention.
- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs` (`handle_outer_payload`): a second `e2e_init` mid-stream silently replaces the established stream key; since ML-KEM decapsulation never fails (implicit rejection), an on-path gateway can inject a bogus `e2e_init` and make all subsequent genuine frames fail decryption. Impact is DoS-only, but rejecting a second `e2e_init` gives a clearer failure mode.
- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs` (EOF comment, ~line 195): the comment claims the route layer does not mint its own `[DONE]`, but `crates/api/src/routes/completions.rs:1497` appends a gateway-minted `[DONE]` after the SSE error frame. Behavior is still safe (the error frame precedes it), but correct the comment so readers don't rely on the stated guarantee.
- `crates/services/src/attestation/chutes.rs` + `verification.rs` (operational): each chat request now performs 4+ sequential network round trips before inference, including a fresh PCCS collateral fetch and an NVIDIA NRAS submission per request. The per-request nonce design makes this intentional, but measure TTFT on staging and consider caching DCAP collateral (which is not request-specific) before flag-on.

Checks run locally: `cargo fmt --all -- --check` clean; `cargo check` across `api`, `services`, `inference_providers`, `config` clean; `cargo clippy --all-targets` clean; `cargo test --lib --bins` 359 passed / 0 failed / 1 ignored (includes the 56+ new chutes tests, fail-closed-ordering, pool pinning, and `verify_tdx_quote` regression tests). GitHub CI (Lint, Test Suite, security_audit) green on the head commit. Not run: the e2e suite (requires PostgreSQL) and live Chutes interop — the PR itself flags the staging round-trip as the remaining gate before flag-on.

Evrard-Nil:
Thanks @PierreLeGuen — round-5 in `84157cf7`.

**Blocking — internal fields no longer leak to Chutes.**
- `request_body()` now strips `ChatCompletionParams.extra` of **all** internal keys before serializing: tracing (`x_request_id`/`x_org_id`/`x_workspace_id`) and the client-E2EE markers (`x_signing_algo`/`x_client_pub_key`/`x_model_pub_key`/`x_encryption_version`/`x_encrypt_all_fields`) — none reach the third party in the encrypted body. (+ `request_body_strips_internal_and_e2ee_keys` asserting they're absent from the request plaintext.)
- A request carrying client-facing-E2EE intent (`x_client_pub_key`) is now **rejected** on the Chutes path (chat + stream), not silently downgraded to a response the client would believe is E2EE but isn't. (+ `rejects_client_facing_e2ee`.) Made `encryption_headers` `pub(crate)` to reuse the canonical key constants rather than hardcoding.

**Non-blocking — all addressed except the two explicitly deferred:**
- **Eviction path pinning** (the 5th site): the inference-url eviction now skips pinned models, so a Blocked DB model sharing a pinned name can't remove the attested provider (which the insert guards would then refuse to restore).
- `get_attestation_report` trims `e2e_pubkey` to match the data path.
- `pick_nonce` uses a wrapping **atomic counter** (infallible + collision-free under contention) instead of an RNG with a `[0]` fallback.
- `e2ee_stream` **rejects a second `e2e_init`** (a gateway-injected re-key would otherwise DoS all later frames via ML-KEM implicit rejection); EOF comment corrected re the route's gateway-minted `[DONE]` (emitted after the error frame, so still safe).

**Deferred (tracked, flag-off):** `chute_id` `OnceCell` 404-invalidation (self-heal on Chutes redeploy); the verified-instance/collateral TTL cache + staging TTFT measurement (the per-request DCAP/NRAS hot path).

60 chutes + 34 services attestation + pinned-pool tests green; fmt + clippy clean across the 4 crates. Re-requesting your review.

PierreLeGuen (changes requested):
This is a well-constructed PR overall: the trust chain is coherent (the Intel-signed quote seals `SHA256(nonce ‖ e2e_pubkey)`, so the gateway can't substitute keys and encapsulation to the attested key makes routing lies harmless), the measurement pin is fail-closed including empty-allowlist and malformed-golden cases, NEAR's verification path is preserved byte-for-byte through the `verify_tdx_quote` extraction, streaming is correctly double-gated off pending the frame-ordering fix, and privacy rules hold (errors carry IDs and failure categories only, the API key has a redacting Debug, internal marker keys are stripped from outbound bodies with a test locking that in). One issue blocks merge.

**Blocking**

- `crates/inference_providers/src/attested/chutes/client.rs:249` — `resp.bytes().await?` buffers the entire non-streaming response into memory *before* the `MAX_RESPONSE_BYTES` check on the next line. A chunked response without `Content-Length` can blow past the 64 MiB cap and exhaust memory before it's rejected, and this is the default Chutes path when the provider is enabled. Read `bytes_stream()` incrementally, track cumulative size, and bail before appending past the cap. Relatedly, `error_for_status` at `client.rs:900` reads the entire error body unbounded via `resp.text().await` and only truncates to 200 chars at Display time — same exposure to a hostile/misbehaving gateway, inconsistent with the success-path cap and the gunzip guard. Bound the read to a few KB before constructing `ChutesClientError::Status`.

**Non-blocking**

- `crates/services/src/attestation/verification.rs:225` + `verify_and_prepare` in `crates/inference_providers/src/attested/chutes/mod.rs` — every chat completion does discovery + `/evidence` + full DCAP + an NRAS round-trip before inference, and `verify_tdx_quote` rebuilds a `CollateralClient` and re-fetches Intel collateral (TCB info, QE identity, CRLs) each time. Per-request quote freshness is a deliberate choice, but the collateral isn't per-quote — a short-TTL collateral cache would cut latency and remove PCCS/NRAS as hard availability dependencies of the data path without weakening freshness. The client-reachable `get_attestation_report` endpoint triggers the same full chain per call; confirm platform rate limits cover that route before flag-on, and load-test before any prod enablement.
- `verify_and_prepare` candidate iteration (skip instance missing from `/evidence`, fall through to next candidate on failure, aggregate error when none verify, nonce-token selection) has no unit coverage — the 60 chutes tests cover crypto, parsing, and gating but not this orchestration. `Config::with_hosts` exists precisely for this; a mock-HTTP "first candidate fails, second succeeds" + "all fail → error mentions last failure" pair would lock in the fallback behavior cheaply.
- `crates/services/src/inference_provider_pool/mod.rs:612` — `register_pinned_provider` has no unpin counterpart: `unregister_provider` removes the mapping but leaves the model in `pinned_models`, so DB discovery can never re-register it until restart. Harmless today; a one-line removal in `unregister_provider` prevents a future foot-gun.
- `PreparedInvoke.measurement_config` (`#[allow(dead_code)]`) is documented as "for logging/annotation" but never logged, so operators can't tell which vetted config or instance served a request during an incident. Log `instance_id` + config label (IDs only, privacy-safe) on successful verification, or drop the field.

For the record, the documented residual risks (golden measurements snapshot an unsigned Chutes endpoint pending the signing ask; streaming's inner-`[DONE]` and frame-ordering assumptions are verify-on-staging items behind the double gate) look acceptable as scoped.

**Checks run locally:** `cargo fmt --all -- --check` clean; `cargo clippy -p inference_providers -p services -p api -p config` clean; `cargo check -p api/-p services/-p inference_providers` pass; `cargo test -p inference_providers attested::chutes` 60/60; `cargo test -p services attestation` 34 passed + 1 pre-existing ignored (network-dependent) + 2 integration tests; `cargo test -p services inference_provider_pool` 52/52 including the two new pinned-provider tests; `git diff --check` clean. CI (Lint, Test Suite, security_audit) is green. Skipped: e2e suite (requires PostgreSQL), vLLM integration tests, and live Chutes interop (the PR itself gates prod enablement on a staging round-trip).

Evrard-Nil:
Thanks @PierreLeGuen — round-6 in `93f45105` (also merged `main`, branch was behind).

**Blocking — fixed.** The non-streaming body is no longer buffered before the size check:
- `invoke_nonstream` now reads the body **incrementally** via `bytes_stream()` with a cumulative cap (`read_body_capped`), bailing *before* appending past `MAX_RESPONSE_BYTES` — so a chunked response without `Content-Length` can't exhaust memory.
- `error_for_status` no longer does an unbounded `resp.text()`; it reads at most `MAX_ERROR_BODY_BYTES` (16 KiB) for the diagnostic prefix — consistent with the success-path cap and the gunzip guard.

**Non-blocking — addressed:**
- `unregister_provider` now also clears `pinned_models` (the missing unpin counterpart), so DB discovery can re-register the name without a restart.
- Successful verification now logs `instance_id` + `measurement_config` + `gpu_verdict` (IDs only — privacy-safe) so an operator can trace which attested config/instance served a request; dropped the now-redundant `PreparedInvoke.measurement_config` field.

**Deferred (tracked, flag-off):**
- Verified-instance + DCAP-collateral TTL cache and staging TTFT measurement; confirming platform rate limits for the `/v1/attestation/report` route before flag-on.
- `chute_id` `OnceCell` 404-invalidation.
- The `verify_and_prepare` candidate-iteration unit test — agreed it's worth locking in; it needs a mock-HTTP dev-dep (none in the crate today), so I'll add `httpmock` + the "first candidate fails → second succeeds / all fail → error" pair as a small follow-up rather than pull a new dep into this PR.

60 chutes + 52 pool + 34 services attestation tests green; fmt + clippy clean across the 4 crates. Re-requesting your review.

PierreLeGuen (changes requested):
This adds Chutes as an attested provider behind a hard-off flag. The trust-chain design is consistently fail-closed, the E2EE protocol code has solid round-trip/tamper tests, the pinned-provider pool guards cover insert, eviction, and stale-removal, and the `verify_tdx_quote` extraction is behavior-preserving per the existing suite. One blocking issue and a couple of important ones below; the rest are pre-flag-on operational items.

**Blocking**

- `crates/services/src/completions/mod.rs:1191-1213, 1347-1369` — `resolve_and_get_model` rejects any model without an active row in the DB `models` table. A `CHUTES_MODELS` entry registered via `register_pinned_provider` has no DB row, so even with `ENABLE_CHUTES=true` and a valid key, every request fails before reaching the pool. The only mitigation is a comment in `env.example`, with no automated check — an easy ops trap. Either bypass/always-allow the DB gate for models in `pinned_models`, or auto-seed the catalog row at startup.

**Important**

- `crates/services/src/attestation/mod.rs:497` — Chutes providers return `supports_chat_signatures() == false`, so signature storage exits `Ok(())` without writing anything, and `/v1/signature/{chat_id}` later returns a plain 404. That makes "unsupported by design" indistinguishable from a missing or failed verification record for an attested model. Return an explicit unsupported result for chat IDs mapped to providers without signature support, or persist a capability marker so the API can report `SIGNATURE_UNSUPPORTED` instead of not-found.
- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:84-97, 187, 1553-1571` — two related gaps in inner-frame handling that must be resolved before `CHUTES_ENABLE_STREAMING` is turned on: (1) the stream only terminates cleanly on an authenticated inner `[DONE]`, while the plaintext outer `[DONE]` is (correctly) discarded as forgeable — if Chutes only emits the outer terminator, every streaming request ends as a truncation error, and the module note admits this is unconfirmed; (2) a decrypted frame that is an SSE comment/keepalive line (e.g. `: ping`, which vLLM backends do emit) is a fatal parse error that kills the stream — skipping lines starting with `:` is a one-line fix. Confirm the terminator behavior on staging and document the resolution in the module note before exposing streaming as attested.

**Minor / pre-flag-on operational**

- `crates/services/src/attestation/verification.rs:3971-3985` and `crates/services/src/attestation/chutes.rs:3718` — every chat completion runs discover + evidence + a fresh PCCS collateral fetch + an NRAS POST (per candidate instance tried). That's multi-second TTFT and hard availability coupling to two rate-limited external services. DCAP collateral is FMSPC-keyed and valid for hours; cache it, and load-measure on staging before prod flag-on.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`verify_and_prepare`) / `client.rs:672-674` — the single-use nonce token is obtained at discovery but `/e2e/invoke` happens only after the multi-second DCAP+NRAS verification; the parsed `nonce_expires_*` fields are never consulted and there's no retry on a rejected token. Check expiry before invoke and retry once with fresh discovery.
- `crates/inference_providers/src/attested/chutes/mod.rs:559-569` (`get_attestation_report`) — diverges from the data path twice: it picks the first E2E-capable instance and fails if that one is absent from `/evidence` or fails attestation (instead of iterating candidates like `verify_and_prepare`), and it calls `resolve_chute_id` directly instead of going through `chute_id_cache`. Reuse the data path's candidate-iteration pattern and the cache.
- `crates/inference_providers/src/attested/chutes/mod.rs` (`request_body`) — `obj.insert("stream_options", json!({"include_usage": true}))` clobbers any client-supplied `stream_options` arriving via `extra`; merge into the existing object instead.
- `crates/services/src/attestation/chutes.rs:210-218` — golden measurements are hardcoded; any Chutes VM image update breaks attestation until a code release. Add a config-driven override path before flag-on so operators can rotate pins without a deploy.
- `crates/inference_providers/Cargo.toml:559` — `ml-kem = "0.3.2"` is a caret requirement despite the comment saying "pinned"; use `=0.3.2` so a future `cargo update` can't float it silently.
- `crates/api/src/lib.rs:679` vs `pool.rs:498` — the Chutes verifier takes `pccs_url` from parsed config while `AttestationVerifier::from_env()` reads `PCCS_URL` from the environment independently; the two can diverge. Consolidate to one source of truth.

**Checks run locally** (CI Lint, Test Suite, and security_audit all green on head 93f4510):
- `cargo fmt --all -- --check`, `cargo check --workspace` — clean.
- `cargo clippy -p inference_providers -p services -p config -p api` — no warnings.
- `cargo test -p inference_providers --lib` — 295 passed (includes all new chutes report_data/measurements/e2ee/e2ee_stream/client/provider tests).
- `cargo test -p services --lib attestation` — 34 passed, 1 ignored (includes the new chutes fail-closed-ordering tests and existing NEAR verification tests).
- `cargo test -p services --lib inference_provider_pool` — 52 passed (includes the new pinned-provider survival/no-overwrite tests).
- Not run: DB-backed e2e suite (requires PostgreSQL) and live Chutes interop (network/staging credentials; the PR itself gates flag-on behind a staging round-trip).

Evrard-Nil:
Thanks Pierre — round-7 pushed in `0eed1def`. Walking your list:

**Blocking — catalog DB gate.** Fixed by auto-seeding, not bypassing: a synthesized `Model` can't be billed because `organization_usage_log.model_id` is a `FK → models(id)` (`V0013`), so the row must genuinely exist. On startup `init_inference_providers` now calls `ensure_chutes_catalog_row` per `CHUTES_MODELS` entry:
- Looks up the **unfiltered** row (`get_by_internal_name`, not the active-only variant) so a deliberately *disabled* row isn't silently re-activated/clobbered.
- **Missing →** INSERT via `upsert_model_pricing` with `attestation_supported/verifiable/is_ready=true`, `provider_type="chutes"`, `text` modalities, and **zero pricing** + a loud `warn!` to set real rates via `PATCH /v1/admin/models`.
- **Exists →** respected verbatim (warns if `is_active=false` or `attestation_supported=false`).
- A `provider_type="chutes"` row with NULL `inference_url` is invisible to both discovery reads (`get_inference_url_models` needs non-null URL; `get_external_models` needs `='external'`), so it's routed purely via the pinned pool — no clobber risk. `env.example` updated.

**Important — signature 404 ambiguity.** `get_chat_signature` now resolves the chat's provider and, if `!supports_chat_signatures()`, returns `SignatureLookupResult::Unavailable { error_code: "SIGNATURE_UNSUPPORTED", … }` (a 200, same shape as `STREAM_DISCONNECTED`) instead of falling through to 404. Symmetric with the store-path guard.

**Important — streaming inner frames.** (2) `inner_event` now skips decrypted SSE comment/keepalive lines (`: ping`) instead of fatally failing the JSON parse (+test). (1) The terminator question is unchanged behaviorally — only the authenticated inner `[DONE]` terminates, the forgeable outer one is ignored — and remains an explicit **verify-on-staging** item in the module note before `CHUTES_ENABLE_STREAMING` is exposed (still default-off).

**Minors:**
- `get_attestation_report` now reuses the memoized `cached_chute_id()` + the same candidate-iteration as `verify_and_prepare` (was: first E2E instance only + direct `resolve_chute_id`).
- `request_body` merges `include_usage` into client-supplied `stream_options` instead of clobbering (+test).
- `ml-kem = "=0.3.2"` (exact pin).
- PCCS: NEAR verifier now built from `config.pccs_url` via `AttestationVerifier::near_with_pccs` — one source of truth with the Chutes verifier; no independent `PCCS_URL` env read.

**Deferred (pre-flag-on, tracked) — explicitly *not* in this PR, gated behind the hard-off flag + the staging-verification gate that already blocks flag-on:**
- DCAP collateral cache (FMSPC-keyed TTL) + staging TTFT load-measurement.
- Nonce-token expiry check + retry-once-with-fresh-discovery.
- Config-driven golden-measurement override (rotate pins without a deploy).

I'd like to land the verified base + the blocking/important fixes here and do those three as focused follow-ups before we ever flip the flag — they each want their own test surface (and #1 wants a real staging load number). Shout if you'd rather see any folded in now. Re-requesting review.

CI: fmt/clippy clean on the 5 touched crates; `inference_providers` chutes 62 passed (+2), `services` attestation 34 (1 ignored), pool 52.

Evrard-Nil:
Follow-up `c4072852` — picked up the still-valid bot findings (the rest were stale: `nonce_token` is already `pick_nonce` not `[0]`; the `0x`-prefixed nonce already falls through to a fresh mint; `norm` already trims once):

- **byte-exact `raw_bytes` (Copilot):** non-stream `chat_completion` now hands back the decrypted plaintext verbatim as `raw_bytes` instead of re-serializing the parsed struct — preserves byte representation *and* any provider fields the typed struct doesn't model (e.g. `hidden_states`). `raw_bytes` is documented as the exact provider bytes, so this fixes a latent contradiction.
- **fail-closed RNG (Copilot):** `random_nonce()` returns `Result` (`E2eeError::Rng`) and is propagated with `?` in `build_request`, instead of `.expect()` on the request path.

**Kept as-is — invalid-nonce in `get_attestation_report` (Copilot):** I left the substitution rather than erroring, because (a) the response map echoes the actual `nonce` used, so it isn't truly silent, and (b) the report route maps *every* error from this path to `503`, so erroring on a malformed caller nonce would surface as a worse "service unavailable" instead of a usable, self-describing report.

PierreLeGuen (changes requested):
This is a well-engineered PR: fail-closed design throughout, strong unit-test coverage (E2EE round-trips, tamper tests, fail-closed ordering, pool pinning), and honest documentation of trust limitations. The hard-off `ENABLE_CHUTES` flag makes it zero-behavior-change by default, and the NEAR attestation refactor (extracted `verify_tdx_quote`) is a pure extraction with all existing tests passing. I verified key invariants locally: `ChatCompletionResponse`/`Choice` carry `#[serde(flatten)]` catch-alls so the decrypt-reparse path drops no provider fields, the `INTERNAL_KEYS` strip list covers everything the completions service injects, the refresh task skips its first tick (no startup race with pinned registration), and the GPU nonce round-trip through `nvidia_payload` serialization is correct.

Two issues should be addressed before merge; the rest are non-blocking.

**Blocking**

- `crates/inference_providers/src/attested/chutes/client.rs:852,866` — the client-wide `read_timeout(READ_TIMEOUT_SECS = 90)` bounds the wait for **response headers**, not just gaps between body reads (verified in the vendored reqwest 0.12.28 source: `PendingRequest::poll` polls `read_timeout_fut` while awaiting the head). For the non-streaming `/e2e/invoke` — the only attested path enabled by default — the instance sends nothing until generation completes, so any completion taking >90s times out even though `request_timeout_secs` defaults to 300 (and the doc comment on `request_timeout_secs` claims the opposite). Long reasoning completions routinely exceed 90s. Fix: drop the client-wide `read_timeout` and use a separate client or per-request override so the non-stream invoke is bounded only by `request_timeout_secs`, keeping the 90s inactivity bound for streaming where it's correct.
- `crates/api/src/lib.rs:459-490` (`ensure_chutes_catalog_row`) — the auto-seeded catalog row is `is_active=true, is_ready=Some(true)` with pricing unset, which `upsert_model_pricing` inserts as zero-cost. The moment the flag is on, paid Chutes traffic is served and billed at $0, guarded only by a startup log warning. Seed `is_ready=false` (admin explicitly readies the model after setting pricing) so the unpriced window is impossible rather than merely warned about, or require a preconfigured priced row before registering the provider.

**Operational — needs an answer before flag-on, not necessarily before merge**

- `crates/inference_providers/src/attested/chutes/mod.rs:2581` (`verify_and_prepare`) — every chat completion does `/e2e/instances` + `/evidence` + an Intel PCCS/DCAP collateral fetch + an NVIDIA NRAS round trip before `/e2e/invoke`, with no verdict caching. That's seconds of added latency per request and a hard per-request availability dependency on PCCS and NRAS (an outage of either correctly fails closed — to 0% traffic). Before production flag-on, consider a short-TTL verified-instance cache keyed by `(instance_id, e2e_pubkey)`, or at minimum latency/failure metrics and alerting on those calls.

**Non-blocking**

- `crates/inference_providers/src/attested/chutes/client.rs:789-791` and `mod.rs:321` — nonce-token lifecycle has two gaps: `nonce_expires_in`/`nonce_expires_at` are deserialized but never consulted (the token is picked at discovery and used only after the multi-second DCAP+NRAS window, so an expired token fails with no retry), and `pick_nonce`'s wrapping counter distributes but does not prevent collisions, so if Chutes strictly enforces single-use tokens, >N concurrent requests to one instance guarantee replayed tokens. Check expiry (or re-discover immediately before invoke) and document/handle the token-consumption semantics expected from Chutes.
- `crates/services/src/attestation/mod.rs:4059-4070` — the `SIGNATURE_UNSUPPORTED` branch keys off the in-memory `chat_id_mapping`, which is lost on restart; after a redeploy, signature lookups for Chutes chats fall back to the bare 404 this branch exists to disambiguate. Consider also keying off catalog metadata (e.g. `provider_type == "chutes"`).
- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:1787-1799` — the outer SSE parser treats each `data:` line as a complete event; per the SSE spec a single event may span multiple `data:` lines to be concatenated. Likely fine for Chutes' framing, but worth confirming on the planned staging round-trip alongside the inner-`[DONE]` question.
- `crates/inference_providers/src/attested/chutes/client.rs:1020-1034` — body-cap violations surface as `ChutesClientError::Status { status: 0 }`, conflating a synthetic value with a real HTTP status; a dedicated `BodyTooLarge` variant would be cleaner.
- `crates/services/src/responses/models.rs:460`, `crates/api/src/models.rs:1447` — commit `684694f` changes `ResponseStatus`/`ResponseItemStatus` serialization from `"inprogress"` to `"in_progress"` (deserialization stays backward-compatible via `#[serde(alias)]`). It references #750, so it appears deliberately carried in, but it changes the wire format for every in-flight Response — call it out in release notes, or land it separately.

**Checks run locally:** `cargo fmt --all -- --check` clean; `cargo clippy` across `inference_providers`, `services`, `api`, `config`, `database` with no warnings; `cargo test -p inference_providers --lib` 297 passed (including all 62 new Chutes tests); `cargo test -p services --lib` attestation (34 passed, 1 pre-existing ignore) and `inference_provider_pool` (52 passed, including the new pinned-provider tests); `cargo test -p api test_response_status` passed. E2E tests skipped (require PostgreSQL, unavailable here). CI: Lint and security_audit green; Test Suite still pending at review time. The reqwest `read_timeout` semantics in the first finding were verified directly against the vendored crate source.

Evrard-Nil:
Thanks Pierre — round-8 in `3ec3b38e`.

**Blocking — `read_timeout` caps the header wait.** Confirmed, and good catch on the vendored-reqwest verification. Split into two clients:
- `http` (non-stream): `connect_timeout` only, **no** `read_timeout`. Discovery / evidence / non-stream `/e2e/invoke` are bounded solely by the per-request total `.timeout(request_timeout_secs)` (300s default) — so a long reasoning completion that sends nothing until it finishes is no longer killed at 90s.
- `stream_http` (streaming only): `connect_timeout` + `read_timeout(90s)` — the inactivity clock resets per SSE frame, which is the case where that bound is correct. `invoke_request` selects the client by `InvokeMode`. Fixed the two contradictory doc comments that claimed `read_timeout` wasn't client-wide.

**Blocking — $0 billing window.** Fixed, with one correction worth flagging: **`is_ready` does not gate serving** — I traced it and `resolve_and_get_model`'s only row gate is `WHERE is_active = true`; `is_ready` is selected but never in any WHERE/if (it's display metadata, `ports.rs:74` "does not affect Cloud API's own listing"). So seeding `is_ready=false` would *not* have closed the window. The field that actually gates is `is_active`, so `ensure_chutes_catalog_row` now seeds **`is_active=false`** (+ `is_ready=false` for honest display). The model can't be served — and therefore can't be billed at the $0 default — until an operator sets real per-token rates **and** flips `is_active=true` in one admin PATCH. The provider is still registered (pinned), so activation is live with no restart.
- One honest caveat: this closes the *auto-seed* window completely, but `is_active`/pricing aren't coupled in the admin path, so an operator could still *manually* activate at $0. Fully guaranteeing that would mean rejecting unpriced models in the shared completions/admin gate — which changes behavior for **all** models incl. NEAR's (some may be intentionally $0), so it's out of scope here. Happy to do it as a separate, explicitly-scoped PR if you want the hard guarantee.

**Operational (pre-flag-on) — per-request DCAP+NRAS, no cache.** Agreed; this is the top item on the pre-flag-on list (a short-TTL verified-instance cache keyed by `(instance_id, e2e_pubkey)` + latency/failure metrics on the PCCS/NRAS calls). It wants a real staging latency number to size the TTL, so I'd like to land it as the first follow-up rather than guess here. Tracked.

**Non-blocking:**
- `Status { status: 0 }` → new dedicated `ChutesClientError::BodyTooLarge { max_bytes, what }`. Done.
- **SIGNATURE_UNSUPPORTED after restart:** deferring with a caveat — `get_chat_signature` only has `chat_id`, and the only `chat_id → provider/model` link is the same in-memory map that's lost on restart, so keying off catalog `provider_type` isn't reachable from `chat_id` alone without persisting a `chat_id → model` record (a larger change). Both states already mean "no signature exists"; the regression is only 200-`SIGNATURE_UNSUPPORTED` vs 404. Tracked as a follow-up alongside the cache work.
- **Nonce expiry / single-use semantics:** tracked in the pre-flag-on list (check `nonce_expires_*` and re-discover-before-invoke, or retry-once on a 403). The `pick_nonce` counter only spreads load; I'll document the consumption contract once confirmed against Chutes on staging.
- **SSE multi-`data:` concatenation:** deferring — Chutes emits exactly one encrypted frame per `data:` line today, so the per-line model is correct; I'll confirm it on the same staging round-trip as the inner-`[DONE]` terminator question and add concatenation if their framing ever changes.
- **`in_progress` serialization (#750):** that came in via the `main` merge (commit `684694f`), not this PR's diff, and it keeps `#[serde(alias = "inprogress")]` for read back-compat. Flagging here for release notes; not re-touching it in this PR.

Re-requesting. CI green locally (fmt/clippy across the 5 crates; chutes 62, services attestation 34/1-ignored, pool 52).

Evrard-Nil:
Filed the residual $0-billing-guarantee gap (activation not coupled to pricing) as nearai/cloud-api#766 for a separately-scoped follow-up — keeps this PR's seed-inactive fix focused while tracking the full invariant (reject activating an unpriced model unless an explicit free-tier flag is set, so NEAR's intentionally-free models aren't broken).

PierreLeGuen (changes requested):
This is a strong PR overall — the E2EE protocol (ML-KEM-768 + HKDF + ChaCha20-Poly1305) round-trips with tamper/wrong-session coverage, the verifier is fail-closed at every binding (TCB floor → TD10 → debug bit, GPU nonce sealed via `report_data[0:32]` and re-checked by NRAS), privacy handling is clean (API key redacted, internal markers stripped, no plaintext in error paths), and the whole feature is hard-off behind `ENABLE_CHUTES`. Two issues in the admin/catalog wiring should be fixed before merge, though.

**Blocking**

- `crates/api/src/routes/admin.rs:455` — the admin PATCH handler unregisters a provider whenever `provider_type` is present in the payload, but the re-registration logic below only covers inference-url and external providers. A PATCH that includes `provider_type: "chutes"` (e.g. an activation/update) removes the pinned Chutes provider and never registers it back, leaving an active catalog row with no serving provider until restart. Either re-register pinned providers from config, or only unregister when the provider type actually changes to a DB-discovered type. Please add a regression test for a PATCH with `provider_type="chutes"`.
- `crates/api/src/lib.rs:689` — the catalog seed is documented as insert-only but goes through `upsert_model_pricing`, whose conflict path updates existing rows. The cold-insert path is fine (`is_active.unwrap_or(true)` correctly yields `Some(false)`, so the seeded row can't serve or bill), but if an operator creates/activates the model concurrently with startup, the conflict path can flip `is_active` back to false and clobber pricing with defaults. Use `INSERT ... ON CONFLICT DO NOTHING` (then re-read) or a no-clobber repository method, with a race regression test.

**Non-blocking observations**

- `crates/inference_providers/src/attested/chutes/mod.rs:2626` — `verify_and_prepare` always picks the first instance that verifies, in `/e2e/instances` order, so all traffic from one gateway process hot-spots a single Chutes instance (and `X-Instance-Id` pinning prevents Chutes-side rebalancing). Rotating the starting index (e.g. reusing the `pick_nonce` counter) would spread load while still verifying every candidate.
- Same function — every chat request performs discovery + `/evidence` + DCAP collateral (PCCS) + an NRAS round trip, coupling availability to PCCS/NRAS uptime and adding multi-second latency. This is the stated per-request-trust design and is flag-gated, but a short-TTL verified-instance cache is worth a deliberate decision before production flag-on.
- `crates/services/src/attestation/chutes.rs:4007` — golden measurements (including runtime RTMR3) are pinned in code, so any Chutes image rollout hard-fails all Chutes requests until a cloud-api deploy. Fail-closed is right; consider making the vetted snapshot config-loadable before flag-on (the comment already acknowledges this).
- `crates/services/src/inference_provider_pool/mod.rs:4357` — the pinned set is cloned before taking the `provider_mappings` write lock, so a racing `register_pinned_provider` could in principle be clobbered by a same-named DB model. Unreachable today given init ordering and the refresh task's first-tick skip, but re-reading the pinned set under the lock would close it for free.
- `crates/api/src/lib.rs:531-577` — if `ENABLE_CHUTES` is later disabled while the catalog row stays `is_active=true`, requests surface as per-request provider errors rather than clean 404s. A startup warning (flag off + active `provider_type='chutes'` row) or an `env.example` note would help operators.

**Checks run**

- PR diff and CI inspected via `gh` — Lint, Test Suite, and security audit green on the head commit; merge state `BEHIND`.
- `cargo fmt --all -- --check`, `cargo clippy` (all targets, affected crates), `cargo check -p api`, `git diff --check` — all clean.
- `cargo test -p inference_providers --lib` — 297 passed (includes the new chutes e2ee/stream/client/measurement/report_data/provider tests).
- `cargo test -p services --lib` for `attestation` (34 passed, 1 pre-existing ignore), `inference_provider_pool` (52 passed), and `pinned_provider` (2 passed: survives refresh, not overwritten by discovery).
- E2E tests skipped locally (require PostgreSQL); they run in the PR's CI, which passed. Live Chutes interop not testable locally; the staging round-trip gate before prod flag-on is an appropriate mitigation.

Evrard-Nil:
Thanks Pierre — round-9 in `464e3cb9`.

**Blocking — admin PATCH unregisters the pinned provider.** Good catch, and it's exactly the path my round-8 fix forces operators down (activate via PATCH). Added `InferenceProviderPool::is_pinned(model)` and the admin unregister loop now **skips pinned models** — they're config-managed and aren't restored by the inference-url/external re-registration. Pricing/metadata still apply (the `batch_upsert_models` above runs regardless); serving stays gated by the catalog `is_active`, and an active pinned model keeps its provider. Regression tests:
- pool unit test `is_pinned_reports_pinned_models_and_guards_admin_unregister` (no DB);
- e2e `admin_patch_with_provider_type_keeps_pinned_provider` — registers a pinned provider, PATCHes with `providerType: "chutes"`, asserts the provider survives.

**Blocking — seed upsert can clobber on a race.** Fixed: added `ModelRepository::seed_model_if_absent` (`INSERT ... ON CONFLICT (model_name) DO NOTHING`, returns `Ok(None)` on conflict) and `ensure_chutes_catalog_row` now uses it instead of `upsert_model_pricing`. A concurrent operator create/activate wins and is left completely untouched. Regression test e2e `seed_model_if_absent_does_not_clobber_existing` — seed (inactive) → operator activates+prices → re-seed is a no-op and `is_active`/pricing are preserved.

**Non-blocking — addressed now:**
- `verify_and_prepare` rotates the starting candidate instance via a wrapping counter, so one process no longer hot-spots the first-listed instance (every candidate is still attempted, just rotated).
- Startup `warn!` when `ENABLE_CHUTES` is off but `CHUTES_MODELS` is still set (leftover active rows would surface provider errors) — covers the flag-disabled case you flagged.

**Non-blocking — deferred (tracked):**
- **Re-read the pinned set under the `provider_mappings` lock** — agreed it closes the (per your note, currently-unreachable) window. It's the same clone-before-lock pattern in three refresh/eviction spots; I'd rather not reorder lock acquisition across the hot refresh path in the same PR as the merge-blocking fixes. I confirmed the reorder is deadlock-safe (the `std::sync` pinned guards are released within their statement, never held across the `provider_mappings` await), so it's a clean follow-up.
- **DCAP/NRAS verified-instance cache** and **config-loadable golden measurements** — both pre-flag-on; the cache wants a real staging latency number to size the TTL. These are the top items on the pre-flag-on list (and #766 already tracks the related pricing-coupling guarantee).

CI green locally (fmt/clippy across the 5 crates; pool 53, chutes 62, database 13; the e2e binary compiles — the two DB-backed regression tests run in CI's Postgres). Re-requesting.

PierreLeGuen (approved):

Issue #650, opened 2026-05-22
OpenAPI spec documents endpoints that 404: /v1/model/{model_name}; two parallel list endpoints


Evrard-Nil:
## Summary

The publicly-served OpenAPI spec at `https://cloud-api.near.ai/api-docs/openapi.json` documents endpoints that don't actually exist. A client generated from the spec hits bare 404.

### `GET /v1/model/{model_name}` — documented but returns 404 size=0

```bash
curl -s -w "HTTP %{http_code} size=%{size_download}\n" \
  -H "Authorization: Bearer $KEY" \
  "https://cloud-api.near.ai/v1/model/openai/gpt-oss-120b"
# HTTP 404 size=0
```

Even though the spec lists this path with a documented schema (in `paths.[/v1/model/{model_name}].get`), the actual gateway returns a bare 404 — same family as #614.

### Two parallel model list endpoints

- `GET /v1/models` — works, OpenAI-compat shape (`{object: "list", data: [...]}`)
- `GET /v1/model/list` — works, custom shape (`{models, limit, offset, total}`)

Both are documented in the spec. The custom endpoint supports pagination; the OpenAI-compat one doesn't (silently ignores `?limit=`). Two parallel endpoints with different shapes is confusing — pick one.

### Inconsistent /v1 prefix

The spec documents:
- `POST /conversations/{conversation_id}/archive` (no `/v1` prefix)
- `GET /v1/conversations/{conversation_id}` (has `/v1` prefix)

Five `/conversations/...` paths are documented without the `/v1` prefix; the rest have it. Probably a routing bug that landed in spec generation. (Didn't actually hit these — couldn't auth as a real user.)

### Health version vs spec version disagree

`/v1/health` reports `"version":"0.0.0"` (already filed in #647), but the OpenAPI spec at `/api-docs/openapi.json` declares `info.version: "1.0.0"`. Both should be the same source of truth.

### Same raw-422 envelope-bypass as #609

`POST /v1/score` with missing `model` field returns:
```
HTTP 422
Failed to deserialize the JSON body into the target type: missing field `model` at line 1 column 13
```
Plain text, not the standard JSON envelope. Same pattern as `/v1/rerank` in #609 and `frequency_penalty="oops"` noted in #622.

## Suggested fix

1. Either implement `GET /v1/model/{model_name}` or remove it from the spec.
[...]

PR #760, opened 2026-06-10
Fix OpenAPI conversation action paths

lloydmak99:
## Problem

Part of #650.

The generated OpenAPI spec documented five conversation action operations without the `/v1` prefix:

- `POST /conversations/{conversation_id}/archive`
- `DELETE /conversations/{conversation_id}/archive`
- `POST /conversations/{conversation_id}/pin`
- `DELETE /conversations/{conversation_id}/pin`
- `POST /conversations/{conversation_id}/clone`

Those unprefixed paths return bare 404s at runtime. The actual router mounts the handlers under `/v1`, so generated clients call the wrong URLs for archive, unarchive, pin, unpin, and clone.

## Change

- Updated the `utoipa::path` annotations for the conversation action handlers to document the runtime `/v1/conversations/...` paths.
- Added an OpenAPI regression test that asserts:
  - the prefixed action paths are present in the generated spec
  - the unprefixed variants are absent

## What Else Is Needed

- CI and reviewer validation.
- #650 also tracks broader cleanup around duplicate model list surfaces and raw/plain-text 422 envelopes. Those are intentionally left for separate PRs because this change is scoped to the confirmed spec/runtime path mismatch.

## Tests

- `cargo test -p api --lib openapi`
- `cargo test --lib --bins`


Evrard-Nil (approved):
Reviewed thoroughly with a local build + spec dump. **LGTM — approving.**

**Verified the core claim is correct:**
- Action handlers are registered *without* `/v1` in `build_conversation_routes()` (`crates/api/src/lib.rs:1230-1239`: `.route("/conversations/{conversation_id}/pin", post(pin).delete(unpin))`, `.../archive`, `.../clone`), and that router is merged under `.nest("/v1", …)` at `lib.rs:930`. So the true runtime path is `/v1/conversations/{conversation_id}/{archive,clone,pin}` — exactly the new annotation strings (param name `{conversation_id}` matches too). The pre-fix bare `/conversations/...` annotations did 404 generated clients. ✅

**No double-prefix:** `ApiDoc` (`openapi.rs`) has no servers/nest/path-prefix and its only modifier (`SecurityAddon`) never touches `paths`. Dumped the real `ApiDoc::openapi()` spec — renders single `/v1/conversations/...`, no `/v1/v1`. ✅

**Consistency:** all 12 utoipa annotations in `conversations.rs` now use `/v1/`; the other 7 already did, these 5 ops (3 unique path keys) were the last stragglers. Grepped every route file for `path = "/` excluding `/v1` → zero matches org-wide. Nothing missed. ✅

**Build/test:** `cargo test -p api --lib openapi` → 3 passed (incl. the new `test_openapi_conversation_action_paths_use_v1_prefix`, which builds the real spec and asserts prefixed-present + bare-absent). `cargo fmt --all -- --check` clean. `cargo clippy -p api --all-targets -- -D warnings` clean. ✅

**Minor (non-blocking, out of scope):** the regression test keys on path strings, not HTTP methods — archive+pin each carry POST+DELETE under one key, so all 5 ops are covered, but a hypothetical regression that kept the path key while dropping the DELETE op would slip through. Extremely low risk; fine to leave as-is.

PierreLeGuen (approved):
This PR correctly fixes the five `utoipa::path` annotations in `crates/api/src/routes/conversations.rs` (pin, unpin, archive, unarchive, clone) that were missing the `/v1` prefix, and adds a regression test.

**Correctness verified:**
- The router registers these handlers at relative paths like `/conversations/{conversation_id}/pin` in `build_conversation_routes` (`crates/api/src/lib.rs:1229-1240`), nested under `/v1` (`crates/api/src/lib.rs:931`), so the runtime paths are indeed `/v1/conversations/...` and the previous spec paths were wrong.
- The OpenAPI doc sets no `servers` base URL or path-modifying modifier (`crates/api/src/openapi.rs:303`; `SecurityAddon` only adds security schemes), so annotations must carry the full `/v1` prefix — no `/v1/v1` double-prefix risk.
- All 12 `utoipa` path annotations in `conversations.rs` now consistently use `/v1/`, and a repo-wide grep confirms no remaining unprefixed `path = "..."` annotations in the api crate — the fix is complete, not partial.
- The new test asserts both the presence of the `/v1`-prefixed path keys and the absence of the unprefixed variants, guarding against regression. Pin/archive each share one path key for POST and DELETE, which the test comment correctly notes.

Scope exclusions (duplicate model list surfaces, 422 envelopes from #650) are documented in the PR body as follow-ups, which is reasonable. No privacy/logging concerns — the change touches only doc annotations and a test.

**Local checks run:**
- `cargo test -p api --lib openapi` — 3 passed, including the new `test_openapi_conversation_action_paths_use_v1_prefix`
- `cargo test --lib --bins` — 742 passed, 1 ignored (pre-existing), 0 failed
- `cargo clippy -p api --all-targets -- -D warnings` — clean
- `git diff --check` — clean
- CI on the PR (Lint, Test Suite, security_audit) is green; integration/vLLM paths weren't run locally as they need services/secrets.

Issue #609, opened 2026-05-19
Rerank: top_n and return_documents silently ignored; multi_modal field leak; missing-query returns raw 422


Evrard-Nil:
## Summary

`POST /v1/rerank` against `Qwen/Qwen3-Reranker-0.6B` works for the happy path but several documented Cohere/Jina-style parameters are silently dropped, and one validation path leaks a raw serde error.

| Bug | Severity |
|---|---|
| `top_n` silently ignored — returns all results regardless | high (changes API contract) |
| `return_documents: false` silently ignored — `document` text still in every result | medium (extra bandwidth + redundant payload) |
| Each result item has `document.multi_modal: null` always-present | low (dead field leak, same pattern as #607's `file_path`) |
| Missing required field (e.g. `query`) returns `422` with raw serde message `Failed to deserialize the JSON body into the target type: missing field query at line 1 column 58` — no JSON error envelope | medium (clients can't parse this consistently) |

## Repro

```bash
KEY=sk-32c0476395fd40c795725fc101f33304

# top_n=2 ignored — gets 3 results back
curl -s https://cloud-api.near.ai/v1/rerank \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-Reranker-0.6B","query":"capital of France","documents":["Paris","Berlin","Tokyo"],"top_n":2}' \
  | jq '.results | length'   # → 3

# return_documents=false ignored — document.text still present
curl -s https://cloud-api.near.ai/v1/rerank \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-Reranker-0.6B","query":"q","documents":["a","b"],"return_documents":false}' \
  | jq '.results[0]'
# {"index":..., "relevance_score":..., "document":{"multi_modal":null,"text":"a"}}

# Missing query → raw plain-text 422
curl -s -w "HTTP %{http_code}\n" https://cloud-api.near.ai/v1/rerank \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3-Reranker-0.6B","documents":["a","b"]}'
# Failed to deserialize the JSON body into the target type: missing field `query` at line 1 column 58
# HTTP 422

# Empty docs → clean 400 with envelope ✓ (works correctly)
[...]

PR #761, opened 2026-06-10
Honor rerank response options

lloydmak99:
## Problem

Part of #609.

`POST /v1/rerank` accepted Cohere/Jina-style response options but ignored them:

- `top_n` was not deserialized, so clients always received every provider result.
- `return_documents: false` was not deserialized, so document text was still returned.
- provider document objects could include `multi_modal: null`, a dead field that leaks into the public response.

This made the endpoint diverge from the documented rerank surface and increased response payloads for callers that explicitly asked to omit documents.

## Change

- Added `top_n` and `return_documents` to `RerankRequest`.
- Validates `top_n=0` as `400 invalid_request_error`.
- Applies rerank response options before returning the response:
  - normalize results to `relevance_score` descending order
  - truncate to `top_n` when provided
  - omit `document` when `return_documents` is `false`
  - strip `multi_modal: null` from document objects when documents are returned
- Added focused unit tests for sorting/truncation, document omission, `multi_modal` stripping, and `top_n=0` validation.

## What Else Is Needed

- CI and reviewer validation.
- This intentionally normalizes result order by relevance score before truncation; callers that consumed raw provider order should treat the response as ranked order.
- #609 also tracks raw plain-text 422 deserialize errors for `/v1/rerank`, plus the issue thread lists the same extractor-envelope bypass on other endpoints. That should be handled in a separate extractor-level cleanup so the fix is consistent across body and query deserialization failures.

## Tests

- `cargo test -p api --lib rerank`
- `cargo test --lib --bins`


Evrard-Nil (approved):
Reviewed locally (fresh clone, full build + tests). **LGTM — approving.**

**Verified the substantive concerns:**

1. **`validate()` is live, not dead code.** The `rerank` handler calls `request.validate()` (`completions.rs:4247`) before `try_rerank`, mapping `Err` → `400 invalid_request_error` (4248-4256). So `top_n=0 → "top_n must be at least 1"` (`models.rs:653-655`) is reachable. ✅

2. **No billing/revenue regression.** `apply_rerank_response_options(&request, &mut response)` (4317) mutates *only* `response.results` (sort/truncate/`document=None`/strip `multi_modal`) and never touches `response.usage`. Billing reads `response.usage.total_tokens` (4323-4327) → `input_tokens: token_count, output_tokens: 0`, fully independent of the truncated list. And `top_n`/`return_documents` are **not** forwarded to the backend (`RerankParams`, 4301-4306), so the provider always reranks+bills all documents — truncation is purely presentation-side. ✅

3. **Unconditional re-sort — benign, one note.** `apply_rerank_response_options` sorts by `relevance_score` desc on *every* call, including the default `(None, None)` path that previously passed through untouched. No provider in `inference_providers` sorts (nearai does `response.json()` verbatim; external delegates), but Cohere/Jina-style rerank backends conventionally return results already sorted desc, so this is a harmless no-op that also hardens against a non-conforming backend. `index` (i32, original-input index) is preserved through sort/truncate, so the back-reference stays correct (test asserts `[0,2]` for top_n=2 over scores 0.9@0/0.4@2/0.1@1 — correct). Worth a one-line note in the PR body if you want the ordering change explicit, but not blocking.

4. `partial_cmp(...).unwrap_or(Equal)` is NaN-panic-safe (sort_by is stable). `multi_modal` stripping is correctly narrow (only `Value::Object` + present + null; non-null kept; string/array docs untouched). `RerankRequest` derives `ToSchema` so the new fields auto-appear in the spec (doc comments present). Minor: `skip_serializing_if` is inert since the struct is `Deserialize`-only — harmless and consistent with the rest of the struct.

**Build/test (isolated target):** `cargo test -p api --lib rerank` → 4/4 pass; `cargo test -p api --lib` → 139 passed, 0 failed, 2 ignored (incl. `test_openapi_spec_generation`); `cargo fmt --all -- --check` clean; `cargo clippy -p api --all-targets -- -D warnings` clean.

Edge cases (`top_n > len` truncate no-op, empty results, string/array documents, non-null multi_modal) are all handled safely though not all directly tested — fine to leave.

Evrard-Nil:
Re-reviewed the new commit (`faebb9f`): `partial_cmp(...).unwrap_or(Equal)` → `total_cmp`. Strict improvement — proper total ordering on f64 with deterministic NaN handling, drops the `unwrap_or` fallback. Rebuilt locally, `cargo test -p api --lib rerank` still 4/4 green. Approval stands.

PierreLeGuen (commented):
This is a well-scoped fix: `POST /v1/rerank` now deserializes `top_n` and `return_documents`, rejects `top_n=0`, and applies the options to the provider response (sort by `relevance_score` descending, truncate, drop `document`, strip `multi_modal: null`) before serializing it to the client. The mutated `inference_providers::RerankResponse` is what the handler serializes directly, so the transformation reaches the actual payload, and `apply_rerank_response_options` runs before usage tracking reads `response.usage` (which it never touches), so truncation cannot under-bill. Using `total_cmp` avoids the NaN panic that `partial_cmp().unwrap()` would risk. The unconditional re-sort is a behavior change for clients relying on raw provider order, but it matches the documented rerank contract, existing tests already assert descending order, and the PR description calls it out as intentional.

A few non-blocking items:

- `crates/api/src/models.rs:626-631` — the new `top_n`/`return_documents` fields carry `#[serde(skip_serializing_if = "Option::is_none")]`, but `RerankRequest` only derives `Deserialize`, so the attributes are dead and misleading. Remove them.
- `crates/api/src/routes/completions.rs:4114-4204` / `crates/api/tests/e2e_all/rerank.rs` — the new unit tests construct `RerankRequest` directly, so nothing exercises JSON deserialization of the new field names or the end-to-end handler path; a field-name typo or serde rename would pass all current tests. Consider an e2e case posting `{"top_n": 2, "return_documents": false}` and asserting result count and absent `document` fields.
- `crates/api/src/routes/completions.rs:4300-4305` (observation) — `top_n`/`return_documents` are not forwarded to the provider, so providers always return full documents and trimming happens locally. Functionally correct and billing-neutral, but it wastes provider→API bandwidth on large requests; passing the options through `RerankParams` could be a follow-up if providers support them.

Checks run locally: `cargo test -p api --lib rerank` (4/4 passed), `cargo fmt --all -- --check` (clean), `cargo clippy -p api --all-features -- -D warnings` (clean). CI is green (Lint, Test Suite, security_audit). The full e2e suite was not run locally (requires a running PostgreSQL instance; one attempt was also blocked by a Cargo build lock) — it's covered by the passing CI Test Suite job.

Issue #614, opened 2026-05-20
Legacy /v1/completions: bare 404 with empty body — return JSON error explaining it's unsupported


Evrard-Nil:
## Summary

`POST /v1/completions` returns `HTTP 404` with an empty body. Many older OpenAI-compatible clients default to this endpoint, so users hit it without warning and get nothing actionable.

```bash
KEY=sk-32c0476395fd40c795725fc101f33304
curl -s -w "HTTP %{http_code} size=%{size_download}\n" https://cloud-api.near.ai/v1/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"openai/gpt-oss-120b","prompt":"Hi","max_tokens":10}'
# HTTP 404 size=0
```

The 404 is identical to a request to a truly nonexistent path (`/v1/this-definitely-does-not-exist` → HTTP 404 size=0). A user can't tell whether they got the URL wrong, are on the wrong API version, or hit a not-yet-implemented endpoint.

## Context

Issue #41 (closed 2025-10-21) confirmed `/v1/completions` is intentionally not implemented; a code comment at `crates/api/src/routes/completions.rs:286` says so. But the resolution there was just closing the issue — the user-facing surface still 404s silently.

## Suggested fix

Either:
1. **Implement a minimal shim**: translate `{prompt: "..."}` into `messages: [{"role":"user","content":"..."}]` and call the existing chat-completions handler. ~20 lines of glue.
2. **Or return a deliberate error envelope**: register the route to respond with `404` (or `410 Gone`) and the standard error JSON:
   ```json
   {
     "error": {
       "message": "/v1/completions is not supported. Use /v1/chat/completions.",
       "type": "endpoint_not_supported",
       "param": null,
       "code": null
     }
   }
   ```
   Same status code, but the client now sees what's wrong.

Option 1 unlocks legacy clients; option 2 is the minimum-viable improvement.

---
Found during routine cloud-api smoke test. Distinct from #41 (which was a 503 from Heroku, since-migrated infra) — that one closed without resolving the empty-body UX.

Evrard-Nil:
Same bare-404 pattern on **`/v1/images/variations`**:

```bash
curl -s -w "HTTP %{http_code} size=%{size_download}\n" https://cloud-api.near.ai/v1/images/variations \
  -H "Authorization: Bearer $KEY" \
  -F model=black-forest-labs/FLUX.2-klein-4B \
[...]

PR #762, opened 2026-06-10
Return JSON errors for unsupported OpenAI endpoints

lloydmak99:
Part of #614

## Problem

Several recognized OpenAI-compatible endpoints currently fall through to Axum's default empty 404/405 response. That makes them indistinguishable from typoed paths and gives OpenAI SDK clients no structured error to inspect.

Issue #614 tracks this broader bare-error surface, including missing routes such as image variations, audio translations, moderations, batches, threads, assistants, list responses, and model retrieve/delete style paths.

## Change

- Adds an explicit unsupported OpenAI-compatible route group for the known missing endpoint surfaces, including wildcard subpaths for batches, threads, assistants, and slash-containing model IDs.
- Returns the standard JSON error envelope with `type: "not_implemented"` and HTTP 501 instead of an empty 404/405.
- Mounts the unsupported routes behind the normal API-key auth and rate-limit middleware.
- Uses `OriginalUri` so error messages preserve the full `/v1/...` path seen by clients.
- Adds regression coverage that each supported placeholder route returns a JSON envelope, wildcard subpaths are caught, and the routes can be merged with existing method-specific routes like `POST /v1/responses` and `GET /v1/models`.

## What Else Is Needed

- This does not implement the endpoints themselves; it only makes unsupported known endpoints fail with a structured response.
- The nginx 413 HTML response for oversized uploads mentioned in #614 is outside this app-router change and needs ingress/nginx configuration work.
- `POST /v1/completions` is not changed here because the endpoint already exists on current `main`.

## Tests

- `cargo test -p api --lib unsupported`
- `cargo test --lib --bins`


Evrard-Nil (approved):
Reviewed locally (fresh clone, full build + router-merge verification). **LGTM — approving**, with one minor cosmetic suggestion below.

I focused the paranoia on the two router-level risks that the committed tests do **not** actually cover:

**1. Startup panic from route overlap — verified CLEAR.** axum is 0.8.9. The two full-app tests (`build_app_with_config`) are `#[ignore]`d (need a live DB), so `cargo test` never exercises the real merge — and `unsupported_routes_merge_with_existing_method_routes` only merges a 2-route stub, not the risky catch-all-vs-literal case. So I reproduced the exact registration (literal `/models` + `POST /models` + `GET/DELETE /models/{*model_id}` + `POST`/`GET /responses`, nested under `/v1`) in a standalone axum-0.8.9 binary → **full merge builds, no panic**. matchit treats literal `/models` and catch-all `/models/{*model_id}` as non-overlapping. Full lib suite: 137 passed, 0 failed, `test_openapi_spec_generation` green.

**2. No live endpoint regressed to 501 — verified.** Existing `/responses` routes on main: `POST /responses`, `GET/DELETE /responses/{id}`, `POST /responses/{id}/cancel`, `GET /responses/{id}/input_items` — there is **no** `GET /responses` (list), so the new 501 is correct. `GET /models` list lives at lib.rs:1115; there is no real `GET/DELETE /models/{id}` or `POST /models` on main; `/model/*` (singular) and `/admin/models/*` are separate surfaces the catch-all doesn't touch.

**3. Catch-all scope — verified by dispatching against the real merged+nested router:** `GET /v1/models` → real list handler (not shadowed); `POST /v1/responses` + `GET /v1/responses/{id}` → real handlers; `GET/DELETE /v1/models/openai/gpt-oss-120b`, `POST /v1/models`, `GET /v1/responses`, `POST /v1/moderations` → 501. Correct.

**4. Auth/ratelimit layering** matches `other_routes` (auth outermost → unauthenticated request gets 401, not a 501 leak). Handler/`ErrorResponse` set `type=not_implemented`, `param/code=null` as asserted.

`cargo test -p api --lib unsupported` 3/3; `cargo test -p api --lib` 137 pass; fmt clean; clippy clean.

---

**Minor (non-blocking) — misleading error message under `/v1` nesting:** `openai_endpoint_not_implemented` builds the message from `request.uri().path()`, but axum nesting strips the mount prefix, so the response reads `"POST /moderations is not implemented…"` while the client actually called `/v1/moderations`. The committed unit test passes only because it exercises `openai_compat_routes()` un-nested. Consider using the `axum::extract::OriginalUri` extractor (or prepending `/v1`) so the message reflects the real path. Cosmetic; safe to address in a follow-up.

**Note:** `unsupported_routes_merge_with_existing_method_routes` proves a 2-route stub merges but not the catch-all-vs-literal `/models` case (the real risk) — the real merge is only behind `#[ignore]`d DB tests. Worth strengthening if you want CI to guard the actual collision.

Evrard-Nil:
Re-reviewed the follow-up commit (`5a2d620`):
- **Path-message fix** — switched to `OriginalUri` + `Method` extractors; the 501 message now reflects the full `/v1/...` path, and the test nests under `/v1` and asserts it. Resolves my earlier nit.
- **Expanded surface** — `/batches`, `/threads`, `/assistants` → `any()` plus `/{*path}` catch-alls, and `/models/{*model_id}` → `any()`. You correctly left `/responses` as `get()`-only and `/models` as `post()`-only, so no collision with the real `POST /responses` / `GET /models`.

Re-verified locally: rebuilt, `cargo test -p api --lib unsupported` 3/3 green (incl. the new nested + catch-all-subpath cases), fmt clean, clippy clean. Re-ran a standalone axum-0.8.9 merge of the **new** route set against the real `/responses`+`/models` routes nested under `/v1` → builds with no panic. Catch-alls only cover unimplemented surfaces, so nothing real is shadowed. Approval stands.

PierreLeGuen (commented):
This adds an explicitly-routed group of unsupported OpenAI endpoints under `/v1` returning a structured `501 {"error": {"type": "not_implemented", ...}}` envelope instead of Axum's bare 404/405. The approach is sound and follows the repo's existing placeholder convention in `routes/responses.rs` (delete/cancel/list_input_items). No route conflicts with the existing surface (`GET /models` + new `POST /models` and `POST /responses` + new `GET /responses` merge cleanly at the method-router level, covered by the new construction test); middleware ordering matches sibling builders with `usage_check`/`body_hash` correctly omitted; CORS preflights are unaffected since the `CorsLayer` sits outside the router; and the handler echoes method+path to the client without logging anything.

Non-blocking observations:

- **SDK retry behavior** — `crates/api/src/routes/unsupported.rs:36-40`: The official OpenAI SDKs (Python and Node) retry any status ≥ 500, so a 501 here means e.g. `client.batches.create()` makes 3 attempts with backoff — burning rate-limit quota and per-attempt auth DB lookups — before surfacing a retryable `InternalServerError` for a permanent condition. That partially undercuts the PR's goal of clean SDK ergonomics (OpenAI itself returns a terminal 404 `invalid_request_error` for unsupported URLs). Since 501 matches the existing convention this isn't blocking, but consider adding the `x-should-retry: false` header (honored by both SDKs) to these responses — and ideally the existing 501s in `routes/responses.rs:669-742` — or switching to 404 + `invalid_request_error` to mirror OpenAI exactly.
- **Coverage gaps** — `crates/api/src/routes/unsupported.rs:12-25`: Other recognized OpenAI surfaces still fall through to the bare empty 404 this PR set out to eliminate, e.g. `POST /v1/audio/speech`, `/v1/fine_tuning/*`, `/v1/vector_stores/*`, `/v1/uploads`, `GET /v1/chat/completions/{id}`. Fine as scoped to #614, but worth either adding them (one line each now) or noting the gap on the issue.
- **OpenAPI spec** — `crates/api/src/lib.rs:940`: The new routes aren't registered in the utoipa spec, so `/docs` won't show that these paths intentionally return 501. Likely acceptable for placeholders, but worth a deliberate call rather than an omission.

Checks run: `cargo fmt --all -- --check` (clean), `cargo clippy -p api --all-targets` (clean), `cargo test -p api --lib unsupported` (3 tests passed), `git diff --check` (clean). Full e2e suite not run locally (requires PostgreSQL), but CI's Test Suite (lib/bins + `e2e_all` against Postgres), Lint (fmt + clippy `-D warnings`, `--all-features`), and security audit all passed on the PR head; the e2e job builds the complete app router, ruling out route-merge panics at startup.

Evrard-Nil:
Concurring with @PierreLeGuen's SDK-retry point — that's the one real ergonomics wrinkle here and I'd missed it. The `openai` Python and Node SDKs both treat any status ≥ 500 as retryable (Node also honors the `x-should-retry` response header explicitly, and Python respects it too), so a 501 on a *permanent* not-implemented path costs ~3 attempts × (backoff + auth DB lookup + rate-limit decrement) before the client sees a terminal-but-mislabeled `InternalServerError`.

Adding `x-should-retry: false` to these responses (and, for consistency, to the existing 501 placeholders in `routes/responses.rs:669-742`) is the lowest-risk fix and keeps the 501 semantics. Switching to `404 + invalid_request_error` would mirror OpenAI exactly but diverges from the repo's current placeholder convention, so the header approach seems preferable.

Non-blocking — my approval stands. Worth a quick follow-up either in this PR or a small one against #614.

lloydmak99:
Addressed the SDK retry feedback in `c164701`.

- Added shared `x-should-retry: false` constants for permanent errors.
- The unsupported OpenAI endpoint 501 handler now returns the header, and the existing unsupported-route test asserts it.
- The existing Responses 501 placeholders (`get_response`, `delete_response`, `cancel_response`) now use the same no-retry helper, with a unit test covering the header.

Left the broader unsupported-surface/OpenAPI coverage notes scoped as non-blocking follow-up for #614.

Validation:
- `cargo test -p api --lib unsupported`
- `cargo test -p api --lib not_implemented_error_disables_sdk_retries`
- `git diff --check`
- `cargo clippy -p api --all-targets -- -D warnings`

PR #763, opened 2026-06-10
Backfill Qwen3.5 reasoning feature metadata

### LINKED ISSUE #703: Qwen/Qwen3.5-122B-A10B metadata missing 'reasoning' in supportedFeatures despite reasoning-by-default behavior
## Summary

`Qwen/Qwen3.5-122B-A10B` reasons by default in production — responses include `reasoning_content` and `usage.reasoning_tokens`, and `chat_template_kwargs.enable_thinking: false` disables it (verified live 2026-06-02, repeatedly). But its catalog metadata lists only:

```
Qwen/Qwen3.5-122B-A10B → supported_features: ['tools', 'structured_outputs']
```

while e.g. `Qwen/Qwen3.6-35B-A3B-FP8` correctly lists `['tools', 'structured_outputs', 'reasoning']`.

## Repro

```bash
curl -s https://cloud-api.near.ai/v1/chat/completions -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen3.5-122B-A10B","messages":[{"role":"user","content":"What is 17*23?"}],"max_tokens":400}'
# → message.reasoning_content populated, usage.reasoning_tokens: 403
```

vs `GET /v1/models` → `supported_features: ["tools", "structured_outputs"]` for that model.

## Fix

One admin PATCH (requires @near.ai admin JWT):

```bash
curl -X PATCH https://cloud-api.near.ai/v1/admin/models \
  -H "Authorization: Bearer $ADMIN_SESSION" -H "Content-Type: application/json" \
  -d '{"modelName":"Qwen/Qwen3.5-122B-A10B","supportedFeatures":["tools","structured_outputs","reasoning"]}'
```

(Field name per the supportedFeatures backfill convention; verify against `GET /v1/admin/models` rather than the cached public endpoint.)

## Context

Surfaced during docs fact-checking: nearai/docs#29 documents Qwen3.5-122B reasoning configuration (live-verified) and tells readers to check `supported_features` — flagged as inconsistent in review. Once this metadata is fixed the docs caveat can be dropped.

lloydmak99:
Fixes #703

## Problem

`Qwen/Qwen3.5-122B-A10B` reasons by default, but issue #703 reported that its catalog metadata did not advertise the `reasoning` supported feature. That makes the model look less capable than Qwen3.6 and forces docs/clients to carry a caveat.

Live check on 2026-06-10: `GET https://cloud-api.near.ai/v1/models` already returns `reasoning` for this model, so production appears to have been patched manually. The database migration is still useful because it makes the fix reproducible for any environment that has not received the manual admin update and repairs the current history snapshot if only `models` was patched directly.

## Change

- Adds migration `V0060__backfill_qwen35_reasoning_feature.sql`.
- Appends `reasoning` to `models.supported_features` for `Qwen/Qwen3.5-122B-A10B` only when it is missing.
- Updates `models.updated_at` when the `models` row is changed.
- Independently repairs the current open `model_history` snapshot when it is missing `reasoning`, even if `models` had already been patched manually.
- Uses `COALESCE(..., '{}')` around feature arrays so the migration remains defensive in non-prod or drifted databases.

## What Else Is Needed

- No manual prod PATCH should be needed if the live catalog remains patched.
- After deployment, verify staging/prod `GET /v1/models` still reports `reasoning` for `Qwen/Qwen3.5-122B-A10B`.

## Tests

- `git diff --check`
- Current main migration numbering checked: `origin/main` ends at `V0059`, so this PR's `V0060` does not collide.
- Prior validation on this PR: `cargo test -p database --lib`, `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --lib --bins`.


Evrard-Nil (approved):
Reviewed locally — applied **all 57 migrations + V0058 against a throwaway Postgres 16** and exercised the backfill. **LGTM — approving.**

(Reviewed at head `0a33f8d` — the follow-up commit "Handle nullable feature arrays" that COALESCEs the array; good hardening.)

**Verified:**
- **Column type is `TEXT[] NOT NULL DEFAULT '{}'`** on *both* `models` and `model_history` (`V0051__add_openrouter_model_fields.sql:36,46`; confirmed live: `udt_name=_text`). So `array_append` / `= ANY` are valid — no JSONB type-error risk — and `model_history` does have the column. ✅
- **Numbering/runner:** refinery 0.9, loaded at runtime via `load_sql_migrations` (not embedded/checksummed at build), highest existing on branch+main is `V0057`, so `V0058` is the correct next number, no collision, naming matches. ✅
- **Idempotent + NULL-safe (proven):** seeded a Qwen row with `{tools}` (no reasoning) →run 1 `UPDATE 1` → `{tools,reasoning}`; run 2 `UPDATE 0` (no dup). NULL-features row → backfills to `{reasoning}` thanks to the COALESCE. ✅
- **History invariant:** the app keeps exactly one open (`effective_until IS NULL`) snapshot per model_id via close-then-insert (`repositories/model.rs:944-955`); the migration's second UPDATE correctly targets only that open row. Verified: closed history row left as `{tools}`, decoy model untouched. ✅
- **model_name exact match** to the live `'Qwen/Qwen3.5-122B-A10B'` (no case/typo no-op). ✅
- `cargo build -p database` / `cargo fmt --all -- --check` / `cargo clippy -p database --all-targets -- -D warnings` all clean. ✅

**Note (non-blocking):** the `COALESCE(supported_features, '{}'::text[])` is technically unreachable given the `NOT NULL DEFAULT '{}'` constraint — harmless defensive hygiene, no change needed.

PierreLeGuen (commented):
This migration is correct, safe, and well-scoped. I validated it functionally: the full migration chain V0001→V0058 applies cleanly to a fresh PostgreSQL database, and on a scratch DB the migration appends `reasoning` exactly once to `Qwen/Qwen3.5-122B-A10B` only, updates only the open (`effective_until IS NULL`) `model_history` row, leaves closed history rows and other models untouched, and is fully idempotent (`UPDATE 0` on re-run). The model name is `UNIQUE` (V0002) so the CTE targets exactly one row, the auto-history trigger was dropped in V0036 so no spurious history rows are created, and V0058 is the correct next number. Production's `/v1/models` already lists `reasoning` for this model, so the migration will correctly no-op there, and `supported_features` is written only via admin endpoints (not the periodic discovery sync), so nothing will overwrite the backfill at runtime.

- `crates/database/src/migrations/sql/V0058__backfill_qwen35_reasoning_feature.sql:7` — The `COALESCE(supported_features, '{}'::text[])` guards are dead code: both `models.supported_features` and `model_history.supported_features` have been `TEXT[] NOT NULL DEFAULT '{}'` since V0051, and `ADD COLUMN ... NOT NULL DEFAULT` backfills pre-existing rows. Harmless, but the "nullable feature array" framing in the second commit implies a schema property that doesn't exist — consider dropping the COALESCEs or fixing the rationale.
- `crates/database/src/migrations/sql/V0058__backfill_qwen35_reasoning_feature.sql:13-17` — Rewriting the open `model_history` row in place diverges from the convention in `record_model_history` (`crates/database/src/repositories/model.rs:944-991`), which closes the open row and inserts a new attributed one. For a "snapshot was wrong all along" backfill that's defensible, but it edits an audit table without attribution — a short comment in the migration explaining why close+insert was deliberately not used would prevent this being copied as a pattern for genuine state changes. Related edge case worth noting in the same comment: the history update only fires when the `models` row changes in the same statement, so a database where `models` was already patched directly via SQL (but history wasn't) would keep its stale history row. Production was patched via the admin endpoint, which maintains history, so this is theoretical.

Checks run:
- `cargo fmt --all -- --check`, `cargo clippy -p database --all-targets -- -D warnings`, `cargo test -p database --lib` (13 passed) — all clean
- Full migration chain V0001→V0058 applied to a fresh PostgreSQL 16 database; functional and idempotency tests of V0058 on a scratch DB
- `git diff --check` on the new SQL file — clean
- Live check of `https://cloud-api.near.ai/v1/models` confirmed the exact model name and that production already lists `reasoning`
- CI: Lint, Test Suite, and security_audit all green

Evrard-Nil:
Heads-up on a migration-number collision: this PR adds `V0058__backfill_qwen35_reasoning_feature.sql`, and #765 (scheduled pricing changes) also introduces `V0058` (+ `V0059`). Both are the correct next version relative to `main` right now, but whichever of the two merges **second** will end up with a duplicate `V0058` and refinery will refuse to load it (versions must be unique). Whichever lands second should renumber (e.g. to `V0060`). No change needed unless this merges after #765.

PierreLeGuen:
Heads-up: #765 (scheduled pricing changes) is approved and merging first with **V0058/V0059**. This PR's `V0058__backfill_qwen35_reasoning_feature.sql` will need to be renumbered to **V0060** before merging, otherwise refinery will refuse to boot on duplicate version numbers.

Evrard-Nil (dismissed):
Converting my prior approval to request-changes — **the migration collision is now LIVE**, not hypothetical.

#765 merged to main at 08:09 (commit 79b8a57), so main now occupies:
- `V0058__add_scheduled_model_pricing_changes.sql`
- `V0059__add_model_pricing_change_email_deliveries.sql`

This PR still adds `V0058__backfill_qwen35_reasoning_feature.sql`. Refinery requires unique version numbers and **refuses to boot on a duplicate**, so merging as-is would break cloud-api startup in prod.

**Required before merge**: rename the migration file
`V0058__backfill_qwen35_reasoning_feature.sql` → `V0060__backfill_qwen35_reasoning_feature.sql`
(V0060 is the next free slot — main tops out at V0059). No content change needed; it's a standalone backfill with no ordering dependency on V0058/V0059. Refinery loads by version, so the rename alone is sufficient.

I'm blocking only to prevent an accidental merge at the duplicate version — the SQL itself is unchanged and still correct (my earlier full review at 0a33f8d stands). I'll re-approve immediately once it's renumbered to V0060.

lloydmak99:
Pushed follow-up commit `61fa97d` for the live migration-version collision.

Change:
- Renamed `V0058__backfill_qwen35_reasoning_feature.sql` to `V0060__backfill_qwen35_reasoning_feature.sql`.
- SQL content is unchanged; this is only the refinery version renumber after main claimed `V0058` and `V0059`.

Validation:
- `git diff --check`
- `cargo build -p database`

PierreLeGuen (commented):
This is a tight, well-scoped change: one 17-line migration (`crates/database/src/migrations/sql/V0060__backfill_qwen35_reasoning_feature.sql`) that appends `reasoning` to `supported_features` for `Qwen/Qwen3.5-122B-A10B` and mirrors the change onto the open (`effective_until IS NULL`) `model_history` snapshot. No Rust changes.

The migration is correct. The `WHERE NOT ('reasoning' = ANY(...))` guard makes it idempotent (verified empirically: rerun is `UPDATE 0`), it touches only the target model and the open history snapshot (matching what the application-level `record_model_history` would do), and the runner picks up the file at runtime with no code changes needed. Numbering checks out: main already has `V0058`/`V0059`, so `V0060` is the next free slot, and the unmerged `V0054` branch touches disjoint concerns so there is no sequencing hazard either way. A live check of `https://cloud-api.near.ai/v1/models` confirms prod already advertises `reasoning` for this model, so the migration is a deliberate no-op there, as the description states.

Non-blocking notes:

- **Root cause not addressed** (`V0060__backfill_qwen35_reasoning_feature.sql:1`): this fixes the database row but not the metadata source that originally omitted `reasoning` (issue #703). The admin update path (`crates/database/src/repositories/model.rs:452`) overwrites `supported_features` whenever a request supplies the array, so a future bulk metadata refresh from the same upstream source could silently regress this. Please confirm the upstream model-metadata source now includes `reasoning` for this model, or note in #703 that only the DB was corrected.
- **Dead COALESCE** (`V0060__backfill_qwen35_reasoning_feature.sql:7`): `COALESCE(supported_features, '{}'::text[])` is unreachable — both `models.supported_features` and `model_history.supported_features` are `NOT NULL DEFAULT '{}'` per `V0051`. Harmless defensive coding, fine to keep, but the commit message's "nullable feature arrays" rationale is inaccurate.
- **Stale branch base**: the branch is based on a main snapshot that lacks `V0058`/`V0059`, so running it directly against a database already migrated past `V0059` hits refinery's missing-migration abort. Harmless once merged; rebasing onto current main avoids dev-environment confusion.

Checks run: executed the migration twice against a scratch PostgreSQL 16 database with the production column shapes (targeted update, history sync limited to the open snapshot, unrelated rows untouched, idempotent rerun); verified the live models endpoint and migration numbering against main; `cargo fmt --all -- --check`, `cargo check -p database --lib`, `cargo clippy -p database --lib -- -D warnings`, and `cargo test -p database --lib` (13 tests) all passed locally; CI (Lint, Test Suite, security_audit) is green.

lloydmak99:
Pushed `17c1d2f` to address Pierre's follow-up notes.

- Rebased the branch onto current `main` (`8b6a19f`), so the branch now includes the merged `V0058`/`V0059` migrations and the latest base commits.
- Removed the dead `COALESCE(..., '{}'::text[])` guards. `supported_features` is `NOT NULL DEFAULT '{}'` by schema, so the migration now relies on that invariant and the commit history no longer carries the inaccurate nullable-array rationale.
- Added an inline migration comment explaining why the currently-open `model_history` snapshot is corrected in place and what happens if the `models` row was already patched independently.
- Checked the root-cause/source question: in the current repo, the external-provider refresh path reads model rows from the DB and loads providers; it does not import upstream/OpenRouter metadata back into `models.supported_features`. The overwrite risk is an explicit future admin/bulk upsert that supplies a stale `supportedFeatures` array. I noted that on #703 so the source/operator metadata concern remains tracked separately from this backfill.

Validation rerun after the final rebase:

- `cargo check -p database --lib`
- `cargo test -p database --lib` (13 passed)

CI is queued on the new SHA.

PierreLeGuen (commented):
Solid, well-scoped backfill migration. `V0060__backfill_qwen35_reasoning_feature.sql` follows the established V0051 pattern: `supported_features` is `TEXT[] NOT NULL DEFAULT '{}'` on both tables, so `array_append` is safe; the `NOT ('reasoning' = ANY(supported_features))` guard makes it idempotent; it's scoped to the exact `model_name`; numbering correctly follows V0059 on main; and routine write paths won't clobber the value (`seed_model_if_absent` is insert-only, admin pricing updates use `COALESCE`). Executed against a scratch Postgres database, the migration behaved as documented across backfill-needed, unrelated-model isolation, idempotent re-run, and already-patched scenarios.

Two non-blocking observations:

1. **Stale open `model_history` snapshot when `models` was patched independently** (`V0060__backfill_qwen35_reasoning_feature.sql:17-21`). The history repair only fires when the `models` UPDATE matched a row. Per the PR body, production already had the `models` row patched manually — if that patch was direct SQL rather than the admin PATCH endpoint, the migration is a no-op there and the open snapshot (`effective_until IS NULL`) stays missing `reasoning`, which would surface in `get_model_state_at_time` and history endpoints. Verified empirically on a scratch DB. Consider making the history update self-guarded and independent, e.g.:
   ```sql
   UPDATE model_history mh
   SET supported_features = array_append(mh.supported_features, 'reasoning')
   FROM models m
   WHERE mh.model_id = m.id
     AND m.model_name = 'Qwen/Qwen3.5-122B-A10B'
     AND mh.effective_until IS NULL
     AND NOT ('reasoning' = ANY(mh.supported_features));
   ```
   The migration's own comment acknowledges this trade-off, so leaving it as-is is defensible if you've confirmed the production patch went through the admin API.

2. **PR description is stale**: it names the file `V0058` (actual: `V0060` after renumbering) and claims "defensive handling for a nullable feature array" that the final SQL neither has nor needs (the column is `NOT NULL`). Worth updating so the merged description matches the SQL.

Checks run: migration executed end-to-end against a throwaway local Postgres replica of the `models`/`model_history` schema (backfill, isolation, idempotency, and drift scenarios); static schema verification against V0002/V0036/V0051 and `record_model_history` (`crates/database/src/repositories/model.rs:1067`); `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --lib --bins` (383 passed) all clean; `git diff --check` clean; PR CI (Lint, Test Suite, security_audit) green.

lloydmak99:
Follow-up after the latest review:

- Renumbering remains `V0060`, following current `origin/main`'s `V0059`.
- Updated the migration so the current open `model_history` snapshot is repaired independently of the `models` row update.
- Added `COALESCE(..., '{}')` guards around both feature-array checks/appends so drifted or non-prod databases with nullable data still backfill safely.
- Updated the PR description to match the final migration filename and behavior.

Validation:

- `git diff --check`
- Confirmed current `origin/main` migration numbering ends at `V0059`, so `V0060` does not collide.

PierreLeGuen (commented):
The migration is correct and well-guarded. I verified it live against PostgreSQL across four scenarios — initial backfill, idempotent re-run, history-only repair after a manual patch of `models`, and a clean no-op when the model row is absent — and all behaved as intended. Context checks confirm the surrounding invariants: `model_name` is UNIQUE (V0002) so the CTE is single-row; `supported_features` is `TEXT[] NOT NULL DEFAULT '{}'` on both `models` and `model_history` (V0051); `reasoning` matches the admin allowlist (`crates/api/src/routes/admin.rs:282`) and OpenRouter's vocabulary; discovery inserts use `ON CONFLICT DO NOTHING` and admin updates use `COALESCE`, so nothing automated will clobber the backfill; V0060 doesn't collide with `main`, whose latest migration is V0059. A concurrent admin update to `model_history` is also safe in all orderings, since the migration only appends the feature and any close-and-insert copies the amended row forward.

- `crates/database/src/migrations/sql/V0060__backfill_qwen35_reasoning_feature.sql:4` — The migration only helps databases where the `Qwen/Qwen3.5-122B-A10B` row already exists. In a fresh environment, V0060 runs at startup before discovery inserts the row, which is then created from upstream metadata that still lacks `reasoning` (`crates/database/src/repositories/model.rs:691-750`, `ON CONFLICT DO NOTHING` with caller-supplied features). The PR body's claim that this "makes the fix reproducible for any environment" is slightly overstated — the root cause is the upstream metadata feeding discovery. Suggest noting the limitation in the migration comment and/or filing a follow-up to fix the discovery-side metadata.
- `crates/database/src/migrations/sql/V0060__backfill_qwen35_reasoning_feature.sql:9-17` — The `updated_model` CTE is never referenced by the final statement; it executes only because Postgres runs data-modifying CTEs even when unreferenced (verified live). This reads like dead code and could be "cleaned up" (i.e., broken) in a future edit. Either split into two plain statements (refinery handles multi-statement files fine) or add a one-line comment stating the CTE executes despite being unreferenced.
- `crates/database/src/migrations/sql/V0060__backfill_qwen35_reasoning_feature.sql:22-27` — The in-place edit of the open `model_history` snapshot bypasses the app's close-and-insert audit pattern (`crates/database/src/repositories/model.rs:1076-1123`) and leaves no `changed_by_*` trace. The migration comment's framing as a metadata repair rather than a state transition is defensible; flagging it so the team consciously accepts that history won't show when or why `reasoning` appeared in the open snapshot.

Checks run locally:
- Live migration test against PostgreSQL 16 in a scratch database with the real V0002/V0051 column types: backfill, idempotent re-run (no double-append), history-only repair, and absent-model no-op all passed; closed snapshots and other models untouched, `updated_at` bumped only on real change.
- `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --lib --bins` (383 passed, 1 ignored), `cargo test -p database --lib` (13 passed), `git diff --check`: all clean.
- Static verification: migration numbering (V0059 exists, no V0060 collision; the V0053→V0055 gap pre-exists on `main`), migrations loaded at runtime via `refinery::load_sql_migrations`, no DB triggers maintain `model_history` since V0036, and CI (lint, test suite, security audit) is green on the head commit.

PierreLeGuen (approved):
This PR adds a single 27-line SQL migration (`V0060__backfill_qwen35_reasoning_feature.sql`, no Rust changes) that appends `reasoning` to `supported_features` for `Qwen/Qwen3.5-122B-A10B` in both the `models` row and the currently-open `model_history` snapshot, with each table guarded independently. The change is correct, idempotent, and safe to land.

What I verified:

- **Schema correctness** — `models.model_name` is `VARCHAR(500) NOT NULL UNIQUE` (V0002), so the exact-match CTE is single-row; `supported_features TEXT[] NOT NULL DEFAULT '{}'` exists on both tables (V0051), so `array_append` and `= ANY` are type-safe. `'reasoning'` is in the admin route's validated feature vocabulary (`crates/api/src/routes/admin.rs:282`).
- **Design consistency** — `record_model_history` (`crates/database/src/repositories/model.rs:1067`) keeps the open snapshot as a mirror of the current `models` row, so repairing it in place (rather than closing it and inserting a new transition) is the right call for a metadata fix. Closed historical snapshots are untouched, and `model_history` has no auto-trigger (dropped in V0036), so no spurious rows.
- **Live execution** — ran the migration against a scratch PostgreSQL database covering: feature missing in both tables (both repaired, `updated_at` bumped), models patched but history drifted (only history repaired, `updated_at` not bumped), model absent (clean no-op), re-run (idempotent), and non-interference with unrelated models and closed snapshots. All passed.
- **Numbering** — `origin/main` ends at V0059, so V0060 doesn't collide; the V0054 gap pre-dates this PR and refinery doesn't require contiguous versions.

Two optional style notes, neither blocking:

- `V0060__backfill_qwen35_reasoning_feature.sql:9` — the `updated_model` CTE is never referenced by the final statement; it only runs because PostgreSQL executes data-modifying CTEs unconditionally. Verified empirically that it does execute, so behavior is correct, but a future editor porting the pattern could break it silently. Consider splitting into two plain `UPDATE` statements (the second already does its own `target_model` lookup) or referencing the CTE in a final `SELECT`.
- `V0060__backfill_qwen35_reasoning_feature.sql:12` — the `COALESCE(supported_features, '{}')` guards are redundant given the columns are `NOT NULL DEFAULT '{}'` since V0051; the commit message frames this as deliberate defensiveness for drifted databases, so fine to keep.

Checks run: `cargo fmt --all -- --check` (clean), `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo test -p database --lib` (13/13 passed), `git diff --check` (clean), live migration testing on a scratch PostgreSQL database (created and dropped for this review), and migration-numbering verification against `origin/main`. CI on the PR head: Lint, Test Suite, and security_audit all passing.

PR #764, opened 2026-06-10
Fix Whisper transcription response validation

### LINKED ISSUE #608: Whisper transcriptions: response_format text/srt/vtt and timestamp_granularities not implemented per OpenAI spec
## Summary

`POST /v1/audio/transcriptions` with `openai/whisper-large-v3` diverges from the OpenAI spec on several documented parameters:

| Parameter | Expected | Actual |
|---|---|---|
| `response_format=text` | plain-text body | JSON envelope (param ignored) |
| `response_format=srt` | SRT subtitle text | HTTP 502 "Audio transcription failed. Please try again later." |
| `response_format=vtt` | WebVTT subtitle text | HTTP 502 same |
| `response_format=verbose_json` | JSON with `segments` (and `words` when granularities set) | works ✓ |
| `timestamp_granularities[]=word` | response includes `words` array | silently ignored — no `words` key |
| `language=en` | language pinned | accepted, no error ✓ |
| (default) | `{"text":"..."}` | `{"text":"...","id":"trans-..."}` (extra `id`) |

The 502 / "try again later" message for `srt` and `vtt` is the same misleading shape as #605 and #606 — a 4xx (or "unsupported format") gets re-wrapped as a 5xx server error, training users to retry uselessly.

`response_format=text` ignoring the parameter is the most insidious — users `curl -F response_format=text ... | grep ...` and get JSON back; the bug is silent.

## Repro

```bash
say "Hello world." -o /tmp/a.aiff && ffmpeg -y -i /tmp/a.aiff -ar 16000 -ac 1 /tmp/a.wav

# text → returns JSON
curl -s https://cloud-api.near.ai/v1/audio/transcriptions \
  -H "Authorization: Bearer $KEY" \
  -F file=@/tmp/a.wav -F model=openai/whisper-large-v3 -F response_format=text

# srt → 502
curl -s https://cloud-api.near.ai/v1/audio/transcriptions \
  -H "Authorization: Bearer $KEY" \
  -F file=@/tmp/a.wav -F model=openai/whisper-large-v3 -F response_format=srt

# vtt → 502 (same)
curl -s ... -F response_format=vtt

# word timestamps → silently dropped
curl -s https://cloud-api.near.ai/v1/audio/transcriptions \
  -H "Authorization: Bearer $KEY" \
  -F file=@/tmp/a.wav -F model=openai/whisper-large-v3 \
  -F response_format=verbose_json -F 'timestamp_granularities[]=word' | jq 'keys'
# ["duration","id","language","segments","text"]  ← no "words"
```

## Suggested fixes

1. **`response_format=text`**: honor it — return `text/plain` body with just `text`, no JSON envelope. (`Content-Type: text/plain; charset=utf-8`.)
2. **`response_format=srt` / `vtt`**: either implement them (build SRT/VTT from segments — straightforward given verbose_json already has `start`/`end`/`text` per segment) or reject upfront with `400 unsupported response_format` instead of 502.
3. **`timestamp_granularities[]=word`**: either implement word-level timestamps (vLLM's whisper supports `word_timestamps=true`) or reject the parameter with a clear `400`.
4. **`id` field**: drop from response (not in OpenAI spec, leaks internal request ID).
5. **Same error-wrapping problem as #605, #606, #607**: any upstream error becomes "5xx server_error, please retry." Triage these together when revisiting error normalization.

## Related

- #605 — embeddings `dimensions` → same 502/"try again later" pattern
- #606 — vision `image_url` UA-blocked → same 502/"try again later" pattern
- #607 — images `size` out-of-range → same 5xx pattern

A consistent boundary that maps upstream client-errors to `400` with the real message would fix the UX across all four.

---

lloydmak99:
## Problem

`POST /v1/audio/transcriptions` did not consistently honor or validate OpenAI-compatible Whisper parameters:

- `response_format=text` still returned a JSON envelope instead of `text/plain`.
- `response_format=srt` and `response_format=vtt` were accepted by validation, forwarded to the provider, and surfaced as retryable 5xx-style failures.
- `timestamp_granularities[]=word` could be silently dropped before reaching real providers.
- Invalid `language` values reached the provider and produced misleading upstream failures.
- Provider transcription responses could include an internal `id` field that is not part of the OpenAI transcription response shape.

Closes #608.

## Change

- Normalize `response_format=text` internally by requesting `verbose_json` from the provider, recording usage as before, and returning only the transcribed text with `Content-Type: text/plain; charset=utf-8`.
- Reject unsupported `srt` and `vtt` response formats at the API boundary with `400 invalid_request_error`.
- Validate timestamp granularities before dispatch:
  - require `response_format=verbose_json` when timestamp granularities are supplied
  - allow `segment` and `word`
- Forward timestamp granularities to both real transcription adapters as repeated multipart `timestamp_granularities[]` fields.
- Validate `language` against the Whisper audio transcription language set while accepting case/region code variants like `EN` and `en-US`, full language names like `english`, and common Whisper aliases like `cantonese`, `yue`, `haw`, and `jw`.
- Convert provider transcription responses into the API response model before JSON serialization so provider-only `id` values are omitted.
- Accumulate repeated `timestamp_granularities[]` multipart fields instead of overwriting earlier values.
- Add unit, e2e, and adapter-level coverage for text responses, unsupported formats, timestamp validation/forwarding, invalid language, language normalization/aliases, provider `id` omission, and sorted language-code/alias lookup.

## Release / Compatibility Notes

- `response_format=srt` and `response_format=vtt` now return `400 invalid_request_error` instead of reaching providers and surfacing retryable 5xx failures. This is intentional; subtitle rendering from `verbose_json` segments can be added separately.
- `language` values are canonicalized before dispatch. Code inputs such as `EN`/`en-US` and full-name inputs such as `english` now forward as their Whisper language codes.

## What Else Is Needed

- The production decoder gaps for `m4a`/`webm` called out in the issue likely need a container/provider image change, not an API validation change.

## Tests

- `git diff --check`
- `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields`
- `cargo test --locked -p api --lib audio_transcription`
- Earlier local runs also covered `cargo test -p api --lib`; database-backed e2e tests compile here but require a local Postgres test DB to run.


Evrard-Nil (changes requested):
Reviewed locally (fresh clone, full unit + e2e run against a local Postgres). The core mechanics are solid, but I'm holding on one **functional regression** the new `language` validation introduces — details below. Everything else verified clean.

## Verified clean ✅
- **`validate()` is wired in** — `audio_transcriptions` calls `request.validate()` and maps `Err` → 400 `invalid_request_error` at `completions.rs:3504-3513`, before model resolution and the provider call. The new checks are live, not dead code.
- **Billing integrity for `response_format=text`** — usage is recorded unconditionally at `completions.rs:3641-3656` (before the text-vs-json branch at 3665), and the provider is asked for `json` so `duration` is populated identically. `text` is billed exactly like `json`. No revenue gap.
- **`From<provider::AudioTranscriptionResponse>` drops only `id`** — the provider struct is exactly `{text, duration, language, segments, words, id}` (no `#[serde(flatten)] extra`, no `task`/`x_groq`/`usage`), so the conversion strips `id` (the #608 fix) and nothing else. `TranscriptionSegment`/`TranscriptionWord` are field-for-field parity. No silent field-drop regression for json/verbose_json callers.
- **timestamp_granularities accumulation** is correct (accumulates across repeated `[]` fields + comma-splits + trims).
- `cargo test -p api --lib audio` 7/7; `cargo test -p api --lib` 143 pass; **e2e `audio_transcription` 18/18** (incl. all 5 new); fmt + clippy clean.

## Please address before merge ⚠️ (functional regression)
**The `language` allowlist is strict ISO-639-1, but Whisper accepts a superset — so this rejects valid inputs that previously worked:**
- `yue` (Cantonese), `haw` (Hawaiian, 3-letter, not ISO-639-1), and the legacy `jw` (Javanese — the list has canonical `jv` but Whisper historically uses `jw`) are all **Whisper-supported yet now return 400**.
- The check is exact `contains(&language)` (case/region-sensitive), so `"EN"`, `"En"`, `"en-US"` now 400 too. Since `language` was previously forwarded unvalidated, this is a breaking change for any caller sending uppercase or BCP-47 region tags.

The PR's goal is to reject *invalid* languages with a clear 400 — rejecting *valid* Whisper languages defeats that. Suggested fixes (either works):
1. **Normalize then check**: `language.to_lowercase()` and strip a `-`/`_` region suffix before lookup, and add `yue`, `haw`, `jw` to the allowlist; **or**
2. **Be lenient on unknowns**: forward unrecognized codes to the provider (which has its own language set) instead of a hard 400 — only 400 on clearly-malformed input. This avoids playing catch-up with Whisper's evolving language list.

## Minor / cosmetic (non-blocking)
- The generic "invalid response_format" message still advertises `srt`/`vtt` in `valid_formats` (`models.rs:611`) even though both are then rejected as "not currently supported". Consider dropping them from the advertised list so the error isn't self-contradicting. (Ordering is fine — srt/vtt correctly hit the "not currently supported" message.)

lloydmak99:
Addressed the review findings in `9f70b7d`:

- Language validation now normalizes case and strips region suffixes before lookup, so `EN`, `En`, `en-US`, and `fr_CA` are accepted and forwarded as the base lowercase code.
- Added the Whisper-compatible aliases called out in review: `yue`, `haw`, and `jw`.
- Updated tests to cover normalized language forms and aliases.
- Updated the invalid `response_format` message so it no longer advertises `srt`/`vtt` as supported values while still returning a clear unsupported-format 400 for those two.
- Confirmed the encrypted-response note is a pre-existing/shared audio path: this endpoint still forwards the same encryption headers into the provider request, and the `text` branch only changes the final API response shape after provider parsing and usage recording.

Latest local checks:
- `git diff --check`
- `cargo test -p api --lib audio_transcription`
- `cargo test -p api --lib`
- `cargo clippy -p api --all-targets --all-features -- -D warnings`
- `cargo test --lib --bins`

PierreLeGuen (changes requested):
## Review

### Blocking

**`timestamp_granularities[]=word` is now rejected globally** — `crates/api/src/models.rs:643`. The base branch accepted `word` and forwarded `timestamp_granularities` in provider params, and both the provider and API response models already carry a `words` field, so this change removes a valid word-timestamp path for providers that support OpenAI/Whisper word timestamps. Either accept `word` and forward it in provider backends that support it, or reject it only after resolving a provider that genuinely cannot supply words. Please add coverage for `verbose_json` + `word` either way.

### Non-blocking

- `crates/api/src/models.rs:600-608` — `srt`/`vtt` now hit the "not currently supported" message while truly unknown formats hit the generic "Invalid response_format" path, so clients see two different errors for unsupported formats. Consider a single rejection path with one unified message listing the valid values.
- `crates/api/src/models.rs:661-689` — In `normalize_audio_transcription_language`, `.split(['-', '_']).next()` always returns `Some` (even for `""`), so the `unwrap_or(language)` is dead code. Harmless — the empty string fails the supported-language check anyway — but worth simplifying for clarity.

A few things checked out cleanly: the multipart parsing now correctly accumulates repeated `timestamp_granularities`/`timestamp_granularities[]` fields instead of overwriting, the language code list is properly sorted for `binary_search`, the `From` impl deliberately dropping the provider `id` is the right call, and routing `text` responses through `json` internally keeps usage recording intact. No new logging of customer data.

### Checks run

- `cargo fmt --check` and `cargo clippy -p api --all-targets --all-features -- -D warnings`: clean
- `cargo test -p api --lib`: 145 passed (2 pre-existing DB-dependent tests ignored)
- `cargo test -p api --test e2e_all audio_transcriptions:: -- --test-threads=1` (isolated DB + mock env): 18 passed
- Workspace test binaries compile (`cargo test --lib --bins --no-run`)
- `git diff --check`: clean; GitHub CI (`Lint`, `Test Suite`, `security_audit`) green

Evrard-Nil (dismissed):
Re-reviewed the latest commit (`9f70b7d`) locally — rebuilt + ran unit (9/9), full lib (145), and e2e `audio_transcription` (18/18), fmt + clippy clean.

**My earlier blockers are resolved — thanks for the quick turnaround:**
- ✅ `normalize_audio_transcription_language()` (`models.rs:680`) lowercases + strips a `-`/`_` region suffix, so `EN`, `En`, `en-US`, `fr_CA` now pass; `yue`, `haw`, `jw` added to the allowlist (kept `jv`). New tests `test_audio_transcription_accepts_normalized_language_forms` + `test_normalize_audio_transcription_language` cover it. The normalized form is forwarded to the provider too.
- ✅ The generic "Invalid response_format" message now lists only `json, text, verbose_json`; srt/vtt hit the dedicated "not currently supported" branch.

I also re-confirmed the previously-clean items survived the force-push: `validate()` still → 400 before dispatch, billing still recorded before the text/json branch, `From<provider>` still drops only `id`.

**Holding, though — I concur with @PierreLeGuen's open blocker on `timestamp_granularities[]=word`.** The hard 400 at `models.rs:643` rejects a value that the rest of the stack is actually wired for: `words` exists on `inference_providers::AudioTranscriptionResponse`, on the API `AudioTranscriptionResponse`, and the new `From` impl maps it. So if the Whisper backend returns word timestamps for `verbose_json` + `timestamp_granularities[]=word`, this PR is removing a capability that would otherwise work — the same "reject a valid input" concern as the language list. Codifying the rejection in a test locks in that behavior rather than resolving the objection.

Suggested resolution: verify against the actual provider whether `verbose_json` + `word` yields populated `words[]`. If yes, accept + forward `word` (with an accept-path test). If the backend genuinely can't produce them, keep the rejection but say so based on a verified provider limitation rather than assumption. Once that's settled I'm happy to approve.

lloydmak99:
Pushed follow-up commit `34a28ed` for the open `timestamp_granularities[]=word` blocker.

Changes:
- Accept `word` alongside `segment` when `response_format=verbose_json`.
- Keep forwarding `timestamp_granularities` unchanged to the provider params.
- Replace the previous word-timestamp rejection tests with acceptance coverage that asserts `words[]` is returned when the provider supplies it.

Validation:
- `git diff --check`
- `cargo test -p api --lib audio_transcription` (8/8 passed)
- `cargo test -p api --test e2e_all test_audio_transcription_word_timestamps_supported` could not run locally because the test DB bootstrap could not connect to Postgres (`Connection refused`).

PierreLeGuen (changes requested):
This is a solid cleanup of `/v1/audio/transcriptions` validation: `text` format now returns a real `text/plain` body, `srt`/`vtt` fail fast with 400 instead of misleading 502s, languages are validated at the boundary, the internal `id` no longer leaks into responses, and repeated `timestamp_granularities[]` fields accumulate correctly. Tests cover those paths well.

**Blocking**

- **`word` timestamp granularity is accepted but not actually delivered** (`crates/api/src/models.rs:640-648`). The second commit re-allows `word`, but neither provider path was changed to forward it correctly: the attested NEAR AI adapter (`crates/inference_providers/src/attested/nearai/mod.rs:1760-1763`) comma-joins granularities into a single form field, which OpenAI-compatible servers generally don't parse as a list, and the external OpenAI-compatible adapter (`crates/inference_providers/src/non_attested/external/openai_compatible.rs:411`) drops `params.timestamp_granularities` entirely when building the multipart form. This is exactly the silent-drop failure from #608 the PR sets out to fix. The e2e test that "proves" word support runs against the mock provider, which returns `words` unconditionally (`mock.rs:1414`), so it can't catch this. Either revert to rejecting `word` with a 400, or make it work end-to-end — send repeated `timestamp_granularities[]` fields in both adapters, add adapter-level tests asserting the outbound multipart fields, and verify against a real Whisper backend.
- **PR description and doc comments contradict the code.** The PR body says `word` is rejected, and the schema doc comment (`crates/api/src/models.rs:498`) says granularities "currently supports 'segment'", but the code accepts `word`. Reconcile these once the decision above is made.

**Non-blocking**

- `crates/api/src/models.rs:494` and `:522`: doc comments (which feed the published OpenAPI schema) still list `srt` and `vtt` as valid response formats, but they now return 400. Update to `json, text, verbose_json`.
- `crates/api/src/models.rs:608-628`: the language allowlist is the full ISO-639-1 set (187 codes), but Whisper supports only ~100 languages, so codes like `aa` or `nv` still reach the provider and reproduce the misleading upstream 5xx. Consider restricting to Whisper's actual set. Also, `binary_search` at `:625-628` is only correct while the array stays sorted — it currently is, but add a unit test asserting `is_sorted()` so a future insertion can't silently break validation.
- `crates/api/src/models.rs:672-677`: `.split(['-', '_']).next()` on a `&str` never returns `None`, so the `unwrap_or(language)` fallback is dead code. Harmless, but worth simplifying.
- Worth a release note: `language` values previously forwarded verbatim (e.g. `english`) now return 400, and region subtags are stripped (`en-US` → `en`). Defensible per the OpenAI spec, but a behavior tightening for lenient callers.

**Checks run locally:** `cargo fmt --all -- --check` and `git diff --check` clean; `cargo clippy -p api --all-targets --all-features -- -D warnings` clean; `cargo test -p api --lib audio_transcription` 8/8; `cargo test -p inference_providers --lib openai_compatible` 26/26; e2e `audio_transcription` suite 18/18 against a fresh database (initial failures were environmental — a stray V58 migration in the shared e2e DB); verified `AUDIO_TRANSCRIPTION_LANGUAGE_CODES` is sorted with no duplicates. CI: Lint, Test Suite, and security_audit all green.

lloydmak99:
Pushed follow-up commit `f916293` for the latest `word` timestamp blocker.

Changes:
- Both real transcription adapters now forward `timestamp_granularities` as repeated multipart `timestamp_granularities[]` fields instead of dropping or comma-joining them:
  - attested NEAR AI adapter
  - external OpenAI-compatible adapter
- Added adapter-level wire tests that inspect the actual outbound multipart body and assert two repeated `timestamp_granularities[]` fields (`word`, `segment`) and no `word,segment` collapsed value.
- Updated OpenAPI/doc comments to remove `srt`/`vtt` from accepted response formats and document `word` support.
- Added a sorted/unique invariant test for the binary-searched language-code allowlist.
- Simplified the now-dead `unwrap_or` in language normalization.
- Updated the PR description so it no longer says `word` is rejected.

Validation:
- `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields`
- `cargo test --locked -p api --lib audio_transcription`
- `git diff --check`

I did not run a live provider request with real Whisper credentials from this environment; the new tests verify the actual outbound multipart wire shape sent by both adapters.

PierreLeGuen (changes requested):
Solid fix overall — the `text` → `json` conversion at the provider boundary is correct (both real adapters unconditionally parse the provider body as JSON, `crates/inference_providers/src/attested/nearai/mod.rs:1816`), usage is recorded before the plain-text body is returned so billing is unaffected, dropping the provider `id` matches issue #608 and nothing in the attestation/signature flow keys off it, and the wiremock tests verify the actual outbound repeated `timestamp_granularities[]` multipart fields on both adapter paths. One issue should be addressed before merge, though.

**Important**

- `crates/api/src/models.rs:603` (`AUDIO_TRANSCRIPTION_LANGUAGE_CODES`) — the allowlist is the full ISO 639-1 set (~185 codes), not Whisper's ~100 supported languages. Codes like `aa`, `kj`, or `nv` pass validation and are forwarded unchanged, reproducing exactly the misleading upstream-failure pattern this PR is meant to prevent (openai-whisper's `LANGUAGES` map in `whisper/tokenizer.py` does not include them). Conversely, full language names (e.g. `english`), which openai-whisper itself accepts, are now rejected with 400 — a contract tightening worth a deliberate decision. Please restrict the list to Whisper's actual language-token set (mapping compatibility aliases before dispatch, as already done for `yue`/`haw`/`jw`) and add tests for unsupported ISO codes like `aa`.

**Minor**

- `crates/api/src/routes/completions.rs:3383` — the utoipa annotation still documents the 200 response only as JSON `body = AudioTranscriptionResponse`, but `response_format=text` now returns `text/plain; charset=utf-8` (line 3667). Add a `text/plain` 200 variant so OpenAPI consumers and generated clients don't expect JSON for text responses.
- `crates/api/tests/e2e_all/audio_transcriptions.rs:506` (`test_audio_transcription_word_timestamps_supported`) — the mock provider (`crates/inference_providers/src/mock.rs:1414`) returns `words` unconditionally, so this test passes even if `timestamp_granularities` is never forwarded; it would have passed before the fix too. The real coverage lives in the two wiremock adapter tests — consider making the mock honor `timestamp_granularities`/`response_format` so the e2e test actually exercises the contract.
- `crates/api/src/routes/completions.rs:3600` (pre-existing, now also inherited by the `text` path) — billing uses `response.duration`, but the OpenAI `json` response shape contains only `text`; an OpenAI-compatible backend that omits `duration` bills 0 seconds. Not a regression, but worth a follow-up since the `text` fix deliberately routes through `json`.

Rejecting `srt`/`vtt` with 400 diverges from the OpenAI spec but is a deliberate, documented choice matching the issue's suggested fix and strictly better than the previous retryable 502 — no change requested.

**Checks run locally:** `cargo fmt --all -- --check` (clean); `cargo clippy --locked -p api -p inference_providers --all-targets -- -D warnings` (clean); `cargo test --locked -p api --lib audio_transcription` (9/9); `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields` (2/2); `cargo test --locked --lib --bins` (356 passed, 1 ignored — pre-existing DB-dependent test). DB-backed e2e tests could not run locally (no usable PostgreSQL/migration state); GitHub CI (Lint, Test Suite, security_audit) is green and covers them.

lloydmak99:
Addressed Pierre's latest review on `8387a85`:

- Replaced the broad ISO-639-1 language table with Whisper's upstream supported language codes, while keeping the existing compatibility aliases (`yue`, `haw`, `jw`) and case/region normalization.
- Added regression coverage for valid ISO codes that Whisper does not support (`aa`, `kj`, `nv`) so these are rejected at the API boundary instead of reaching the provider.
- Updated the audio transcription OpenAPI response annotation to document both `application/json` and `text/plain` 200 responses.
- Tightened the mock transcription provider so `words` only appear when `response_format=verbose_json` and `timestamp_granularities[]` includes `word`; segment timing remains the default verbose JSON timing output.

Validation:

- `cargo test --locked -p api --lib audio_transcription`
- `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields`
- `git diff --check`

I also re-reviewed the PR diff after the patch. The latest requested changes are addressed, so I'm re-requesting review.

lloydmak99:
Pushed `85cdc1e` to fix the failed GitHub Lint check.

Change:

- Replaced the mock provider's `map_or(true, ...)` with `is_none_or(...)` per the CI clippy failure.

Validation:

- `cargo test --locked -p api --lib audio_transcription`
- `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields`
- `git diff --check`

No behavior change from the review fix; this only clears the CI lint failure.

PierreLeGuen (commented):
The implementation looks solid: validation happens before dispatch, the text/plain conversion runs only after usage is recorded, the provider→API conversion correctly strips the internal `id` (the only response-shape change), and `wiremock` is properly a dev-dependency. The findings below are non-blocking observations.

**Findings**

- `crates/api/src/routes/completions.rs:3563` — For `response_format=text` the provider is asked for `json`, and billing then reads `response.duration.unwrap_or(0.0)` (completions.rs:3602). OpenAI-style `json` transcription bodies typically contain only `{"text": ...}` with no `duration`, so text-format requests may be billed as 0 seconds. Not a regression (plain `json` already behaves this way, and `text` previously failed outright), but since the forwarded format is already rewritten here, requesting `verbose_json` instead would guarantee a real `duration` for duration-based billing while still returning only `response.text` to the client.
- `crates/api/src/models.rs:601` — `language` is now validated against a hardcoded Whisper-large-v3 code list, applied model-agnostically at the API boundary. This rejects full language names (`english`) and ISO-639-3 codes (`eng`) that some OpenAI-compatible backends accept, and would falsely constrain a future non-Whisper transcription model. Aligned with vLLM today, so acceptable — but consider scoping the check per model if other backends are added. Relatedly, region/script subtags are silently dropped before forwarding (completions.rs:3571: `en-US` → `en`, `zh-TW` → `zh`); for Chinese that loses a distinction some providers honor.
- `crates/api/src/routes/completions.rs:3472` — The accumulation fix for repeated `timestamp_granularities[]` fields is correct, but comma-splitting a single value (`"word,segment"` → two granularities) is a lenient divergence from OpenAI, which would treat that as one invalid value. Pre-existing and fine to keep, but worth documenting the accepted parameter forms in the OpenAPI schema, since mixing both forms merges values.
- API contract note — Rejecting `srt`/`vtt` with a 400 is an intentional divergence from OpenAI compatibility (these previously surfaced as retryable 5xx). The "not currently supported" error message is the right shape; just make sure public API docs/changelog mention it.

**Checks run**

- `cargo fmt --all -- --check` and `git diff --check` — clean.
- `cargo clippy --locked -p api -p inference_providers --all-targets` — no warnings.
- `cargo test --locked -p api --lib audio_transcription` — 9/9 unit tests pass (validation, language normalization, sorted-list invariant, `id` omission).
- `cargo test --locked -p inference_providers audio_transcription` — 2/2 wiremock adapter tests pass (repeated `timestamp_granularities[]` multipart fields for both Fleet and OpenAI-compatible adapters).
- `cargo test --locked -p api --test e2e_all audio_transcription` against a fresh test database — 18/18 e2e tests pass, including the new text/plain, srt/vtt-rejection, word-timestamp, granularity-validation, and invalid-language tests. (A run against the shared e2e database failed for an environmental reason — stale V58 migration history from another branch — unrelated to this PR.)
- CI: Lint, security_audit, and Test Suite all passed.

lloydmak99:
Follow-up after the latest review:

- Changed `response_format=text` forwarding from provider `json` to provider `verbose_json`, then still return only `response.text` as `text/plain` at the API boundary.
- This keeps provider parsing stable while preserving `duration` for usage/billing before the plain-text response is sent.

Validation:

- `cargo test --locked -p api --lib audio_transcription`
- `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields`
- `git diff --check`

I also tried `cargo test --locked -p api --test e2e_all audio_transcription_text_response_format_returns_plain_text`, but local Postgres was not running (`Connection refused` while bootstrapping the admin database), so that e2e check was environment-blocked here.

PierreLeGuen (commented):
This change is solid: `response_format=text` now returns real `text/plain` (while requesting `verbose_json` internally so duration-based billing keeps working), `srt`/`vtt` fail fast with a 400 instead of surfacing as provider 5xx, `timestamp_granularities[]` is validated and accumulated across repeated multipart fields, both adapters forward it as repeated fields on the wire, language codes are validated against the Whisper set with case/region normalization, and the internal `id` no longer leaks into responses. I traced the handler, service, and both provider adapters; the usage-recording path is unchanged and the dropped provider `id` is unused in the service layer. The wiremock assertions on the actual multipart wire format are a nice touch.

Non-blocking observations:

- `crates/api/src/routes/completions.rs:3562` — the `text` → `verbose_json` mapping fixes zero-duration billing for `text`, but `response_format=json` and the default are still forwarded as-is. An OpenAI-style plain-JSON provider response (`{"text": ...}`) has no `duration`, so `duration_seconds` at `completions.rs:3604` falls back to 0 and the request bills as 0 seconds. Pre-existing, not introduced here, but worth a follow-up applying the same provider-side mapping (or a provider usage field) to the `json`/default path.
- `crates/api/src/models.rs:496` — the OpenAPI schema names the field `timestamp_granularities`, while OpenAI clients (and the new e2e tests) send `timestamp_granularities[]`. The handler accepts both, so this is docs-only; consider naming the schema property to match real client traffic.
- `crates/api/tests/e2e_all/audio_transcriptions.rs:495` — `test_audio_transcription_word_timestamps_supported` runs against the mock provider, which returns `words` unconditionally (`crates/inference_providers/src/mock.rs:1414`), so it would pass even without the forwarding fix. The real coverage lives in the wiremock adapter tests, which is fine — a short comment marking this as a smoke test would prevent it from being mistaken for contract coverage.
- `crates/api/src/routes/completions.rs:3475` — values are also comma-split (`"word,segment"` parses as two granularities), which is more lenient than OpenAI's one-granularity-per-field multipart semantics. Not wrong, just be aware a value containing a literal comma can never reach validation intact.
- Worth a release-note line: previously-accepted inputs now return 400 (`srt`/`vtt` formats; non-Whisper language values such as `english`, `nb`, `iw`, with region subtags like `en-US` silently truncated to `en`). Intentional and strictly better than the old misleading 5xx, but clients relying on the old leniency will start failing fast.

On the language allowlist itself: I checked `crates/api/src/models.rs:657` against Whisper's supported set — it matches (including the `yue`/`haw`/`jw` additions, correctly sorted), and `test_audio_transcription_rejects_invalid_language` confirms codes outside that set (e.g. `nb`, `iw`) are rejected, so the validation does not re-create the upstream-failure problem for unsupported ISO codes.

Checks run locally: `cargo fmt --all -- --check` and `git diff --check` clean; `cargo clippy --locked -p api -p inference_providers --all-targets` clean; `cargo test --locked -p api --lib audio_transcription` 9/9; `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields` 2/2; `cargo test --locked -p api --test e2e_all audio_transcription` 18/18 against a fresh local PostgreSQL database mirroring CI's env. GitHub CI (Lint, Test Suite, security_audit) all green on the PR head.

lloydmak99:
Follow-up commit `880102b` addresses the remaining actionable notes from Pierre's latest non-blocking review:

- Documented the accepted multipart shapes for `timestamp_granularities`: repeated `timestamp_granularities[]` fields, repeated `timestamp_granularities` fields, and comma-separated values.
- Clarified that `test_audio_transcription_word_timestamps_supported` is an API-boundary smoke test; the real provider multipart forwarding contract remains covered by the `inference_providers` wire tests.

I also checked the review threads through GraphQL: all threads are resolved. The PR still shows `CHANGES_REQUESTED` because earlier review decisions remain until reviewers re-review the latest head.

Local validation for this follow-up:

- `git diff --check`

GitHub CI has started on `880102b`; it was still running when I posted this.

lloydmak99:
CI update for `880102b`: all required checks are green.

- `security_audit`: passed
- `Lint`: passed
- `Test Suite`: passed

Review requests are still active for `Evrard-Nil` and `PierreLeGuen`.

PierreLeGuen (commented):
The change is well-scoped and well-tested. Validation tightening matches the PR description; the `text` format path preserves duration-based billing by requesting `verbose_json` upstream and converting at the boundary after usage is recorded; the repeated-multipart-field fix (`extend` instead of overwrite) is verified with wire-level wiremock tests against both adapters (the Chutes provider delegates to the same path, so it's covered); and the `From` conversion correctly drops the provider-internal `id` field, with a unit test pinning it. The language allowlist is sorted (so `binary_search` is safe, enforced by a test), validation errors return 400 without logging user-supplied values, and the OpenAPI spec now documents both `application/json` and `text/plain` response content types. No security or privacy concerns.

A few non-blocking observations, mostly forward-looking:

- `crates/api/src/models.rs:660` — the language allowlist hardcodes Whisper's language set into the model-agnostic `/v1/audio/transcriptions` endpoint. A future non-Whisper transcription model with broader coverage would have valid requests rejected with 400 at the API boundary, and some real-world ISO-639-1 codes Whisper doesn't use (e.g. `nb` vs Whisper's `no`) are rejected rather than mapped. Consider mapping common synonyms (`nb`→`no`, `iw`→`he`) in `normalize_audio_transcription_language`, and treating per-model language validation as a follow-up if such models land.
- `crates/api/src/models.rs:670`, `crates/api/src/routes/completions.rs:3574` — region subtags are silently stripped before forwarding (`en-US`→`en`). Reasonable for Whisper, but it's an undocumented transformation of client input; worth a note in the `AudioTranscriptionRequestSchema.language` doc comment.
- `crates/api/src/routes/completions.rs:3562` — implementing `response_format=text` via upstream `verbose_json` means any future backend without `verbose_json` support would fail `text` requests it could otherwise serve. Fine today (all adapters target Whisper-compatible servers) and the billing rationale is documented in a comment; just an operational note.

Checks run locally: `cargo fmt --all -- --check`, `git diff --check`, `cargo clippy --locked -p api -p inference_providers --all-targets -- -D warnings`, `cargo test --locked -p api --lib audio_transcription` (9 passed), and `cargo test --locked -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields` (2 passed) — all clean. DB-backed e2e tests were skipped locally (no Postgres); CI's Test Suite job runs the `e2e_all` suite against Postgres and passed on the PR head, along with Lint and security_audit.

PierreLeGuen (commented):
This change is in good shape and I have no blocking concerns. The validation tightening at the API boundary, the `text` → `verbose_json` provider rewrite (which preserves duration-based billing, per the inline comment), the repeated `timestamp_granularities[]` multipart fields (verified at the wire level for both the Fleet and OpenAI-compatible adapters), and the conversion layer that stops leaking the provider-internal `id` all check out. Rejecting `timestamp_granularities` without `verbose_json` with a 400 matches OpenAI's documented behavior, and the sorted language table is enforced by a unit test so the `binary_search` stays correct.

Two minor, non-blocking observations:

- `crates/api/src/models.rs:668` (`is_supported_audio_transcription_language`): the new validation only accepts ISO-639-1-style codes (plus case/region variants), so clients sending Whisper-accepted full language names like `english` — which previously passed through and generally worked on vLLM Whisper backends — will now get a 400. ISO-639-1 is the documented contract, so this is defensible, but it's a client-visible tightening. Consider a small alias map from full names to codes, or confirm from logs that no production traffic sends full names and call it out in release notes.
- `crates/api/src/models.rs:617`: rejecting `srt`/`vtt` with a 400 narrows the OpenAI-compatible surface rather than implementing subtitle output. It's a clear improvement over the prior retryable 5xx, and the PR documents it as intentional, but clients requesting these formats will see the error class change from 5xx to 4xx, which can alter retry behavior. Worth a changelog note; rendering subtitles from `verbose_json` segments later is the right follow-up.

Local checks run: `cargo fmt --all -- --check` (clean); `cargo clippy --locked -p api -p inference_providers --all-targets -- -D warnings` (clean); `cargo test --locked -p api --lib audio_transcription` (9/9 passed — validation, language normalization, sorted-table invariant, provider-`id` omission); `cargo test --locked -p inference_providers audio_transcription` (2/2 wiremock adapter tests passed). The database-backed e2e tests weren't runnable locally (no working Postgres setup in the review environment), but CI's Test Suite runs them against Postgres and is green on the PR head, along with Lint and security_audit.

lloydmak99:
Addressed Pierre's latest observations in `bb0376d`:

- Added Whisper full-language-name aliases so requests like `language=english`, `language=FRENCH`, `language=haitian creole`, `language=cantonese`, and `language=burmese` validate and are canonicalized to provider language codes before dispatch.
- Kept the existing code/region normalization path (`EN`, `en-US`, `fr_CA`) intact.
- Added alias-table tests so language aliases stay sorted and only point at supported Whisper codes.
- Updated the PR description with an explicit release/compatibility note that `response_format=srt` and `response_format=vtt` now return `400 invalid_request_error` instead of retryable provider failures, and corrected the `text` forwarding note to `verbose_json`.

Validation:

- `cargo test --locked -p api --lib audio_transcription`
- `git diff --check`

PierreLeGuen (changes requested):
The substance of the fix is in good shape — earlier concerns (language allowlist, `word` granularity, adapter forwarding) were resolved in prior commits, and the new alias map is sorted, test-enforced, and correctly normalized. However, the latest commit `bb0376d` introduces two deterministic CI failures (confirmed red on the head commit's Lint and Test Suite jobs), each a one-line fix:

- `crates/api/src/models.rs:803` — `entry.0.cmp(&alias_candidate.as_str())` trips `clippy::needless_borrow` under CI's `-D warnings`, failing the Lint job. Fix: `entry.0.cmp(alias_candidate.as_str())`.
- `crates/api/tests/e2e_all/audio_transcriptions.rs:617-620` — `test_audio_transcription_invalid_language` still asserts the old error substring "supported audio transcription language code"; the message was changed to "…supported Whisper language code or name…" and only the unit test was updated, so this e2e test fails deterministically in CI. Fix: update the assertion to match the new message.

Local checks: `cargo fmt --all -- --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` reproduces the `needless_borrow` error on the committed code (clean once fixed); `cargo test -p api --lib audio_transcription` 10/10 passed; `cargo test -p inference_providers audio_transcription_sends_repeated_timestamp_granularity_fields` 2/2 passed; DB-backed e2e tests skipped locally (no Postgres) — the e2e failure is verified from the CI logs on the head commit.

lloydmak99:
CI follow-up is complete on latest head `02150a2`:

- `ab546e8` fixed the Clippy lint from the alias binary search.
- `02150a2` updated the stale e2e invalid-language assertion to the widened `code or name` error message.

Latest GitHub checks are green:

- `Lint`: passed
- `Test Suite`: passed
- `security_audit`: passed

PierreLeGuen (approved):
This fixes Whisper transcription validation at the `/v1/audio/transcriptions` boundary: `text` format now returns real `text/plain` (still requesting `verbose_json` upstream so duration-based billing is preserved), `srt`/`vtt` fail fast with 400, `timestamp_granularities[]` is validated and forwarded as repeated multipart fields by both adapters with wire-level test coverage, `language` is validated against the Whisper set with normalization and aliases, and the provider-internal `id` is stripped from responses. All concerns from prior review rounds — including the clippy `needless_borrow` lint and the stale e2e assertion (`crates/api/tests/e2e_all/audio_transcriptions.rs:620`) — are resolved at the head commit. No issues found; validation errors echo user input only in the response body, never in logs.

Checks: `cargo fmt --check`, `git diff --check`, and `cargo clippy --locked -p api -p inference_providers --all-targets -- -D warnings` clean; `cargo test -p api --lib audio_transcription` and the `inference_providers` wiremock granularity tests pass locally; DB-backed e2e covered by green CI (Lint, Test Suite, security_audit) on `02150a2`.

PR #765, opened 2026-06-11
feat(admin): scheduled model pricing changes with notifications and auto-apply

PierreLeGuen:
## Summary

Admins can now **schedule** pricing changes for one or more models (each with its own effective date), **notify** affected users by email at confirm time, and have the new pricing **apply automatically** when each effective date is reached.

### New endpoints (admin auth)

| Method | Path | Purpose |
|---|---|---|
| POST | `/v1/admin/models/pricing-changes/preview` | Recipient/org counts + per-model old→new breakdown, no mutation |
| POST | `/v1/admin/models/pricing-changes/confirm` | Persist schedule + send notification emails (idempotent per client-supplied `batchId`) |
| GET | `/v1/admin/models/pricing-changes?status=` | List scheduled changes |
| DELETE | `/v1/admin/models/pricing-changes/{id}` | Cancel a pending change |

### Notifications

Mirrors the deprecation flow (Resend sender, per-(user, org) delivery audit table), with one key difference to **minimize email volume**: each recipient gets exactly **one email per batch**, listing only the models their org(s) used in the last 30 days, each with its own effective date and old → new pricing for the **changed fields only**. Recipients are owner/admin members of orgs with recent usage (aliases resolved).

### Auto-apply

New `ModelPricingScheduler` background task (env `PRICING_CHANGE_APPLY_INTERVAL_SECS`, default 60s, `0` disables; wired into coordinated shutdown):

- Claims due rows with an atomic `UPDATE … FOR UPDATE SKIP LOCKED`, so **multiple API instances behind the LB never double-apply** (covered by an e2e test racing two schedulers)
- Applies via the existing `upsert_model_pricing` → `model_history` audit entry (`"Scheduled pricing change (batch …)"`) + models cache invalidation
- Bounded retries (5 attempts) then parks as `failed` with the error recorded; stale `applying` claims from crashed instances are recovered automatically

### Guard rails

- Pricing snapshot captured at confirm time (the "old" price shown in emails)
- Effective date must be ≥ 1 hour out (users are notified before the switch)
- At most **one open change per model**, enforced by a partial unique index → 409 on conflict (explicit cancel + reschedule, no silent supersede of announced pricing)
- Max 50 models per batch; pricing fields only (input/output/cache-read per token, per image)

### Schema

- `V0058` — `scheduled_model_pricing_changes` (new/old pricing, effective_at, status lifecycle pending→applying→applied/cancelled/failed, batch grouping, audit columns)
- `V0059` — `model_pricing_change_email_deliveries` (batch-scoped delivery audit, unique per (batch, user, org))

## Testing

- `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean
- New unit tests for the consolidated email renderer (multi-model, changed-fields-only, HTML escaping, Resend payload)
- New e2e suite `admin_pricing_changes.rs` (7 tests): schedule/list, validation, 409 conflict + idempotent same-batch retry, cancel lifecycle, scheduler apply (pricing switch + history audit), two schedulers racing over 5 due rows (each applied exactly once), recipient consolidation (one delivery row covering both used models)
- Full `--lib --bins` suite and all existing admin e2e tests pass

Companion UI PR: nearai/nearai-cloud-admin-ui (batch pricing editor + pending-changes view).


PierreLeGuen:
Addressed the automated review feedback in 291bb06:

- **Concurrent same-batch idempotency**: the existence check now runs inside the insert transaction behind a transaction-scoped advisory lock keyed on `batchId`, so two concurrent confirms of the same batch serialize and both return 200 with the same rows (previously the loser hit the open-change unique index and got a spurious 409). Added an e2e test firing the same confirm twice concurrently.
- **Pricing drift visibility**: the scheduler now logs a warning (IDs only) when the live pricing differs from the snapshot captured at confirm time (i.e. a direct `PATCH /v1/admin/models` edit happened while the change was pending). The announced change remains the source of truth and is applied regardless.

The sync-email-in-handler and duplicate-history-on-crash-recovery observations are left as-is per the review (consistent with the existing deprecation flow / value-idempotent).

PierreLeGuen:
Addressed the Codex review findings in 291bb06 and 48bd92f:

- **Concurrent same-batch confirms** → existence check now runs inside the insert transaction behind a batch-keyed advisory lock; both racing confirms return 200 with the same rows (e2e-covered).
- **Reused batchId with different changes** → confirm now verifies the request body against the persisted batch and returns 409 on mismatch; emails are built from the persisted rows, never the request (e2e-covered).
- **Retry blocked by lead-time validation** → confirm loads the existing batch first and skips new-schedule validation for retries, so resuming email delivery near/past the effective date works (e2e-covered).
- **Duplicate model_history on crash-recovery re-apply** → the scheduler now detects pricing already at the scheduled values and marks the row applied without re-running the upsert (e2e-covered).
- **Non-USD currency silently billed as USD** → the endpoints now reject any currency other than USD with a 400 (e2e-covered).

Pricing-change e2e suite is now 11 tests; full admin suite + units green, clippy -D warnings clean.

Evrard-Nil (approved):
Reviewed locally (fresh clone, full build + e2e against a local Postgres). **LGTM — approving.** This is a defensively-engineered, billing-and-admin-sensitive feature and it holds up to a paranoid pass. One cross-PR coordination flag below (not a code defect).

**Verified:**
- **AuthZ** — all 4 new routes sit on the admin Router behind `admin_middleware` (`lib.rs:1699-1703`), every handler takes `Extension<AdminUser>`; a normal API key can't reach them.
- **Auto-apply is multi-replica-safe & idempotent** — DB-polled worker claims due rows via `UPDATE … WHERE id IN (SELECT … WHERE status='pending' AND effective_at<=NOW() FOR UPDATE SKIP LOCKED)`, `mark_applied` guarded on `status='applying'`, plus a value-idempotent re-apply guard so a crash between upsert and mark doesn't write a duplicate `model_history` row. Proven by e2e `test_concurrent_schedulers_apply_each_change_once`. TIMESTAMPTZ + `NOW()` (UTC), 5-retry→`failed`, stale-`applying` recovery after 600s. Persisted across restart.
- **Pricing correctness** — applies through the existing `upsert_model_pricing` (same path as `PATCH /v1/admin/models`), so `models` + `model_history` invariant is maintained and the nano-dollar units match the live billing read path.
- **Notifications** (Resend) are decoupled from apply (best-effort; a notify failure never rolls back the price), HTML-escaped, no secrets logged, no unwrap/panic in the prod path.
- **Validation** — past/<1h-lead rejected, negative prices rejected at route + service, nonexistent/inactive model → 404, one-open-change-per-model → 409 (partial unique index), batch dup/empty/>50/non-USD all 400.
- `cargo test -p services --lib` 347/0, `cargo test -p api --test e2e_all admin_pricing_changes` 11/0 (incl. the concurrency races), fmt + clippy clean.

**⚠️ Cross-PR migration collision (please coordinate — not a defect in this PR):** this PR adds `V0058__…pricing…` (+ `V0059`), but the already-approved #763 also adds `V0058__backfill_qwen35_reasoning_feature.sql`. Both are correct relative to `main` today, but whichever merges **second** will have a duplicate `V0058` and refinery will refuse to load (it requires unique versions). Whichever lands second should renumber to `V0060`(+`V0061`). Flagging on #763 as well.

**Minor/notes:** `already_at_target` guard skips a `model_history` entry when a scheduled price equals the current price (correct no-op, but worth knowing for audit completeness); consolidated email uses canonical `model_id` for the code label (cosmetic). Pre-existing (not this PR): `admin_middleware` debug-logs the raw bearer token — worth a separate cleanup.

PierreLeGuen:
Thanks for the thorough review! Re the V0058 collision: this PR keeps **V0058/V0059** and merges first; #763 will need to renumber its migration before merging (briefly tried renumbering ours in fd83fea, reverted in 3f671cb). Pricing e2e suite re-verified after the revert. Merging once CI is green.

PierreLeGuen:
**Final status: merged** as `79b8a57e`.

- CI (Lint, Test Suite, security_audit) green on every revision; pricing-change e2e suite grew to 11 tests over the course of review.
- All automated review findings addressed: transaction-scoped advisory lock makes concurrent same-batch confirms idempotent; reused batchIds with different content are rejected (409) and emails are built from persisted rows only; retries of a persisted batch skip new-schedule lead-time validation; crash-recovery re-apply is value-idempotent (no duplicate model_history); non-USD currencies rejected; pricing-drift warning at apply time.
- Approved by @Evrard-Nil after independent local build + e2e verification.
- Migration numbering: this PR keeps **V0058/V0059**; #763 needs to renumber its V0058 to V0060 before merging (noted there).
- Verified end-to-end against a live local server: schedule → consolidated notification audit → automatic apply by the background scheduler → model_history audit; cancel and all guard rails exercised over HTTP.

Rollout: staging first per repo workflow. Migrations run on boot; the scheduler starts on every instance (`PRICING_CHANGE_APPLY_INTERVAL_SECS`, default 60s, multi-instance-safe via FOR UPDATE SKIP LOCKED). Emails send only where Resend is configured.

PR #768, opened 2026-06-11
feat(chutes): tiered NEAR-primary/Chutes-fallback across all TEE models

Evrard-Nil:
Builds on #758 (Chutes as an attested provider). Turns Chutes from a single pinned provider (registered under its raw `-TEE` slug) into a **fallback fabric across all Chutes TEE models**, under **canonical** ids, with **NEAR-primary** routing and a hard **"verifiable models never serve plaintext"** rule.

## Requirements (from product)
1. Register **all** Chutes TEE models, not one.
2. List them under **canonical** ids — the NEAR-served id when NEAR also serves the model (`zai-org/GLM-5.1-FP8`, **not** `…-TEE`), else the OpenRouter id.
3. Route **NEAR first, Chutes fallback** when NEAR serves the model; **Chutes-primary** when NEAR doesn't.
4. **Do not fall back to a non-attested provider for verifiable models.**
5. (Streaming) enable E2EE streaming — gated on a staging check (below).

### 1. Canonical id vs chute slug — decouple them
The Chutes provider used `model_name` as **both** the catalog/public id **and** the upstream `model` it sends to Chutes. We split that:
- **`CHUTES_MODELS`** is now comma-separated **`canonical_id=chute_slug`** pairs (a bare entry → `canonical == slug`), parsed into `Vec<ChutesModelEntry>`.
  - Example: `CHUTES_MODELS=zai-org/GLM-5.1-FP8=zai-org/GLM-5.1-TEE,moonshotai/Kimi-K2.6-TEE`
  - `zai-org/GLM-5.1-FP8` (NEAR-served) is the public/route id; `zai-org/GLM-5.1-TEE` is the upstream chute slug; `moonshotai/Kimi-K2.6-TEE` (Chutes-only) is canonical==slug for now (swap to its OpenRouter id when known).
- The provider is still built with the **chute slug** (`request_body` pins it + `cached_chute_id` resolves it upstream), but seeded/registered under the **canonical id**. So a client calling `zai-org/GLM-5.1-FP8` reaches Chutes with the right `-TEE` slug.

### 2. Wire `ProviderTier` into selection
`ProviderTier{Near, Attested3p, NonAttested}` existed but carried no behavior. Added `fn tier(&self)` to the `InferenceProvider` trait (default `NonAttested`); `nearai::Provider → Near`, `chutes::Provider → Attested3p`.

### 3. Coexistence — Chutes as a *secondary* provider (highest-risk change)
`register_pinned_provider` **overwrote** a model's provider list, and the discovery guards made a pinned name **mutually exclusive** with DB discovery — so NEAR + Chutes couldn't share a canonical id. New `register_pinned_secondary_provider` **pushes** instead, and records the provider in a `pinned_providers` map (the source of truth for the pinned tier). The refresh merge now rebuilds each model's list as **`[fresh discovered NEAR..] + [re-attached pinned Chutes]`** — discovery refreshes/adds the NEAR side but can **never drop or overwrite** the per-request-verified Chutes provider (the security invariant from #758 is preserved: discovery can only ADD an attested provider, never substitute it).

### 4. Tier-ordered selection = NEAR-primary / Chutes-fallback (per-request)
`get_providers_with_fallback` stable-sorts by tier (`Near < Attested3p < NonAttested`) **within each health group**. The existing `retry_with_fallback` then walks that ordered list, so a retryable NEAR failure (5xx/429/connection) falls through to Chutes **within the same request** (a non-retryable client 4xx still fast-returns without spending Chutes quota). A Chutes-only id has a single provider, so it's primary automatically — no special case. NEAR demotion (10 consecutive failures) already moves it below Chutes.

### 5. Verifiable models never fall back to plaintext
If a model has **any** attested provider (`Near` or `Attested3p`), **all `NonAttested` providers are dropped from selection**. A verifiable request thus only ever tries attested providers and **fails closed** rather than silently downgrading to a plaintext backend — defending the case where a Chutes-only canonical id also has a stray external (OpenAI/OpenRouter) row. Pure non-attested models (no attested provider) are unaffected.

### 6. Catalog seeding under the canonical id
Seed/register under the canonical id. A NEAR-served canonical id (e.g. `zai-org/GLM-5.1-FP8`) already has an active, attested row → it's respected verbatim and just gains Chutes as a fallback provider. A Chutes-only canonical id is auto-seeded **inactive / $0** (operator activates + prices via `PATCH /v1/admin/models`, unchanged from #758).

## Streaming
Decoder already implements the exact protocol (single `e2e_init` ML-KEM encapsulation → one HKDF stream key → per-chunk ChaCha20-Poly1305). It stays gated behind `CHUTES_ENABLE_STREAMING` pending a **staging verification** that Chutes emits the `[DONE]` terminator **inside** the encrypted channel (our truncation guard depends on it). Per-chunk AEAD authenticates every byte; frame **reorder** remains undetectable (no sequence numbers) — accepted residual, truncation is caught. No code change needed to flip it on.

## Tests
- **config**: `canonical=slug` pair parsing (+ bare entries, whitespace, dropped empties); empty when unset.
- **pool**: `verifiable_model_prefers_near_then_chutes_and_never_plaintext`; `chutes_only_model_serves_as_single_attested_primary`; `non_verifiable_model_is_served_by_its_non_attested_provider`; `pinned_secondary_chutes_coexists_with_db_near_provider`. Existing `pinned_provider_survives_refresh` still green.
- 297 inference_providers + config tests, 57 pool tests, clippy + fmt clean.

## Follow-ups
- Flip `CHUTES_ENABLE_STREAMING` after the staging inner-`[DONE]` check.
- Supersedes the standalone `zai-org/GLM-5.1-TEE` staging row activated for #758 verification — that should be deactivated; Chutes attaches under `zai-org/GLM-5.1-FP8`.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -796,10 +796,14 @@ pub async fn init_inference_providers(
                     services::attestation::chutes::vetted_golden_measurements(),
                     pccs_url,
                 ));
-                for model in &config.external_providers.chutes_models {
+                for entry in &config.external_providers.chutes_models {
+                    // The provider talks to Chutes with the chute SLUG (request_body
+                    // pins it + cached_chute_id resolves it); we expose/route under
+                    // the CANONICAL id (the NEAR-served id when NEAR also serves the
+                    // model, else the OpenRouter id) — never the raw `-TEE` slug.
                     let cfg = inference_providers::attested::chutes::Config::new(
                         api_key.clone(),
-                        model.clone(),
+                        entry.chute_slug.clone(),
                         config.external_providers.timeout_seconds,
                     )
                     .with_streaming(allow_streaming);
@@ -808,18 +812,34 @@ pub async fn init_inference_providers(
                         verifier.clone(),
                     ) {
                         Ok(provider) => {
-                            // Ensure a catalog row exists so the data plane resolves
-                            // the model (and usage can be billed against a real id).
-                            ensure_chutes_catalog_row(&models_repo, model).await;
-                            // Pinned: registered out-of-band (not DB discovery), so
-                            // excluded from the refresh's stale-removal; also skips the
-                            // signing-key discovery Chutes never satisfies.
-                            pool.register_pinned_provider(model.clone(), Arc::new(provider))
-                                .await;
-                            tracing::info!(model = %model, "Registered Chutes attested provider");
+                            // Ensure a catalog row exists under the canonical id so the
+                            // data plane resolves the model (and usage bills against a
+                            // real id). If NEAR already serves this id, its row is left
+                            // untouched and we just add Chutes as a fallback provider.
+                            ensure_chutes_catalog_row(&models_repo, &entry.canonical_id).await;
+                            // Pinned SECONDARY: pushed onto the canonical id's provider
+                            // list (coexists with NEAR's own providers) and excluded from
+                            // discovery's stale-removal/overwrite. Tier ordering puts NEAR
+                            // first and Chutes as fallback; a Chutes-only id has just this
+                            // provider, so it serves as primary.
+                            pool.register_pinned_secondary_provider(
+                                entry.canonical_id.clone(),
+                                Arc::new(provider),
+                            )
+                            .await;
+                            tracing::info!(
+                                canonical = %entry.canonical_id,
+                                chute_slug = %entry.chute_slug,
+                                "Registered Chutes attested provider (fallback tier)"
+                            );
                         }
                         Err(e) => {
-                            tracing::warn!(model = %model, error = %e, "Failed to build Chutes provider");
+                            tracing::warn!(
+                                canonical = %entry.canonical_id,
+                                chute_slug = %entry.chute_slug,
+                                error = %e,
+                                "Failed to build Chutes provider"
+                            );
                         }
                     }
                 }
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -830,6 +830,7 @@ pub async fn init_inference_providers(
                         entry.chute_slug.clone(),
                         config.external_providers.timeout_seconds,
                     )
+                    .with_canonical_id(entry.canonical_id.clone())
                     .with_streaming(allow_streaming);
                     match inference_providers::attested::chutes::Provider::new(
                         cfg,
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -745,6 +745,30 @@ pub async fn init_inference_providers(
     let models_source =
         models_repo.clone() as Arc<dyn services::inference_provider_pool::ExternalModelsSource>;
 
+    // Fail-closed reservation (MUST run before external/discovery loads below):
+    // when Chutes is enabled, reserve EVERY configured canonical id as a pinned
+    // (verifiable) model up front — even before we try to build the providers. This
+    // guarantees a plaintext external/OpenRouter row sharing a canonical id can
+    // never register for it, even if the Chutes provider fails to build (missing
+    // key / construction error). A reserved id then serves only its attested
+    // provider(s) or fails closed (404); it can never silently serve plaintext for
+    // a model an operator configured as verifiable.
+    if config.external_providers.enable_chutes {
+        let canonical_ids: Vec<String> = config
+            .external_providers
+            .chutes_models
+            .iter()
+            .map(|e| e.canonical_id.clone())
+            .collect();
+        if !canonical_ids.is_empty() {
+            pool.reserve_pinned_models(&canonical_ids);
+            tracing::info!(
+                count = canonical_ids.len(),
+                "Reserved Chutes canonical ids as verifiable (fail-closed) before external load"
+            );
+        }
+    }
+
     // Load inference_url models (our own vLLM/SGLang backends)
     match models_source.fetch_inference_url_models().await {
         Ok(models) if !models.is_empty() => {
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -1255,12 +1255,24 @@ pub async fn delete_model(
             }
         })?;
 
-    // Unregister external provider if it was registered
-    // This is a no-op for vLLM models (they are discovered, not registered)
-    app_state
-        .inference_provider_pool
-        .unregister_provider(&model_name)
-        .await;
+    // Unregister external provider if it was registered.
+    // No-op for vLLM models (discovered, not registered). Skip pinned models
+    // (e.g. a NEAR-served canonical id with a Chutes fallback): unregistering would
+    // tear down the config-pinned provider too, recoverable only by restart. The
+    // catalog delete already 404s the model via resolve_and_get_model's is_active
+    // gate; config remains the source of truth for the pinned provider.
+    if app_state.inference_provider_pool.is_pinned(&model_name) {
+        tracing::warn!(
+            model = %model_name,
+            "delete_model: skipping unregister of a pinned (config-managed) provider; \
+             the catalog row is removed (model 404s), pinned provider left intact"
+        );
+    } else {
+        app_state
+            .inference_provider_pool
+            .unregister_provider(&model_name)
+            .await;
+    }
 
     Ok(StatusCode::NO_CONTENT)
 }
@@ -1342,10 +1354,22 @@ pub async fn deprecate_model(
     // resolver will route requests with the deprecated model_id to the
     // successor on subsequent calls — so this prevents in-flight resolution
     // from picking up a stale provider reference for the now-inactive model.
-    app_state
-        .inference_provider_pool
-        .unregister_provider(&model_id)
-        .await;
+    // Skip pinned models (e.g. a NEAR-served canonical id with a Chutes fallback):
+    // unregistering would tear down the config-pinned provider too (recoverable
+    // only by restart), and the alias already routes the deprecated id to the
+    // successor, so the pinned provider isn't reached.
+    if app_state.inference_provider_pool.is_pinned(&model_id) {
+        tracing::warn!(
+            model = %model_id,
+            "deprecate_model: skipping unregister of a pinned (config-managed) provider; \
+             the alias routes the deprecated id to its successor, pinned provider left intact"
+        );
+    } else {
+        app_state
+            .inference_provider_pool
+            .unregister_provider(&model_id)
+            .await;
+    }
 
     let to_api = |model_name: String, m: services::admin::ModelPricing| ModelWithPricing {
         model_id: model_name,
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -460,6 +460,15 @@ pub async fn batch_upsert_models(
         // model — would leave an active catalog row with no serving provider until
         // restart. Pricing/metadata still applied via batch_upsert_models above;
         // serving is gated by the catalog `is_active`, not by this in-memory entry.
+        //
+        // NOTE: with tiered fallback this now also covers a NEAR-served canonical id
+        // that has a Chutes fallback (it's `is_pinned`). So a PATCH that changes its
+        // `inference_url` or deactivates it skips the eager `unregister_provider`
+        // cleanup here: the NEW url is still re-registered below (the merge keeps
+        // Chutes), but the stale OLD-url NEAR provider + its pubkey/failure-counter
+        // entries linger until the next periodic refresh prunes them. Behavior stays
+        // safe (pubkey intersection + catalog `is_active` gating); it just isn't
+        // instantaneously clean for a runtime url change on a Chutes-fallback model.
         if app_state.inference_provider_pool.is_pinned(model_name) {
             continue;
         }
--- a/crates/api/tests/e2e_all/chutes_catalog.rs
+++ b/crates/api/tests/e2e_all/chutes_catalog.rs
@@ -19,7 +19,7 @@ async fn admin_patch_with_provider_type_keeps_pinned_provider() {
     let model = format!("zai-org/GLM-5.1-TEE-pin-{}", uuid::Uuid::new_v4());
 
     // Register a pinned provider exactly as init_inference_providers does for Chutes.
-    pool.register_pinned_provider(
+    pool.register_pinned_secondary_provider(
         model.clone(),
         Arc::new(inference_providers::mock::MockProvider::new()),
     )
--- a/crates/config/src/types.rs
+++ b/crates/config/src/types.rs
@@ -980,6 +980,59 @@ mod tests {
         std::env::remove_var("RESEND_API_KEY_FILE");
         std::env::remove_var("CLOUD_UI_BASE_URL");
     }
+
+    #[test]
+    #[serial]
+    fn chutes_models_parses_canonical_slug_pairs_and_bare_entries() {
+        // `canonical=chute_slug` pairs; a bare entry means canonical == slug.
+        // Surrounding whitespace is trimmed; empty/half-empty tokens dropped.
+        std::env::set_var(
+            "CHUTES_MODELS",
+            "zai-org/GLM-5.1-FP8=zai-org/GLM-5.1-TEE , moonshotai/Kimi-K2.6-TEE ,, =bad, alsobad=",
+        );
+        let cfg = ExternalProvidersConfig::from_env();
+        std::env::remove_var("CHUTES_MODELS");
+
+        assert_eq!(
+            cfg.chutes_models,
+            vec![
+                ChutesModelEntry {
+                    canonical_id: "zai-org/GLM-5.1-FP8".to_string(),
+                    chute_slug: "zai-org/GLM-5.1-TEE".to_string(),
+                },
+                // Bare entry: canonical == slug.
+                ChutesModelEntry {
+                    canonical_id: "moonshotai/Kimi-K2.6-TEE".to_string(),
+                    chute_slug: "moonshotai/Kimi-K2.6-TEE".to_string(),
+                },
+            ],
+            "pairs split on '='; bare => canonical==slug; empty/half-empty tokens dropped"
+        );
+    }
+
+    #[test]
+    #[serial]
+    fn chutes_models_empty_when_unset() {
+        std::env::remove_var("CHUTES_MODELS");
+        let cfg = ExternalProvidersConfig::from_env();
+        assert!(cfg.chutes_models.is_empty());
+    }
+}
+
+/// One Chutes model to register, parsed from a single `CHUTES_MODELS` token.
+///
+/// The token is `canonical_id=chute_slug` (e.g. `zai-org/GLM-5.1-FP8=zai-org/GLM-5.1-TEE`),
+/// or a bare `name` meaning `canonical_id == chute_slug`. We deliberately keep
+/// the two ids distinct: the **canonical id** is what we expose in `/v1/models`
+/// and route under (the NEAR-served id when NEAR also serves the model, else the
+/// OpenRouter id) — never the raw `-TEE` chute slug; the **chute slug** is the
+/// internal upstream identity we send to Chutes and resolve to a `chute_id`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChutesModelEntry {
+    /// User-facing / catalog model id (e.g. `zai-org/GLM-5.1-FP8`).
+    pub canonical_id: String,
+    /// Chutes chute slug sent upstream (e.g. `zai-org/GLM-5.1-TEE`).
+    pub chute_slug: String,
 }
 
 /// External providers configuration for third-party AI providers
@@ -1001,9 +1054,10 @@ pub struct ExternalProvidersConfig {
     pub enable_chutes: bool,
     /// Chutes API key (`cpk_...`), from `CHUTES_API_KEY[_FILE]`. A secret.
     pub chutes_api_key: Option<String>,
-    /// Comma-separated Chutes model ids to register (e.g. `zai-org/GLM-5.1-TEE`),
-    /// from `CHUTES_MODELS`.
-    pub chutes_models: Vec<String>,
+    /// Chutes models to register, from `CHUTES_MODELS` (comma-separated
+    /// `canonical_id=chute_slug` pairs; a bare entry means the two are equal).
+    /// See [`ChutesModelEntry`].
+    pub chutes_models: Vec<ChutesModelEntry>,
     /// Expose Chutes **streaming** as an attested path (`CHUTES_ENABLE_STREAMING`,
     /// default off). Off because Chutes' stream protocol has no authenticated
     /// frame sequence numbers, so an on-path gateway could drop/reorder frames
@@ -1078,11 +1132,24 @@ impl ExternalProvidersConfig {
         // An empty key is not a key — treat "" as absent so a misconfigured
         // secret can't pass as Some("") and silently fail at request time.
         .filter(|s| !s.is_empty());
+        // `canonical_id=chute_slug` per comma-separated token; a bare token means
+        // canonical_id == chute_slug. Drop tokens missing either side.
         let chutes_models = env::var("CHUTES_MODELS")
             .unwrap_or_default()
             .split(',')
-            .map(|s| s.trim().to_string())
+            .map(|s| s.trim())
             .filter(|s| !s.is_empty())
+            .map(|tok| match tok.split_once('=') {
+                Some((canonical, slug)) => ChutesModelEntry {
+                    canonical_id: canonical.trim().to_string(),
+                    chute_slug: slug.trim().to_string(),
+                },
+                None => ChutesModelEntry {
+                    canonical_id: tok.to_string(),
+                    chute_slug: tok.to_string(),
+                },
+            })
+            .filter(|e| !e.canonical_id.is_empty() && !e.chute_slug.is_empty())
             .collect();
         let chutes_enable_streaming = env::var("CHUTES_ENABLE_STREAMING")
             .ok()
--- a/crates/config/src/types.rs
+++ b/crates/config/src/types.rs
@@ -1017,6 +1017,34 @@ mod tests {
         let cfg = ExternalProvidersConfig::from_env();
         assert!(cfg.chutes_models.is_empty());
     }
+
+    #[test]
+    #[serial]
+    fn chutes_models_dedup_duplicate_canonical_ids_first_wins() {
+        // Duplicate canonical id (even with a different slug) is dropped so a
+        // misconfig can't register redundant fallback providers — first wins.
+        std::env::set_var(
+            "CHUTES_MODELS",
+            "zai-org/GLM-5.1-FP8=zai-org/GLM-5.1-TEE,zai-org/GLM-5.1-FP8=zai-org/GLM-5-TEE,other/model",
+        );
+        let cfg = ExternalProvidersConfig::from_env();
+        std::env::remove_var("CHUTES_MODELS");
+
+        assert_eq!(
+            cfg.chutes_models,
+            vec![
+                ChutesModelEntry {
+                    canonical_id: "zai-org/GLM-5.1-FP8".to_string(),
+                    chute_slug: "zai-org/GLM-5.1-TEE".to_string(),
+                },
+                ChutesModelEntry {
+                    canonical_id: "other/model".to_string(),
+                    chute_slug: "other/model".to_string(),
+                },
+            ],
+            "duplicate canonical id dropped (first wins); the second slug is ignored"
+        );
+    }
 }
 
 /// One Chutes model to register, parsed from a single `CHUTES_MODELS` token.
@@ -1133,24 +1161,38 @@ impl ExternalProvidersConfig {
         // secret can't pass as Some("") and silently fail at request time.
         .filter(|s| !s.is_empty());
         // `canonical_id=chute_slug` per comma-separated token; a bare token means
-        // canonical_id == chute_slug. Drop tokens missing either side.
-        let chutes_models = env::var("CHUTES_MODELS")
-            .unwrap_or_default()
-            .split(',')
-            .map(|s| s.trim())
-            .filter(|s| !s.is_empty())
-            .map(|tok| match tok.split_once('=') {
-                Some((canonical, slug)) => ChutesModelEntry {
-                    canonical_id: canonical.trim().to_string(),
-                    chute_slug: slug.trim().to_string(),
-                },
-                None => ChutesModelEntry {
-                    canonical_id: tok.to_string(),
-                    chute_slug: tok.to_string(),
-                },
-            })
-            .filter(|e| !e.canonical_id.is_empty() && !e.chute_slug.is_empty())
-            .collect();
+        // canonical_id == chute_slug. Drop tokens missing either side, and dedup by
+        // canonical id (first wins) so a misconfig can't register duplicate
+        // fallback providers that every refresh would re-attach.
+        let chutes_models = {
+            let raw = env::var("CHUTES_MODELS").unwrap_or_default();
+            let mut seen: std::collections::HashSet<String> = std::collections::HashSet::new();
+            let mut entries: Vec<ChutesModelEntry> = Vec::new();
+            for tok in raw.split(',').map(|s| s.trim()).filter(|s| !s.is_empty()) {
+                let entry = match tok.split_once('=') {
+                    Some((canonical, slug)) => ChutesModelEntry {
+                        canonical_id: canonical.trim().to_string(),
+                        chute_slug: slug.trim().to_string(),
+                    },
+                    None => ChutesModelEntry {
+                        canonical_id: tok.to_string(),
+                        chute_slug: tok.to_string(),
+                    },
+                };
+                if entry.canonical_id.is_empty() || entry.chute_slug.is_empty() {
+                    continue;
+                }
+                if !seen.insert(entry.canonical_id.clone()) {
+                    eprintln!(
+                        "WARN: duplicate CHUTES_MODELS canonical id '{}' ignored (first wins)",
+                        entry.canonical_id
+                    );
+                    continue;
+                }
+                entries.push(entry);
+            }
+            entries
+        };
         let chutes_enable_streaming = env::var("CHUTES_ENABLE_STREAMING")
             .ok()
             .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -51,7 +51,7 @@ use crate::{
     ImageGenerationError, ImageGenerationParams, ImageGenerationResponseWithBytes,
     InferenceProvider, ListModelsError, ModelInfo, ModelsResponse, PrivacyClassifyError,
     RerankError, RerankParams, RerankResponse, SSEEvent, ScoreError, ScoreParams, ScoreResponse,
-    StreamingResult,
+    StreamChunk, StreamingResult,
 };
 
 /// Sane fallback when a non-positive timeout is supplied; a non-positive value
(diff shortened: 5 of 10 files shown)


PierreLeGuen (changes requested):
This is a well-structured change — the canonical-id/slug split is threaded correctly (slug goes upstream in the request body, canonical id drives routing/catalog), the verifiable-never-plaintext filter sits on the single selection path all completion ops use, and the new config format is well documented in `env.example`. Three issues need addressing before merge, though.

**Blocking / important**

1. **Fail-closed gap when Chutes registration fails** — `crates/services/src/inference_provider_pool/mod.rs:1427`. The plaintext filter only kicks in once an attested provider is registered. With the canonical IDs, an active external/OpenRouter row can share a model ID with a configured Chutes model; if Chutes is enabled but its API key is missing or provider construction fails, the model never gets pinned, the external row survives `load_external_providers` (the `pinned_models` guard there only protects models that actually got pinned), and plaintext is served for a model configured as verifiable. Reserve configured Chutes canonical IDs before external provider loading, or make Chutes config/registration failures fatal for those IDs — and add a regression test for the collision case.

2. **Streaming requests regress while `CHUTES_ENABLE_STREAMING=false` (the default)** — mod.rs:1383-1517 with `crates/inference_providers/src/attested/chutes/mod.rs:481-488`. Selection has no streaming-capability filter, so the Chutes provider is included for streaming requests it categorically refuses. Its refusal is a `CompletionError` matching no retry keyword, and since Chutes is tried last it overwrites both `last_error` and `last_retry_decision` in `retry_with_fallback`: a transient NEAR 5xx that previously got exponential-backoff retries now gets none, and the client sees "set CHUTES_ENABLE_STREAMING" instead of the real NEAR error. Filter providers that can't serve the requested operation at selection time, or treat capability refusals as skip-without-clobbering.

3. **Legacy `register_pinned_provider` silently lost its overwrite protection** — mod.rs:3394-3407 with mod.rs:651-662. The refresh merge now re-attaches only providers recorded in `pinned_providers`, but `register_pinned_provider` writes only `pinned_models`, so a same-name DB `inference_url` row overwrites a legacy-pinned provider on the next refresh — the exact substitution the removed guard existed to prevent. It's latent (production no longer calls the legacy function after this PR), but it's a broken security invariant on a still-public API that the e2e tests still use. Either have it populate `pinned_providers` too, or delete it and migrate its callers; either way, add an overwrite test on the `load_inference_url_models` path.

**Non-blocking**

- mod.rs:686-691 — `register_pinned_secondary_provider` pushes into `pinned_providers` unconditionally, so duplicate canonical ids in `CHUTES_MODELS` accumulate duplicate providers that every refresh re-attaches. Dedup by Arc pointer there too, or reject duplicate canonical ids during `CHUTES_MODELS` parsing.
- mod.rs:3394-3407 / 3483-3498 — if an operator drops the NEAR `inference_url` row but keeps the Chutes config, the model's entry is never rebuilt (absent from `model_providers`) and never stale-removed (pinned), so dead NEAR providers stay tier-first until per-provider failure demotion and are only pruned at restart. Consider rebuilding pinned models absent from fresh discovery as pinned-only.
- mod.rs:1456 — round-robin rotation is applied over the full provider list before the stable tier sort, so with multiple NEAR providers plus a fallback the rotation skews same-tier balancing (e.g. with two NEAR providers and one Chutes, one NEAR provider leads two cycles out of three). Tier-group first and rotate within each tier.
- E2EE/pubkey-routed requests get no Chutes fallback, since pinned secondaries are never added to `pubkey_to_providers`. That looks intentional (Chutes has no signing keys), but a comment stating it would help.
- The refresh-merge re-attachment — the riskiest part of the change — has no direct test; the four new pool tests cover registration and selection ordering only. A test driving the actual `load_inference_url_models` merge with a pinned secondary would lock in the "discovery can never drop the Chutes provider" invariant.

**Checks run locally:** `cargo fmt --all -- --check` clean; `cargo clippy` clean across services/config/inference_providers/api; `cargo test -p config --lib` 25 passed (incl. the two new `CHUTES_MODELS` parsing tests); `cargo test -p services --lib inference_provider_pool` 57 passed (incl. the four new tier/coexistence tests); `cargo test -p inference_providers --lib` 297 passed; `cargo test -p api --no-run` compiles all test targets against the new `Vec<ChutesModelEntry>` type. The `chutes_catalog` e2e tests fail for environmental reasons only (they need a dstack TEE unix socket absent outside a CVM; all e2e tests in this checkout are blocked identically). GitHub CI (Lint, Test Suite, security audit) was green at last read.

Evrard-Nil:
Thanks for the thorough review, Pierre — all three blocking issues + the five non-blocking ones are addressed in `2f89b1a1`. I also ran an independent adversarial review pass (multi-dimension, refute-verified) that confirmed the verifiable-never-plaintext invariant holds on every path it probed; it flagged the same streaming/stale/dedup issues you did (and refuted a handful of false positives), and your **#1 fail-closed gap was the most valuable catch** — fixed.

**Blocking**

1. **Fail-closed when Chutes registration fails** — new `InferenceProviderPool::reserve_pinned_models`, called in `init_inference_providers` **before** the external/discovery load, reserves every configured Chutes canonical id. A colliding plaintext external/OpenRouter row can no longer register for a configured-verifiable id even if the Chutes provider fails to build (missing key / `Provider::new` error); the id then serves only attested providers or fails closed (404). Regression test `reserved_canonical_id_blocks_plaintext_external_collision`.

2. **Streaming regresses with `CHUTES_ENABLE_STREAMING=false`** — `InferenceProvider::supports_streaming()` (default true; Chutes returns `allow_streaming`). `retry_with_fallback` now filters streaming-incapable providers for streaming ops when a capable sibling exists (extracted to `filter_streaming_capable`), so a retryable NEAR 5xx no longer falls through to Chutes' refusal (which clobbered `last_retry_decision` and surfaced the wrong error). If *no* provider can stream (Chutes-only, streaming off), the list is kept so the clear error still surfaces. Unit test `filter_streaming_capable_prefers_streaming_providers`.

3. **Legacy `register_pinned_provider` lost overwrite protection** — it now also records in `pinned_providers`, so the refresh merge re-attaches it (a same-name DB inference_url row can't overwrite it). The refresh merge itself is extracted to a pure `merge_discovered_and_pinned` helper with a direct unit test (`merge_discovered_and_pinned_never_drops_pinned_and_prunes_stale`) — the riskiest path you flagged as untested.

**Non-blocking**

- **Dedup duplicate canonical ids** in `CHUTES_MODELS` parsing (first wins + `WARN`), plus symmetric Arc-pointer dedup in `register_pinned_secondary_provider`. Test `chutes_models_dedup_duplicate_canonical_ids_first_wins`.
- **Stale-NEAR prune** — a pinned canonical id absent from a (complete) discovery cycle is rebuilt pinned-only, so a dropped NEAR backend isn't tried tier-first until demotion. (Safe because `load_inference_url_models` early-returns on empty and `fetch_inference_url_models` is all-or-error — documented on the helper.)
- **Round-robin within the leading tier** — providers are ordered by `(health, tier)` first, then the leading group is rotated, fixing the same-tier balance skew.
- **pubkey/E2EE fallback** — added the comment: pinned secondaries are intentionally absent from `pubkey_to_providers` (Chutes has no signing key; its integrity is the ML-KEM AEAD channel), so the E2EE path has no Chutes fallback by design.

All green locally: `cargo fmt`/`clippy` clean across the four crates; config 25, pool 60, inference_providers 297. Re-requesting review.

PierreLeGuen (changes requested):
The tiered NEAR-primary/Chutes-fallback design is sound and well-tested on the happy path: `reserve_pinned_models` runs before any external load so plaintext rows can't claim a canonical id (fail-closed), the verifiable-tier filter prevents silent downgrade to non-attested providers, and discovery merges re-append pinned providers rather than overwriting them. However, the merge logic has real gaps around its "complete discovery set" precondition that affect existing call paths.

**Blocking**

- `crates/services/src/inference_provider_pool/mod.rs:1888-1893`, caller `crates/api/src/routes/admin.rs:501` — `merge_discovered_and_pinned` rebuilds every pinned canonical id absent from `discovered` as pinned-only, but the admin `PATCH /v1/admin/models` runtime path calls `load_inference_url_models` with only the models in the PATCH batch, violating the documented complete-set precondition. Any admin PATCH carrying an `inference_url` overwrites the `model_to_providers` entry of every Chutes-enabled, NEAR-served model *not* in the batch to `[Chutes only]`. Until the next periodic refresh (up to `refresh_interval_secs`, default 900s; indefinitely if refresh is disabled), traffic for those models silently shifts to the Chutes fallback — inverting the PR's routing goal — and E2EE pubkey-routed requests hard-fail with `NoPubKeyProvider`, since Chutes intentionally has no `pubkey_to_providers` entry. Fix: gate the absent-pinned rebuild on an explicit `complete_set: bool` (true only on init and the discovery/refresh path), or give the admin path a variant that merges pinned providers only for the models actually in the batch. Please add a regression test for a partial-batch call with an unrelated pinned model — the new merge tests only cover complete sets.

**Important**

- `crates/services/src/inference_provider_pool/mod.rs:2884` — the inverse edge case: `load_inference_url_models` returns early on an empty discovery set, so when the last NEAR backend disappears the pinned merge never runs at all. Since `remove_stale_providers` exempts pinned names, a Chutes-pinned model that previously had NEAR+Chutes keeps the stale NEAR provider in `model_to_providers` until restart, breaking the "pinned id absent from discovery is rebuilt pinned-only" invariant. Treat an empty fetch as a complete empty set: still run the merge with `HashMap::new()` and include the dropped providers in cleanup.
- `crates/services/src/inference_provider_pool/mod.rs:3468-3478` — related cleanup gap: when the merge rebuilds a pinned id as pinned-only, the dropped NEAR provider Arcs are never pruned from `pubkey_to_providers` (`old_provider_ptrs` only covers keys present in this cycle's discovery, and `remove_stale_providers` skips pinned names). Orphaned entries keep stale pubkeys routable and surface as `NoPubKeyProvider` instead of a clean miss. Collect the pointers dropped by the absent-pinned branch and prune them in the same atomic update.

**Minor**

- `crates/inference_providers/src/attested/nearai/mod.rs:1068` — `Fleet` implements `InferenceProvider` but doesn't override `tier()`, so it reports `NonAttested`. Not registered in the pool today, but if it ever is, the verifiable filter would classify NEAR's own fleet as plaintext. A one-line `tier()` override closes the latent footgun.
- `crates/services/src/inference_provider_pool/mod.rs:1842` — the streaming-capability filter keys off `operation_name.ends_with("_stream")`. Only `chat_completion_stream` exists today, but a future streaming op that breaks the suffix convention silently skips the filter. An explicit `is_streaming` flag on `retry_with_fallback`, or a const list of streaming ops asserted in a test, would make this fail loudly.

**Checks run locally**: CI (Lint, Test Suite, security_audit) green on the PR; `cargo fmt --all -- --check` clean; clippy clean on the touched crates; `cargo test -p config` 26/26 (incl. new `CHUTES_MODELS` parsing tests); `cargo test -p services` pool suite 60/60 (incl. new tier/merge/streaming-filter/reservation tests); `cargo test -p inference_providers` 297/297; `cargo test -p api` 137/137; combined services/inference_providers/config run 373/373. Skipped: e2e (needs PostgreSQL) and vLLM integration tests (need a live inference server). Note the blocking and important findings above are not exercised by any current test.

Evrard-Nil:
Round 2 addressed in `e2e92028` — you were right, the F2 stale-prune I added had a real partial-batch regression. Thanks for catching it before merge.

**Blocking — admin-PATCH partial batch**
- `merge_discovered_and_pinned` is now **re-append only** and touches only the models in `discovered`, so a partial batch (the admin `PATCH /v1/admin/models` runtime registration) can no longer reset unrelated NEAR+Chutes models to Chutes-only.
- The stale-prune moved to a separate **complete-set-only** `prune_stale_pinned`, invoked ONLY from `sync_inference_url_models` (the periodic refresh, which passes the full fetch). The admin path calls `load_inference_url_models` directly → re-append only, no prune. Regression test `merge_discovered_and_pinned_reappends_and_leaves_absent_pinned_untouched` asserts an absent pinned id is left untouched.

**Important — empty discovery**
- `sync_inference_url_models` now captures the (possibly empty) complete name set and still runs `prune_stale_pinned`. So when the *last* NEAR backend disappears, a Chutes-pinned model is rebuilt pinned-only (`load_inference_url_models` early-returns on empty; the prune does the rebuild). Test `prune_stale_pinned_rebuilds_absent_pinned_only` covers both "still served → coexist" and "dropped → pinned-only".

**Important — cleanup**
- `prune_stale_pinned` collects the dropped NEAR Arc pointers and prunes them from `pubkey_to_providers` (and the failure counters), so they don't linger as orphaned / `NoPubKeyProvider`-routable entries.

**Minor**
- `nearai::Fleet` now overrides `tier() -> Near` (defensive — `Provider` is what's pooled, but no misclassification if a `Fleet` is ever pooled directly).
- Streaming ops are now an explicit `STREAMING_OPERATIONS` const (not a `_stream` suffix), with `streaming_operations_list_is_explicit` asserting `chat_completion_stream` is registered so a rename/new op can't silently bypass the filter.

pool 62, inference_providers 297, config 26; clippy + fmt clean. Re-requesting.

PierreLeGuen (changes requested):
The tiered NEAR-primary/Chutes-fallback design is sound and well-tested: NEAR `inference_url` backends stay tier `Near` and pass the verifiable filter, `merge_discovered_and_pinned` is partial-batch safe for the admin PATCH path, `prune_stale_pinned` only runs on complete-set refreshes, the model-pubkey (E2EE) routing path deliberately gets no Chutes fallback, and the streaming-capability filter correctly covers `chat_completion_stream`, the only streaming op routed through `retry_with_fallback`. New log lines carry only IDs/counts. Two issues should be addressed before merge.

**Blocking**

- `crates/services/src/inference_provider_pool/mod.rs:2030` — provider filtering before fallback only handles streaming capability, not client-side E2EE. A request with `x-client-pub-key` but no `x-model-pub-key` is valid on NEAR but rejected by Chutes at `crates/inference_providers/src/attested/chutes/mod.rs:397` with a non-retryable "client-facing E2EE is not supported" error. If NEAR fails with a retryable 5xx/connection error, fallback reaches Chutes and that error masks the NEAR failure and suppresses further retry. Add a client-E2EE capability filter (mirroring the streaming one) or exclude Chutes when `x-client-pub-key` is present, plus a regression test.

**Important**

- `crates/services/src/inference_provider_pool/mod.rs:1907,3645` + `crates/api/src/lib.rs:764` — a *reserved-only* pinned id (Chutes enabled but provider construction failed, or `CHUTES_API_KEY` missing — the reservation at lib.rs:764 runs before the key check) lands in `pinned_models` but not `pinned_providers`. `remove_stale_providers` skips everything in `pinned_models`, while `prune_stale_pinned` iterates only `pinned_providers` (early return at mod.rs:1913). So if NEAR later drops such a model from discovery, its dead `nearai::Provider` mapping, `pubkey_to_providers` entries, and failure counters linger for the process lifetime. The catalog `is_active` gate bounds the blast radius to staleness rather than wrong serving, but it breaks cleanup in exactly the failure mode the reservation was added for. Fix: in `prune_stale_pinned`, also process ids in `pinned_models` with no `pinned_providers` entry that are absent from the complete set — remove the mapping entirely (true fail-closed 404) — or make `remove_stale_providers` skip only ids present in `pinned_providers`.

**Non-blocking**

- `crates/services/src/inference_provider_pool/mod.rs:1907-1953` — `prune_stale_pinned` strips dropped provider Arcs from `pubkey_to_providers` globally by pointer. Since `inference_url_providers` is keyed by URL, two model names sharing one inference URL share one Arc; pruning a stale backend for one pinned id also removes the pubkey entries the other, still-discovered model relies on, transiently breaking its pubkey-routed selection until the next refresh backfills. This mirrors the pre-existing pattern in `remove_stale_providers` and self-heals, but a guard (skip pointers still referenced by other models' `model_to_providers`) would be strictly correct.
- `crates/api/src/routes/admin.rs:463` — the `is_pinned` skip-unregister guard now covers every NEAR-served canonical id with a Chutes fallback, not just dedicated Chutes ids. A PATCH that changes/clears `inference_url` or deactivates such a model no longer runs `unregister_provider` cleanup, leaving stale `pubkey_to_providers` entries, failure counters, and the `inference_url_providers` cache entry until the next refresh tick; a runtime `provider_type` transition for these ids is silently ignored. Behavior stays safe (pubkey intersection plus catalog gating), but update the comment at admin.rs:456-462 and ideally let the URL-change case refresh cleanly.
- `crates/inference_providers/src/attested/chutes/mod.rs:705` — under one canonical id, a Chutes-fallback-served response has `supports_chat_signatures = false`, so signature availability becomes nondeterministic per request for a model whose NEAR-served responses are normally signable. This is the documented #758 trade-off, but it's now per-request visible on an existing model id and deserves client-facing documentation.

**Checks run locally**
- `cargo clippy --all-targets` / `--no-deps` across `services`, `config`, `inference_providers`, `api` — clean
- `cargo test -p services --lib inference_provider_pool` — 62 passed (includes new tier/merge/prune/streaming-filter/reservation tests)
- `cargo test -p config` — 26 passed (includes `CHUTES_MODELS` parsing/dedup tests)
- `cargo test -p inference_providers --lib` — 297 passed
- `cargo test --lib --bins` — 375 passed, 1 ignored
- `cargo check -p api`, `git diff --check origin/main...HEAD` — clean
- GitHub CI (security_audit, Lint, Test Suite) green. E2E suites skipped locally — they require a running PostgreSQL instance; the unit tests cover the new logic directly.

Evrard-Nil:
Round 3 addressed in `a459509b`.

**Blocking — client-side E2EE fallback gap**
Same class as the streaming one, fixed the same way:
- `InferenceProvider::supports_client_e2ee()` (default true; Chutes returns false).
- `retry_with_fallback_caps` applies `filter_client_e2ee_capable`: when a request carries `x_client_pub_key` and a capable (NEAR) sibling exists, the non-capable Chutes provider is dropped — so a retryable NEAR failure can't fall through to Chutes' non-retryable "client E2EE not supported" rejection (which masked the NEAR error + suppressed retry). `chat_completion`/`chat_completion_stream` compute the flag from `params.extra` and call the `_caps` variant; the other 15 callers go through a thin `retry_with_fallback` wrapper (`needs_client_e2ee=false`) and are unchanged. Test `filter_client_e2ee_capable_prefers_capable_providers`.

**Important — reserved-only pinned cleanup**
`remove_stale_providers` now skips ids present in **`pinned_providers`** (an actual pinned provider), not all of `pinned_models`. So a *reserved-only* id (reserved for the fail-closed external block but with no provider because Chutes failed to build / the key was missing) is no longer exempt: if NEAR also drops it, it's removed entirely (fail-closed 404) with full pubkey/lb/failure-counter cleanup, instead of lingering as a dead NEAR mapping. The reservation still blocks plaintext external (that check uses `pinned_models`, untouched). Test `reserved_only_pinned_id_is_removed_when_stale_but_real_pinned_survives`.

**Non-blocking**
- `admin.rs`: documented that the `is_pinned` skip-unregister now also covers NEAR+Chutes coexisting ids, so a runtime `inference_url` change leaves the stale old-url NEAR provider until the next refresh prunes it (self-heals; pubkey-intersection + catalog gating keep it safe).
- `chutes`: documented the per-request signature-availability trade-off for a model listing both a NEAR (signable) and a Chutes (unsigned) provider.
- The shared-Arc pubkey-prune guard: I left `prune_stale_pinned` mirroring `remove_stale_providers`' existing global-by-pointer prune (it self-heals on the next refresh, exactly as you noted) rather than adding the guard to one path only — happy to add it to both as a separate cleanup if you'd prefer.

pool 64, inference_providers 297, config 26; clippy + fmt clean. Re-requesting.

PierreLeGuen (changes requested):
Solid iteration overall — the fail-closed reservation before external load, the partial-batch-safe `merge_discovered_and_pinned`, the verifiable→never-plaintext filter at selection time, and the capability filters that only drop a provider when a capable sibling exists are all well constructed, and the new pool tests cover them well. The #758 invariant (discovery can never substitute the attested provider) is preserved, and the pubkey-routed E2EE path is unaffected since Chutes never enters `pubkey_to_providers`. One substantive gap remains, which is why I'm requesting changes.

**Blocking**

- `crates/inference_providers/src/attested/chutes/mod.rs:433`, with pass-through at `:465` (non-streaming `raw_bytes`) and `:516` (each decrypted stream chunk): the provider receives the internal chute slug and pins it into the upstream request, then returns the decrypted upstream body verbatim — so `response.model` carries the slug (e.g. `zai-org/GLM-5.1-TEE`) instead of the canonical id the client requested. `crates/api/src/routes/completions.rs:1601` and `:3299` forward these bytes unmodified. For a NEAR-primary model this happens on every fallback-served response; for a Chutes-only model with `canonical_id≠chute_slug` (the documented OpenRouter-id case) it happens on every response. This breaks the PR's own canonical-id contract ("never the raw `-TEE` slug") and the expectation that `response.model` matches an id listable in `/v1/models`; response shape also ends up depending on which tier served the request. Since `supports_chat_signatures` is `false` for Chutes, rewriting `model` in the decrypted plaintext doesn't sacrifice any hash/signature guarantee. Suggested fix: carry both the canonical id and the chute slug — use the slug only for resolution and the upstream call, and rewrite `model` to the canonical id in `chat_completion` and in the stream decoder before returning.

**Non-blocking**

- `crates/services/src/inference_provider_pool/mod.rs` (`prune_stale_pinned`): the `needs_rewrite` check uses `cur.len() != pinned.len()` as a proxy for set equality. That's only correct while the current list is guaranteed a superset of the pinned list; a future partial overwrite would silently skip the rebuild. Compare the Arc-pointer sets directly — you already compute them for `dropped`.
- Same function, cleanup path: `dropped` prunes `pubkey_to_providers` and failure counters by Arc pointer without checking whether the same Arc still backs another discovery-served model (providers are cached per-URL; see the shared-Arc comment near mod.rs:3263). If two model names share an `inference_url` and only one is in `CHUTES_MODELS`, that model leaving discovery breaks pubkey-routed requests for the sibling until the next refresh (~one cycle). Consider excluding pointers still referenced by other `model_to_providers` entries.
- `crates/api/src/routes/admin.rs:463-474`: the `is_pinned` skip-unregister guard means a runtime `inference_url` change on a pinned model never registers the new URL's provider during the PATCH — the periodic refresh eventually applies it, but the window is unbounded if `refresh_interval_secs` is long or refresh is disabled, leaving the model de facto Chutes-only until restart. Add an unregister→re-register path for the URL-change case inside the guard, or document the window.
- `crates/config/src/types.rs:258`: the duplicate-`CHUTES_MODELS` warning uses `eprintln!`; use `tracing::warn!` so it reaches the logging pipeline in containerized deployments.
- `crates/services/src/inference_provider_pool/mod.rs:651`: `register_pinned_provider` no longer has production callers (tests and `crates/api/tests/e2e_all/chutes_catalog.rs` only); production uses `register_pinned_secondary_provider` exclusively. Two registration paths with replace-vs-push semantics is a maintenance hazard — migrate the tests and remove it, or mark it test-only.

**Checks run locally**: `cargo fmt --all -- --check` and `cargo clippy` (config, services, inference_providers, api) clean; `cargo test -p config --lib` 26 passed (incl. 3 new `CHUTES_MODELS` parsing tests); `cargo test -p inference_providers --lib` 297 passed; `cargo test -p services --lib` 377 passed, 1 ignored (incl. all 64 `inference_provider_pool` tests and the new tier/merge/prune/capability tests). Skipped: e2e tests (need PostgreSQL) and vLLM integration tests (need a live server). GitHub CI on the PR head: Lint, Test Suite, and security_audit all green.

Evrard-Nil:
Round 4 addressed in `ce397b2c`.

**Blocking — `response.model` leaked the chute slug**
The provider now carries a `canonical_id` (set in api init via `Config::with_canonical_id(entry.canonical_id)`). When it differs from the chute slug, `chat_completion` and the stream decoder rewrite `model` to the canonical id via a shared `set_model_in_json` helper — so `response.model` matches an id listable in `/v1/models` and the `-TEE` slug never reaches a client, on both the non-stream `raw_bytes` and each decrypted stream chunk. The rewrite preserves all other (incl. unmodeled) fields and sacrifices no signature guarantee (`supports_chat_signatures == false` for Chutes). When `canonical == slug` the byte-exact path is untouched. Tests: `set_model_in_json_rewrites_model_and_preserves_other_fields`, `rewrite_sse_event_model_rewrites_data_chunk_but_passes_control_through`.

**Non-blocking**
- `prune_stale_pinned`: the rewrite decision now compares Arc-pointer **sets** (not lengths), and a **shared-Arc guard** excludes any dropped pointer still referenced by another served model before pruning `pubkey_to_providers`/counters — so a sibling sharing a per-URL Arc isn't transiently broken.
- Removed `register_pinned_provider` (no production callers; the replace-vs-push duplication was the maintenance hazard you flagged). Migrated the unit + e2e tests to `register_pinned_secondary_provider` and fixed the doc references.
- The duplicate-`CHUTES_MODELS` warning: I kept `eprintln!` deliberately — `config` has no `tracing` dependency and `from_env` runs before the tracing subscriber is initialized (a `tracing::warn!` there would be dropped), whereas stderr is captured by the container log pipeline. It's also consistent with the sibling `CHUTES_API_KEY_FILE` warning. Happy to add a `tracing` dep + switch both if you'd prefer it routed through the pipeline instead.

inference_providers 299, pool 64, config 26; clippy + fmt clean; e2e binary compiles. Re-requesting.

PierreLeGuen (changes requested):
This is a careful, well-tested change — the tier ordering, partial-batch-safe merge, complete-set-only stale prune, consistent lock ordering, and canonical-id billing all check out, and the new test coverage is solid. A few real issues remain before merge, one of which breaks the PR's own core invariant.

**Blocking**

- `crates/inference_providers/src/attested/chutes/mod.rs:366-396` — `rewrite_sse_event_model` rewrites only `raw_bytes` and leaves `SSEEvent.chunk` carrying the chute slug, on the assumption that clients always receive `raw_bytes`. That assumption fails on the route's re-serialization paths: in `crates/api/src/routes/completions.rs:1361-1380`, when `auto_redact_enabled` or `alias_served` is true, byte-exact passthrough is disabled and each event is re-serialized from `event.chunk`, including `model`. The auto-redact flush-chunk template at `completions.rs:1385-1396` also caches the slug. Once `CHUTES_ENABLE_STREAMING` is flipped on (the PR's stated plan, "no code change needed"), auto-redacted or alias-served streams will emit the internal `-TEE` slug in every chunk — violating the "slug never reaches a client" invariant. Fix is small: also rewrite the parsed chunk (`if let Some(StreamChunk::Chat(c)) = &mut ev.chunk { c.model = canonical.to_string(); }`) and extend the unit test to assert the chunk's model is canonical, covering the reserialization path.

**Important**

- `crates/inference_providers/src/attested/chutes/mod.rs:745` — the public attestation report emits `self.model_name` (the upstream chute slug). `/v1/attestation/report?model=<canonical>` returns `model_attestations` verbatim, so canonical-routed Chutes models leak a mismatched `-TEE` id in a client-visible surface. Emit `self.canonical_id` here.
- `crates/services/src/inference_provider_pool/mod.rs:583` with `crates/api/src/routes/admin.rs:1262` and `admin.rs:1347` — the admin PATCH path correctly guards pinned models (`admin.rs:472`), but delete and deprecate call `unregister_provider` unconditionally, which also strips `pinned_models`/`pinned_providers`. Deleting or deprecating a NEAR-served canonical id with a Chutes fallback tears down the fallback too, and only a restart recovers it. These handlers should apply the same `is_pinned` guard (or at minimum log loudly).

**Non-blocking**

- `crates/inference_providers/src/attested/chutes/mod.rs:471-482` — `models()` still advertises the chute slug as the public id. No production caller today (the client-facing list comes from the DB catalog), but it's the provider's listing surface; return `self.canonical_id`.
- `crates/api/src/routes/admin.rs:460-474` — because every NEAR-served id with a Chutes fallback is `is_pinned`, a runtime PATCH that changes `inference_url` or deactivates the model skips eager cleanup, leaving the stale NEAR provider and its pubkey/failure-counter entries until the next periodic refresh. This is documented as accepted and safe in practice (`resolve_and_get_model` filters on `is_active`), but consider a follow-up that scopes the skip to the pinned providers themselves rather than the whole model entry.

**Local checks**

- `cargo fmt --all -- --check` — clean; `cargo clippy --all-targets -- -D warnings` — clean; `cargo check` — pass.
- `cargo test -p config --lib` — 26 passed (includes new `CHUTES_MODELS` pair-parsing/dedup tests).
- `cargo test -p inference_providers --lib` — 299 passed (includes new rewrite/SSE tests).
- `cargo test -p services --lib inference_provider_pool` — 64 passed (tier ordering, merge/prune, capability filters).
- E2E suite not exercised locally (requires PostgreSQL and a dstack socket/signing keys unavailable in this environment); GitHub CI (Lint, Test Suite, security_audit) is green on the PR head.

Evrard-Nil:
Round 5 addressed in `75f8c06e` — all of these were the same "slug must never surface" theme; thanks for chasing the re-serialization path.

**Blocking — streamed chunk leaked the slug on re-serialization**
`rewrite_sse_event_model` now rewrites the parsed `chunk` too, not just `raw_bytes`: `if let Some(StreamChunk::Chat(c)) = &mut ev.chunk { c.model = canonical }`. So the auto-redact / alias-served paths (which re-serialize from `event.chunk` per `completions.rs:1361-1396`) emit the canonical id once streaming is enabled. The unit test now asserts the chunk's `model` is canonical, covering that path.

**Important**
- `get_attestation_report` now emits `self.canonical_id` (was `self.model_name`) — `/v1/attestation/report?model=<canonical>` no longer returns a mismatched `-TEE` id.
- `delete_model` and `deprecate_model` now apply the same `is_pinned` guard as the PATCH path and log loudly, instead of unconditionally `unregister_provider`-ing — so deleting/deprecating a NEAR-served canonical id no longer tears down its Chutes fallback (recoverable only by restart). The catalog delete still 404s the model via the `is_active` gate; deprecate's alias routes the id to its successor, so the pinned provider isn't reached.

**Non-blocking**
- `models()` now returns `self.canonical_id` (was the slug) on the provider's listing surface.
- The admin url-change cleanup scoping (skip scoped to the pinned providers rather than the whole entry) — left as the documented, accepted follow-up you noted.

inference_providers 299, pool 64, config 26; `clippy --all-targets -D warnings` + fmt clean. Re-requesting.

PierreLeGuen (commented):
This is a solid landing of the tiered NEAR-primary/Chutes-fallback design. I traced the security-critical paths and they hold: `reserve_pinned_models` runs before the external/discovery loads in `crates/api/src/lib.rs:756-770` and `load_external_providers` checks `pinned_models`, so a plaintext provider can't claim a pinned canonical id even if the Chutes build fails; the verifiable filter in `get_providers_with_fallback` (mod.rs:1439) drops `NonAttested` providers whenever an attested one exists; `merge_discovered_and_pinned` is partial-batch safe while `prune_stale_pinned` only runs on the complete-set refresh path, with the shared-Arc guard and pubkey/failure-counter cleanup correct (mod.rs:1926-1956); and the canonical id is rewritten on every client surface — non-stream raw bytes + struct, both `raw_bytes` and `chunk.model` on the SSE path (every `e2ee_stream` chunk-bearing event is a `data:` chat chunk, so the rewrite covers all shapes), `models()`, and the attestation report. The `is_pinned` guards on delete/deprecate/unregister are in place, and all findings from the earlier review rounds are resolved at the current head.

- **Attestation evidence is unreachable for fallback-served responses** — `get_attestation_report` (crates/services/src/inference_provider_pool/mod.rs:2358) returns the first successful provider's report (NEAR's, given list order). A response served by Chutes fallback has neither a chat signature (documented trade-off) nor a client-addressable attestation report once the old `-TEE` id is retired. Worth documenting for clients of dual-provider models, or adding a provider/tier selector to the attestation endpoint in a follow-up.
- **Streaming fallback is inert by default** — with `CHUTES_ENABLE_STREAMING=false`, `filter_streaming_capable` (mod.rs:1825) removes Chutes whenever NEAR is present, and Chutes-only models can't stream at all. Since most chat traffic streams, the fallback fabric does little for streaming until the flag flips after the staging `[DONE]` check. Deliberate, but worth stating in the rollout/ops notes so the fallback isn't assumed live.
- **`rerank`/`embeddings`/`score` lack per-op capability filtering** (mod.rs:2700) — on a NEAR+Chutes canonical id, a NEAR failure falls through to Chutes' "unsupported" error, masking the real NEAR error — the same masking the streaming/client-E2EE filters fix for chat. Edge case today; consider extending the capability filter if these ops ever share canonical ids in practice.
- **Admin PATCH on a pinned model's `inference_url` skips the eager unregister** (crates/api/src/routes/admin.rs:473), so the stale old-URL provider lingers until the next periodic refresh. Documented in-code and safe (catalog `is_active` gating), flagging only so operators know runtime URL changes on pinned models aren't instantaneous.

Local checks: `cargo fmt --all -- --check` clean; `cargo clippy --no-deps -p config -p inference_providers -p services -p api` clean; `cargo test -p config --lib` 26 passed (new CHUTES_MODELS parsing tests), `cargo test -p inference_providers --lib` 299 passed (canonical-id rewrite/SSE tests), `cargo test -p services --lib inference_provider_pool` 64 passed (tier/merge/prune/filter/reservation tests), full `cargo test --lib --bins` 377 passed / 1 pre-existing ignore; `cargo check --tests` compiles across crates. GitHub CI (Lint, Test Suite, security_audit) green on the PR head. Not run locally: e2e suite (needs PostgreSQL) and vLLM/dstack integration tests (need live servers).

Evrard-Nil:
Thanks Pierre — appreciate the thorough trace across all the security-critical paths. Capturing your four non-blocking observations:

1. **Attestation/signature unreachable for fallback-served responses** → filed **#769** (document the dual-provider limitation and/or add a provider/tier selector to `/v1/attestation/report`).
2. **Streaming fallback inert by default** → it stays gated behind `CHUTES_ENABLE_STREAMING` pending the staging inner-`[DONE]` verification; I'll call this out explicitly in the rollout/ops notes so the streaming fallback isn't assumed live until the flag flips.
3. **`rerank`/`embeddings`/`score` lack per-op capability filtering** → filed **#770** (not reachable today since Chutes is chat-only and isn't a fallback for those canonical ids; extend the capability-filter pattern if that ever changes).
4. **Admin PATCH `inference_url` change on a pinned model skips eager unregister** → documented in-code and safe via the catalog `is_active` gate; the scoped-skip cleanup is the accepted follow-up.

All four are post-merge follow-ups (none reachable in the current default posture). The slug-leak surfaces and the fail-closed / verifiable invariants are all covered at the current head. Thanks again for the five rounds.

PierreLeGuen (commented):
The tiered NEAR-primary/Chutes-fallback design holds up well end to end. I traced the new pool mechanisms against the surrounding code and confirmed the load-bearing pieces:

- **Tier classification is complete**: DB-discovered backends (`nearai::Provider`, pool `mod.rs:3059`) report `Near`, Chutes reports `Attested3p`, and `non_attested::external::ExternalProvider` keeps the `NonAttested` default — the verifiable filter can't misclassify a production NEAR backend.
- **Fail-closed reservation ordering is correct**: `reserve_pinned_models` runs before any external/discovery load, `load_external_providers` skips reserved ids, and the only unguarded overwrite paths (`register_provider`/`register_providers`) are test-only.
- **Merge/prune split is sound**: `merge_discovered_and_pinned` is re-append-only and partial-batch safe (admin PATCH path), `prune_stale_pinned` runs only on the complete-set refresh path, and shared-Arc pubkey/counter pruning is guarded.
- **Canonical-id rewrite covers all client surfaces** (non-stream `raw_bytes`, stream `raw_bytes` + parsed `chunk`, `models()`, attestation report), and the E2EE `CLIENT_PUB_KEY` detection matches exactly what `completions.rs:57` and `responses/service.rs:1380` insert into `params.extra`.
- **Backward compatible**: bare `CHUTES_MODELS` entries parse as canonical==slug, so deploying without env changes is a no-op.

Non-blocking observations:

1. **`crates/inference_providers/src/attested/chutes/mod.rs:426-448`** (`rewrite_sse_event_model`) — `chunk.model` is mutated before the `raw_bytes` rewrite is attempted; if the raw rewrite fails (non-UTF-8 / non-object data line), the event is returned with `chunk` carrying the canonical id while `raw_bytes` still leaks the `-TEE` slug. Unreachable today since the decoder builds `raw_bytes` from the same JSON it parses into `chunk`, but the function should fail atomically: rewrite the raw bytes first, and only mutate `chunk` on success.
2. **`crates/config/src/types.rs:330`** — the duplicate-canonical-id warning uses `eprintln!`, bypassing structured log collection in production. Prefer `tracing::warn!` from `init_inference_providers` (which has tracing initialized and already iterates the parsed entries), or add a comment explaining why stderr is the only option here.
3. **`crates/api/src/routes/admin.rs:108-116`** — the NOTE comment overstates the lingering: when a PATCH changes a pinned model's `inference_url`, the atomic update already prunes the replaced provider's `pubkey_to_providers` entries and the merge drops it from `model_to_providers`; only the failure-counter entry lingers until the next refresh. Worth correcting so a future reader doesn't hunt for a cleanup gap that mostly doesn't exist.
4. **Operational note (no code change)** — adding a canonical id to `CHUTES_MODELS` that currently has an active plaintext external row hard-cuts that model to attested-only at the next restart (and 404s if the Chutes provider fails to build). That's the intended fail-closed rule and it warn-logs, but it deserves a line in deploy/runbook notes since it can silently take an externally-served model offline. Relatedly, the PR body's follow-up to deactivate the standalone staging `zai-org/GLM-5.1-TEE` row needs to actually happen at deploy time, or the old slug stays publicly routable alongside the canonical id.

Local checks run against the PR head (75f8c06): `cargo check --lib` across `config`/`inference_providers`/`services`/`api` plus the `e2e_all` test target; `cargo clippy` clean; `cargo fmt --check` clean; `cargo test` for `config` (26 passed, incl. 3 new `CHUTES_MODELS` parsing tests), `inference_providers` (299 passed, incl. new rewrite/SSE/tier tests), and `services` (377 passed, 1 pre-existing ignore, incl. the 9 new tier/merge/prune/filter/reservation tests). The `chutes_catalog` e2e tests passed in one environment and failed only for environmental reasons (missing dstack TEE socket) in another; CI Lint, Test Suite, and security_audit are all green on the same commit.

PierreLeGuen (approved):

lloydmak99 (approved):
Approve. PierreLeGuen's thorough multi-round review (5 rounds, 4 blocking issues resolved) covers the critical paths — verifiable-never-plaintext invariant, fail-closed reservation before external load, client E2EE capability filtering, response.model canonicalization, and stale-prune correctness. The four non-blocking items (#769, #770, streaming flag doc, rerank capability) are tracked as follow-ups. No new blocking issues from my read of the diff.

Issue #770, opened 2026-06-11
Chutes tiered fallback: extend per-op capability filter to rerank/embeddings/score


Evrard-Nil:
Follow-up from #768 review (non-blocking; edge case today).

#768 added capability filters in `retry_with_fallback_caps` so a NEAR failure doesn't fall through to Chutes for operations Chutes can't serve — `filter_streaming_capable` (streaming) and `filter_client_e2ee_capable` (client-facing E2EE). Without the filter, Chutes' "unsupported" error overwrites the retryable NEAR error and suppresses retry.

`rerank` / `embeddings` / `score` (`crates/services/src/inference_provider_pool/mod.rs` ~2700) have the **same masking risk**: on a NEAR+Chutes canonical id, a retryable NEAR failure would fall through to Chutes' "unsupported" error.

**Not reachable today** — Chutes is chat-only and isn't configured as a fallback for rerank/embeddings/score canonical ids. If that ever changes, extend the capability-filter pattern (e.g. a `supports_<op>()` capability) to those paths.

Issue #769, opened 2026-06-11
Chutes tiered fallback: no client-addressable attestation/signature for fallback-served responses


Evrard-Nil:
Follow-up from #768 review (non-blocking).

On a dual-provider canonical id (NEAR primary + Chutes `Attested3p` fallback), a response actually served by the **Chutes fallback** has:
- **no chat signature** — `supports_chat_signatures == false` for Chutes (documented #758 trade-off; integrity is the ML-KEM AEAD channel), and
- **no client-addressable attestation report** — `get_attestation_report` (`crates/services/src/inference_provider_pool/mod.rs` ~2358) returns the FIRST successful provider's report, which given tier order is NEAR's. Once the standalone `…-TEE` id is retired, a client can't fetch the Chutes instance's attestation for a fallback-served response.

So for a fallback-served response the client can't independently re-verify *which* attested backend served it.

**Options:**
- Document the limitation for clients of dual-provider models, and/or
- Add a `provider`/`tier` selector to `GET /v1/attestation/report` so a client can request the Chutes instance's report explicitly, and/or
- Surface the serving tier in the response so a client knows whether to expect a NEAR signature vs a Chutes E2EE-channel attestation.

Context: docs/chutes-attestation.md; the canonical-id rewrite already keeps the slug off `response.model`/`models()`/the report.

PR #771, opened 2026-06-11
fix(chutes): post-merge review polish (#768 follow-up)

Evrard-Nil:
Addresses the four non-blocking observations from Pierre's final review on #768 (merged). No behavior change in the default posture; all unreachable-today but worth tightening.

- **`rewrite_sse_event_model` atomicity** — compute the rewritten `raw_bytes` first and return the event unchanged on any failure, mutating the parsed `chunk` only after that succeeds. Prevents an event ever carrying `chunk` = canonical id while `raw_bytes` still leaks the `-TEE` slug (only reachable if the raw rewrite failed on a chunk it parsed, which the decoder makes impossible — but the function now fails atomically regardless).
- **config `eprintln!` rationale** — comment why the duplicate-`CHUTES_MODELS` warning uses `eprintln!` not `tracing::warn!`: `config` has no `tracing` dep and `from_env` can run before the subscriber is installed (a `tracing::warn!` would be dropped); stderr is captured by the container log pipeline; consistent with the sibling `CHUTES_API_KEY_FILE` warning.
- **admin `is_pinned`-guard comment corrected** — a runtime `inference_url` change on a pinned model re-registers the new url, and `load_inference_url_models`' atomic update already drops the replaced provider from `model_to_providers` and prunes its `pubkey_to_providers` entries; only the per-provider failure-counter lingers until the next refresh. The previous comment overstated the lingering.

The two larger follow-ups are tracked separately: #769 (attestation/signature for fallback-served responses) and #770 (per-op capability filter for rerank/embeddings/score).

`cargo test -p inference_providers --lib` 299 · clippy `-D warnings` + fmt clean.

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -465,11 +465,10 @@ pub async fn batch_upsert_models(
         // that has a Chutes fallback (it's `is_pinned`). So a PATCH that changes its
         // `inference_url` skips the eager `unregister_provider` here, but the NEW url
         // is still re-registered below and `load_inference_url_models`' atomic update
-        // already drops the replaced provider from `model_to_providers` and prunes
-        // its `pubkey_to_providers` entries — only the per-provider failure-counter
-        // entry lingers until the next periodic refresh. Behavior stays safe (pubkey
-        // intersection + catalog `is_active` gating); the residual is just that one
-        // stale counter, not the provider or its routing.
+        // drops the replaced provider from `model_to_providers`, prunes its
+        // `pubkey_to_providers` entries, and prunes its per-provider failure-counter
+        // entry — so the refresh leaves no stale routing or counter state behind.
+        // Behavior stays safe (pubkey intersection + catalog `is_active` gating).
         if app_state.inference_provider_pool.is_pinned(model_name) {
             continue;
         }
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -463,12 +463,13 @@ pub async fn batch_upsert_models(
         //
         // NOTE: with tiered fallback this now also covers a NEAR-served canonical id
         // that has a Chutes fallback (it's `is_pinned`). So a PATCH that changes its
-        // `inference_url` or deactivates it skips the eager `unregister_provider`
-        // cleanup here: the NEW url is still re-registered below (the merge keeps
-        // Chutes), but the stale OLD-url NEAR provider + its pubkey/failure-counter
-        // entries linger until the next periodic refresh prunes them. Behavior stays
-        // safe (pubkey intersection + catalog `is_active` gating); it just isn't
-        // instantaneously clean for a runtime url change on a Chutes-fallback model.
+        // `inference_url` skips the eager `unregister_provider` here, but the NEW url
+        // is still re-registered below and `load_inference_url_models`' atomic update
+        // already drops the replaced provider from `model_to_providers` and prunes
+        // its `pubkey_to_providers` entries — only the per-provider failure-counter
+        // entry lingers until the next periodic refresh. Behavior stays safe (pubkey
+        // intersection + catalog `is_active` gating); the residual is just that one
+        // stale counter, not the provider or its routing.
         if app_state.inference_provider_pool.is_pinned(model_name) {
             continue;
         }
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -465,10 +465,12 @@ pub async fn batch_upsert_models(
         // that has a Chutes fallback (it's `is_pinned`). So a PATCH that changes its
         // `inference_url` skips the eager `unregister_provider` here, but the NEW url
         // is still re-registered below and `load_inference_url_models`' atomic update
-        // drops the replaced provider from `model_to_providers`, prunes its
-        // `pubkey_to_providers` entries, and prunes its per-provider failure-counter
-        // entry — so the refresh leaves no stale routing or counter state behind.
-        // Behavior stays safe (pubkey intersection + catalog `is_active` gating).
+        // drops the *replaced* NEAR provider from `model_to_providers`, prunes its
+        // `pubkey_to_providers` entries, and prunes its per-provider failure counter
+        // (the prune is filtered against still-live pointers, so the coexisting
+        // pinned Chutes fallback keeps its counter) — no stale routing or counter
+        // state for the replaced provider is left behind. Behavior stays safe (pubkey
+        // intersection + catalog `is_active` gating).
         if app_state.inference_provider_pool.is_pinned(model_name) {
             continue;
         }
--- a/crates/config/src/types.rs
+++ b/crates/config/src/types.rs
@@ -1183,6 +1183,12 @@ impl ExternalProvidersConfig {
                     continue;
                 }
                 if !seen.insert(entry.canonical_id.clone()) {
+                    // `eprintln!` (not `tracing::warn!`) on purpose: the `config`
+                    // crate has no `tracing` dependency, and `from_env` runs during
+                    // startup config parsing — potentially before the tracing
+                    // subscriber is installed, where a `tracing::warn!` would be
+                    // dropped. stderr is always captured by the container log
+                    // pipeline. Consistent with the sibling CHUTES_API_KEY_FILE warn.
                     eprintln!(
                         "WARN: duplicate CHUTES_MODELS canonical id '{}' ignored (first wins)",
                         entry.canonical_id
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -376,13 +376,14 @@ const INTERNAL_KEYS: &[&str] = {
 /// alias-served paths — so leaving the slug there would still leak it). Only
 /// touches chunk-bearing data lines; control events ([DONE], blanks, the keyed
 /// init) have no chunk and pass through unchanged. On any parse failure the event
-/// is returned as-is (never drop a chunk over a rewrite).
+/// is returned as-is (never drop a chunk over a rewrite). The rewrite is ATOMIC:
+/// we compute the rewritten `raw_bytes` first and bail (returning the event
+/// unchanged) on any failure, mutating the parsed `chunk` only once that succeeds —
+/// so the event can never end up with `chunk` carrying the canonical id while
+/// `raw_bytes` still leaks the slug.
 fn rewrite_sse_event_model(mut ev: SSEEvent, canonical: &str) -> SSEEvent {
     if ev.chunk.is_none() {
         return ev;
-    }
-    if let Some(StreamChunk::Chat(c)) = &mut ev.chunk {
-        c.model = canonical.to_string();
     }
     let Ok(s) = std::str::from_utf8(&ev.raw_bytes) else {
         return ev;
@@ -394,6 +395,10 @@ fn rewrite_sse_event_model(mut ev: SSEEvent, canonical: &str) -> SSEEvent {
     let Ok(json) = String::from_utf8(rewritten) else {
         return ev;
     };
+    // raw_bytes rewrite succeeded — now mutate the parsed chunk to match.
+    if let Some(StreamChunk::Chat(c)) = &mut ev.chunk {
+        c.model = canonical.to_string();
+    }
     SSEEvent {
         raw_bytes: bytes::Bytes::from(format!("data: {json}\n\n")),
         chunk: ev.chunk,
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -947,6 +947,39 @@ mod tests {
     }
 
     #[test]
+    fn rewrite_sse_event_model_leaves_chunk_bearing_event_untouched_on_rewrite_failure() {
+        use crate::{ChatCompletionChunk, StreamChunk};
+        // Pins the ATOMIC invariant: if the raw_bytes rewrite fails (here the
+        // payload isn't a JSON object so set_model_in_json returns None), the event
+        // is returned fully unchanged — raw_bytes AND the parsed chunk's model both
+        // keep their original value. Guards against a future refactor reintroducing
+        // a half-rewritten state (chunk mutated to canonical while raw_bytes still
+        // leaks the slug). This failure path is unreachable in practice (the decoder
+        // only sets chunk: Some when raw_bytes parsed as valid JSON) but is cheap to
+        // pin and the only thing the atomicity reorder protects.
+        let chunk: ChatCompletionChunk = serde_json::from_value(json!({
+            "id":"c","object":"chat.completion.chunk","created":0,
+            "model":"zai-org/GLM-5.1-TEE","choices":[]
+        }))
+        .unwrap();
+        let ev = SSEEvent {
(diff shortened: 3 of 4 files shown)


PierreLeGuen (commented):
Small, well-scoped follow-up to #768 — one code change and two comment corrections, all verified against the implementation. No blocking issues; two non-blocking notes below.

**Verified correct:**

1. **`rewrite_sse_event_model` atomicity** (`crates/inference_providers/src/attested/chutes/mod.rs:384`) — moving the `chunk.model` mutation after the `raw_bytes` rewrite succeeds is a strict improvement: previously a raw-rewrite failure could return an event with a mutated `chunk` but unmodified `raw_bytes`. The failure path is in practice unreachable: the decoder (`e2ee_stream.rs:60-86`) only sets `chunk: Some` when the exact `raw_bytes` content parsed as valid JSON, so `from_utf8`/`set_model_in_json` cannot fail on a chunk-bearing event, and `chutes/mod.rs:609` is the only call site. The reconstructed event preserves all three fields including `raw_passthrough`.
2. **Config `eprintln!` rationale comment** (`crates/config/src/types.rs:1186`) — accurate; the config crate has no `tracing` dependency and `from_env` may run before subscriber install.
3. **Admin `is_pinned`-guard comment** (`crates/api/src/routes/admin.rs:464`) — traced against `load_inference_url_models` (`crates/services/src/inference_provider_pool/mod.rs:2989`): the atomic update replaces the model's `model_to_providers` entry, prunes the replaced provider's `pubkey_to_providers` entries by Arc pointer, and the URL cache replacement drops the old URL; `provider_failure_counts` is the only residual until the periodic refresh cleans it (`mod.rs:3685`). The corrected comment matches the runtime behavior.

- `crates/inference_providers/src/attested/chutes/mod.rs:384` — the new atomic-failure invariant isn't pinned by a test; `rewrite_sse_event_model_rewrites_data_chunk_but_passes_control_through` (mod.rs:908) covers only the success and control paths. Adding one case (event with `chunk: Some` and non-JSON `raw_bytes` → returned fully unchanged) would prevent a future refactor from silently reintroducing the half-rewritten state.
- Pre-existing, worth a follow-up issue (untouched by this diff): `batch_upsert_models` (`crates/api/src/routes/admin.rs:511`) calls `load_inference_url_models` with only the PATCHed models, which wholesale-replaces `inference_url_providers` (`mod.rs:3641`) and prunes `inference_url_fingerprint_states` to that subset (`mod.rs:3656`). Routing stays correct (other models' provider Arcs remain in `model_to_providers`), but the next periodic refresh rebuilds every other URL's provider from `Bootstrap` — full re-attestation, and fail-closed replacement if discovery transiently fails. Churn/availability risk, not correctness. Suggested fix: have `load_inference_url_models` merge rather than replace when given a partial set, leaving full-set reconciliation to `sync_inference_url_models`.

**Checks run locally:** `cargo fmt --check` clean; `cargo clippy -p inference_providers -p api -p config --all-targets -- -D warnings` clean; `cargo test -p inference_providers --lib` 299/299 passed; `cargo test -p config --lib` 26/26 passed; targeted `services` tests for the pinned-model guard and Chutes/NEAR provider coexistence passed; `git diff --check` clean. CI lint was green; the full DB-backed e2e suite was still in progress but nothing in this 76-line, mostly-comment diff touches it.

Evrard-Nil:
Thanks @PierreLeGuen — both non-blocking notes addressed:

1. **Atomic-failure invariant now pinned by a test** (a8631b80). Added `rewrite_sse_event_model_leaves_chunk_bearing_event_untouched_on_rewrite_failure`: a `chunk: Some` event whose `raw_bytes` can't be rewritten (non-JSON object → `set_model_in_json` returns `None`) must come back fully unchanged — `raw_bytes`, the parsed chunk's `model` (stays `…-TEE`, not mutated to canonical), and `raw_passthrough` all preserved. This is exactly the half-rewritten state the reorder protects against, so a future refactor that reintroduces it will now fail the suite. Local: `inference_providers` lib 300/300 (was 299), `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

2. **`batch_upsert_models` partial-replace churn** — agreed this is pre-existing and out of scope for this comment-polish PR, so I filed it as a follow-up: #775 (with your suggested merge-vs-replace fix and the `mod.rs:3641/:3656` refs). Leaving it untouched here.

Re-requesting review for the approving pass — no behavior change, just the added test.

PierreLeGuen (commented):
Small, well-scoped follow-up (3 files, +55/−10) and the substantive change is correct: `rewrite_sse_event_model` now computes the `raw_bytes` rewrite before mutating the parsed chunk, so a parse failure can no longer leave an event half-rewritten (canonical id in the chunk while `raw_bytes` still leaks the `-TEE` slug), and the new test pins exactly that invariant. The `config` `eprintln!` rationale comment is also accurate — `crates/config/Cargo.toml` has no `tracing` dependency.

One non-blocking observation:

- **`crates/api/src/routes/admin.rs:469-470`** — the corrected comment claims the stale per-provider failure counter "lingers until the next periodic refresh", but no refresh path actually prunes it while the model remains in the DB: `remove_stale_providers` only covers models dropped from the DB entirely, `prune_stale_pinned` skips ids still in the discovery set, and `unregister_provider` is skipped by the very `is_pinned` guard this comment annotates. The entry lingers indefinitely. Since `provider_failure_counts` is keyed by raw Arc pointer address (`crates/services/src/inference_provider_pool/mod.rs:213`), a future provider allocated at a recycled address can inherit a demotion-level count and start deprioritized — for a tiered model, potentially below the Chutes fallback until its first success resets it. Either reword to "lingers until the entry is overwritten or the model is removed", or better, prune `old_provider_ptrs` from `provider_failure_counts` inside `load_inference_url_models`'s atomic update (mod.rs:3601-3608 already builds the ptr set), which removes the residual entirely and makes the comment true. Pre-existing behavior and a comment-only change here, so not blocking — but a PR aimed at comment precision shouldn't trade one overstatement for another.

Local checks run: `cargo test -p inference_providers --lib` (300 passed, 0 failed, including the new `rewrite_sse_event_model_leaves_chunk_bearing_event_untouched_on_rewrite_failure` test), `cargo fmt --all -- --check` clean, `cargo clippy -p inference_providers -p config` no warnings, `cargo check -p api` clean, `git diff --check` clean. CI on the PR head: Lint and security_audit green; Test Suite green at last check (was briefly still in progress earlier). Full e2e suite skipped locally (requires PostgreSQL; covered by CI).

Evrard-Nil:
Thanks Pierre — addressed your note in `00ee7d0c`, and took the *better* option you offered rather than just rewording.

`load_inference_url_models`' atomic update now prunes `provider_failure_counts` for the replaced providers using the same `old_provider_ptrs` set it already uses to prune `pubkey_to_providers` (right after that block). This:
- removes the residual entirely, so the comment is now true rather than trading one overstatement for another;
- closes the latent hazard you flagged — a future provider at a recycled Arc address inheriting a demotion-level count, which for a tiered model could start a fresh NEAR provider below its Chutes fallback;
- makes the field's own `// cleaned up on refresh` doc accurate for the replace-in-place case (`remove_stale_providers` only covered models dropped entirely).

Reworded the `admin.rs:469` comment to match (prunes routing **and** the failure counter → no stale state left behind).

No dedicated test: the inference-url load path requires live attestation/TLS-pinning against real URLs (the unit harness only exercises the *external* path), so the prune is covered by symmetry with the adjacent `pubkey_to_providers` prune and the existing `remove_stale_providers` counter cleanup, plus the 64 pool tests passing. Happy to extract a testable seam if you'd prefer explicit coverage.

`cargo fmt --all --check`, `cargo clippy -p services -p api --all-targets --all-features -- -D warnings`, and `cargo test -p services --lib inference_provider_pool` (64 pass) all clean locally.

PierreLeGuen (changes requested):
The SSE rewrite reorder and its new test, the `config` `eprintln!` rationale comment, and the admin.rs comment rewrite are all solid — the rewrite is now genuinely atomic (`chutes/mod.rs:384-407`), the new test pins the right invariant (including the `as_object_mut` non-object edge case), and the config crate confirmed has no `tracing` dependency. However, the new failure-counter prune has a real problem that needs fixing before merge.

**Important — pinned fallback's failure counter is wiped on every refresh** (`crates/services/src/inference_provider_pool/mod.rs:3617-3620`)

The new `provider_failure_counts` prune is keyed on `old_provider_ptrs`, which over-approximates "replaced providers":

- `reused_provider_ptrs` (mod.rs:3567) is built only from the `inference_url_providers` URL cache (mod.rs:2997-3003). Pinned Chutes providers are registered via `register_pinned_secondary_provider` (mod.rs:679) and never enter that cache.
- After any refresh, `model_to_providers` for a tiered model holds `[NEAR_arc, Chutes_arc]` (merge at mod.rs:3595-3599), so the collection loop (mod.rs:3575-3584) puts the pinned Chutes pointer into `old_provider_ptrs` on **every** periodic refresh — NEAR is excluded as reused, Chutes is not.
- The pre-existing `pubkey_to_providers` prune was a harmless no-op for pinned providers (they never populate pubkey mappings, mod.rs:673-678). The counter prune is not: it resets the Chutes provider's consecutive-failure counter every ~5-minute refresh cycle.

Consequence: a persistently failing Chutes fallback can never stay demoted (demotion needs ≥10 consecutive failures, mod.rs:1464,1483) — each refresh re-promotes it above demoted NEAR providers in selection ordering. That contradicts the PR's "no behavior change in the default posture" claim, and the in-code comment ("replaced providers", "mirrors the pubkey_to_providers prune") mis-describes what the prune actually touches. It also lacks the shared-Arc guard the analogous prune in `prune_stale_pinned` deliberately has (mod.rs:1932-1942), which matters on the admin partial-batch path where one Arc can back a model outside the batch.

Suggested fix: exclude pinned provider pointers when building `old_provider_ptrs` (mirroring the `reused_provider_ptrs` exclusion, sourced from `self.pinned_providers`), or better, follow `prune_stale_pinned`'s pattern and filter the prune set against pointers still referenced by the post-merge `model_to_providers` — that also covers the shared-Arc case.

Minor:

- Same hunk: no test covers the new prune — neither "a replaced provider's counter is gone after refresh" nor "a pinned provider's counter survives refresh." The latter would have caught the issue above; given this PR adds a dedicated test for the SSE invariant, the counter prune (which runs on every production refresh) deserves the same.
- `crates/api/src/routes/admin.rs:463-472`: the rewritten comment ("leaves no stale routing or counter state behind") is accurate for the replaced NEAR provider, but re-check the wording once the prune is fixed so it stays true for the pinned fallback as well.

Checks run locally: `cargo fmt --all -- --check` clean; `cargo clippy -p services -p inference_providers -p config -p api` clean; `cargo test -p inference_providers --lib` 300/300 passed (includes the new SSE atomicity test); `cargo test -p config --lib` 26/26 passed; `cargo test -p services --lib inference_provider_pool` 64/64 passed (none exercise the new counter prune); targeted runs of `rewrite_sse_event_model`, `test_reused_providers_keep_pubkey_mapping`, and `prune_stale_pinned_rebuilds_absent_pinned_only` all passed. CI lint and security audit green. Full e2e/DB-backed suite not run (requires PostgreSQL; the diff touches unit-tested library code only).

Evrard-Nil:
Good catch — you're right, that was a real regression. Fixed in `1f863930`.

The prune now filters against the **post-merge live pointers**, mirroring `prune_stale_pinned`'s shared-Arc guard: a counter is dropped only when its provider is in `old_provider_ptrs` **and** no longer referenced by any model. So:
- a re-merged pinned **Chutes** fallback (in `old_provider_ptrs` every refresh, but still in `model_to_providers`) → **counter kept** → it can stay demoted;
- a genuinely-replaced NEAR provider (gone from `model_to_providers`) → pruned;
- an Arc shared by another still-served model (your partial-batch case) → kept.

`live_ptrs` is built from the full post-merge `model_to_providers` (all models, not just the batch), so the shared-Arc case is covered. Lock order is unchanged (`provider_mappings.write()` then `provider_failure_counts.write()`, same as `unregister_provider`).

Per your minor note, I extracted the logic into a pure `prune_replaced_failure_counts(counts, candidate_ptrs, live_ptrs)` and added `prune_replaced_failure_counts_keeps_live_pinned_and_shared_drops_replaced`, which asserts the pinned/shared counters survive and the replaced one is pruned — **it fails on the previous (00ee7d0c) version**, so it pins exactly the invariant that was broken. Also reworded the `admin.rs` comment so it stays true (the pinned fallback keeps its counter).

`cargo test -p services --lib inference_provider_pool` → 65/65; `cargo clippy -p services -p api --all-targets --all-features -- -D warnings` clean.

lloydmak99 (approved):
Approve. The blocking issue from PierreLeGuen's round-3 review is correctly fixed: the failure counter prune now filters against the post-merge live pointers rather than the over-approximate old_provider_ptrs set. A Chutes fallback's counter is kept as long as it's still referenced by any model in model_to_providers — so a persistently-failing Chutes fallback can now stay demoted across refresh cycles. The fix correctly handles the shared-Arc (partial admin PATCH) case too. The pure helper + unit test that fails on the pre-fix version is the right way to pin this invariant. The admin.rs comment is accurately reworded. 65/65 tests pass.

PierreLeGuen (approved):
Clean, well-scoped follow-up that addresses all four observations from the #768 review. No issues found.

**What I verified:**

- `crates/inference_providers/src/attested/chutes/mod.rs` — the `rewrite_sse_event_model` reorder is correct: every bail point (`from_utf8`, `set_model_in_json`, `String::from_utf8`) now precedes the `chunk` model mutation, so the half-rewritten state (canonical model in `chunk`, slug in `raw_bytes`) is impossible. The new test pins this invariant via a non-object JSON payload. The path is unreachable today (the decoder only sets `chunk` when `raw_bytes` parsed as a valid JSON object), but consistent failure is the right posture.
- `crates/services/src/inference_provider_pool/mod.rs` — the failure-counter prune correctly filters against post-merge live pointers, so a still-served pinned Chutes fallback (which lands in the over-approximate `old_provider_ptrs` every ~5-minute refresh) keeps its demotion counter instead of being reset each cycle. The extracted `prune_replaced_failure_counts` helper is pure and unit-tested across all four cases (replaced/pruned, pinned-live/kept, shared-Arc/kept, untouched/kept), mirroring the shared-Arc guard pattern in `prune_stale_pinned`.
- Lock safety: the prune takes the std `provider_failure_counts` write lock while holding the tokio `provider_mappings` write lock. That nesting already exists in `unregister_provider` (mod.rs:614-618), no path acquires them in reverse order, and the std guard is statement-scoped and never held across an `.await` — no deadlock risk.
- Prune semantics: candidates are removed from `pubkey_to_providers` before the prune, so no provider stays live via pubkey routing with a pruned counter; a pruned counter for in-flight sticky requests simply re-creates on next failure.
- The comment-only changes in `crates/api/src/routes/admin.rs` and `crates/config/src/types.rs` are accurate against the current code.

**Checks run locally:**

- `cargo test -p inference_providers --lib` — 300 passed (includes the new atomic-failure test)
- `cargo test -p services --lib inference_provider_pool` — 65 passed (includes the new prune test; the new case fails on the pre-fix version and passes on the fix)
- `cargo test -p config chutes --lib` — 3 passed
- `cargo clippy -p services -p inference_providers -p config -p api --all-targets -- -D warnings` — clean
- DB-backed e2e suite skipped (requires external PostgreSQL; the diff doesn't touch routes, repositories, or migrations behaviorally, and CI's `security_audit`, `Lint`, and `Test Suite` are green)

PR #772, opened 2026-06-11
test(chutes): live probe confirms inner-[DONE] streaming terminus

Evrard-Nil:
Resolves the open question that gated `CHUTES_ENABLE_STREAMING` (flagged across #758/#768 and in the `e2ee_stream` module note): **does Chutes terminate a stream with an authenticated inner `[DONE]`, or only a forgeable outer plaintext one?**

Our decoder only accepts an **inner** `[DONE]` (decrypted from an `e2e` frame) as a clean terminus — it ignores the outer plaintext `[DONE]` (a gateway could inject it after dropping frames) and treats EOF-without-inner-`[DONE]` as a fatal truncation. So if Chutes only sent the outer one, attested streaming would always error.

**Verified it doesn't.** New `#[ignore]` live probe `live_chutes_streaming_done_probe` streams a real request to `zai-org/GLM-5.1-TEE` over the full E2EE path (stub verifier — testing the *stream protocol*, not re-verifying attestation; the encaps pubkey still comes from the live discovered instance):

```
PROBE: events=22 inner_done=true err=None
PROBE last_event: data: [DONE]
```

22 frames, clean inner-`[DONE]` terminus, no error → **Chutes emits the terminator inside the encrypted channel**, so the decoder completes cleanly. The `e2ee_stream` module note now records this with a *narrowed* claim: the probe only establishes that EOF **without** an inner `[DONE]` is detected (surfaced as a truncation error). It does **not** make streaming tamper-evident — drop, reorder, replay, and truncation-*with*-a-forwarded-inner-`[DONE]` all remain **undetectable** until Chutes adds sequence numbers or a transcript MAC, and stay on the tracked Chutes asks alongside signed measurements.

Run the probe (model overridable via `CHUTES_PROBE_MODEL`, defaults to `zai-org/GLM-5.1-TEE`):
`CHUTES_API_KEY=cpk_... cargo test -p inference_providers --lib attested::chutes::tests::live_chutes_streaming_done_probe -- --ignored --nocapture`

Docs-only + an ignored live test (CI skips it). Flipping `CHUTES_ENABLE_STREAMING` on staging is a separate cvm-ansible vars change (note: a separate measurement-allowlist bug, #773, currently 502s non-H200 Chutes models before any stream starts, so end-to-end staging streaming has not yet been exercised).

### DIFF
--- a/crates/inference_providers/src/attested/chutes/e2ee_stream.rs
+++ b/crates/inference_providers/src/attested/chutes/e2ee_stream.rs
@@ -21,11 +21,16 @@
 //! as a clean terminus; a *plaintext outer* `[DONE]` is forgeable (the gateway
 //! could inject it after dropping frames) and is ignored, so a truncated stream
 //! surfaces an error instead of a fake success. Frame *ordering* is still not
-//! cryptographically guaranteed. **Open (verify on staging):** confirm Chutes
-//! emits the terminator *inside* the encrypted channel; if it only sends the
-//! outer plaintext `[DONE]`, the terminator is forgeable — add it to the tracked
-//! Chutes asks (alongside missing frame sequence numbers + unsigned measurements)
-//! and reconsider exposing streaming as attested.
+//! cryptographically guaranteed.
+//!
+//! **Terminator confirmed (2026-06-11):** a live round-trip against GLM-5.1-TEE
+//! (see the `live_chutes_streaming_done_probe` test in `super`) showed Chutes
+//! emits the terminator *inside* the encrypted channel — the stream ends on an
+//! authenticated inner `[DONE]`, so the inner-terminus assumption holds and
+//! streaming can be exposed as attested. The remaining residual is frame
+//! *reorder* (undetectable without sequence numbers — an accepted limitation;
+//! truncation IS caught). Sequence numbers + signed measurements stay on the
+//! tracked Chutes asks.
 
 use async_stream::try_stream;
 use base64::Engine;
--- a/crates/inference_providers/src/attested/chutes/e2ee_stream.rs
+++ b/crates/inference_providers/src/attested/chutes/e2ee_stream.rs
@@ -26,11 +26,15 @@
 //! **Terminator confirmed (2026-06-11):** a live round-trip against GLM-5.1-TEE
 //! (see the `live_chutes_streaming_done_probe` test in `super`) showed Chutes
 //! emits the terminator *inside* the encrypted channel — the stream ends on an
-//! authenticated inner `[DONE]`, so the inner-terminus assumption holds and
-//! streaming can be exposed as attested. The remaining residual is frame
-//! *reorder* (undetectable without sequence numbers — an accepted limitation;
-//! truncation IS caught). Sequence numbers + signed measurements stay on the
-//! tracked Chutes asks.
+//! authenticated inner `[DONE]`. This narrowly establishes that an EOF *without*
+//! an inner `[DONE]` is detected (surfaced as a truncation error). It does **not**
+//! make streaming tamper-evident: because content frames carry no sequence
+//! numbers, an on-path gateway can still **drop** encrypted content frames and
+//! forward the (authenticated) inner `[DONE]`, or **reorder** / **replay** frames,
+//! and the decoder cannot detect any of it. So drop, reorder, replay, and
+//! truncation-*with*-a-forwarded-inner-`[DONE]` all remain **undetectable** until
+//! Chutes adds sequence numbers or a transcript MAC — these stay on the tracked
+//! Chutes asks. The only guarantee added here is "no inner `[DONE]` ⇒ error".
 
 use async_stream::try_stream;
 use base64::Engine;
@@ -104,8 +108,8 @@ fn handle_outer_payload(
         // frames to fake a successful-but-truncated stream. Ignore it: only the
         // **authenticated inner** `[DONE]` (decrypted from an `e2e` frame, handled
         // in `inner_event`) is a valid terminus. If no inner `[DONE]` arrives, the
-        // stream ends as truncation. (If staging shows Chutes terminates only with
-        // the outer `[DONE]`, that's a forgeable terminator — see the module note.)
+        // stream ends as truncation. Chutes is confirmed to emit the inner
+        // terminator (see the module note), so ignoring the outer one is safe.
         return Ok(None);
     }
     let v: serde_json::Value = match serde_json::from_str(payload) {
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -552,10 +552,11 @@ impl InferenceProvider for Provider {
         _request_hash: String,
     ) -> Result<StreamingResult, CompletionError> {
         // Streaming is off by default: Chutes' stream protocol has no
-        // authenticated frame ordering (a gateway could drop/reorder AEAD-valid
-        // frames undetectably), so it isn't honestly attested until Chutes adds
-        // sequence numbers and the inner-terminator behavior is confirmed on
-        // staging. Opt in with CHUTES_ENABLE_STREAMING. Non-streaming is attested.
+        // authenticated frame ordering (content frames carry no sequence numbers),
+        // so an on-path gateway could drop, reorder, or replay AEAD-valid frames
+        // undetectably. It isn't honestly attested until Chutes adds sequence
+        // numbers or a transcript MAC. Opt in with CHUTES_ENABLE_STREAMING.
+        // Non-streaming is attested.
         if !self.allow_streaming {
             return Err(CompletionError::CompletionError(
                 "Chutes streaming is not enabled as an attested path (frame ordering is not \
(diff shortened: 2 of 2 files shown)


PierreLeGuen (changes requested):
The core claim of this PR checks out: in `decrypt_e2ee_sse`, an outer plaintext `[DONE]` is swallowed (`handle_outer_payload`, e2ee_stream.rs:101-110), the only clean termination is after yielding the decrypted inner done marker (e2ee_stream.rs:197-200), and EOF without it always errors (e2ee_stream.rs:212-215). So the probe's "no error + stream ended" observation genuinely proves Chutes emits the terminator inside the encrypted channel, and the `#[ignore]` gating is correct.

One important issue with how that result is documented, plus a few minor cleanups:

**Important**

- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:26-33` — the new module note overstates what the probe proves. It says streaming can be exposed as attested and that truncation is caught, but without sequence numbers a gateway can still drop encrypted content frames and forward the (authenticated) inner `[DONE]`, or reorder/replay frames, and the decoder cannot detect it. The probe only establishes that EOF *without any* inner `[DONE]` is caught. As written, the note could lead operators to enable `CHUTES_ENABLE_STREAMING` under a false truncation guarantee. Please narrow the claim to "Chutes emits an inner terminator, so EOF without an inner `[DONE]` is detected," and keep drop/reorder/replay/truncation-with-DONE explicitly listed as undetectable until Chutes adds sequence numbers or a transcript MAC.

**Minor**

- `crates/inference_providers/src/attested/chutes/mod.rs:554-558` — the gating comment in `chat_completion_stream` still says streaming isn't honestly attested until "the inner-terminator behavior is confirmed on staging," which this PR's module note now declares confirmed. Reword the gate comment to cite only the remaining residual (no authenticated frame ordering) — this dovetails with the fix above.
- `crates/inference_providers/src/attested/chutes/e2ee_stream.rs:107-108` — the parenthetical in `handle_outer_payload` still phrases the terminator question as open; drop or rephrase it now that the module note resolves it.
- `crates/inference_providers/src/attested/chutes/mod.rs:1110-1113` — the probe sets `inner_done` via `s.contains("[DONE]")` on raw bytes, so model text containing "[DONE]" could false-positive. The conclusion is still sound (the decoder already filters outer `[DONE]`, and `err.is_none()` plus completion implies the inner marker by construction), but `ev.is_done_marker()` is the precise predicate — use it.

**Checks run locally**

- `cargo clippy -p inference_providers --all-targets --all-features -- -D warnings` — no warnings
- `cargo test -p inference_providers --lib` — 299 passed, 1 ignored (the new probe; needs `CHUTES_KEY` + network, so the pasted probe output is taken at face value)
- `cargo test -p inference_providers --lib --no-run` — the `#[ignore]` probe compiles cleanly
- `git diff --check` — clean; single commit, no merge skew
- CI at review time: Lint and claude-review green; Test Suite and security_audit still in progress

Evrard-Nil:
Thanks @PierreLeGuen — addressed all four in f7dc55c7:

**Important** — `e2ee_stream.rs` module note: narrowed the claim. It now states only that the probe establishes "EOF *without* an inner `[DONE]` is detected (surfaced as a truncation error)" and that this does **not** make streaming tamper-evident. Drop, reorder, replay, and truncation-*with*-a-forwarded-inner-`[DONE]` are all explicitly listed as **undetectable** until Chutes adds sequence numbers or a transcript MAC. No more "streaming can be exposed as attested" / "truncation IS caught" overstatement.

**Minor (a)** — `chat_completion_stream` gate comment (`mod.rs`): reworded to cite only the residual (no authenticated frame ordering — content frames carry no sequence numbers, so a gateway could drop/reorder/replay undetectably). Dropped the "confirmed on staging" wording entirely; staging streaming was **not** verified end-to-end (a separate measurement-allowlist bug, #773, 502s non-H200 Chutes models before any stream starts). The honest claim is the direct live probe confirmed the inner-`[DONE]` terminus.

**Minor (b)** — `handle_outer_payload` parenthetical: rephrased as settled — "Chutes is confirmed to emit the inner terminator (see the module note), so ignoring the outer one is safe."

**Minor (c)** — live probe now detects the terminus with `ev.is_done_marker()` (the `pub` predicate on `SSEEvent`) instead of `raw_bytes.contains("[DONE]")`, with a comment noting the decoder already filters the forgeable outer `[DONE]` so the only yielded done-marker is the authenticated inner one.

Also picked up Copilot's adjacent nit: the probe now reads the canonical `CHUTES_API_KEY` env var (matching `config::types`) instead of `CHUTES_KEY`, so the pasted run command works as-is.

Local checks (isolated target dir): `cargo fmt -p inference_providers` clean, `cargo clippy -p inference_providers --lib` no warnings, `cargo test -p inference_providers --lib chutes` → 64 passed, 1 ignored (the new probe).

PierreLeGuen (commented):
This PR is what it claims: documentation updates in `e2ee_stream.rs` and `chutes/mod.rs` recording the result of a live probe, plus one `#[ignore]`d live test `live_chutes_streaming_done_probe` that CI skips. No production code paths change.

I verified the doc claims against the actual decoder behavior: `handle_outer_payload` (e2ee_stream.rs:105-113) ignores the plaintext outer `[DONE]`, only an inner decrypted `[DONE]` yields a done event, and EOF without one errors as truncation (e2ee_stream.rs:216-219). The updated module note correctly *narrows* the security claim (drop/reorder/replay/truncation-with-forwarded-inner-`[DONE]` remain undetectable) rather than overstating what the probe proved — the right posture for this codebase. The test's use of `is_done_marker()` (sse_parser.rs:36-44) is sound: it checks `raw_bytes` for the `data: [DONE]` line and returns false for content chunks, and since the decoder filters the forgeable outer `[DONE]`, the only done-marker the stream can yield is the authenticated inner one — no false positive on model text. The probe also fails in the safe direction long-term: if Chutes ever stops emitting the inner terminator, production streams fail closed with a truncation error rather than silently succeeding.

Two minor, non-blocking notes:

- **PR description (not in the diff):** the body's run command uses `CHUTES_KEY=cpk_...`, but the test at crates/inference_providers/src/attested/chutes/mod.rs:1085 reads `CHUTES_API_KEY` and panics otherwise — copy-pasting the documented command fails immediately. The in-code doc comment and `#[ignore]` reason are already correct; just edit the PR body.
- **crates/inference_providers/src/attested/chutes/mod.rs:1087:** the probe hardcodes `zai-org/GLM-5.1-TEE`, so it silently rots when that model is decommissioned and re-running the verification requires a code edit. Consider reading the model from an env var with this value as the default, e.g. `std::env::var("CHUTES_PROBE_MODEL").unwrap_or_else(|_| "zai-org/GLM-5.1-TEE".into())`.

Local checks run:
- `cargo clippy -p inference_providers --all-targets -- -D warnings` — clean
- `cargo check -p inference_providers --all-targets` — passed
- `cargo test -p inference_providers --lib` (chutes tests) — all passed, 1 ignored (the live probe; needs network + `CHUTES_API_KEY`, so it was not executed live)
- `git diff --check` — clean
- CI: Lint, security_audit, and Test Suite all green

Evrard-Nil:
Thanks @PierreLeGuen — both addressed.

- **PR description run command:** fixed the body to use `CHUTES_API_KEY=cpk_...` (matching the in-code doc comment and `#[ignore]` reason). Also tightened the body's prose to match the narrowed module note — dropped the old "streaming can be exposed as attested" / "truncation IS caught" wording and now list drop/reorder/replay/truncation-with-forwarded-inner-`[DONE]` as the undetectable residual, and noted the staging flip hasn't been exercised e2e (the #773 measurement-allowlist bug 502s non-H200 Chutes models before any stream starts).

- **Hardcoded probe model (mod.rs:1087):** done in f58ce185 — the model now reads `CHUTES_PROBE_MODEL` and falls back to `"zai-org/GLM-5.1-TEE"`, applied to both the `Config` and the request `"model"` so they stay in sync. Re-running the verification after that model is decommissioned no longer needs a code edit.

Local checks (isolated target dir): `cargo fmt -p inference_providers` clean, `cargo clippy -p inference_providers --lib` no warnings, `cargo test -p inference_providers --lib chutes` → 64 passed, 1 ignored (the probe).

PierreLeGuen (approved):
This is a documentation-plus-test change with no production code paths modified: comment updates in `crates/inference_providers/src/attested/chutes/e2ee_stream.rs` and one `#[ignore]`d live probe test (`live_chutes_streaming_done_probe`).

Verified by inspection:

- The probe's assertion logic is sound. `handle_outer_payload` (e2ee_stream.rs:105-113) returns `Ok(None)` for the outer plaintext `[DONE]`, and `done_event()` is only synthesized from a decrypted inner frame — so a true `is_done_marker()` in the test necessarily came through the authenticated channel. Using `is_done_marker()` (requires `chunk: None` plus an exact trimmed `data: [DONE]` line) rather than a raw-bytes scan correctly avoids false positives from model text containing "[DONE]".
- The updated module note (e2ee_stream.rs:26-37) is precisely scoped — it dates the observation, names the model, and keeps drop/reorder/replay/truncation-with-forwarded-`[DONE]` listed as undetectable — and matches the decoder's actual behavior, including the fail-closed EOF-without-inner-`[DONE]` error at e2ee_stream.rs:216-219.
- All issues from prior review rounds are addressed in the final state: narrowed security claims, the `is_done_marker()` predicate, the `CHUTES_KEY` → `CHUTES_API_KEY` fix (PR body now matches), and the `CHUTES_PROBE_MODEL` override.
- The probe's `eprintln!` of the last raw event is local-only test output under `--nocapture`, not production logging.

One non-blocking note: the inline comment at e2ee_stream.rs:111-112 says Chutes "is confirmed to emit the inner terminator" based on a single live run against `zai-org/GLM-5.1-TEE`. A Chutes model that emitted only the outer `[DONE]` would fail every stream with a truncation error — fail-closed, so not a correctness risk, but consider scoping the comment like the module note does ("confirmed on GLM-5.1-TEE, 2026-06-11") and re-running the probe via `CHUTES_PROBE_MODEL` when onboarding new Chutes models before enabling `CHUTES_ENABLE_STREAMING`.

Checks run locally:
- `cargo test -p inference_providers --lib` — 299 passed, 0 failed, 1 ignored (the new live probe, as intended)
- `cargo clippy -p inference_providers --all-targets -- -D warnings` — clean
- `git diff --check` — clean
- Live probe not run (needs `CHUTES_API_KEY` + network); the PR body's captured output (`events=22 inner_done=true err=None`) plus code inspection cover it
- CI: Lint and security_audit green; Test Suite still pending at review time, but the diff is confined to the crate tested locally

lloydmak99 (approved):
Approve. The live probe correctly answers the gating question: Chutes emits the inner [DONE] inside the encrypted channel, confirming that EOF without an inner terminus surfaces as a truncation error. The module note is appropriately scoped — it accurately states the remaining undetectable failure modes (drop, reorder, replay, truncation-with-forwarded-inner-[DONE]) and that these await sequence numbers or a transcript MAC. The #[ignore] test is the right pattern for a network-dependent live probe. PierreLeGuen approved after two rounds.

Issue #646, opened 2026-05-22
Streaming default emits 'usage' field in every chunk; should be null until stream_options.include_usage=true


Evrard-Nil:
## Summary

`POST /v1/chat/completions` with `stream: true` (and no `stream_options`) includes a populated `usage` block in **every** SSE chunk. Per OpenAI spec, the default should be no usage (or `usage: null`) in chunks unless the caller opts in via `stream_options.include_usage: true`, in which case only the final chunk before `[DONE]` carries the populated usage.

Three behaviors observed vs OpenAI spec:

| Request | OpenAI behavior | cloud-api |
|---|---|---|
| `stream: true` (no `stream_options`) | `usage: null` in each chunk, no final usage chunk | **`usage: {populated}` in every chunk** ❌ |
| `stream: true, stream_options: {"include_usage": false}` | same as default | `usage` key omitted entirely from chunks ❌ |
| `stream: true, stream_options: {"include_usage": true}` | `usage: null` in each chunk, one final chunk with populated usage | only final chunk has `usage` (other chunks have no `usage` key) ⚠️ partially compliant |

## Repro

```bash
KEY=sk-32c0476395fd40c795725fc101f33304

# Default streaming — every chunk has a non-null usage block
curl -sN https://cloud-api.near.ai/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"openai/gpt-oss-120b","messages":[{"role":"user","content":"hi"}],"max_tokens":50,"stream":true}' \
  | grep -c '"usage":{"prompt_tokens"'
# ~37 (one per chunk)

# With include_usage:true — only the final chunk has usage
curl -sN ... '{"stream_options":{"include_usage":true},...}' | grep -c '"usage":{"prompt_tokens"'
# 1 (final chunk only — correct)
```

Sample chunk shapes (`/tmp/log` above):
```
# Default — every chunk:
{"choices":[{"delta":{"reasoning":"The"}}],"usage":{"prompt_tokens":74,"completion_tokens":4,"total_tokens":78}}

# include_usage:false (would be expected as default):
{"choices":[{"delta":{"reasoning":"The"}}]}    # no usage key
```

## Why it matters

1. **Bandwidth**: every chunk carries ~80 extra bytes of usage info. For long streams (1k-token responses → ~250 chunks) that's ~20 KB of redundant data.
[...]

PR #777, opened 2026-06-11
Fix streaming chat include_usage chunks

lloydmak99:
## Problem

`POST /v1/chat/completions` streaming responses need an OpenAI-compatible `stream_options.include_usage=true` path: intermediate chunks should expose `usage: null`, and the client should receive one final populated usage chunk before `[DONE]`.

The earlier version of this PR also normalized default public streams for attestation-supported models. Pierre flagged that as a signer-contract change because it replaced the provider TEE response signature with a gateway signature on the default streaming path.

Refs #646. Default attested streams intentionally remain byte-exact passthrough in this PR; fully normalizing that path still needs the separate security/product decision Pierre called out.

## Changes

- Narrow public stream usage rewriting to explicit `stream_options.include_usage=true` requests.
- Preserve raw passthrough for default streams, `include_usage=false`, `continuous_usage_stats=true`, E2EE streams, non-text modality streams, and streams whose cached model metadata is unavailable.
- Keep provider-side chat signature storage on the default attested streaming path.
- For explicit rewritten attested streams, store a gateway signature over the exact SSE bytes returned to the client, including forwarded comments, synthetic final usage, and `[DONE]`.
- Treat `stream_options.continuous_usage_stats=true` as a passthrough opt-out instead of a 400.
- Preserve unknown vendor keys inside well-formed `stream_options` with a flattened extra map.
- Forward rewritten SSE comment events with their blank-line separator so line-splitting clients do not glue comments to the next event.
- Move gateway signature creation metrics off the verification metric names and dedupe the signature creation log line.
- Keep malformed `stream_options` values in flattened `extra` so upstream/provider validation still sees them.
- Keep DB-backed e2e coverage for default-stream billing, and keep include-usage coverage for the rewritten public response/signature path.

## Security / Verification Note

Default attested streaming responses remain byte-exact provider passthrough and keep the provider TEE signing behavior.

Only explicit `stream_options.include_usage=true` public text streams are rewritten. For attestation-supported models on that opt-in path, the gateway signs the exact rewritten response bytes. Verification clients that use `include_usage=true` must validate the gateway signer for that response; clients that need provider-key verification should use the default passthrough path or `continuous_usage_stats=true`.

## Validation

Passed locally:

- `cargo test --locked -p api --lib stream_options`
- `cargo test --locked -p api --lib include_usage`
- `cargo test --locked -p api --lib continuous_usage_stats_opts_out_of_rewrite`
- `cargo test --locked -p api --lib rewritten_control_events`
- `cargo test --locked -p api --lib test_chat_completion_request_conversion`
- `cargo test --locked -p services --lib extract_stream_options_consumes_typed_options`
- `cargo test --locked -p inference_providers --lib stream_options`
- `cargo test --locked -p api --lib`
- `cargo test --locked -p services --lib`
- `cargo test --locked -p inference_providers --lib`
- `cargo check --locked --all-targets`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`
- `git diff --check`

DB-backed e2e tests were compile-checked by `cargo check --locked --all-targets`; this local environment still does not have the Postgres test database required to run them end to end.

## Rollout Notes

- Default attested streaming behavior and provider signatures are preserved.
- `include_usage=true` on public text streams now opts into normalized final-usage chunks and gateway signing on attestation-supported models.
- `continuous_usage_stats=true` is passthrough and keeps existing provider behavior.
- Billing still sees upstream usage before any public response shaping.
- Partial/error rewritten streams do not emit a synthetic final usage chunk and do not store a gateway signature.
- No schema migration or config change is needed.


lloydmak99:
Addressed the review findings in `742b6cd` and `b6b0dcf`.

- Preserve byte-exact passthrough for attested/verifiable chat streams, so the existing provider-signed streaming signature path is not invalidated by gateway reserialization.
- Limit public usage rewriting to non-E2EE, non-attested, text-only streams.
- Store a gateway signature for rewritten public streams using the gateway-emitted SSE bytes, so rewritten non-attested streams still have a matching chat signature.
- Preserve usage on converter-backed terminal choice chunks when `stream_options.include_usage=true`; dedicated usage-only chunks are still supported, and default streams continue to expose `usage:null` while dropping provider usage-only chunks.

Validation:
- `cargo test --locked -p api --lib include_usage_keeps_only_final_usage_chunk_populated`
- `cargo test --locked -p api --lib chat_stream`
- `cargo check --all-targets`
- `git diff --check`

I also tried the two CI-failing e2e targets locally:
- `cargo test --locked -p api --test e2e_all usage_chat_completions::test_chat_completions_stream_include_usage_true_emits_final_usage_only`
- `cargo test --locked -p api --test e2e_all signature_verification::test_streaming_chat_completion_signature_verification`

Both failed before running application code because local Postgres admin DB bootstrap could not connect (`Connection refused`).

lloydmak99:
Pushed `5f6dcb4` for the failed Test Suite run on `0ad1faf`.

Root cause:
- The public stream rewrite path preserved usage on a terminal choice chunk, then also preserved a later dedicated usage-only chunk.
- That made `usage_chat_completions::test_chat_completions_stream_include_usage_true_emits_final_usage_only` see two final usage objects.

Fix:
- Added per-stream usage state to the public rewrite path.
- The first final usage-bearing chunk is kept.
- Later duplicate usage-only chunks are dropped, and later duplicate choice-bearing usage chunks are serialized with `usage:null`.
- The converter-backed terminal-choice usage regression remains covered by the new unit test.

Validation:
- `cargo test --locked -p api --lib include_usage_drops_duplicate_usage_only_chunk_after_terminal_usage`
- `cargo test --locked -p api --lib chat_stream`
- `cargo check --all-targets`
- `git diff --check`

I retried the exact failing e2e locally, but this environment still cannot connect to the local Postgres admin DB (`Connection refused`) before application code runs. CI should exercise it on the pushed head.

lloydmak99:
Final review-loop status:

- Addressed the automated review finding for converter-backed providers by preserving final usage on terminal choice chunks.
- Addressed the CI duplicate-usage failure by tracking final usage per rewritten stream and suppressing later duplicate usage chunks.
- Addressed the stream-signature mismatch by signing the exact gateway-emitted SSE bytes for rewritten text streams.
- Kept E2EE and non-text modality streams on raw passthrough.

Latest GitHub checks on `5f6dcb4` are green:

- Lint
- Test Suite
- security_audit

Requested review from Evrard and Pierre.

PierreLeGuen (changes requested):
This is a worthwhile fix for the streaming usage contract, but the signature-handling path and a couple of usage-shaping edge cases need work before merge.

**Blocking / important**

1. **Redundant and racy signature storage, and `[DONE]` now blocks on it** — `crates/api/src/routes/completions.rs:1640-1687` + `crates/services/src/completions/mod.rs:109-150`. With rewriting now the default for public text chat streams, the service layer still fetches and stores the inference-TEE signature (two upstream round trips plus DB writes, up to 5s), and the route tail then overwrites it with the gateway signature via the `add_chat_signature` upsert — after *also* awaiting the store before emitting `data: [DONE]`. Consequences: worst-case `[DONE]` latency is now ~10s (5s TEE fetch + 5s gateway store), every stream appears hung at completion for DB/signing latency, and if the gateway store fails or times out (the branches at completions.rs:1664-1678 only log), the DB permanently retains the TEE signature whose `response_hash` covers the original upstream bytes — `/v1/signature/{chat_id}` then silently fails verification against the rewritten bytes the client actually received. Fix: skip `store_chat_signature_from_provider` when the route will store a gateway signature (plumb a flag through the service request), compute the hash synchronously, and spawn the timeout-wrapped store like `responses.rs` does so the `[DONE]` bytes return immediately.

2. **Final usage can be an undercounted mid-stream snapshot** — `crates/api/src/routes/completions.rs:368-401`. The NEAR provider forces `continuous_usage_stats: true` upstream (`crates/inference_providers/src/attested/nearai/mod.rs:1405`), so the finish_reason chunk always carries a usage snapshot, and `prepare_chat_stream_chunk_for_client_with_state` treats that chunk as the final usage carrier while dropping vLLM's authoritative usage-only chunk. With `n>1`, choice 0's finish_reason arrives while other choices are still generating, so the client's only populated usage undercounts what billing records. Strip usage from all choice-bearing chunks, track the last usage seen, and synthesize the canonical empty-choices usage chunk in the stream tail before `[DONE]` — this also covers converter providers that never emit a usage-only chunk.

3. **Client-requested `continuous_usage_stats` is silently ignored** — `crates/api/src/routes/completions.rs:331-343,368-401`. A client sending `stream_options: {include_usage: true, continuous_usage_stats: true}` now gets intermediate usage nulled because shaping only consults `include_usage`, even though the public `StreamOptions` type models the field and the Chutes provider explicitly preserves it (chutes/mod.rs:1004). Either honor it (skip intermediate stripping) or reject/strip it at the request boundary rather than accepting a no-op option.

4. **Stored gateway signatures are unreachable for some providers** — `crates/services/src/attestation/mod.rs:569`. `get_chat_signature` returns `SIGNATURE_UNSUPPORTED` before checking the repository when the mapped provider has `supports_chat_signatures() == false`. The PR now writes gateway-generated signatures for public streams, but Chutes-style providers make them unreachable. Check the repository first, or only return unsupported when no stored signature is found.

**Non-blocking**

- `crates/api/src/routes/completions.rs:1298-1299`: `rewrite_public_stream_usage` doesn't check whether the provider is attested, so gateway signatures are now stored for non-attested providers (OpenAI/Anthropic/Gemini) where none previously existed, changing what `/v1/signature/{chat_id}` presence implies, and `METRIC_VERIFICATION_SUCCESS` is inflated (it also fires twice per attested stream). Gate on `attestation_supported` unless this is intended.
- `crates/api/src/routes/completions.rs:1437,1577-1581`: `public_signature_bytes` buffers the whole response stream just to hash it. Use an incremental `Sha256` hasher updated per chunk — same hash, O(1) memory.
- `crates/api/src/routes/completions.rs:1494`: with rewriting active, upstream SSE control events (keepalive comments, `event:`/`id:` lines, parsed as `chunk: None`) are silently dropped where passthrough forwarded them; vanishing keepalives can trip idle timeouts on long generations. Consider forwarding `event.raw_bytes` for these and including them in the signed hash.
- `/v1/completions` text streams still leak the forced continuous usage on every chunk (`nearai/mod.rs:1666`) — the same contract problem this PR fixes for chat. Fine as a follow-up, but worth tracking.

**Checks run locally:** `cargo fmt --all -- --check` passed; `cargo test -p api --lib` (163 passed), `cargo test -p services --lib` (386 passed), `cargo test -p inference_providers --lib` (300 passed), including the new chunk-shaping, `extract_stream_options`, and attestation tests; `cargo clippy -p api --all-targets --all-features -- -D warnings` clean. DB-backed e2e tests couldn't run locally (no PostgreSQL / dstack signing socket); CI on the head commit is green across Lint, Test Suite, and security_audit.

lloydmak99:
Pushed `a8831a9` to address Pierre's latest requested changes.

Changes:
- Skips provider-side `store_chat_signature_from_provider` when the route rewrites public stream bytes and stores a gateway signature instead.
- Computes the gateway response hash synchronously over the emitted SSE bytes, then spawns the timeout-wrapped signature store so `[DONE]` is not held behind DB/signing latency.
- Changes public usage shaping to "last usage wins": all emitted provider chunks serialize `usage:null`, usage-only provider chunks are suppressed, and the route emits one canonical empty-choices final usage chunk before `[DONE]`.
- Rejects client-requested `stream_options.continuous_usage_stats=true` with a 400 instead of accepting a no-op option.
- Looks up stored chat signatures before returning `SIGNATURE_UNSUPPORTED`, so gateway signatures remain reachable for providers that do not expose provider-side signatures.

Validation:
- `cargo check --all-targets`
- `cargo test --locked -p api --lib chat_stream`
- `cargo test --locked -p api --lib include_usage`
- `cargo test --locked -p api --lib`
- `cargo test --locked -p services --lib`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`
- `git diff --check`

I retried `cargo test --locked -p api --test e2e_all usage_chat_completions::test_chat_completions_stream_include_usage_true_emits_final_usage_only`, but local Postgres is still unavailable here (`Connection refused` during admin DB bootstrap), so CI is the DB-backed gate for that target.

lloydmak99:
CI update for `a8831a9`: all required checks are green.

- Lint
- Test Suite
- security_audit

Review remains requested from Evrard and Pierre.

PierreLeGuen (changes requested):
This PR's usage-shaping logic is sound and well-tested, but routing every default (non-E2EE, text-only) chat stream through the rewrite path changes signature semantics and adds a memory cost on the hottest path. A few things need to be addressed before merge.

**Blocking / important**

1. **Signature availability race** — `crates/api/src/routes/completions.rs:1715`. The gateway chat signature is stored via a fire-and-forget `tokio::spawn` while the final SSE bytes (including `[DONE]`) are flushed. The previous path (`crates/services/src/completions/mod.rs:108-116`) explicitly awaited `store_chat_signature_from_provider` in the `Finalizing` state so the signature was guaranteed available when the client received `[DONE]`. A client calling `/v1/signature/{chat_id}` immediately after stream completion can now get a NotFound for a successful stream. Await `store_chat_signature` (with the existing 5s timeout) in the tail future before yielding the final bytes, and cover with a no-sleep signature-lookup test.

2. **Signature path not gated on attestation** — `crates/api/src/routes/completions.rs:1307-1309`. `rewrite_public_stream_usage` (and therefore gateway-signature storage) depends only on `!e2ee_active && text-only`, with no check of `model.attestation_supported`. Two consequences: (a) for attested models, the inference-TEE per-response signature is no longer fetched/stored on the default streaming path — clients verifying `signing_address` against the inference TEE's attestation will break or silently verify against the gateway key; (b) non-attested providers (OpenAI/Anthropic/Gemini), which previously returned `SIGNATURE_UNSUPPORTED`, now get a stored gateway signature, which can overclaim integrity to verifying clients. Gate `public_signature_enabled` on `attestation_supported`, or record signer provenance in the stored signature/response, and document the signer change.

3. **Unbounded memory buffering** — `crates/api/src/routes/completions.rs:1447,1593-1596,1700-1706`. `public_signature_bytes` accumulates the entire response's SSE bytes in a `Vec<u8>` solely to compute one final SHA-256. That's O(response size) memory per stream on the default chat path; under concurrent long generations this is an availability risk. Replace with an incremental `Sha256` hasher (`update()` per chunk, `finalize()` in the tail) — constant memory, same digest.

4. **Control-only SSE events dropped in rewrite mode** — `crates/api/src/routes/completions.rs:1494-1507`. `event.chunk?` short-circuits on events with no chunk, so upstream keep-alive comments and `event:`/`id:` lines are no longer forwarded (raw passthrough previously preserved them). Clients relying on pings for liveness during long generation pauses may hit idle timeouts. Forward control events as raw bytes (and include them in the signed hash), or emit a gateway keep-alive.

5. **Synthetic usage chunk may under-report with `n > 1`** — `crates/api/src/routes/completions.rs:1642-1656`. The final usage chunk captures usage from the chunk that carried it upstream, but with multiple choices, choice 0's finish-reason chunk can arrive while other choices are still streaming, so the captured usage is a mid-stream snapshot. Either wait for all choices to finish or use the authoritative trailing usage-only chunk vLLM emits.

**Non-blocking**

- `crates/api/src/routes/completions.rs:1277-1287` — rejecting `stream_options.continuous_usage_stats` with a 400 is a client-visible behavioral change (previously forwarded upstream; the Chutes provider explicitly preserved it). Defensible, but make sure it's called out in the rollout notes.
- `crates/services/src/attestation/mod.rs:405-424` — `store_gateway_signature` now increments `METRIC_VERIFICATION_SUCCESS` for a different event than dashboards previously measured, and emits two near-duplicate `info!` lines per stream. Consider a distinct metric/tag and fixing the log.
- `crates/api/src/routes/completions.rs:1700` — skipping signature storage when `error_count_final != 0` leaves partially-delivered streams without a verifiable signature; worth documenting in the `/v1/signature/{chat_id}` contract.

The repository-first lookup reordering in the signature route is correct, and removing `skip_serializing_if` on `usage` matches OpenAI's streaming shape with no deserialization regressions.

**Checks run locally**: `cargo fmt --all -- --check` and `git diff --check` clean; `cargo test -p api --lib` (164 passed, incl. new `chat_stream_*` tests), `cargo test -p services --lib` (386 passed), `cargo test -p inference_providers --lib` (300 passed); `cargo clippy -p api -p services -p inference_providers --all-targets` clean; the streaming-signature e2e compiles but DB-backed e2e tests were skipped locally (no PostgreSQL) — GitHub CI (Lint, Test Suite, security_audit) is green on the PR head.

lloydmak99:
Pushed `1681035` after reviewing the latest requested changes.

I took the conservative signature-preserving route instead of extending the gateway chat-signature path:

- Removed the route-level gateway chat-signature storage path and the service flag that skipped provider signature storage.
- Public stream usage rewriting now runs only for non-E2EE, non-attested, text-only chat streams.
- Attested/verifiable streams stay byte-exact and continue through the existing `store_chat_signature_from_provider` path, so clients keep the inference-TEE signature semantics.
- Non-attested rewritten streams do not mint gateway chat signatures, preserving the prior `SIGNATURE_UNSUPPORTED` behavior instead of overclaiming integrity.
- Rewrite mode now forwards control-only SSE events as raw bytes; it only holds upstream `[DONE]` when the tail must emit a synthetic final usage chunk first.
- The usage shaper still uses last-usage-wins and emits one canonical empty-choices usage chunk, so a trailing provider usage-only chunk wins over earlier choice snapshots.
- Added a dedicated non-verifiable Qwen-like e2e fixture for the stream-shaping tests so the tests cover the rewrite path without mutating the shared verifiable Qwen model row.

Rollout note: the earlier 400 for `stream_options.continuous_usage_stats=true` remains intentional and client-visible; callers should use `stream_options.include_usage` for the single final usage chunk.

Validation:

- `cargo check --all-targets`
- `cargo test --locked -p api --lib chat_stream`
- `cargo test --locked -p api --lib include_usage`
- `cargo test --locked -p services --lib`
- `cargo test --locked -p api --lib`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`
- `git diff --check`

I also retried `cargo test --locked -p api --test e2e_all usage_chat_completions::test_chat_completions_stream`, but this environment still cannot connect to the local Postgres admin DB (`Connection refused` during test DB bootstrap). The e2e target compiled before failing at DB bootstrap, so CI remains the DB-backed gate.

lloydmak99:
CI update for `1681035`: all required checks are green.

- `security_audit`: passed
- `Lint`: passed
- `Test Suite`: passed

Review remains requested from Evrard and Pierre.

PierreLeGuen (changes requested):
The chunk-shaping logic itself (`prepare_chat_stream_chunk_for_client_with_state`, synthetic final usage chunk before `[DONE]`) is correct and well unit-tested, and the E2EE / non-text passthrough paths are preserved. But the last commit ("Preserve stream signature semantics") scoped the fix down in a way that leaves the originally reported bug unfixed for the platform's main model class, and several pieces of the PR no longer match what the code does.

**Blocking**

- `crates/api/src/routes/completions.rs:1296-1303` — usage normalization is gated on `!model_attestation_supported`, but `attestation_supported` defaults to `true` for vLLM models (`crates/database/src/repositories/model.rs:535-538`) and the NEAR attested provider force-sets `continuous_usage_stats: true` upstream (`crates/inference_providers/src/attested/nearai/mod.rs:1405-1408`). So attested-model streams still emit populated `usage` on every chunk — exactly the behavior reported in #646 (repro model `openai/gpt-oss-120b` is a TEE/vLLM model). The PR's own e2e tests had to create a special `attestationSupported: false` model to observe the fix. Either normalize all non-E2EE text chat streams (re-signing rewritten attested streams over the exact emitted SSE bytes, with coverage via the existing verifiable `setup_qwen_model` path), or state the trade-off explicitly, re-scope/keep #646 open, and document which models are actually affected. As written, "Closes #646" is incorrect.

- PR description / `crates/services/src/attestation/mod.rs:570-573` — the description claims rewritten streams get a gateway chat signature over the public response hash, but that mechanism was deleted in commit 1681035 (`compute_sha256`, `skip_provider_chat_signature`, public-signature storage are gone; `store_gateway_signature` only serves the pre-existing Responses path). The comment justifying the signature-lookup reorder references behavior that no longer exists, and the reorder now adds a repository query before every `SIGNATURE_UNSUPPORTED` short-circuit for no reason. Update the description and either revert the reorder or fix the comment.

- `crates/api/src/routes/completions.rs:1296-1301` — `model_attestation_supported` is fetched via `get_model_by_name`, an uncached per-request DB query (`crates/services/src/models/mod.rs:98-104`) run on every chat completion including non-streaming where it's unused. The `unwrap_or(false)` failure default means a transient DB error on an attested model silently enables rewriting and breaks byte-exact response-hash verification for that response. Fail toward passthrough (don't rewrite on lookup failure) and use the cached model list, as `resolve_alias_cached` already does.

- `crates/api/src/routes/completions.rs:1265-1275` — the new unconditional 400 for `stream_options.continuous_usage_stats: true` is an undocumented breaking change and applies even to paths this PR deliberately keeps on raw passthrough, where the option previously worked — the Chutes provider explicitly merges and preserves it (`crates/inference_providers/src/attested/chutes/mod.rs:430-444`, test at line 1004), and that path is now unreachable for streams. Restrict the rejection to requests where `rewrite_public_stream_usage` is true, or document the removal; either way add a route-level test.

**Non-blocking**

- `crates/api/src/conversions.rs:89-97` — the `From<ChatCompletionRequest>` impl removes `stream_options` from `extra` unconditionally, so a malformed value is silently dropped rather than forwarded for upstream validation, inconsistent with `extract_stream_options_from_extra` (`crates/services/src/completions/mod.rs:304-314`), which deliberately preserves malformed values. Only remove the key on successful parse.
- No test pins the new invariant that attestation-supported models stay on raw passthrough. Since the last commit exists specifically to protect signature semantics, a route-level or e2e test there would prevent a silent regression.

One note in the PR's favor: the typed `stream_options` field on `ChatCompletionParams` serializes correctly through `serde_json::to_value(params)`, so Chutes still sees `stream_options` after the service-layer extraction from `extra` — that part of the plumbing is sound.

**Local checks:** `cargo fmt --all -- --check` clean; `cargo clippy -p api --all-targets` no warnings; `cargo test -p api --lib` 164 passed (incl. new `chat_stream_*` / `include_usage_*` tests); `cargo test -p services --lib` 386 passed (incl. `extract_stream_options_*`); `cargo test -p inference_providers --lib` 300 passed; `git diff --check` clean. DB-backed e2e suite skipped locally (no PostgreSQL); GitHub Actions Lint / Test Suite / security_audit all green on the head commit.

lloydmak99:
Pushed `b870374` addressing the latest review.

What changed:

- Normalization is back on for all known, non-E2EE, text-only chat streams, including attestation-supported vLLM/TEE models. The e2e stream-shaping tests are back on the normal verifiable `setup_qwen_model` path.
- Rewritten attested streams now skip provider-side chat signature storage and store a gateway chat signature over the exact SSE bytes returned to the client. The hash is incremental, includes forwarded control events and synthetic tail bytes, and the store is awaited with the existing 5s timeout before final tail bytes are yielded.
- Model metadata for the rewrite decision now comes from the cached model list after alias resolution. Cache lookup failure or missing metadata fails toward raw passthrough.
- `continuous_usage_stats=true` is rejected only when the API is rewriting the public text stream. E2EE, non-text, and metadata-missing passthrough paths are not rejected; added route-level unit coverage for that invariant.
- Malformed `stream_options` are now preserved in `extra` unless typed parsing succeeds.
- The PR description and rollout notes now describe the gateway-signing behavior and passthrough cases accurately.

Local validation:

- `git diff --check`
- `cargo test --locked -p api --lib chat_stream_usage_mode`
- `cargo test --locked -p api --lib continuous_usage_stats`
- `cargo test --locked -p api --lib chat_stream`
- `cargo test --locked -p api --lib include_usage`
- `cargo test --locked -p api --lib malformed_stream_options_remain_in_extra`
- `cargo test --locked -p api --lib`
- `cargo test --locked -p services --lib`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`

I also tried `cargo test --locked -p api --test e2e_all usage_chat_completions::test_chat_completions_stream`; it compiled, but the local test environment has no Postgres admin DB listening, so the matched tests failed during bootstrap with `Connection refused`.

PierreLeGuen (changes requested):
This PR correctly implements the OpenAI final-usage contract from #646, and extracting `stream_options` into a typed struct fixes a latent duplicate-JSON-key hazard where a client-supplied `stream_options` in `extra` could clobber the gateway-forced `include_usage` upstream. The streaming logic itself is sound, but there are a few issues to resolve before merge.

**Blocking / important**

- `crates/api/src/routes/completions.rs:1336` + `crates/services/src/attestation/mod.rs:328` — `skip_provider_chat_signature = gateway_signature_enabled` means that for every public (non-E2EE) streaming chat on an attestation-supported model, the inference TEE's per-response signature is no longer fetched or stored; the gateway signs `request_hash:response_hash` with its own keys. This reverses the design documented in `store_chat_signature_from_provider` ("avoids gateway-side signature synthesis and ensures signing_address reflects the provider/model identity") and disables #701's byte-exact provider-TEE verifiability on this path. `signing_address` now differs between streaming (gateway key) and non-streaming (provider key) responses for the same model, which will break verification clients that pin the provider signing address. The PR body acknowledges the trade-off, but please get explicit security/product sign-off recorded on the PR and document the gateway-signed mode for verification clients (which key, how to attest it against the gateway's attestation report). If a narrower rewrite scope is feasible, consider it.
- `crates/api/src/routes/completions.rs:1542-1768` — the core invariant of the new gateway-signature path (stored `response_hash` == sha256 of the exact SSE bytes the client received, including control events, alias warnings, auto-redact flushes, the synthetic usage chunk, and `[DONE]`) has no automated coverage. The hashing is spread across two closures plus the tail, and a silent regression would produce permanently unverifiable responses. Add an e2e or route-level test with a mock provider that hashes the received bytes and asserts they match `GET /v1/signature/{chat_id}`; cover the alias-warning and `include_usage` variants.
- `crates/api/src/routes/completions.rs:1685-1750` — the tail appends the synthetic populated usage chunk and `[DONE]` before `stream_error_count` is checked, so a rewritten `include_usage=true` stream that already emitted SSE error frames still sends an ordinary-looking final usage object built from the last partial snapshot. That makes failed/partial streams look successfully accounted to clients and violates the final-usage contract. Load `error_count_final` before this block and skip the synthetic usage chunk when it's nonzero (matching the existing rule that error streams don't get signatures); add a test for a stream that emits usage-bearing chunks and then errors.

**Non-blocking**

- `crates/api/src/routes/completions.rs:1726-1733` — when `include_usage=true` but no provider usage was observed, the final usage chunk is silently omitted (the exact contract violation this PR fixes) and the `(None, _)` arm doesn't log. At minimum warn there; `InterceptStream`'s `last_usage_stats` could serve as a fallback source.
- `crates/api/src/routes/completions.rs:1550-1567` — bare blank-line control events from the lossless parser are forwarded raw while the gateway also emits `\n\n` framing per serialized chunk, producing duplicated blank lines between SSE events. Spec-legal and absorbed into the signed hash, but non-canonical; forward only `:`-prefixed comment/keepalive lines and drop bare separators.
- `crates/services/src/completions/mod.rs:636-647`, `crates/api/src/conversions.rs:89-99` — typing `stream_options` as `include_usage` + `continuous_usage_stats` only silently drops other vendor keys clients previously passed through via flattened `extra`. Add a flattened extra map to `StreamOptions` or merge unrecognized keys back when forwarding.
- `crates/api/src/routes/completions.rs:1340-1349` — rejecting `stream_options.continuous_usage_stats=true` with a 400 on the rewritten public path is a breaking change for clients using this vLLM extension. Intentional and documented in the PR body, but make sure it lands in user-facing release notes.
- `crates/api/src/routes/completions.rs:1322` + `crates/services/src/models/mod.rs:93-96` — `get_models_with_pricing()` deep-clones the full model catalog on every chat request, including non-streaming where the result is unused. Gate the lookup on `request.stream == Some(true)` or add a clone-free `get_model_attestation_supported(&name)`.

**Checks run locally**: `cargo fmt --all -- --check`, `git diff --check`, and `cargo clippy --locked -p api -p services --all-targets -- -D warnings` all pass. `cargo test --locked -p api --lib` (169 passed), `-p services --lib` (386 passed), `-p inference_providers --lib` (300 passed), plus targeted `chat_stream` (8), `include_usage` (3), `continuous_usage_stats` (2), `malformed_stream_options_remain_in_extra` (1), and services `stream_options` (2) all pass. DB-backed e2e tests could not run locally (no Postgres/dstack socket); the GitHub CI Test Suite, Lint, and security_audit on the PR head are all green.

lloydmak99:
Pushed `5acb89c` addressing the concrete code findings from the latest review.

Changes:

- Loads `stream_error_count` before building the rewritten stream tail.
- Suppresses the synthetic final usage chunk when any stream error frame was emitted, matching the existing rule that failed/partial streams do not get gateway signatures.
- Added an e2e regression for a mock provider that emits usage-bearing chunks and then errors.
- Added an include-usage streaming signature e2e that asserts the stored `response_hash` matches `sha256` of the exact SSE body returned to the client, including the synthetic usage chunk and `[DONE]`.
- Drops bare blank-line control separators on rewritten streams while preserving `:` comment/keepalive control events in the emitted bytes and signed hash.
- Warns when `include_usage=true` but no provider usage was observed.
- Gates the cached model metadata lookup to streaming requests so non-streaming chat requests do not clone the catalog for stream shaping.
- Added unit coverage for final usage tail construction and rewritten control-event handling.

Validation:

- `cargo test --locked -p api --lib final_usage_chunk`
- `cargo test --locked -p api --lib rewritten_control_events`
- `cargo test --locked -p api --lib`
- `cargo test --locked -p services --lib`
- `git diff --check`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`

I also ran `cargo test --locked -p api --test e2e_all test_chat_completions_stream_error_does_not_emit_final_usage`; it compiled, then failed before application code because this local environment still has no Postgres admin DB listening (`Connection refused` during DB bootstrap). `cargo clippy --all-targets` compiled the e2e target.

Outstanding from the review: the gateway-signed mode for rewritten attestation-supported streams still needs explicit security/product sign-off, or the PR needs to be re-scoped. I updated the PR body to call out the signer change and verification-client expectation directly instead of presenting that as a code-only fix.

PierreLeGuen (changes requested):
The streaming rewrite itself is carefully built — billing capture happens in `InterceptStream` upstream of the public rewrite, the gateway hash covers the exact client bytes (serialized chunks, forwarded comments, and the synthetic tail), error streams skip both the synthetic usage chunk and the signature store, and the passthrough fallbacks (E2EE, non-text modalities, metadata miss) are correctly scoped, with safe failure toward passthrough when the model lookup misses. The blockers are contract/process issues rather than stream-handling bugs.

**Blocking**

- `crates/api/src/routes/completions.rs:369`, `1801-1848`, `crates/services/src/completions/mod.rs:113` — On the default streaming path for attestation-supported models, `skip_provider_chat_signature` means the provider-TEE signature is never fetched; the gateway signs the rewritten bytes instead. This changes `signing_address` for streamed responses while non-streaming responses from the same model keep the provider key, so verification clients that pin the provider attestation will break or silently over-trust a gateway-signed response. The PR body itself gates this signer change on a security/product sign-off that has not been recorded. Don't merge until that sign-off lands, or narrow the rewrite scope (e.g. only rewrite when the client actually sent `stream_options`, keeping byte-exact provider-signed passthrough as the default). Either way, give clients a way to distinguish the signer — a signer-type field on `GET /v1/signature/{chat_id}` or explicit docs on the gateway signing address.
- `crates/api/src/routes/completions.rs:1383-1392` — `stream_options.continuous_usage_stats=true` now returns 400 on the default public text-stream path, where it previously worked (forwarded to vLLM, which honors it). The rejection is correctly gated on the rewrite path and doesn't interfere with the NEAR provider force-setting the flag internally, but it is still a breaking API change for existing clients and needs the same sign-off plus changelog/client communication. Alternative: treat it as a passthrough opt-out (skip rewriting, keep the provider signature) instead of rejecting.

**Important**

- `crates/api/src/conversions.rs:17-27`, `89-99` and `crates/services/src/completions/mod.rs:636-654` — Parsing `stream_options` into the typed two-field `StreamOptions` silently drops any other keys a client sent inside a well-formed `stream_options` object; previously the raw object was forwarded verbatim via `extra`, so future vLLM stream options get swallowed. Merge unknown keys back (a flattened extra map), or fall back to raw passthrough on unknown fields, matching the existing malformed-value behavior. Relatedly, `stream_options` is now extracted at two layers with subtly different semantics (the conversion removes the key on successful parse; the service-layer `extract_stream_options_from_extra` then never sees it). The net behavior is correct today, but the duplication is fragile — extract once.
- `crates/services/src/attestation/mod.rs:405-424` — `store_gateway_signature` increments `METRIC_VERIFICATION_SUCCESS` and records `METRIC_VERIFICATION_DURATION` for what is a signature *creation* event, inflating verification dashboards. Use a separate metric or a differentiating tag; also dedupe the two near-identical `info!` lines per signing.

**Minor**

- `crates/api/src/routes/completions.rs:443-455` — On rewritten streams, SSE comments are forwarded as bare `: keepalive\n` while the following blank-line separator is dropped, so the comment glues onto the next event block. Spec-compliant parsers ignore it, but clients that split on `\n\n` will mis-parse. Append `\n` when forwarding comment events; also worth a comment noting that `event:`/`id:`/`retry:` field lines are silently dropped (acceptable for OpenAI-format upstreams).
- `crates/api/src/routes/completions.rs:1726-1733` — When `include_usage=true` but no provider usage was observed, the final usage chunk is silently omitted, violating the client contract. At minimum log a warning; consider falling back to `InterceptStream`'s accumulated usage stats.
- `crates/api/src/routes/completions.rs:1361-1378` — `get_models_with_pricing()` clones the entire cached model+pricing `Vec` per streaming request to read one `attestation_supported` bool. Add a single-model cached lookup to avoid the per-request O(models) allocation.
- `crates/api/tests/e2e_all/usage_chat_completions.rs:124-139` — Switching `test_chat_completions_stream_records_usage_in_history` to `include_usage:true` removed the only e2e coverage that a *default* stream (usage chunks now stripped from the public response) still records usage in billing history. Since "rewrite must not affect billing" is the core invariant here, add a default-stream variant asserting a non-zero usage-history entry.

**Checks run locally:** `cargo fmt --all -- --check` and `cargo clippy --locked -p api -p services --all-targets -- -D warnings` clean; unit tests pass (`api` 171, `services` 386, `inference_providers` 300); `git diff --check` clean. With local Postgres in `DEV=true` ephemeral-key mode (matching CI), the e2e slice across streaming usage, signature verification, chat encryption, auto-redact, model alias transparency, client disconnect, and external/streaming provider-error tests passed 98/98 (1 ignored) — including `test_chat_completions_stream_error_does_not_emit_final_usage`, which the PR author was blocked on, and `test_streaming_chat_include_usage_signature_hashes_client_bytes`. CI (Lint, Test Suite, security_audit) is green.

lloydmak99:
Pushed `f4f57a6` addressing Pierre's latest review.

I took the narrower-scope route from the review:

- Default attestation-supported streaming responses now stay byte-exact passthrough and keep provider-side chat signature storage.
- Public stream rewriting is limited to explicit `stream_options.include_usage=true` requests.
- `stream_options.continuous_usage_stats=true` now opts out to passthrough instead of returning 400.
- The PR title/body now use `Refs #646` and call out that default attested stream normalization is intentionally deferred pending the security/product decision.

Also addressed the concrete follow-ups:

- Added flattened `StreamOptions.extra` so well-formed vendor stream option keys survive typed parsing and re-serialization.
- Moved gateway signature creation onto separate metrics and deduped the gateway signature log line.
- Forwarded SSE comments with the blank-line separator on rewritten streams.
- Restored default-stream billing-history e2e coverage while keeping the `include_usage=true` response/signature tests for the rewritten path.

Validation:

- `cargo test --locked -p api --lib stream_options`
- `cargo test --locked -p api --lib include_usage`
- `cargo test --locked -p api --lib continuous_usage_stats_opts_out_of_rewrite`
- `cargo test --locked -p api --lib rewritten_control_events`
- `cargo test --locked -p api --lib test_chat_completion_request_conversion`
- `cargo test --locked -p services --lib extract_stream_options_consumes_typed_options`
- `cargo test --locked -p inference_providers --lib stream_options`
- `cargo test --locked -p api --lib`
- `cargo test --locked -p services --lib`
- `cargo test --locked -p inference_providers --lib`
- `cargo check --locked --all-targets`
- `cargo clippy --locked -p api -p services --all-targets -- -D warnings`
- `git diff --check`

DB-backed e2e tests compile via `cargo check --locked --all-targets`, but I still cannot run them end to end locally because this environment has no Postgres test DB listening.

lloydmak99:
CI update for `f4f57a6`: all required checks are green.

- `security_audit`: passed
- `Lint`: passed
- `Test Suite`: passed

Review is requested from Evrard and Pierre.

PierreLeGuen (approved):
Narrowing the rewrite to explicit `stream_options.include_usage=true` requests while keeping default attested streams byte-exact passthrough resolves the prior blocking concerns: the signature store is now awaited with a timeout before `[DONE]`, error streams skip the synthetic usage chunk and signature, the gateway hash is built incrementally over exactly the bytes the client receives, and `continuous_usage_stats=true` became a passthrough opt-out instead of a 400. Approving, with a few optional follow-ups:

- `crates/services/src/completions/mod.rs:113` — when `skip_provider_chat_signature` is set on gateway-signed streams, the provider chat mapping unpin that normally happens inside `store_chat_signature_from_provider` may be skipped; worth confirming the finalizer/tail still unpins `signature_buckets`/`signature_rotation` entries on that path (including store errors/timeouts) so they don't accumulate.
- `crates/inference_providers/src/models.rs:432` — removing `skip_serializing_if` on `ChatCompletionChunk.usage` means all re-serialized streams (auto-redact, alias-served, converter-normalized) now emit `usage: null` even when the client sent no `stream_options`, where OpenAI omits the field. Intentional and harmless for standard SDKs, but it's a wire-format change beyond the opt-in rewrite path.
- Since the fix is now scoped to explicit `include_usage=true`, default attested streams still carry provider usage on every chunk — #646 should stay open or be re-scoped accordingly.

Checks: `cargo fmt --check`, `git diff --check`, and `cargo clippy -p api -p services` clean; `cargo test --locked` unit suites pass (api 174, services 386, inference_providers 300, plus targeted stream-options/usage tests); CI green on head `f4f57a6` (Lint, Test Suite incl. DB-backed e2e, security_audit) — DB e2e not run locally.

PR #785, opened 2026-06-12
fix(chutes): cache /e2e/instances discovery + consume single-use nonces (#774)

### LINKED ISSUE #774: Chutes /e2e/instances 429 (rate limit) is masked as a generic 502; no discovery caching
## Symptom
Under any burst of Chutes-routed requests, cloud-api returns 502 *"The model is currently unavailable."* The masked upstream cause (visible only in logs as `provider_message`) is:

```
Provider failed for model '…': discover instances: Chutes returned HTTP 429:
{"detail":"Rate limit exceeded. Try again later."}
```

Observed on `cloud-stg-api` 2026-06-11 while bringing up the full Chutes inventory — firing ~10+ models in quick succession tripped it; spacing requests ~13s apart still occasionally hit it (concurrent infra-test traffic also calls Chutes).

## Two problems
1. **Wrong status mapping.** A 429 from Chutes `/e2e/instances` is a transient, retryable condition. The Chutes provider returns it as `CompletionError::CompletionError("discover instances: Chutes returned HTTP 429: …")`, which the completions classifier maps to a generic **502** (`bad_gateway`, non-retryable from the client's view). It should surface as a retryable **503 / ServiceOverloaded** (like the existing `(503, _) => ServiceOverloaded` arm) so clients back off instead of treating it as a hard failure.
2. **No instance-discovery caching.** Every request performs a fresh `/e2e/instances` discovery against `api.chutes.ai`, so request volume maps 1:1 onto the rate-limited endpoint. A short-TTL cache of the discovered E2E-capable instances per chute (respecting the freshness needs of the per-request attestation) would cut discovery calls dramatically and avoid self-inflicted 429s.

## Where
- `crates/inference_providers/src/attested/chutes/mod.rs` — `verify_and_prepare()` → `discover instances` (the 429 origin); error variants are plain `CompletionError::CompletionError(String)` with no status carried.
- `crates/services/src/completions/mod.rs` — the error classifier that maps these to 502.

## Suggested fix
- Carry the upstream status (or detect 429 in the message, consistent with the existing keyword approach) and classify Chutes 429 → `ServiceOverloaded` (503, retryable).
- Add a short-TTL per-chute instance-discovery cache; on a cache miss or attestation failure, refresh.

## Severity
Not a trust-chain issue (fail-closed still holds). It degrades availability/UX for attested Chutes models under load and makes real failures harder to diagnose (everything looks like a flat 502). Separate from the measurement-allowlist fix in #773.

Evrard-Nil:
## What

Closes #774 (the discovery-caching half; #778 already shipped the status-mapping half).

Every Chutes-routed request performed a fresh `GET /e2e/instances` against `api.chutes.ai`, so request volume mapped **1:1** onto a rate-limited endpoint — any burst self-inflicted `429`s. Observed data: one discovery response carries ~5 instances × ~10 single-use nonce tokens (≈50) with `nonce_expires_in: 60`, so **one** discovery call can serve ~50 requests over its nonce lifetime.

This adds a short-TTL, per-chute cache of the `/e2e/instances` discovery response and consumes nonce tokens one at a time, cutting the discovery-call rate dramatically.

## Design (security/concurrency-sensitive)

- `Provider.instances_cache: Arc<std::sync::Mutex<HashMap<String, Arc<tokio::sync::Mutex<CachedInstances>>>>>`. The **outer** `std::sync::Mutex` is held **only** for a fast get-or-create of the per-chute cell — never across an `await`. Each per-chute `tokio::sync::Mutex` serializes refresh + nonce consumption for **that chute only** (single-flight per chute → no thundering herd; other chutes are unaffected).
- `discover_cached()` refreshes via a real discovery call **only** when the cached entry is expired **or** carries no usable (E2E-capable + nonce-bearing) instance. TTL is derived from `nonce_expires_in` (`>0`, clamped `≤120s`, shaved `5s`; fallback `30s`), and it returns an **owned** snapshot for candidate selection + verification.
- `take_nonce()` `.pop()`s a token from the cache — the **atomic single-use point** that prevents reuse. A token drained by a concurrent request between snapshot and commit just moves to the next candidate; a fully drained chute refreshes on the next `discover_cached`.

### Attestation is UNCHANGED
Every request still mints a fresh `boot_nonce`, fetches `/evidence`, and verifies the chosen instance end-to-end. **Only `/e2e/instances` discovery is cached** — evidence and verification are never cached.

## Also carries the 5xx-body strip from #778's review
The preserved-status arm of `map_client_error` now emits `"<stage>: Chutes returned HTTP <status>"` instead of the raw upstream body, so a 5xx body can't trip the pool's `is_client_media_fetch_error` keyword scan and get misclassified as a client error. The masked (4xx/non-status) arm keeps its full message.

## Tests
Added cache-level unit tests (no network): single-use nonce consumed at most once / pool drains; missing instance → `None`; the empty/expired seed and an all-drained-but-fresh snapshot both signal refresh-needed; `chute_cache` returns a stable per-chute cell and distinct cells across chutes; and a `map_client_error` test pinning that the preserved-status message omits the upstream body. Full crate: `323 passed; 0 failed; 1 ignored`. `cargo fmt` + `clippy -D warnings` clean.

### DIFF
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -171,6 +171,47 @@ pub struct Provider {
     /// Memoized model→`chute_id` (the mapping is static), so we don't re-fetch
     /// `/v1/models` on every request.
     chute_id_cache: tokio::sync::OnceCell<String>,
+    /// Short-TTL cache of `/e2e/instances` discovery PER chute (#774). One
+    /// discovery response carries ~5 instances × ~10 single-use nonce tokens
+    /// (≈50 tokens) valid for `nonce_expires_in` seconds, so caching it and
+    /// consuming nonces one-at-a-time serves ~50 requests from a single
+    /// rate-limited discovery call — cutting the call rate that self-inflicts the
+    /// `/e2e/instances` 429s.
+    ///
+    /// The OUTER `std::sync::Mutex` guards ONLY a fast get-or-create of the
+    /// per-chute cell (never held across an `await`). Each chute's
+    /// `Arc<tokio::sync::Mutex<CachedInstances>>` serializes refresh + nonce
+    /// consumption for THAT chute only (single-flight per chute; other chutes are
+    /// unaffected). ATTESTATION IS UNCHANGED — only `/e2e/instances` discovery is
+    /// cached; every request still mints a fresh boot nonce, fetches `/evidence`,
+    /// and verifies the chosen instance.
+    instances_cache: Arc<
+        std::sync::Mutex<
+            std::collections::HashMap<String, Arc<tokio::sync::Mutex<CachedInstances>>>,
+        >,
+    >,
+}
+
+/// A cached `/e2e/instances` discovery snapshot for one chute. `instances` is the
+/// owned discovery result; nonce tokens are POPPED from it as they're consumed
+/// (single-use), so the same token is never handed out twice. `expires_at` is the
+/// freshness deadline derived from the upstream `nonce_expires_in` (with a safety
+/// margin); past it — or once an instance's nonce pool empties — the next use
+/// refreshes via a fresh discovery call.
+struct CachedInstances {
+    instances: Vec<client::E2eInstance>,
+    expires_at: std::time::Instant,
+}
+
+impl CachedInstances {
+    /// An empty, already-expired entry: its first use triggers a refresh (the
+    /// `or_insert_with` seed for a chute we haven't discovered yet).
+    fn empty_expired() -> Self {
+        Self {
+            instances: Vec::new(),
+            expires_at: std::time::Instant::now(),
+        }
+    }
 }
 
 /// Everything needed to invoke a verified instance: the targeting headers, the
@@ -203,6 +244,7 @@ impl Provider {
             canonical_id: config.canonical_id,
             allow_streaming,
             chute_id_cache: tokio::sync::OnceCell::new(),
+            instances_cache: Arc::new(std::sync::Mutex::new(std::collections::HashMap::new())),
         })
     }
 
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Caches `/e2e/instances` discovery per chute with a short, clamped TTL and consumes nonces atomically from the shared cache (single-use via per-chute mutex pop), plus trims upstream 5xx bodies out of preserved-status error messages. The concurrency design is sound — the outer std mutex is never held across an await, the per-chute tokio mutex gives single-flight refresh and atomic nonce consumption — and attestation flow is unchanged.

Non-blocking observation:
- The comment at `crates/inference_providers/src/attested/chutes/mod.rs:515` calls the all-candidates-drained error retryable, but the pool classifier treats it as non-retryable. It's a rare, self-healing burst condition already flagged in earlier review comments; worth aligning the comment (or the error mapping) in a follow-up.

Checks: `cargo test -p inference_providers --lib` (323 passed, including the new cache/nonce and error-message tests), `cargo clippy -p inference_providers --lib` clean, `cargo fmt --check` clean, and PR CI (Lint, Test Suite, security_audit) all green.

Evrard-Nil:
Addressed both review themes in 63bc2ac1 — they converge on the same root cause (the cache introduces a snapshot→commit nonce-drain window), and the fix handles both ends:

**1. Drain exhaustion is no longer a non-retryable hard failure** (@claude's production-safety flag). The candidate loop now distinguishes an *all-pools-drained* exhaustion (the recoverable race) from a real verification failure via a new `PrepareError::AllDrained` outcome:
- `verify_and_prepare` runs `prepare_from_snapshot` and, on `AllDrained`, forces **one** fresh-snapshot retry (`discover_cached(.., force=true)`) — the forced refresh refills the pools so it resolves in-place, exactly the inline-refresh-and-retry you suggested.
- A genuine *double*-drain on the forced pass maps to a **retryable 429** (`HttpError { status_code: 429, is_external: true }` → `retryable_http_429`), so the pool's outer retry/backoff still engages instead of the keyword-less terminal string.
- A non-drain failure (attestation/evidence/E2EE) clears the `drained_only` flag and is surfaced verbatim — we never force a pointless refresh that wouldn't help.

**2. No more discovery tight loop on an empty/all-drained upstream** (@gemini-code-assist's stampede flag — correct catch). Freshness is now strictly **TTL-gated**, not usability-gated:
- `discover_cached` refreshes only when `expires_at <= now` (or `force`). A refresh that yields no usable nonces installs a short `EMPTY_REFRESH_COOLDOWN` (5s) instead of the full TTL — quick recovery, no per-request hammering.
- `take_nonce` proactively sets `expires_at = now` when it pops the **last** usable nonce, so a normal drain still refreshes promptly on the next request (your proactive-expire suggestion). Because a failed refresh installs the 5s cooldown, this expiry can't become a tight loop.

Invariants preserved: single-use nonces still consumed via `pop` under the per-chute tokio Mutex (no reuse); no std Mutex across an await; single-flight per chute; **attestation stays per-request** (only discovery is cached); usable-branch TTL still never outlives `nonce_expires_in`. The attestation-report path stays uncached (lower volume).

Tests: +3 (`take_nonce_expires_cell_when_last_nonce_drains`, `all_drained_maps_to_retryable_429`, `hard_error_passes_through_unchanged`); the two stale `!usable`-predicate tests were rewritten for the TTL-gated contract. 90 chutes unit tests pass, clippy `-D warnings` clean.

Issue #781, opened 2026-06-12
Chutes OpenAI-compat gaps: request validation, supported_features, max_tokens:0, SSE usage, error envelope, attestation param


Evrard-Nil:
Found in the Chutes prod-readiness bug-hunt (staging). Bundle of correctness/compat gaps (verdict: ship-then-fix, except H2 which is strongly-recommended). Checklist:

- [ ] **H2 — request validation masked by upstream errors.** `max_tokens:-5` and context-too-long payloads never produce a distinct 4xx (`context_length_exceeded`, invalid max_tokens) — they collapse into the opaque upstream failure, so clients retry forever instead of fixing the request. `temperature:5` and empty `messages` *are* validated (clean 400s). Add client-side `max_tokens`/context-length validation before forwarding. (Partly unmasked once #774 makes the 429 retryable, but bounds checks should be client-side regardless.)
- [ ] **M1 — empty `supported_features` / `supported_sampling_parameters`** in `/v1/models` for Chutes models. Silently disables tools for OpenRouter-style routers that gate on these. Prioritize if router consumers are in scope (ties to the OpenRouter-listing effort).
- [ ] **M2 — `max_tokens:0`** returns 200 with null content and self-inconsistent `usage`. Should validate or normalize.
- [ ] **L1 — SSE emits `usage` on intermediate chunks** even when `stream_options.include_usage` is unset (should only be on the final chunk when requested).
- [ ] **L2 — malformed JSON / missing required field** returns raw axum/serde plain-text, not the OpenAI error envelope. (Cross-cutting, not Chutes-specific.)
- [ ] **L3 — attestation endpoint silently ignores an unknown `request_nonce` param** (correct param is `nonce`) — param-name hygiene.

Source: 6-agent Chutes feature probe, 2026-06-12. No secrets/hostnames/stack-traces leaked in any error body (explicitly checked).

Evrard-Nil:
[...]

Issue #780, opened 2026-06-12
Chutes responses leak prompt_sha256 (user-prompt fingerprint) + serving internals to clients


Evrard-Nil:
Found in the Chutes prod-readiness bug-hunt (staging). The genuine privacy item.

## Symptom
Chutes responses carry top-level/extra fields absent from non-Chutes (Anthropic/OpenAI) responses on the same endpoint, passed through verbatim by cloud-api (the canonical==slug path is byte-exact; the rewrite path preserves unmodeled fields):
- **`prompt_sha256`** — deterministic SHA-256 of the rendered user prompt, returned in the response. A stable fingerprint of user content: links identical prompts and lets an observer confirm/guess prompt contents. **Privacy concern.**
- `template_sha256`, `choices[].matched_stop` (sglang stop-token id), `metadata.weight_version` — serving internals.
- `chutes_verification` — possibly an intentional attestation receipt, but non-standard + Chutes-branded inside an OpenAI-shaped body.

## Repro
`POST /v1/chat/completions {"model":"qwen/qwen3-32b","messages":[{"role":"user","content":"hi"}],"max_tokens":20}` → top-level keys include `prompt_sha256`/`template_sha256`/`chutes_verification`/`metadata`; choices include `matched_stop`. An Anthropic-model response on the same endpoint has none of these.

## Fix direction
Decide intentional-vs-leak per field on the Chutes response path:
- **Strip `prompt_sha256`** (or document the privacy model explicitly) — highest priority.
- Strip `template_sha256` / `matched_stop` / `metadata`.
- Keep `chutes_verification` as a documented attestation hook or move it out of the OpenAI body.
Goal: match the clean response shape returned for first-party / Anthropic / OpenAI models. Note this interacts with the hidden-states passthrough design — the strip must be deliberate on the Chutes path.

Evrard-Nil:
**Follow-up from post-deploy re-verification (user test + fleet, 2026-06-12):** the deployed strip works for the 4 named fields (`prompt_sha256`/`template_sha256`/`metadata`/`choices[].matched_stop` — confirmed gone via JSON-walk + raw-byte grep across 6 bodies, stream + non-stream). BUT the strip is a fixed **denylist**, and the internal field set **varies by model / sglang config**. `moonshotai/kimi-k2.6` (and likely others) emit additional top-level/choice keys NOT stripped:
[...]

PR #787, opened 2026-06-12
fix(chutes): allowlist response fields + gate streamed usage (#781)

Evrard-Nil:
## What

Two related Chutes response/stream-normalization fixes in `crates/inference_providers/src/attested/chutes/`. They touch the same chunk-processing code, so one PR. Addresses two items from #781 (and follows up #780/#782).

### (A) Allowlist the response shape (#780 follow-up)

#782 stripped Chutes internals via a **denylist** of 4 named keys. But model-specific serving internals vary per model — kimi-k2.6 alone adds `templated_prompt`, `prompt_token_ids`, `prompt_logprobs`, `kv_transfer_params`, `choices[].token_ids`, `choices[].stop_reason`, a `reasoning` dup, … — so a denylist silently leaks any new internal.

Switched to an **allowlist**: anything not on it is dropped by default.

- **Top-level kept:** `id, object, created, model, choices, usage, system_fingerprint, service_tier` (the latter two **are** standard OpenAI, not internals) plus the deliberately-kept extras `chutes_verification` (attestation receipt) and `hidden_states` (passthrough).
- **Choice-level kept:** `index, delta, message, finish_reason, logprobs`.
- **Message/delta kept:** `role, content, reasoning_content, reasoning, tool_calls, function_call, refusal, annotations, name, tool_call_id` (both reasoning fields kept — we never drop a legit reasoning field).

Conservative under-strip: when unsure a field is kept; the known internals above are deliberately absent and dropped.

**Chunk side:** the parsed chunk is now **re-derived from the sanitized bytes** rather than patched field-by-field. This guarantees the chunk and `raw_bytes` carry the exact same allowlisted shape and — crucially — also drops **typed** internal slots (`prompt_token_ids`, `modality`, per-choice `token_ids`) that a `chunk.extra`-only strip would miss, so the re-serialized route paths (auto-redact / alias) don't leak them either.

### (B) Gate streamed `usage` to the OpenAI spec (#781 L1)

Streamed responses emitted a populated `usage` on every intermediate chunk and ignored `stream_options.include_usage`. Per spec, intermediate chunks must carry `usage: null`; `usage` appears only on the final chunk, and only when requested.

- Intermediate chunk (**non-empty** `choices`) → `usage` always stripped (covers vLLM `continuous_usage_stats`).
- Final usage-only chunk (**empty** `choices`) → `usage` kept iff the request set `stream_options.include_usage` (or `continuous_usage_stats`), else dropped.

The client's intent is threaded into the stream processing. We still force `include_usage` **upstream** so billing always sees usage; the gate touches only the client-facing `raw_bytes`, and the parsed `chunk.usage` is preserved so `InterceptStream` keeps billing on it.

## Tests

New unit tests (all green):
- allowlist drops unknown + known internals (top/choice/message level) while every standard field + `chutes_verification`/`hidden_states`/`reasoning_content`/`reasoning` survives — non-stream and per-chunk, incl. the typed slots a chunk.extra-only strip would miss;
- `include_usage` unset → no usage on intermediate or final chunk; set → usage only on the final chunk;
- `chunk.usage` preserved for billing even when the client didn't request usage.

`cargo fmt`, `cargo clippy -p inference_providers --all-targets --all-features -D warnings`, and the full `inference_providers` suite (93 chutes tests + 328/8/1 overall, live probes ignored) all pass.

Refs #781.

### DIFF
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -559,47 +559,87 @@ const INTERNAL_KEYS: &[&str] = {
     ]
 };
 
-/// Chutes-internal / serving fields that must be stripped from the decrypted
-/// response before it reaches a client (#780). `prompt_sha256` is the genuine
-/// privacy concern — a deterministic SHA-256 fingerprint of the rendered user
-/// prompt — while the others are sglang/serving internals that a NEAR/Anthropic/
-/// OpenAI response on the same endpoint never carries.
+/// Standard OpenAI chat-response / chunk fields kept at the TOP LEVEL, plus the two
+/// extras this crate deliberately passes through. This is an ALLOWLIST (#780 →
+/// #781 follow-up): model-specific serving internals vary too much to denylist
+/// reliably (e.g. kimi-k2.6 emits `templated_prompt`, `prompt_token_ids`,
+/// `prompt_logprobs`, `kv_transfer_params`, … alongside the already-known
+/// `prompt_sha256`/`template_sha256`/`metadata`), so anything NOT on this list is
+/// dropped by default.
 ///
-/// `chutes_verification` (the attestation receipt) is deliberately NOT listed —
-/// it's kept untouched. We strip ONLY these named keys, so unmodeled fields the
-/// repo intentionally passes through (e.g. `hidden_states`) survive.
-const STRIPPED_TOP_LEVEL_FIELDS: &[&str] = &["prompt_sha256", "template_sha256", "metadata"];
+/// Kept:
+/// - the standard OpenAI top-level fields — incl. `system_fingerprint` and
+///   `service_tier`, which ARE standard (do NOT confuse them with serving internals);
+/// - `chutes_verification` — the attestation receipt, deliberately surfaced;
+/// - `hidden_states` — the intentional unmodeled passthrough.
+const ALLOWED_TOP_LEVEL_FIELDS: &[&str] = &[
+    "id",
+    "object",
+    "created",
+    "model",
+    "choices",
+    "usage",
+    "system_fingerprint",
+    "service_tier",
+    "chutes_verification",
+    "hidden_states",
+];
 
-/// Per-choice serving internal stripped from every element of `choices` (#780):
-/// sglang's matched stop-token id.
-const STRIPPED_CHOICE_FIELDS: &[&str] = &["matched_stop"];
+/// Standard OpenAI fields kept on each element of `choices` (both the non-stream
+/// `message`-bearing shape and the stream `delta`-bearing shape). Anything else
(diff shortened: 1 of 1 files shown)


lloydmak99 (approved):
Approve. The allowlist approach (#781 follow-up to #780) is clearly more robust than the denylist — model-specific internals vary unpredictably per model, so dropping-by-default is the right posture. ALLOWED_TOP_LEVEL/CHOICE/MESSAGE_FIELDS cover the standard OpenAI shape plus the intentional extras (chutes_verification, hidden_states, reasoning_content/reasoning). The gate_stream_usage logic correctly identifies the final chunk by empty-choices and gates accordingly, preserving usage in the parsed chunk for billing.

One nit not in prior comments: if serde_json::from_value::<ChatCompletionChunk>(v.clone()) fails at line ~257 (the chunk re-derivation step), ev.chunk retains the old pre-sanitization value while raw_bytes is already sanitized — a potential mismatch on the re-serialized route path. In practice this would only happen if the sanitized JSON is structurally valid but not parseable as ChatCompletionChunk, which is an extremely unlikely edge case post-allowlist filtering. Not blocking.

PierreLeGuen (approved):
Allowlist approach and the streamed-usage gate are sound: re-deriving the chunk from sanitized bytes is safe (all required `ChatCompletionChunk` fields are allowlisted), and `chunk.usage` is preserved for billing via `InterceptStream`. Two things to follow up on, neither blocking:

- `crates/inference_providers/src/attested/chutes/mod.rs:592,599` — `ALLOWED_CHOICE_FIELDS` and `ALLOWED_MESSAGE_FIELDS` omit `hidden_states`, but per #739 sglang emits it at `choices[].hidden_states` and `choices[].delta.hidden_states` (see the regression test at `crates/inference_providers/src/models.rs:1303`), not at the top level where this PR allowlists it. The passthrough the PR says it preserves is dropped where it would actually appear. Non-blocking since #739 notes the feature is a no-op until an sglang backend ships with `--enable-return-hidden-states`, but please add `"hidden_states"` to both lists (with a test on the choice-level shape) before that lands.
- `crates/inference_providers/src/attested/chutes/mod.rs:1037` — worth double-checking that `client_wants_usage` sees real requests: `ChatCompletionRequest` flattens unknown fields into `extra`, and `crates/api/src/conversions.rs:115` appears to leave client `stream_options` there with the typed field set to `None`, in which case a client sending `stream_options: {"include_usage": true}` would have its final usage chunk stripped. If that's the case, derive the flag from both the typed field and `extra["stream_options"]` (or parse it into the typed field during conversion).

Checks run locally: `cargo test -p inference_providers` (328 passed, incl. 93 chutes tests), `cargo test -p api conversions` (9 passed), `cargo clippy -p inference_providers --all-targets` clean, `cargo fmt --check` clean.

PR #788, opened 2026-06-12
fix(api): pre-flight reject invalid max_tokens/n with 400 instead of masked 502 (#786)

### LINKED ISSUE #786: Chutes: invalid client params (max_tokens<0, n:0, context-too-long) return 502 instead of 4xx
Split out from #781 (which auto-closed when M1/#783 merged) — this HIGH item is **not yet fixed** and the re-verify fleet flagged it as the one thing warranting a fix **before/with prod**.

## Symptom (deterministic, staging)
On Chutes models, these all return `502 {"type":"bad_gateway","message":"The model is currently unavailable..."}` instead of a proper 4xx:
- `max_tokens: -5`
- `n: 0`
- context-too-long (~500k-token payload)

The gateway only validates `temperature`/`top_p` up front; other invalid params fall through to upstream and the rejection collapses into the generic 502.

## Why it matters for prod
- **Mis-signals retryability** — clients see a 5xx and retry a permanently-bad request forever.
- **Pollutes outage metrics/alerting** — real backend outages and client mistakes look identical (both generic 502).
- No upstream details leak (generic message) → this is a **classification bug, not disclosure**.

## Fix
Pre-flight validation of `max_tokens` (>=1 or per-model cap), `n` (>=1), and context length → return proper `400 invalid_request_error` / `context_length_exceeded`, before forwarding to Chutes. (Likely shared with other providers — scope to where request params are validated.)

Found in the post-fix re-verification, 2026-06-12. Related still-open follow-ups (from the now-closed bundles): H1 allowlist hardening (#780 — model-specific internal fields like `templated_prompt` not in the fixed strip-list), L1 (SSE `usage` on every chunk + `include_usage` no-op), L2 (serde/axum errors bypass the OpenAI error envelope), B3 stream-path `<think>` extraction (#779 — non-stream only shipped).

Evrard-Nil:
## Summary

Closes #786 (HIGH). Invalid client sampling params currently fall through to upstream and the rejection collapses into a generic masked `502 bad_gateway` instead of a proper `4xx`. This mis-signals retryability (clients retry a permanently-bad request) and pollutes outage metrics (real backend outages and client mistakes look identical).

Today only `temperature`/`top_p` are validated up front. This PR adds the **same-style** pre-flight validation for the params that fall through, on the existing validation seam, for **both** `/v1/chat/completions` and `/v1/completions` (cross-cutting across all providers, not Chutes-specific):

- `max_tokens`: reject `< 1` (negative or zero) -> `400 invalid_request_error`, `param: "max_tokens"`.
- `n`: reject `< 1` -> `400 invalid_request_error`, `param: "n"`.

A new `validate_request()` method wraps the existing message-only `validate()` (temperature/top_p/stop/logprobs, unchanged) and adds the param-bearing checks, returning the full OpenAI error envelope `{error:{message,type,param,code}}`. The two handlers now call `validate_request()` and return that envelope directly. The existing `validate()` and all its call sites/tests are untouched, so valid requests and all other endpoints are unaffected.

## Context length: intentionally skipped (per issue)

The issue's third sub-item (reject prompts exceeding a model's max context with `context_length_exceeded`) is **not** implemented: cloud-api has no per-model tokenizer, so the prompt token count is not cleanly knowable pre-dispatch. A character-based estimate would risk rejecting valid requests, violating the "do not change behavior for valid requests" constraint. Over-length prompts continue to be rejected by the upstream provider. This is noted in the code comments.

## Tests

Added unit tests (all green):
- valid (unset / positive) `max_tokens` + `n` pass;
- `max_tokens: -5` and `0` rejected with `400` + `param: "max_tokens"`;
- `n: -1` and `0` rejected with `400` + `param: "n"`;
- existing checks (out-of-range temperature) still surface via the new method;
- the completions (`/v1/completions`) path covered too.

`cargo fmt`, `cargo clippy -p api -p services --all-targets --all-features -- -D warnings`, and the relevant unit tests all pass locally.

### DIFF
--- a/crates/api/src/models.rs
+++ b/crates/api/src/models.rs
@@ -1056,7 +1056,61 @@ use crate::consts::{
 };
 use crate::routes::common::{validate_max_length, validate_non_empty_field};
 
+/// Reject `max_tokens` values the OpenAI API requires to be `>= 1`.
+///
+/// `None` (unset) is valid. Negative or zero values are rejected with a 400
+/// `invalid_request_error` carrying `param: "max_tokens"` instead of falling
+/// through to upstream and surfacing as a masked 502 (nearai/cloud-api #786).
+fn validate_max_tokens(max_tokens: Option<i64>) -> Result<(), ErrorResponse> {
+    if let Some(value) = max_tokens {
+        if value < 1 {
+            return Err(ErrorResponse::with_param(
+                "max_tokens must be at least 1".to_string(),
+                "invalid_request_error".to_string(),
+                "max_tokens".to_string(),
+            ));
+        }
+    }
+    Ok(())
+}
+
+/// Reject `n` values below 1. `None` (unset) is valid.
+fn validate_n(n: Option<i64>) -> Result<(), ErrorResponse> {
+    if let Some(value) = n {
+        if value < 1 {
+            return Err(ErrorResponse::with_param(
+                "n must be at least 1".to_string(),
+                "invalid_request_error".to_string(),
+                "n".to_string(),
+            ));
+        }
+    }
+    Ok(())
+}
+
 impl ChatCompletionRequest {
+    /// Pre-flight validation that returns a ready-to-serialize OpenAI error
+    /// envelope (`{error:{message,type,param,code}}`) on failure.
+    ///
+    /// This wraps the message-only [`Self::validate`] (temperature/top_p/stop/
+    /// logprobs) and adds the param-bearing sampling checks (`max_tokens`, `n`)
+    /// that would otherwise fall through to upstream and surface as a masked
+    /// 502 instead of a 4xx (nearai/cloud-api #786). All checks reject only
+    /// genuinely-invalid values; valid requests are untouched.
+    ///
+    /// NOTE: context-length validation is intentionally NOT performed here.
+    /// cloud-api has no per-model tokenizer, so the prompt token count is not
+    /// cleanly knowable pre-dispatch; a character-based estimate would risk
+    /// rejecting valid requests. Over-length prompts are still rejected by the
+    /// upstream provider.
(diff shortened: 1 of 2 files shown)


Evrard-Nil:
Closes #786. Cross-cutting pre-flight validation on the existing `validate()` seam for `max_tokens` (>=1) and `n` (>=1) on both `/v1/chat/completions` and `/v1/completions`. Context-length sub-item deliberately skipped (no per-model tokenizer in cloud-api -> not cleanly knowable pre-dispatch; char-estimate would risk rejecting valid requests). Valid requests are unchanged.

lloydmak99 (approved):
Approve. Clean, well-scoped pre-flight validation: validate_max_tokens and validate_n both reject values < 1 with a properly-shaped OpenAI error envelope (type + param fields), validate_request() chains the existing validate() plus these new checks, and both routes now call validate_request(). Tests cover valid params, rejection with correct error fields, and that existing checks (temperature range) still surface. Context-length intentionally omitted — documented rationale is sound.

PierreLeGuen (approved):
Small, well-scoped fix: requests with `max_tokens < 1` or `n < 1` on `/v1/chat/completions` and `/v1/completions` now get a 400 `invalid_request_error` (with `param`, matching the OpenAI envelope) instead of being forwarded upstream and surfacing as a masked 502. The new `validate_request()` wrapper delegates to the existing `validate()`, so previously-rejected requests keep the same error contract, `None` values remain valid, and the remaining `.validate()` call sites (image gen, rerank, score, responses) are on request types without these fields — no gaps or regressions found.

Local checks: `cargo test -p api --lib validate_request` (5/5 new tests pass), `cargo clippy -p api --lib -- -D warnings` clean, `cargo fmt --check` clean, `git diff --check` clean.

PR #792, opened 2026-06-13
fix(audio): treat provider 4xx as client error, not a paging server error

PierreLeGuen:
## What

Audio transcription requests that the vLLM whisper backend rejects with a **4xx** (e.g. `Invalid or unsupported audio file`) are **client-input** errors, not service faults. Today they:

- log at **ERROR** at both the provider adapter (`inference_providers::vllm`) and the route handler (`api::routes::completions`), and
- return to the client as `server_error` + "Audio transcription failed. Please try again later."

A sustained stream of bad uploads from one workspace therefore spikes the Cloud-API error monitor and **pages on-call**, and the "try again later" message encourages clients to retry payloads that can never succeed.

## Changes

- **vLLM adapter** (`crates/inference_providers/src/attested/nearai/mod.rs`): log provider **4xx at WARN**, keep **5xx at ERROR**.
- **`audio_transcriptions` route** (`crates/api/src/routes/completions.rs`): map provider **4xx → `invalid_request_error`** (429 → `rate_limit_error`) with an actionable message; keep **5xx → `server_error`**. Log client 4xx at WARN, 5xx at ERROR.
- Add `organization_id` / `workspace_id` (**IDs only**, per the logging policy) to the audio error logs so a bad-input burst is **attributable to its source** — these logs currently carry no client identifier.

This mirrors the existing chat-completions behavior in `completion_stream_error_openai_type` (`400..=499 => invalid_request_error`).

## Why now

Triggered by a prod page on 2026-06-13: a sustained ~8 failed transcriptions/min, **all** HTTP 400 `Invalid or unsupported audio file`, with roughly half of all transcription requests failing; a single workspace accounted for ~87% of transcription volume. The endpoint itself was healthy throughout (valid audio returned 200). This pattern accounted for ~99% of Cloud-API's error volume during the page.

## Out of scope / follow-ups

- The pool-level `All providers failed for model` log (`services::inference_provider_pool`) still emits at ERROR for these 4xx. It is a shared path across all operations (and a provider `401`/`403` there is a real infra problem we *do* want at ERROR), so it is left untouched here. Options: make it status-aware for non-retryable 4xx, or exclude it on the monitor.
- Datadog monitor `14604516` ("High number of errors in Cloud-Api") counts `@level:ERROR`; after this lands the audio 4xx no longer count toward it.

## Testing

- `cargo check --workspace` passes; `cargo fmt --all` clean.
- Verified against a prod repro: valid audio → `200`; empty file → `400 invalid_request_error` (cloud-api's own validation, unchanged); the backend's own 4xx now surfaces to the client as `invalid_request_error` instead of `server_error`.


PierreLeGuen:
Thanks for the reviews — all three caught real things. Pushed a follow-up (and rebased on `main`) that addresses them:

**🔴 The "fix is ineffective" P1 (Codex) — correct, and the important one.** Confirmed by prod telemetry: the failing requests logged `retry_decision: non_retryable_no_keyword_match` (the *stringified* `CompletionError` branch), and clients were seeing `502`, never `400`. `inference_provider_pool::audio_transcription` flattened the adapter's `AudioTranscriptionError::HttpError` into `CompletionError::CompletionError(e.to_string())`, so the service mapper's per-status arm never ran and every failure collapsed to `502`.

Fix: the pool now carries the structured status through `retry_with_fallback` as `CompletionError::HttpError { .., is_external: false }` and re-wraps it as `AudioTranscriptionError::HttpError` on the way out. The service mapper (`completions/mod.rs`) then does the per-code mapping it was always meant to: `400 → client 400`, `401/403 → 500`, `5xx → 502`, `503 → ServiceOverloaded`, `429 → RateLimitExceeded`.

**Unreachable `429` branch.** Removed — the service maps a provider `429 → RateLimitExceeded` before the route's `ProviderError` arm, so it could never carry `429`. Documented in a comment.

**401/403 handling (Gemini).**
- *Route handler:* not needed there — the service mapper remaps `401/403 → 500` upstream (`completions/mod.rs`), so the route never sees them as 4xx. Added a comment so it isn't surprising.
- *vLLM adapter log:* valid, applied — `401/403` now stay at `error` (not `warn`) so a backend-credentials failure still alerts.

**Tests added** (`crates/api/tests/e2e_all/audio_transcriptions.rs`, via a new `MockProvider::set_audio_transcription_error_override` hook):
- provider `400` → client `400` / `invalid_request_error`
- provider `500` → client `502` / `server_error`

These lock in exactly the behavior the P1 flagged. The `404`/`408` message-precision nit is left as-is (non-blocking, rare for the whisper backend).

PierreLeGuen:
@codex review — the P1 (provider 4xx collapsing to 502) is fixed by preserving the structured `HttpError` through the pool; please re-check the merged commit.

PierreLeGuen:
Addressed the three P2s from the latest pass. They all stem from the new status-preservation, so I added a single shared predicate `is_client_audio_input_status` (400/413/415/422) as the source of truth for "genuine bad audio" vs "provider infra", and applied it at the pool, route, and adapter:

- **408 → bad-audio message:** 408 (and any non-client-input 4xx) now take the server-error path — `server_error`, not `invalid_request_error` with "file is valid…" guidance.
- **404 logged as client input:** the adapter now logs only client-input 4xx at `warn`; `401/403/404/408` stay at `error` so a missing-route / creds / timeout fault still alerts.
- **401/403 skipping failover:** the pool drops `401/403/404/408` to a generic provider failure *before* the 4xx fast-return, so they fall through to the remaining providers (and surface as 5xx) instead of short-circuiting. Moot for whisper today (single provider) but correct for any future multi-provider audio model.

New test `test_audio_transcription_provider_infra_4xx_maps_to_server_error` covers `401/403/404/408 → 5xx server_error`; the existing `400 → invalid_request_error` and `500 → 502` tests still pass (21 audio e2e tests green locally + in CI).

PierreLeGuen:
@codex review

PierreLeGuen:
@Evrard-Nil — when you get a chance, this is ready for your review 🙏

It's green and all automated feedback is resolved:
- **CI passing** (Lint, Test Suite, security_audit) on the latest commit
- **Codex** P1 + 3×P2 addressed (re-review now clean), **Gemini** and **claude-review** points addressed too
- 21/21 audio e2e tests, branch up to date with `main`

Just needs your approval to merge — should be a quick one. Thanks!

Evrard-Nil (approved):

PR #794, opened 2026-06-15
fix(db): tolerate not-yet-ready members in Patroni cluster discovery

Evrard-Nil:
## Symptom
Adding a new postgres replica triggered this in **both cloud-api and chat-api** (shared `database` crate):
```
ERROR database::patroni_discovery: Failed to refresh cluster state:
Failed to parse cluster info: error decoding response body
```

## Exact root cause (confirmed empirically)
Patroni lists a node as a `/cluster` member **as soon as it registers in the DCS — before it publishes its `conn_url`**. So a replica mid-creation (state `creating replica`, or an uninitialized node) appears **without `host`/`port`/`api_url`**. Those fields are **required** on `ClusterMember`, so a strict `Vec<ClusterMember>` parse fails the *entire* response.

Reproduced against the real struct in a unit test:
```
PARSE ERROR: missing field `host` at line 3 column 84
```
One half-registered member poisons the whole cluster view → discovery refresh errors for every consumer until the new node finishes registering (and startup discovery would fail outright during that window). This is the **same class** as the prior `lag:"unknown"` poison — different field.

## Fix
Parse members **individually** (`deserialize_members`) and drop any that don't fully deserialize — a member with no connection info isn't a usable leader/replica target anyway, and the next refresh picks it up once it finishes registering. `ClusterMember` stays rigid, so `cluster_manager` is untouched (every surviving member still has `host`/`port`). Same philosophy as the existing `deserialize_lag` resilience.

## Tests
- `test_initializing_member_does_not_poison_parse` — leader + one `creating replica` (no host/port) → parses, bad member dropped.
- `test_full_cluster_with_initializing_replica_parses` — real staging `/cluster` shape + an initializing replica → 3 ready members survive, leader discoverable.
- All existing `patroni_discovery` tests still pass (`6 passed; 0 failed`).

## Validation
Verified in dev (deterministic repro against the real struct). End-to-end validation: deploy this, then add a staging postgres replica and confirm discovery stays healthy through the member's initialization window. **Unblocks deploying new postgres instances.**

### DIFF
--- a/crates/database/src/patroni_discovery.rs
+++ b/crates/database/src/patroni_discovery.rs
@@ -44,8 +44,43 @@ pub struct ClusterMember {
     pub timeline: Option<i64>,
 }
 
+/// Patroni lists a node as a member as soon as it registers in the DCS — which
+/// happens *before* it publishes its `conn_url`. So a replica mid-creation (or
+/// an uninitialized node) appears without `host`/`port`/`api_url`. Those fields
+/// are required on `ClusterMember`, so a strict `Vec<ClusterMember>` parse fails
+/// the ENTIRE `/cluster` response with `missing field \`host\`` over one
+/// half-registered member — freezing topology discovery for every consumer
+/// (cloud-api + chat-api) whenever a new postgres instance is being added.
+///
+/// Parse each member independently and drop any that don't fully deserialize; a
+/// member with no connection info is not a usable leader/replica target anyway,
+/// and the next refresh picks it up once it finishes registering. Same
+/// resilience philosophy as `deserialize_lag` — one bad member must not poison
+/// the whole cluster view.
+fn deserialize_members<'de, D>(deserializer: D) -> Result<Vec<ClusterMember>, D::Error>
+where
+    D: Deserializer<'de>,
+{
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
The fix isolates member-level parse failures so a single not-yet-ready node can't poison the whole `/cluster` topology response, mirroring the existing `deserialize_lag` strategy. Small, correct, and covered by two new regression tests.

Non-blocking follow-ups:
- The `warn!` for dropped members logs only member name + serde error, no customer data (`patroni_discovery.rs:60-79`) — confirmed safe; downstream consumers in `cluster_manager.rs` only see fully-deserialized members.
- Doc comment in `test_full_cluster_with_initializing_replica_parses` says "4 ready members" but the fixture has 3 (the assertion correctly checks `len == 3`) — cosmetic.
- `cargo fmt --check` fails on two changed assertions in `patroni_discovery.rs`, and CI `Lint` is red for the same diff — worth a quick `cargo fmt` before merge.

Checks: `cargo test -p database patroni_discovery` (6/6 passed, incl. 2 new regression tests) and `cargo clippy -p database --lib` clean; `cargo fmt --check` / CI `Lint` failed on formatting.

PR #795, opened 2026-06-16
test(pool): end-to-end NEAR→Chutes per-request fallback

Evrard-Nil:
Adds deterministic integration tests for the **per-request NEAR→Chutes fallback** through `chat_completion` → `retry_with_fallback_caps`.

## Why
We tried to validate the fallback on staging by draining the NEAR backend at the model-proxy. It **couldn't** be validated that way: cloud-api holds **sticky keepalive connections** to NEAR backends through the L4 proxy, so a registry change never makes the *live* request fail — the fallback only fires on a genuine **request-level** failure. That's deterministic to inject with `MockProvider`, so it belongs in a unit/integration test rather than a fragile infra experiment.

## Tests (all green, 3.7s)
- **`near_5xx_falls_back_to_chutes_within_one_request`** — NEAR primary returns a retryable upstream `503`; the request falls through to the Chutes (`Attested3p`) provider and **succeeds**. Asserts NEAR was attempted first, Chutes served (response body check), and the fallback is **immediate** (`< 5s`, i.e. next-provider-in-same-round, no client-facing backoff — addresses the timeout we saw in the staging poke).
- **`all_providers_5xx_surfaces_error`** — NEAR + Chutes both `503` → error **surfaces** (no false success).
- **`healthy_near_serves_without_invoking_chutes`** — healthy NEAR serves itself; Chutes is **never invoked** (no needless fallback/billing).

Complements the existing component tests (`verifiable_model_drops_nonattested_and_orders_tiers`, `filter_streaming_capable_*`, `merge_discovered_and_pinned_*`) with the end-to-end behavior.

### DIFF
--- a/crates/database/src/patroni_discovery.rs
+++ b/crates/database/src/patroni_discovery.rs
@@ -44,8 +44,43 @@ pub struct ClusterMember {
     pub timeline: Option<i64>,
 }
 
+/// Patroni lists a node as a member as soon as it registers in the DCS — which
+/// happens *before* it publishes its `conn_url`. So a replica mid-creation (or
+/// an uninitialized node) appears without `host`/`port`/`api_url`. Those fields
+/// are required on `ClusterMember`, so a strict `Vec<ClusterMember>` parse fails
+/// the ENTIRE `/cluster` response with `missing field \`host\`` over one
+/// half-registered member — freezing topology discovery for every consumer
+/// (cloud-api + chat-api) whenever a new postgres instance is being added.
+///
+/// Parse each member independently and drop any that don't fully deserialize; a
+/// member with no connection info is not a usable leader/replica target anyway,
+/// and the next refresh picks it up once it finishes registering. Same
+/// resilience philosophy as `deserialize_lag` — one bad member must not poison
+/// the whole cluster view.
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):
Test-only additions (end-to-end NEAR→Chutes per-request fallback) plus an already-validated Patroni deserializer resilience fix. The production change parses each cluster member independently and drops un-deserializable ones, mirroring the existing `deserialize_lag` pattern; `warn!` logs only the member name and serde error (topology, not customer data), so it's privacy-safe (`patroni_discovery.rs:62`). Tier ordering (NEAR primary → Attested3p fallback) makes the "NEAR tried first" assertion deterministic, and a 503 maps to `retryable_http_5xx`, so fallback happens within the same retry round — matching the tests' immediate-fallback assertion. No correctness, privacy, or contract issues found.

Optional follow-up:
- `cargo fmt --check` fails on a rustfmt-only assertion wrap at `crates/services/src/inference_provider_pool/mod.rs:6738`, which also fails the GitHub `Lint` job. Worth a quick `cargo fmt` before merge.

Checks: new tests (`near_5xx_falls_back_to_chutes_within_one_request`, `all_providers_5xx_surfaces_error`, `healthy_near_serves_without_invoking_chutes`) and the 6 Patroni discovery tests pass; clippy clean; `cargo fmt --check` fails as noted above.

PR #796, opened 2026-06-16
feat: add model-consumption-timeseries and performance-timeseries admin endpoints

Evrard-Nil:
## Summary

- Adds `GET /admin/platform/model-consumption-timeseries`: top-N per-model consumed cost timeseries with server-side "Other" bucket collapse. Groups by `model_id` (UUID) rather than `model_name` — rename-safe. Caps: hour→7d, day→366d, week→1096d, month→1826d.
- Adds `GET /admin/platform/performance-timeseries`: platform-wide TTFT percentiles (P50/P95/P99 via `PERCENTILE_CONT`), output token throughput, and error rate per time bucket. Supports optional `model_name` filter. `ttft_sample_count` exposes the streaming-only denominator.
- New covering migration `V0062__add_analytics_covering_index.sql`: `CONCURRENTLY` index on `organization_usage_log(created_at DESC) INCLUDE(model_name, total_cost, total_tokens, output_tokens, ttft_ms, stop_reason)` + secondary `(model_name, created_at DESC)` for the filtered path.
- `allowlisted_date_trunc()` helper returns `&'static str` — SQL injection via granularity parameter is structurally impossible.
- `parse_metrics_range()` gets per-granularity absolute caps to prevent DoS full-table scans.

## Security

- Granularity validated with `allowlisted_date_trunc()` — user string never reaches `format!()` SQL.
- `top_n` clamped to `[1, 20]` in the handler; query always has an explicit `LIMIT`.
- Date range caps per granularity prevent unbounded table scans.

## DB cost

No schema changes. Covering index built `CONCURRENTLY` (no table lock). Write overhead: one additional index entry per `INSERT` into `organization_usage_log` — append-only table, negligible.

## Test plan

- [ ] `cargo check` passes
- [ ] `GET /admin/platform/model-consumption-timeseries?start=...&end=...&granularity=day&top_n=10` returns `model_labels` + bucketed `data`
- [ ] `GET /admin/platform/performance-timeseries?start=...&end=...&granularity=day` returns TTFT percentiles and error_rate
- [ ] Invalid `granularity=foo` returns 400 with descriptive error
- [ ] `top_n=0` and `top_n=99` are clamped to 1 and 20 respectively

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1839,11 +1839,12 @@ pub fn build_admin_routes(
         batch_upsert_models, cancel_model_pricing_change, confirm_model_deprecation,
         confirm_model_pricing_changes, create_admin_access_token, create_service,
         delete_admin_access_token, delete_model, deprecate_model, get_admin_organization_balance,
-        get_billing_summary, get_infra_summary, get_model_history, get_model_revenue,
-        get_org_revenue, get_organization as get_admin_organization,
-        get_organization_concurrent_limit, get_organization_limits_history,
-        get_organization_metrics, get_organization_timeseries, get_platform_metrics,
-        get_platform_timeseries, list_admin_access_tokens, list_invitation_email_deliveries,
+        get_billing_summary, get_infra_summary, get_model_consumption_timeseries,
+        get_model_history, get_model_revenue, get_org_revenue,
+        get_organization as get_admin_organization, get_organization_concurrent_limit,
+        get_organization_limits_history, get_organization_metrics, get_organization_timeseries,
+        get_performance_timeseries, get_platform_metrics, get_platform_timeseries,
+        list_admin_access_tokens, list_invitation_email_deliveries,
         list_model_pricing_changes, list_models as admin_list_models, list_organization_members,
         list_organizations, list_users, preview_model_deprecation, preview_model_pricing_changes,
         resend_invitation_email, update_organization_concurrent_limit, update_organization_limits,
@@ -1988,6 +1989,14 @@ pub fn build_admin_routes(
         .route(
             "/admin/platform/infra-summary",
             axum::routing::get(get_infra_summary),
+        )
+        .route(
+            "/admin/platform/model-consumption-timeseries",
+            axum::routing::get(get_model_consumption_timeseries),
+        )
+        .route(
+            "/admin/platform/performance-timeseries",
+            axum::routing::get(get_performance_timeseries),
         )
         .route(
             "/admin/invitation-email-deliveries",
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1844,9 +1844,9 @@ pub fn build_admin_routes(
         get_organization as get_admin_organization, get_organization_concurrent_limit,
         get_organization_limits_history, get_organization_metrics, get_organization_timeseries,
         get_performance_timeseries, get_platform_metrics, get_platform_timeseries,
-        list_admin_access_tokens, list_invitation_email_deliveries,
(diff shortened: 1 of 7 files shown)


PierreLeGuen (changes requested):
The new admin timeseries endpoints look correct, but the accompanying index migration will fail at runtime and block the server from booting.

- **Blocking** — `crates/database/src/migrations/sql/V0062__add_analytics_covering_index.sql:19,28`: both `CREATE INDEX CONCURRENTLY` statements run inside a transaction. Refinery applies each migration in its own transaction by default (`run_async` in `crates/database/src/migrations/mod.rs`), and PostgreSQL rejects `CREATE INDEX CONCURRENTLY` inside a transaction block (`ERROR: CREATE INDEX CONCURRENTLY cannot run inside a transaction block`). CI's Test Suite reproduces this exact failure, and since migrations run on startup the server won't boot in any environment that hasn't already applied it. Fix: drop `CONCURRENTLY` (plain `CREATE INDEX IF NOT EXISTS`, consistent with the existing non-concurrent indexes) and accept the build-time lock, or build the index out-of-band and reduce this migration to a no-op guard.

- `crates/database/src/repositories/analytics.rs` — `model_labels_ordered` is built in first-seen order from a result sorted by bucket then cost, so it reflects the first bucket's top costs rather than the true top-N across the whole period. Either reword the "Ordered list" doc comment or compute the ordering by total `cost_nano` per label.

Note: the previously raised nullable-`model_id` and double-NULL `COALESCE` panic concerns don't apply — V0013 makes both `model_id` and `model_name` `NOT NULL` on `organization_usage_log`.

Checks: `cargo check`/`cargo clippy` pass; `cargo fmt --all -- --check` reports diffs in `crates/api/src/lib.rs` and `crates/database/src/repositories/analytics.rs`; e2e suite not run (no local PostgreSQL).

Evrard-Nil (commented):
Thanks for the review Pierre.

**Blocking (CONCURRENTLY)** — fixed in commit 4f09d563 (dropped CONCURRENTLY from the migration, added a comment explaining how to pre-build manually if needed before deploying to a large table).

**Optional (model_labels_ordered first-bucket ordering)** — fixed in cce4cc8d: accumulate total `cost_nano` per label across all buckets, then sort by global period total DESC with "Other" pinned last.

Re-requesting your review.

PierreLeGuen (approved):
Solid addition of the two admin timeseries endpoints. Prior blocking feedback is resolved: `CONCURRENTLY` was dropped from `V0062__add_analytics_covering_index.sql` so the migration runs inside Refinery's transaction, and `model_labels_ordered` now sorts by global period cost with "Other" pinned last. SQL injection surface is safe (`granularity` goes through the `allowlisted_date_trunc()` static allowlist, everything else parameterized), and NULL panics are ruled out since `model_id`/`model_name` are `NOT NULL`.

- `crates/api/src/openapi.rs:159` — the new handlers are routed/annotated but not added to `ApiDoc` `paths(...)`, so they'll be missing from `/api-docs/openapi.json` and generated clients. Worth registering them.
- `crates/database/src/migrations/sql/V0062__add_analytics_covering_index.sql:23` — covering index omits `model_id`, which `get_model_consumption_timeseries` groups/joins on, so the consumption query can't go fully index-only. Pure perf; add `model_id` to `INCLUDE` if desired.
- PR description says a `hour→7d` cap but the code enforces 31 days (`admin.rs:3531`/`3600`, `common.rs:220`). Description wording only.

Checks: `cargo fmt --all -- --check`, `cargo check -p api -p database -p services`, `cargo clippy --all-targets`, and `cargo test`/openapi lib tests all pass; e2e/integration suite skipped (no local PostgreSQL).

PR #798, opened 2026-06-16
fix(pool): fall back to Chutes on provider model-not-found (#797)

### LINKED ISSUE #797: Staging: z-ai/glm-5.1 intermittently 400s for tool calls when CHUTES_MODELS fallback is configured
## Summary

GLM-5.1 tool call requests in staging fail intermittently with:

```json
{"error":{"message":"Provider failed for model 'z-ai/glm-5.1': The model 'z-ai/glm-5.1' does not exist.","type":"invalid_request_error","param":"model","code":null}}
```

This error occurs for ~30-50% of tool call requests. Simple chat completions (without `tools`) work consistently.

## Environment

- **Environment**: Staging (`cloud-stg-api.near.ai`)
- **Config**: `ENABLE_CHUTES=true`, `CHUTES_MODELS=z-ai/glm-5.1=zai-org/GLM-5.1-TEE,...`

## Reproduction

```bash
for i in 1 2 3; do
  curl -s -H "Authorization: Bearer <STG_KEY>" \
    -d '{"model": "zai-org/GLM-5.1-FP8", "messages": [{"role": "user", "content": "Weather?"}], "max_tokens": 64, "tools": [{"type":"function","function":{"name":"get_weather","description":"Get weather","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}], "tool_choice": "auto"}' \
    "https://cloud-stg-api.near.ai/v1/chat/completions" | python3 -c "import json,sys; d=json.load(sys.stdin); print('OK' if 'choices' in d else 'ERR: ' + d.get('error',{}).get('message','?')[:80])"
done
# Run 1: ERR: Provider failed for model 'z-ai/glm-5.1': ...
# Run 2: OK
# Run 3: OK
```

## Behavior

- **Regular chat** (no `tools`): always succeeds via Chutes fallback
- **Tool calls**: ~30-50% success rate; failures are fast (<1s), suggesting the error is returned before reaching Chutes
- **Chutes directly** with `zai-org/GLM-5.1-TEE` + tools: always succeeds

## Root Cause Hypothesis

When a tool call request arrives for `z-ai/glm-5.1`:
1. cloud-api first tries the inference-proxy endpoint (which exists in the VPC for the production GLM CVM)
2. The production inference-proxy sometimes returns a fast "model not found" response (it doesn't know the model as `z-ai/glm-5.1`)
3. cloud-api propagates this error as "Provider failed: model z-ai/glm-5.1 doesn't exist" **without falling back to CHUTES_MODELS**
4. For regular chat, the error is not fast-propagated (or the inference-proxy takes longer → timeout → Chutes fallback)

## Expected Behavior

When the inference-proxy returns "model not found" for a model that is in `CHUTES_MODELS`, cloud-api should fall back to Chutes instead of propagating the error.

## Infra-tests Impact

- `test_cloud_stg_api::test_tool_calling[GLM-5.1-FP8]` — intermittent
- `test_cloud_stg_api::test_tool_result_roundtrip[GLM-5.1-FP8]` — intermittent
- `test_stg_tool_call_generic::test_tool_call_returns_structured_tool_calls[GLM-5.1-FP8]` — intermittent

Tracked in nearai/infra-tests CI (runs every 30min).

Evrard-Nil:
Fixes #797.

## Problem
After renaming GLM-5.1's canonical id to `z-ai/glm-5.1`, ~30–50% of (tool-call) requests intermittently failed with `Provider failed for model 'z-ai/glm-5.1': The model 'z-ai/glm-5.1' does not exist.` instead of falling back to the configured Chutes provider.

A NEAR/inference-proxy backend that doesn't recognize a model id answers with an OpenAI-compatible **404/400 "model does not exist"**. The inner retry loop classified *all* non-429/408 4xx as "client error, not retrying" and **fast-returned**, so the request never advanced to the Chutes fallback under the same canonical id. (It's intermittent because it only happens when the request lands on a NEAR backend that hasn't picked up the new id.)

## Fix
Add `is_model_not_found_error(status_code, message)` (404/400 + "model … does not exist / not found" / `model_not_found`) and **exclude it from the 4xx fast-return**. A model-not-found now **falls through to the next provider in the same round** — exactly like the existing `408` handling:
- It stays **non-retryable** for the round (the same NEAR provider isn't re-hit; the failure counter isn't bumped).
- If **no** provider serves the model, the error still **propagates** after the loop (no masking).
- A **genuine client 4xx** (bad params) still fast-returns without wasting the fallback.

## Tests (5 green locally, clippy + fmt clean)
- `near_model_not_found_falls_back_to_chutes` — NEAR `404` **and** `400` "model does not exist" → Chutes serves; NEAR tried first.
- `genuine_client_4xx_does_not_fall_back` — a real bad-params `400` propagates; Chutes untouched (guards against over-reach).
- Existing fallback tests (`near_5xx_falls_back…`, `all_providers_5xx_surfaces_error`, `healthy_near_serves…`) still pass.

Note: not currently reproducible on staging (the GLM backends presently recognize `z-ai/glm-5.1`), but this hardens the path against the reported failure mode regardless of which backend a request lands on.

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -1686,6 +1686,27 @@ impl InferenceProviderPool {
         None
     }
 
+    /// A provider's "this model isn't served here" response — OpenAI-compatible
+    /// engines (vLLM/SGLang) answer an unknown model with a 404/400 like
+    /// "The model `X` does not exist." Unlike a genuine client 4xx (bad params),
+    /// this is NOT the request's fault: a *different* provider for the same
+    /// canonical id (e.g. a Chutes fallback) may serve the model. So the inner
+    /// loop falls through to the next provider instead of fast-returning, while
+    /// staying non-retryable for the round (re-hitting the same provider would
+    /// just repeat the answer, and it's not a backend-health signal).
+    ///
+    /// Fixes nearai/cloud-api#797: a NEAR backend that didn't recognize a renamed
+    /// canonical id (`z-ai/glm-5.1`) 4xx'd and the request never reached the
+    /// configured Chutes fallback.
+    fn is_model_not_found_error(status_code: u16, message: &str) -> bool {
+        if status_code != 400 && status_code != 404 {
+            return false;
+        }
+        let m = message.to_ascii_lowercase();
+        m.contains("model_not_found")
+            || (m.contains("model") && (m.contains("does not exist") || m.contains("not found")))
+    }
+
     /// Single source of truth for the retry decision: the inner retry loop
     /// gates on `starts_with("retryable_")`, and the terminal error log emits
     /// the label directly so the rationale is visible in production logs.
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Straightforward, well-scoped fix: a model-not-found 4xx now skips the fast-return and falls through to the next provider (Chutes) in the same round, while genuine client 4xx still short-circuits as before. Logic and error propagation check out.

- Non-blocking — `crates/services/src/inference_provider_pool/mod.rs:1705-1712`: `is_model_not_found_error` matches any 400/404 whose message contains "model" plus "not found"/"does not exist". This is intentionally broad, but a false positive only costs one wasted fallback attempt and the error still propagates if no provider serves the model, so it's not a correctness risk.

Verified: model-not-found classifies as non-retryable, the failure counter isn't bumped, and `last_error` propagates after the loop if no provider serves the model. `CompletionError::HttpError.message` is a non-optional `String` and is populated at all construction sites, so the new destructuring is safe.

Checks: `cargo test -p services` (both new tests `near_model_not_found_falls_back_to_chutes` and `genuine_client_4xx_does_not_fall_back` plus existing related tests pass), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean.

PR #801, opened 2026-06-16
Fix streaming: emit usage:null in intermediate chunks per OpenAI spec (fixes #646)

### LINKED ISSUE #646: Streaming default emits 'usage' field in every chunk; should be null until stream_options.include_usage=true

Evrard-Nil:
## Summary

- Streaming responses were forwarding upstream (vLLM/sglang) SSE bytes verbatim, which included populated `usage` in every intermediate chunk — violating the OpenAI spec
- Per spec, `usage` must be `null` in all intermediate chunks; a populated usage chunk only appears as a final extra chunk when `stream_options.include_usage: true` is explicitly requested
- The `include_usage: true` path already added the final usage chunk correctly via `rewrite_public_stream_usage`; the bug was that intermediate chunks in ALL other paths (default, `include_usage: false`) still had populated usage

## Changes

**`ChatStreamUsageMode` gains `strip_intermediate_usage`** — a new flag that triggers re-serialisation of chunks with `usage` nulled out, without adding a final usage chunk. Active for default streaming and `include_usage: false`. Mutually exclusive with `rewrite_public_stream_usage`.

**Passthrough bypass** — the byte-exact passthrough shortcut (`raw_passthrough && !rewrite_public_stream_usage`) now also checks `!strip_intermediate_usage`, forcing chunks through the re-serialisation path when usage needs to be stripped.

**Re-serialisation path** — alongside the existing `rewrite_public_stream_usage` branch, a new `else if strip_intermediate_usage` branch calls `prepare_stream_chunk_for_client` with `include_usage: false` to null usage and drop usage-only chunks.

Exceptions (passthrough preserved): E2EE (chunks are opaque), `continuous_usage_stats: true` (explicit per-chunk usage request), non-text modalities.

## Test plan

- [x] `cargo test --lib --bins` — 393 tests pass (0 failures)
- [x] New unit tests for all `ChatStreamUsageMode` combinations, including `strip_intermediate_usage` flag
- [x] New e2e test: default streaming — all chunks carry `usage: null`
- [x] New e2e test: `include_usage: false` — all chunks carry `usage: null`
- [x] Existing e2e test: `include_usage: true` — exactly 1 final chunk with populated usage

Closes #646

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -177,6 +177,26 @@ jobs:
         RUST_LOG: debug
         DEV: "true"
 
+  # Aggregates all test jobs into a single required status check. Branch
+  # protection requires "Test Suite"; this job satisfies that requirement
+  # without running any extra work.
+  test-suite:
+    name: Test Suite
+    runs-on: [self-hosted, infra]
+    needs: [lint, unit-test, integration-test, e2e-test]
+    if: always()
+    steps:
+      - name: Check all tests passed
+        run: |
+          if [[ "${{ needs.lint.result }}" != "success" || \
+                "${{ needs.unit-test.result }}" != "success" || \
+                "${{ needs.integration-test.result }}" != "success" || \
+                "${{ needs.e2e-test.result }}" != "success" ]]; then
+            echo "One or more test jobs failed"
+            exit 1
+          fi
+          echo "All test jobs passed"
+
   # Release build, only on main pushes. Split into its own job so it does
   # not block the test jobs and does not pull in PostgreSQL or test env vars
   # (there is no build.rs; cargo build --release needs none of them).
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -2969,7 +2969,10 @@ mod tests {
 
         assert!(!mode.rewrite_public_stream_usage);
         assert!(!mode.gateway_signature_enabled);
-        assert!(mode.strip_intermediate_usage, "default stream must strip intermediate usage");
+        assert!(
+            mode.strip_intermediate_usage,
+            "default stream must strip intermediate usage"
+        );
     }
 
     #[test]
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -387,10 +387,20 @@ fn chat_stream_usage_mode(
 
     // For default streaming (no include_usage or include_usage:false), strip populated
     // usage from all chunks so intermediate ones carry `usage: null` per OpenAI spec.
-    // Skipped when the client explicitly requested continuous stats, uses E2EE (chunks
-    // are opaque), has non-text modalities, or when rewrite_public_stream_usage already
-    // handles usage shaping.
+    //
+    // Only applied for non-attested (external) models. Attested models (e.g. nearai,
+    // chutes) sign the raw upstream bytes; stripping usage before forwarding to the
+    // client would break byte-exact TEE signature verification because the client
+    // would hash different bytes than what the inference proxy signed. For those
+    // models the provider-signed stream MUST be forwarded verbatim (the
+    // `rewrite_public_stream_usage` path handles `include_usage:true` via the
+    // separately-computed gateway signature).
+    //
+    // Also skipped when: the client requested continuous stats (explicit per-chunk
+    // usage), E2EE is active (chunks are opaque), non-text modalities are in use,
+    // or `rewrite_public_stream_usage` already handles usage shaping.
     let strip_intermediate_usage = request.stream == Some(true)
+        && model_attestation_supported == Some(false)
         && !rewrite_public_stream_usage
         && !chat_stream_continuous_usage_requested(request)
         && !e2ee_active
@@ -2961,25 +2971,40 @@ mod tests {
     }
 
     #[test]
-    fn default_attested_stream_strips_intermediate_usage() {
-        // Default streaming (no include_usage): per OpenAI spec intermediate
-        // chunks must carry `usage: null`; no final usage chunk is emitted.
+    fn default_attested_stream_preserves_passthrough_for_tee_verification() {
+        // Default streaming for ATTESTED models: byte-exact passthrough is preserved
+        // so the client can verify the inference TEE's provider signature.
+        // Stripping usage would change the bytes the client sees vs what the provider
+        // signed, breaking TEE signature verification.
         let request = chat_request_with_include_usage(None);
         let mode = chat_stream_usage_mode(&request, Some(true), false);
 
         assert!(!mode.rewrite_public_stream_usage);
         assert!(!mode.gateway_signature_enabled);
         assert!(
-            mode.strip_intermediate_usage,
-            "default stream must strip intermediate usage"
+            !mode.strip_intermediate_usage,
+            "attested streams must preserve raw passthrough for TEE verification"
         );
     }
 
     #[test]
-    fn include_usage_false_strips_intermediate_usage() {
-        // Explicit include_usage:false: same as default — null out intermediate usage.
+    fn include_usage_false_attested_stream_preserves_passthrough_for_tee_verification() {
+        // Explicit include_usage:false for ATTESTED model: same as default,
+        // preserve passthrough for TEE signature verification.
         let request = chat_request_with_include_usage(Some(false));
         let mode = chat_stream_usage_mode(&request, Some(true), false);
+
+        assert!(!mode.rewrite_public_stream_usage);
+        assert!(!mode.gateway_signature_enabled);
+        assert!(!mode.strip_intermediate_usage);
+    }
+
+    #[test]
+    fn default_non_attested_stream_strips_intermediate_usage() {
+        // Non-attested (external) models: no TEE signature to preserve.
+        // Strip usage so intermediate chunks carry `usage: null` per OpenAI spec.
+        let request = chat_request_with_include_usage(None);
+        let mode = chat_stream_usage_mode(&request, Some(false), false);
 
         assert!(!mode.rewrite_public_stream_usage);
         assert!(!mode.gateway_signature_enabled);
@@ -2987,8 +3012,8 @@ mod tests {
     }
 
     #[test]
-    fn default_non_attested_stream_strips_intermediate_usage() {
-        let request = chat_request_with_include_usage(None);
+    fn include_usage_false_non_attested_stream_strips_intermediate_usage() {
+        let request = chat_request_with_include_usage(Some(false));
         let mode = chat_stream_usage_mode(&request, Some(false), false);
 
         assert!(!mode.rewrite_public_stream_usage);
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -365,6 +365,12 @@ fn chat_stream_has_non_text_modalities(request: &ChatCompletionRequest) -> bool 
 struct ChatStreamUsageMode {
     rewrite_public_stream_usage: bool,
     gateway_signature_enabled: bool,
+    /// Strip `usage` from intermediate chunks without adding a final usage chunk.
+    /// Active for the default case (no `include_usage`) and `include_usage: false`.
+    /// Per OpenAI spec, intermediate chunks must carry `usage: null`; populated
+    /// usage only appears in the one synthetic final chunk emitted when
+    /// `include_usage: true` is explicitly requested.
+    strip_intermediate_usage: bool,
 }
 
 fn chat_stream_usage_mode(
@@ -379,10 +385,22 @@ fn chat_stream_usage_mode(
         && !e2ee_active
         && !chat_stream_has_non_text_modalities(request);
 
+    // For default streaming (no include_usage or include_usage:false), strip populated
+    // usage from all chunks so intermediate ones carry `usage: null` per OpenAI spec.
+    // Skipped when the client explicitly requested continuous stats, uses E2EE (chunks
+    // are opaque), has non-text modalities, or when rewrite_public_stream_usage already
+    // handles usage shaping.
+    let strip_intermediate_usage = request.stream == Some(true)
+        && !rewrite_public_stream_usage
+        && !chat_stream_continuous_usage_requested(request)
+        && !e2ee_active
+        && !chat_stream_has_non_text_modalities(request);
+
     ChatStreamUsageMode {
         rewrite_public_stream_usage,
         gateway_signature_enabled: rewrite_public_stream_usage
             && model_attestation_supported.unwrap_or(false),
+        strip_intermediate_usage,
     }
 }
 
@@ -1374,6 +1392,7 @@ async fn chat_completions_inner(
     let usage_mode = chat_stream_usage_mode(&request, model_attestation_supported, e2ee_active);
     let rewrite_public_stream_usage = usage_mode.rewrite_public_stream_usage;
     let gateway_signature_enabled = usage_mode.gateway_signature_enabled;
+    let strip_intermediate_usage = usage_mode.strip_intermediate_usage;
     service_request.skip_provider_chat_signature = gateway_signature_enabled;
 
     // Auto-redact (opt-in via x-auto-redact header or auto_redact body field).
@@ -1536,6 +1555,7 @@ async fn chat_completions_inner(
                         let upstream_done = upstream_done_forwarded.clone();
                         let include_stream_usage_in_response = include_stream_usage_in_response;
                         let rewrite_public_stream_usage = rewrite_public_stream_usage;
+                        let strip_intermediate_usage = strip_intermediate_usage;
                         let gateway_signature_enabled = gateway_signature_enabled;
                         let public_signature_hasher = public_signature_hasher.clone();
                         let public_signature_chat_id = public_signature_chat_id.clone();
@@ -1546,7 +1566,7 @@ async fn chat_completions_inner(
                                     // Byte-exact passthrough (issue #701): when no public
                                     // chunk rewriting is active, forward the upstream wire
                                     // bytes untouched. Explicit include_usage shaping needs
-                                    // parsed-chunk serialization; encrypted, default, and
+                                    // parsed-chunk serialization; encrypted and
                                     // continuous-usage streams preserve passthrough.
                                     //
                                     // Disabled for alias-served responses: those inject
@@ -1554,10 +1574,15 @@ async fn chat_completions_inner(
                                     // and are non-byte-verifiable anyway since the TEE
                                     // signs under the canonical model name, not the alias
                                     // the client requested.
+                                    //
+                                    // Also disabled when strip_intermediate_usage is active:
+                                    // per OpenAI spec intermediate chunks must carry
+                                    // `usage: null`, so we re-serialize to null it out.
                                     if event.raw_passthrough
                                         && !auto_redact_enabled
                                         && !alias_served
                                         && !rewrite_public_stream_usage
+                                        && !strip_intermediate_usage
                                     {
                                         if event.is_done_marker() {
                                             upstream_done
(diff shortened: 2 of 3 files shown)


PierreLeGuen (changes requested):
The streaming fix is correct for the common path, but there's an unaddressed regression on the committed PR HEAD: when a client explicitly requests `include_usage: true` but model metadata is unavailable, usage is silently dropped entirely.

- **Blocking** — `crates/api/src/routes/completions.rs:393`: `strip_intermediate_usage` becomes `true` whenever `rewrite_public_stream_usage` is false, which includes the `include_usage: true` case when `model_attestation_supported == None` (model absent from the cached catalog, or `get_models_with_pricing()` errors — the documented "preserving raw passthrough" fallback at `completions.rs:1373-1390`). The strip branch then calls `prepare_stream_chunk_for_client(..., include_usage=false, ...)`, which nulls usage on every chunk *and* drops the final usage-only chunk. A client that explicitly asked for usage gets none — strictly worse than the prior byte-exact passthrough. Add `&& !chat_stream_include_usage_requested(request)` to the `strip_intermediate_usage` condition so explicit usage requests always fall back to passthrough.
- Note: a corrective change matching the above (plus a `include_usage_true_with_metadata_lookup_failure_preserves_passthrough` regression test) currently exists only as uncommitted local edits — it is not part of commit `2dcda5eb` on #801. The fix needs to be committed and pushed to actually resolve this.

Checks run: `cargo fmt --check` flagged a formatting issue in `completions.rs`; `cargo test --lib` passes against the committed HEAD; `cargo clippy --all-targets` clean. E2E tests requiring PostgreSQL were not run.

Evrard-Nil (commented):
Fixed both blocking issues from the review:

**Regression guard**: Added `&& !chat_stream_include_usage_requested(request)` to the `strip_intermediate_usage` condition. A client that explicitly requests `include_usage: true` now always falls back to byte-exact passthrough, regardless of whether model metadata is available. The previous `model_attestation_supported == Some(false)` guard already prevented the `None` case from activating strip, but the explicit `include_usage` check makes the intent unambiguous and adds a second layer of defence.

**Regression tests** (commit 73a5b184):
- `include_usage_true_with_metadata_lookup_failure_preserves_passthrough` — verifies that `model_attestation_supported == None` (catalog miss / pricing error) + `include_usage: true` → both `rewrite_public_stream_usage` and `strip_intermediate_usage` are false (byte-exact passthrough)
- `include_usage_true_non_attested_does_not_strip` — verifies that `model_attestation_supported == Some(false)` + `include_usage: true` → `rewrite_public_stream_usage` is true, `strip_intermediate_usage` is false (rewrite path handles it, strip does not fire)

**Formatting**: `cargo fmt --check` is clean (no diff).

All 391 lib tests pass.

PierreLeGuen (approved):
The streaming fix correctly emits `usage: null` on intermediate chunks and drops the usage-only final chunk for non-attested models, while preserving byte-exact passthrough for attested/TEE streams and respecting the e2ee, continuous-usage, and non-text exclusions. The prior blocking regression — `strip_intermediate_usage` firing when `include_usage: true` was requested but model metadata was unavailable — is resolved on HEAD `73a5b184` via the `model_attestation_supported == Some(false)` and `!chat_stream_include_usage_requested(request)` guards, now backed by dedicated regression tests.

No blocking findings.

Checks run: `cargo fmt --check` clean, `cargo clippy --all-targets` clean, and the relevant lib/unit tests pass (including the two regression tests for the `include_usage:true` + missing-metadata case). E2E tests requiring PostgreSQL/dstack attestation were skipped for environmental reasons; that logic was reviewed by reading the source.

lloydmak99 (dismissed):
Requesting changes for a privacy regression in the new reserialization path:

1. `crates/api/src/routes/completions.rs:1599` disables raw passthrough when `strip_intermediate_usage` is active, so default and `include_usage:false` non-attested streams now go through the parsed/re-serialized branch. That branch logs the full serialized chunk body at debug level on `crates/api/src/routes/completions.rs:1725`. Those chunks can contain customer prompt/completion content, whereas the previous passthrough path did not log the body. Please remove that per-chunk body log or suppress it when `strip_intermediate_usage` is active.

2. The new passthrough guards for non-attested E2EE and non-text modality streams are not directly tested; the current unit cases exercise those guards with `model_attestation_supported=Some(true)`, where stripping is already disabled. Please add a regression for `Some(false)` with E2EE enabled and `Some(false)` with non-text modalities.

Validation run locally:

Evrard-Nil (commented):
Fixed the privacy regression: removed the chunk-body debug log entirely from the re-serialization path.

The original guard (`if !auto_redact_enabled`) was insufficient — `auto_redact_enabled` being false is the normal non-PII case, so the log still fired for regular non-attested streams when `strip_intermediate_usage` forced them into the re-serialization path. `json_data` always contains AI response text deltas, which must never appear in logs at any level per the project logging policy.

The fix removes the `tracing::debug!` call entirely and replaces it with a comment explaining why logging is forbidden here. No functional change; 391 unit tests pass.

Evrard-Nil (commented):
Addressed all review comments. Per-chunk body debug log removed from re-serialization path. Regression tests added for E2EE and non-text modality guards.

Evrard-Nil:
Addressed lloydmak99's feedback:
1. Removed the per-chunk debug body log (which could expose customer content when strip_intermediate_usage is active).
2. Added regression tests for the passthrough guards with model_attestation_supported=Some(false): E2EE enabled case and non-text modality case both correctly skip strip_intermediate_usage.

PierreLeGuen (approved):
Clean fix for the streaming usage spec issue. The `strip_intermediate_usage` path is correctly guarded—mutually exclusive with `include_usage`, `continuous_usage_stats`, E2EE, non-text modalities, and attested models (`model_attestation_supported == Some(true)` or `None` keep byte-exact passthrough, so TEE signature verification is unaffected). Since `usage` has no `skip_serializing_if`, content chunks serialize as `usage: null` as the spec requires, and chat billing runs through `InterceptStream` independent of the client-facing stream, so stripping usage from client chunks doesn't affect billing.

No blocking issues found.

Local checks: `cargo test --lib --bins` (392 passed), `cargo check -p api --tests` / `cargo check --test e2e_all` (compile clean), `cargo clippy --bin api` (no warnings), `cargo fmt --check` and `git diff --check` passed. E2E suite not run locally (requires PostgreSQL); CI shows Test Suite, Lint, and security_audit green.

PierreLeGuen (approved):
Streaming now emits `usage: null` in intermediate chunks per the OpenAI spec, and the prior blocking issues from earlier reviews have been resolved. The change is sound and ready to merge.

- `crates/api/src/routes/completions.rs:405-411` — The `strip_intermediate_usage` gate is correctly narrowed: it only fires for `stream==true` with a non-attested model, and falls back to passthrough when `include_usage:true` and metadata lookup fails, so usage is never silently dropped. Attested-model byte-exact passthrough is preserved.
- `crates/api/src/routes/completions.rs:1719-1724` — The per-chunk `tracing::debug!` that could reflect user/AI content in the re-serialization path has been removed; billing is unaffected since chat usage is recorded separately via `InterceptStream`.
- `usage: Option<TokenUsage>` on `ChatCompletionChunk` has no `skip_serializing_if`, so stripped chunks correctly serialize as `usage: null`, and `#[serde(flatten)] extra` preserves unknown upstream fields.

Checks: `cargo test -p api` unit/lib suites pass (streaming, usage, E2EE, non-text-modality, and `include_usage:true` metadata-failure regression guards), `cargo clippy --bin api` clean, `cargo fmt --check` clean. E2E suite requiring PostgreSQL was not run; that path was reviewed statically.

PierreLeGuen (approved):
Clean, well-contained fix. Nulling `usage` in the client-facing output doesn't affect billing, which is captured separately via `InterceptStream`, and the strip path only fires for non-attested models, preserving byte-exact passthrough for attested/TEE-signed streams. The `Option<TokenUsage>` field has no `skip_serializing_if`, so `None` correctly serializes as `usage: null` per spec, and the `!chat_stream_include_usage_requested` + `None`-metadata fallback (with regression tests) ensures a client that asks for usage never has it silently dropped.

Non-blocking observations:
- `completions.rs:405-411` — `strip_intermediate_usage` is scoped to non-attested models, so attested-model streams still forward populated `usage` in intermediate chunks; the #646 spec violation is only fixed for external models. This is a deliberate trade-off to preserve TEE signature verification — flagging only to confirm partial coverage is acceptable.

Checks run locally: `cargo fmt --check` clean, `cargo clippy --bin api` clean, `cargo check -p api --tests` passed, `cargo test --lib --bins` green (392 passed). E2E tests requiring PostgreSQL were reviewed statically only. Confirmed the per-chunk debug log was removed from the re-serialization path.

PierreLeGuen (approved):
The change correctly nulls `usage` on content chunks and drops usage-only chunks for non-attested streaming responses, satisfying the #646 spec fix. Both prior blocking issues (the `include_usage:true` + missing-metadata regression and the debug-log content leak) are resolved on HEAD, and the logic verifies end-to-end.

Non-blocking observations:
- `completions.rs:405-411`: `strip_intermediate_usage` is gated on `model_attestation_supported == Some(false)`, so attested/TEE models still emit populated `usage` in intermediate chunks — the spec fix applies only to non-attested models. This is a deliberate trade-off to preserve byte-exact TEE signature verification; acceptable as-is.

Checks: `cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test --lib --bins` (392 passed) all clean. PostgreSQL-dependent e2e suite was not run locally (environmental); e2e logic reviewed statically.

PierreLeGuen (approved):
Correctly emits `usage: null` on intermediate streaming chunks per the OpenAI spec. The change is well-contained and the gating is sound — `strip_intermediate_usage` only fires for `model_attestation_supported == Some(false)` and explicitly excludes `include_usage:true`, continuous-usage, E2EE, non-text modalities, and the attested passthrough/rewrite paths, so byte-exact TEE passthrough is preserved and usage is never dropped when the client asked for it.

Notes (non-blocking):
- `ChatCompletionChunk.usage` (models.rs:434) has no `skip_serializing_if`, so `None` serializes as `usage: null` as required — relies on this staying put; worth a comment if anyone touches that struct.
- `[DONE]` and control lines still forward raw in the strip path, and billing is unaffected (captured separately via `InterceptStream`); both verified.

Checks: `cargo fmt --check`, `git diff --check`, `cargo clippy -p api`, `cargo check -p api --tests`, and targeted `cargo test -p api` (`chat_stream`, `usage` — 32 tests) all passed. CI Lint and security_audit green; Test Suite still running at review time.

PierreLeGuen (approved):
The fix correctly emits `usage: null` on intermediate streaming chunks per the OpenAI spec. The `strip_intermediate_usage` path re-serializes via `prepare_stream_chunk_for_client(..., include_usage=false, ...)`, and since `usage: Option<TokenUsage>` has no `skip_serializing_if`, `None` serializes as `"usage": null` as intended.

Non-blocking notes:
- Gating is sound: the strip path only fires for `model_attestation_supported == Some(false)` and is excluded for `include_usage:true`, `continuous_usage_stats`, E2EE, and the `rewrite_public_stream_usage` path, so attested/TEE streams keep byte-exact passthrough and signature verification is unaffected (`completions.rs:1595,1670`).
- Billing is unaffected — usage is recorded independently via `InterceptStream`.
- The two previously-raised blocking issues (the `include_usage:true` + missing-metadata regression, guarded by `chat_stream_include_usage_requested` at `completions.rs:408`; and the debug-log content leak) are both resolved on HEAD.

Local checks: `cargo fmt --check`, `cargo clippy --bin api`, and `cargo test --lib --bins` (392 passed) all clean; e2e target (`e2e_all`) and `inference_providers` compile clean.

PierreLeGuen (approved):
Fixes #646 by emitting `usage: null` on intermediate streaming chunks per the OpenAI spec. The change is well-contained and the core logic is correctly guarded; both previously-raised blocking issues (the `include_usage:true` missing-metadata regression and the per-chunk debug-log content leak) are resolved on HEAD and covered by regression tests.

- `completions.rs:405-411`: `strip_intermediate_usage` is gated on `model_attestation_supported == Some(false)`, so attested/TEE models still emit populated `usage` in intermediate chunks. This is a deliberate trade-off to preserve byte-exact TEE signature verification — flagging only to confirm the partial #646 coverage is intentional.

The strip path is correctly mutually exclusive with `rewrite_public_stream_usage` and gated against E2EE, `continuous_usage_stats`, non-text modalities, and explicit `include_usage:true`; since `ChatCompletionChunk.usage` has no `skip_serializing_if`, `None` serializes as `usage: null` as required. Billing is unaffected (recorded separately via `InterceptStream`).

Checks: `cargo test --lib -p api` (incl. new `chat_stream_usage_mode` / `default_chat_stream_*` and the `include_usage_true...missing_metadata` regression tests) passed; `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` clean. E2E tests not run (require PostgreSQL); that path reviewed statically.

PierreLeGuen (approved):
The fix is correct and well-scoped: intermediate streaming chunks now emit `usage: null` per the OpenAI spec, gated so usage is never dropped when the client asks for it.

- `strip_intermediate_usage` is properly gated on `model_attestation_supported == Some(false)` and is mutually exclusive with `include_usage`/`continuous_usage_stats`/E2EE/non-text modalities/rewrite, so attested (TEE) and E2EE streams keep byte-exact passthrough.
- `ChatCompletionChunk.usage` has no `skip_serializing_if` (`crates/inference_providers/src/models.rs:462`), so `None` serializes as `"usage": null` rather than being omitted — matching the spec and the e2e assertion.
- Usage-only upstream chunks are dropped and terminal finish-reason chunks have usage nulled while preserving the choice; both covered by tests.
- Billing/usage accounting is unaffected — it taps the upstream stream via `InterceptStream`, independent of the reshaped client-facing bytes.
- The per-chunk debug log was removed from the re-serialization path, which is a genuine privacy improvement since that path now carries response deltas for non-attested models.

Checks: `cargo fmt --check`, `cargo clippy --bin api -- -D warnings`, `cargo test --lib --bins` (392 passed), and `cargo check --test e2e_all` all pass. PostgreSQL-backed e2e tests were not run (environmental).

PierreLeGuen (approved):
The streaming fix is well-isolated and correct: `strip_intermediate_usage` only fires for non-attested models, preserving byte-exact passthrough for attested/TEE streams, E2EE, non-text modalities, `continuous_usage_stats`, and explicit `include_usage:true` (including the metadata-lookup-failure `None` case), all covered by tests.

- The strip path uses `prepare_stream_chunk_for_client(include_usage=false)`, which nulls `usage` on content chunks and drops usage-only chunks while still forwarding `[DONE]` via the control-event branch (completions.rs:1675).
- `usage: Option<TokenUsage>` (chunk struct line 462) has no `#[serde(skip_serializing_if)]`, so `None` correctly serializes as `"usage": null` per the OpenAI spec.
- Billing is unaffected — chat usage is recorded separately via `InterceptStream` at the service layer (completions.rs:1430), not from this client-facing byte stream.

Checks run: `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean, `cargo test --lib --bins` (392 passed) plus targeted `include_usage_true` / `strip_intermediate_usage` / `non_attested_stream` tests passing, `cargo check --test e2e_all` compiles. E2E suite requiring PostgreSQL was skipped; CI Lint and security_audit passed.

PierreLeGuen (approved):
Well-scoped fix: for non-attested OpenAI-format upstreams, intermediate streaming chunks now emit `usage: null` per the OpenAI spec, gated correctly and mutually exclusive with the other usage-shaping paths. The gating logic and serialization are sound and billing is unaffected.

- `crates/api/src/routes/completions.rs:1595-1682`: default/`include_usage:false` streams for non-attested OpenAI upstreams now leave byte-exact passthrough for a parse-then-re-serialize path. Correct (`#[serde(flatten)] extra` preserves unknown fields, `usage: Option` serializes as `usage: null`), but slightly costlier for external models — flagged for awareness.
- `crates/api/src/routes/completions.rs:1719-1724`: the per-chunk `tracing::debug!` of chunk content was removed — a good privacy improvement, since re-serialized chunks can contain user/response text.

Verified: `strip_intermediate_usage` is scoped to non-attested models, and billing is recorded via `InterceptStream` independently of forwarded chunk bytes, so stripping `usage` cannot affect accounting.

Checks: `cargo fmt --check`, `cargo clippy --bin api -- -D warnings`, and `cargo test --lib` (incl. `strip_intermediate_usage`/attested/E2EE/modality/metadata-failure tests) all pass. E2E tests (PostgreSQL-dependent) were not run locally.

Evrard-Nil:
Branch updated to main (merge commit cb965fb4).

@lloydmak99 — both issues from your review have been addressed on earlier commits:
1. **Per-chunk debug log** (privacy regression): removed entirely from the re-serialization path in commit 2c041472 — `json_data` no longer appears in logs at any level.
2. **E2EE + non-text modality regression tests for `Some(false)`**: added in commit a130bf23 — guards are now tested with `model_attestation_supported=Some(false)` for both E2EE-enabled and non-text-modality cases.

Could you take another look? CI (Lint, Test Suite, security_audit) is green.

PR #802, opened 2026-06-16
Fix batch_upsert_models: merge not replace on partial PATCH (fixes #775)

### LINKED ISSUE #775: batch_upsert_models: partial inference_url PATCH wholesale-replaces provider state, forcing full re-attestation churn
Follow-up from PR #771 review (PierreLeGuen non-blocking observation). Pre-existing — not introduced by #771.

### Problem
`batch_upsert_models` (`crates/api/src/routes/admin.rs:511`) calls `load_inference_url_models` with **only the PATCHed models**. That call wholesale-replaces `inference_url_providers` (`crates/services/src/inference_provider_pool/mod.rs:3641`) and prunes `inference_url_fingerprint_states` down to that subset (`mod.rs:3656`).

Routing stays **correct** — other models' provider `Arc`s remain in `model_to_providers`, so live traffic isn't misrouted. But the next periodic refresh rebuilds every *other* URL's provider from `Bootstrap`:
- full re-attestation of providers that were never touched by the PATCH, and
- fail-closed replacement if discovery transiently fails at that moment.

This is a **churn / availability risk, not a correctness bug**.

### Suggested fix
Have `load_inference_url_models` **merge** rather than replace when given a partial set, leaving full-set reconciliation to `sync_inference_url_models`.

### Refs
- PR #771 (chutes post-merge polish) review thread
- `crates/api/src/routes/admin.rs:511`
- `crates/services/src/inference_provider_pool/mod.rs:3641`, `:3656`

Evrard-Nil:
## Summary

When `PATCH /v1/admin/models` was called with a subset of models, `load_inference_url_models` was wholesale-replacing `inference_url_providers` and pruning `inference_url_fingerprint_states` down to that subset — evicting provider state for all untouched models and forcing full re-attestation on the next periodic refresh. If discovery transiently failed during that refresh, those models would fail-close despite not having been touched.

The fix adds a `partial: bool` parameter to `load_inference_url_models`:

- **`partial = true`** (admin PATCH path, `batch_upsert_models`): merge-upserts only the provided models into the existing URL-provider cache and fingerprint-state map. Untouched entries are preserved.
- **`partial = false`** (periodic sync via `sync_inference_url_models`, startup load in `lib.rs`): existing replace-all behaviour is preserved so stale entries are pruned as expected.

All three existing call sites outside the periodic sync path (`batch_upsert_models`, startup load in `lib.rs`, and three unit-test direct calls) are updated to compile.

## Test plan
- [x] `cargo build` clean (no errors, no new warnings)
- [x] `cargo test --lib --bins` passes (new test `test_partial_load_preserves_untouched_providers` added)
- [ ] `PATCH /v1/admin/models` on a single model does not evict provider state for other models (staging verification)

Closes #775

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -858,7 +858,7 @@ pub async fn init_inference_providers(
     match models_source.fetch_inference_url_models().await {
         Ok(models) if !models.is_empty() => {
             tracing::info!(count = models.len(), "Loading inference_url models");
-            pool.load_inference_url_models(models).await;
+            pool.load_inference_url_models(models, false).await;
         }
         Ok(_) => {
             tracing::info!("No inference_url models found in database");
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -519,7 +519,7 @@ pub async fn batch_upsert_models(
         );
         app_state
             .inference_provider_pool
-            .load_inference_url_models(inference_url_models)
+            .load_inference_url_models(inference_url_models, true)
             .await;
     }
 
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -3133,12 +3133,13 @@ impl InferenceProviderPool {
     /// * `models` - List of (model_name, inference_url) tuples
     /// Load (or refresh) a set of inference_url models into the provider pool.
     ///
-    /// When `partial = true` (called from the admin PATCH path), only the
-    /// provided models are upserted — existing state for untouched models is
-    /// preserved.  When `partial = false` (called from the periodic sync path
-    /// via `sync_inference_url_models`), the URL-provider cache and fingerprint
-    /// state map are replaced wholesale, pruning entries for models that are no
-    /// longer in the active set.
+    /// The `partial` flag controls whether this is a merge or a replace:
+    ///
+    /// - `partial = true` (admin PATCH path): only the provided models are
+    ///   upserted — existing state for untouched models is preserved.
+    /// - `partial = false` (periodic sync / startup): the URL-provider cache
+    ///   and fingerprint-state map are replaced wholesale, pruning entries for
+    ///   models that are no longer in the active set.
     pub async fn load_inference_url_models(&self, models: Vec<(String, String)>, partial: bool) {
         if models.is_empty() {
             return;
@@ -7046,7 +7047,7 @@ mod tests {
 
         // Pre-seed the URL cache with two "existing" models as if they were
         // loaded during a prior full sync.
-        let untouched_model = "model-untouched".to_string();
+        let _untouched_model = "model-untouched".to_string();
         let untouched_url = "https://untouched.completions.near.ai".to_string();
         let untouched_provider = Arc::new(MockProvider::new());
 
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -3131,7 +3131,15 @@ impl InferenceProviderPool {
     ///
     /// # Arguments
     /// * `models` - List of (model_name, inference_url) tuples
-    pub async fn load_inference_url_models(&self, models: Vec<(String, String)>) {
+    /// Load (or refresh) a set of inference_url models into the provider pool.
+    ///
+    /// When `partial = true` (called from the admin PATCH path), only the
+    /// provided models are upserted — existing state for untouched models is
+    /// preserved.  When `partial = false` (called from the periodic sync path
+    /// via `sync_inference_url_models`), the URL-provider cache and fingerprint
+    /// state map are replaced wholesale, pruning entries for models that are no
+    /// longer in the active set.
+    pub async fn load_inference_url_models(&self, models: Vec<(String, String)>, partial: bool) {
         if models.is_empty() {
             return;
         }
@@ -3808,23 +3816,42 @@ impl InferenceProviderPool {
             "pubkey_to_providers state after update"
         );
 
-        // Update the URL→provider cache
-        *self.inference_url_providers.write().await = new_url_cache;
+        // Update the URL→provider cache.
+        //
+        // Partial mode (admin PATCH): merge new entries into the existing map
+        // so that untouched models keep their warm providers and fingerprint
+        // state.  Full mode (periodic sync): replace the map entirely so that
+        // URLs for models that have been removed are pruned.
+        {
+            let mut cache = self.inference_url_providers.write().await;
+            if partial {
+                // Merge: only insert/replace the URLs we just processed.
+                for (url, provider) in new_url_cache {
+                    cache.insert(url, provider);
+                }
+            } else {
+                *cache = new_url_cache;
+            }
+        }
 
         // Rebuild the URL → FingerprintState map so its key set exactly
         // matches the active inference_url set:
         // - Newly-created URLs take the freshly-discovered state.
         // - Reused URLs keep their existing entries (cumulative discovery
         //   mutated the Arc in place).
-        // - URLs no longer in the active set are pruned — prevents a slow
-        //   leak of stale per-URL state as models are added and removed.
+        // - URLs no longer in the active set are pruned on a full sync —
+        //   prevents a slow leak of stale per-URL state as models are added
+        //   and removed.  On a partial sync we only insert new states and
+        //   leave existing ones untouched.
         {
             let mut states = self.inference_url_fingerprint_states.write().await;
             for (url, state) in new_fingerprint_states {
                 states.insert(url, state);
             }
-            let active_urls = self.inference_url_providers.read().await;
-            states.retain(|url, _| active_urls.contains_key(url));
+            if !partial {
+                let active_urls = self.inference_url_providers.read().await;
+                states.retain(|url, _| active_urls.contains_key(url));
+            }
         }
 
         info!(
@@ -3848,7 +3875,7 @@ impl InferenceProviderPool {
         // load_inference_url_models directly (partial batch, no prune).
         let complete_names: std::collections::HashSet<String> =
             models.iter().map(|(name, _)| name.clone()).collect();
-        self.load_inference_url_models(models).await;
+        self.load_inference_url_models(models, false).await;
         self.prune_stale_pinned(&complete_names).await;
     }
 
@@ -5670,7 +5697,7 @@ mod tests {
 
         // Call load_inference_url_models — the provider should be reused and
         // the self-healing path should detect missing pubkeys and re-fetch them.
-        pool.load_inference_url_models(vec![(model_id.clone(), url)])
+        pool.load_inference_url_models(vec![(model_id.clone(), url)], false)
             .await;
 
         // Verify pubkeys were recovered
@@ -6228,7 +6255,7 @@ mod tests {
         mock_provider.set_fail_attestation(true);
 
         // Load — the provider is reused, pubkeys are missing, re-fetch fails
-        pool.load_inference_url_models(vec![(model_id.clone(), url.clone())])
+        pool.load_inference_url_models(vec![(model_id.clone(), url.clone())], false)
             .await;
 
         // The URL should have been evicted from the cache
@@ -6303,7 +6330,7 @@ mod tests {
                 .insert("pretend-pubkey".to_string(), vec![mock.clone()]);
         }
 
-        pool.load_inference_url_models(vec![(model_id.clone(), url.clone())])
+        pool.load_inference_url_models(vec![(model_id.clone(), url.clone())], false)
             .await;
 
         // Blocked URL evicted from URL cache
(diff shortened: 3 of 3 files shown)


PierreLeGuen (approved):
The fix correctly scopes the wholesale cache replacements behind a new `partial` flag: in PATCH mode the URL cache is merge-inserted (`mod.rs:3785`) and the fingerprint-state prune is skipped (`mod.rs:3806`), so untouched models no longer lose provider/attestation state. All four call sites pass the right value (PATCH→`true` in `admin.rs:521`; startup `lib.rs:848`, periodic sync `mod.rs:3838`, and existing tests→`false`), and per-model `model_to_providers`/`pubkey_to_providers` updates already had merge semantics, so there's no partial-mode hole. The core change is sound.

One thing to clean up before merge:
- `crates/services/src/inference_provider_pool/mod.rs:6934`: the new test binds `untouched_model` but never uses it. CI compiles with `-D warnings`, so both Lint and Test Suite are currently failing on this (and on a `doc_lazy_continuation` lint at `mod.rs:3094`). Prefix with `_` or drop the binding to get CI green — trivial, not a correctness issue.

Checks run: `cargo check -p services` / `cargo build` clean; `cargo test -p services --lib test_partial_load_preserves_untouched_providers` passes; `cargo clippy --all-targets` surfaces the two lints above; inspected failed CI run 27650181060 confirming the lint/test failures.

lloydmak99 (dismissed):
Requesting changes. The partial merge fixes the main eviction bug, but URL-change cleanup is incomplete:

1. `crates/services/src/inference_provider_pool/mod.rs:3788` merges new URL-cache entries in partial mode without removing the old URL entry for a model whose `inference_url` changed. The admin route unregisters the model mapping before reloading, but `unregister_provider` does not remove `inference_url_providers` or `inference_url_fingerprint_states`, so repeated admin URL changes can leave old provider Arcs, connection pools, and fingerprint state alive until a later full refresh, or indefinitely if full refresh is disabled. Please prune the old cache/state entries for replaced patched models before inserting the new URL.

2. `crates/api/src/routes/admin.rs:521` is the route-level call that makes this behavior matter, but the new test only exercises `load_inference_url_models` directly. Please add an admin PATCH regression test that proves a one-model inferenceUrl update preserves unrelated cached providers and cleans up the replaced URL.

Validation run locally:

Evrard-Nil (commented):
Fixed: partial mode now evicts the old URL key from `inference_url_providers` and `inference_url_fingerprint_states` when a model's `inference_url` changes.

Two-pronged fix:
1. `unregister_provider` now also cleans up URL-cache and fingerprint-state entries for providers no longer referenced by any model. This covers the common non-pinned path where unregister is called before `load_inference_url_models`.
2. `load_inference_url_models` in partial mode collects old provider pointers still in `model_to_providers` (for pinned models that skip unregister) and evicts their URL entries if not in the new URL set.

New test `test_partial_load_evicts_old_url_on_url_change` verifies old URL is evicted and new URL is present after a URL-change PATCH. All 393 tests pass.

Evrard-Nil (commented):
Addressed all review comments. Old URL eviction on model URL change is now handled in both unregister_provider and partial load paths.

Evrard-Nil:
Addressed: removed unused `_untouched_model` variable binding from the test. CI should be green now.

PierreLeGuen (approved):
Adds a `partial` flag to `load_inference_url_models` so admin PATCH (`batch_upsert_models`) merge-upserts instead of wholesale-replacing the URL→provider cache and fingerprint-state map, fixing the over-eviction in #775. The two-pronged old-URL eviction (`unregister_provider` for non-pinned + the partial-path `stale_url_cache_entries` for pinned) cleanly handles the URL-change edge case from prior review, and the change looks correct.

One optional follow-up (non-blocking):
- `crates/services/src/inference_provider_pool/mod.rs:3761-3801` — the partial-mode `stale_url_cache_entries` eviction lacks the "still referenced by another model" guard that `unregister_provider` has (`still_live_ptrs`, lines 619-631). Since co-aliased models sharing one `inference_url` reuse a single provider Arc (line 3184), a PATCH that changes one alias's URL while its co-alias is absent from the batch could evict a still-live URL/fingerprint entry. Narrow config, so not blocking — worth mirroring the `still_live_ptrs` guard here.

Local checks: `cargo clippy -p services --all-targets` and `cargo clippy -p api --all-targets` clean; `cargo check -p services` clean; `cargo test -p services --lib test_partial_load` 2/2 pass; `git diff --check` clean. Remote Lint passed (Test Suite / security_audit pending at review time).

PierreLeGuen (approved):
The `partial` flag correctly scopes the wholesale cache replacement and fingerprint-state prune to full-sync only, so admin PATCH now merges instead of replacing. Stale-URL eviction on URL change is handled on both the non-pinned path (`unregister_provider`) and the pinned path (`stale_url_cache_entries`), addressing the earlier old-URL leak. Looks good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:657-658`: `unregister_provider` awaits the `inference_url_providers` read lock while holding the `inference_url_fingerprint_states` write lock. Not a deadlock risk — every site in the file uses the same `fingerprint_states` → `inference_url_providers` ordering — but the await-under-write-lock could be tightened by snapshotting the live URL set before taking the states write lock.

Local checks: `cargo test -p services --lib partial_load` (2/2 passed), `cargo check`/`cargo clippy -p services` clean, `git diff --check` clean; remote CI (security_audit, Lint, Test Suite) green. Full e2e suite not run (requires PostgreSQL; not relevant to this in-memory pool change).

PierreLeGuen (approved):
Well-scoped fix for #775: only `inference_url_providers` and `inference_url_fingerprint_states` were being wholesale-replaced, and those are exactly the two maps now gated behind `partial`. The earlier old-URL leak flagged on `inference_url` changes is handled by the two-pronged eviction (`unregister_provider` cleanup for non-pinned + `stale_url_cache_entries` for pinned). Logic looks correct; good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:3761-3801`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard that `unregister_provider` has (lines 622-631). If two pinned aliases share one provider/`inference_url` and a PATCH changes one alias's URL while the co-alias is absent from the batch, a still-live URL/fingerprint entry could be evicted. Narrow co-alias config, self-heals on next full sync — mirror the `still_live_ptrs` filter before building the stale set if you want to close it.

Local checks: `cargo check/clippy -p services` and `-p api` clean; `cargo test -p services --lib partial_load` 2/2 passed; `git diff --check` clean. Remote CI green on Lint and security_audit; Test Suite pending.

PierreLeGuen (approved):
The `partial` flag cleanly gates merge-vs-replace so admin PATCH merges into the two URL maps (`inference_url_providers`, `inference_url_fingerprint_states`) instead of wholesale-replacing them, and old-URL eviction on URL change is handled on both the non-pinned (`unregister_provider`) and pinned (`stale_url_cache_entries`) paths. Earlier blocking feedback on stale-URL eviction has been addressed. Looks good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:3761-3798`: the partial-mode `stale_url_cache_entries` eviction doesn't apply the alias-sharing guard (`still_live_ptrs`) that `unregister_provider` uses. If one old provider Arc were shared across models at the same URL and only one is PATCHed, its URL entry could be evicted while another model still references it — impact is limited to an unnecessary re-attestation on the next refresh, not a serving break.

Checks: `cargo clippy --all-targets` clean; `cargo test -p services --lib partial_load` 2/2 pass; `cargo check -p services --tests` clean. Remote CI: security audit and lint passed, test suite still running at review time.

PierreLeGuen (approved):
The core merge-not-replace fix for #775 is sound: the `partial` flag cleanly gates the two wholesale-replaced maps (`inference_url_providers`, `inference_url_fingerprint_states`) behind PATCH semantics while other shared state stays keyed per-model. Old-URL eviction on `inference_url` change is handled in two complementary paths — `unregister_provider` for non-pinned models and `stale_url_cache_entries` in `load_inference_url_models` for pinned ones — and all prior blocking feedback (old-URL leak, clippy lints) has been addressed.

- `crates/services/src/inference_provider_pool/mod.rs:3760-3801` — the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard used in `unregister_provider` (~620-631). Worst case if two pinned aliases share a provider Arc and only one is PATCHed is a single unnecessary re-attestation on next refresh, which self-heals; not a serving break.

CHECKS:
- `cargo clippy -p services --lib --all-features` / `cargo clippy -p api --all-targets`: clean.
- `cargo test -p services --lib`: 394 passed, 1 ignored, including the 2 new tests (`test_partial_load_preserves_untouched_providers`, `test_partial_load_evicts_old_url_on_url_change`).
- `git diff --check`: clean. Remote CI was pending/green at review time.

PierreLeGuen (approved):
Clean, well-scoped fix: partial PATCH now merges into existing models instead of replacing them, and the old-URL leak flagged earlier was addressed. Prior blocking feedback and clippy lints are resolved; both new tests pass.

- `crates/services/src/inference_provider_pool/mod.rs:3761-3798`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard that `unregister_provider` uses (lines 622-631). If two co-aliased models share one provider Arc at the same `inference_url` and a PATCH changes one alias's URL while the co-alias is absent from the batch, the old URL key could be evicted while still referenced. Impact is bounded to one unnecessary re-attestation on the next periodic refresh (self-heals); serving is uninterrupted since `model_to_providers` stays intact. Optional fix: mirror the `still_live_ptrs` filter before building `old_ptrs`.

Checks run: `cargo clippy -p services --lib --all-features` / `--all-targets` clean; `cargo test -p services --lib partial_load` 2/2 passing (full `-p services --lib` suite 394 passed, 1 ignored); `git diff --check` clean. E2e suite not run (requires PostgreSQL, irrelevant to this in-memory change).

PierreLeGuen (approved):
The core fix is well-scoped and correct: the two maps that were being wholesale-replaced (`inference_url_providers` and `inference_url_fingerprint_states`) are exactly the two now gated behind `partial`, while per-model state (`model_to_providers`, `pubkey_to_providers`, failure counts) is untouched, so untouched models are unaffected. The in-function eviction path only iterates URLs derived from the input `models` set, so partial mode cannot evict an untouched model, and the old-URL-change leak that earlier blocked this PR is now handled via `unregister_provider` cleanup and `stale_url_cache_entries`. All four call sites pass the correct `partial` value.

- `crates/services/src/inference_provider_pool/mod.rs:3760-3801`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard used by `unregister_provider` (lines 622-631). If two co-aliased models share one provider Arc at the same `inference_url` and a PATCH changes one alias's URL while the co-alias is absent from the batch, the old URL/fingerprint entry could be evicted while still referenced. Impact is bounded: `model_to_providers` stays intact so serving is uninterrupted, and the entry self-heals via one re-attestation on the next periodic refresh. Could be addressed by mirroring the `still_live_ptrs` filter before building `old_ptrs`.

Local checks: `cargo clippy` (services/api/all-targets) clean, `cargo build`/`cargo check` clean, `cargo test -p services --lib` 394 passed (incl. both new partial-load tests). CI Lint and security_audit passed; Test Suite was still in progress at review time.

PierreLeGuen (approved):
The fix is well-scoped: only the two maps that were being wholesale-replaced/pruned (`inference_url_providers`, `inference_url_fingerprint_states`) are gated behind the new `partial` flag, while per-model state (`model_to_providers`, `pubkey_to_providers`, failure counts) is untouched. Call sites are correct (PATCH→`true` at `admin.rs:521`; startup→`false` at `lib.rs:861`; periodic sync→`false` at `mod.rs:3979`; tests→`false`), old-URL eviction on URL change is handled via both `unregister_provider` and `stale_url_cache_entries`, and lock ordering is unchanged. Prior blocking feedback has been addressed.

- `crates/services/src/inference_provider_pool/mod.rs:3761-3798`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard that `unregister_provider` uses (lines 622-631). For the rare case of two co-aliased pinned models sharing a provider Arc where only one is PATCHed to a new URL, the old URL entry could be evicted while the sibling alias still references it, forcing a transient re-attestation on the next refresh. If pursued, filter `old_ptrs` against pointers still present in `model_to_providers` after the mapping update.

Checks: `cargo clippy -p services --all-targets` and `-p api` clean, `cargo build`/`cargo check` clean, and `cargo test -p services --lib partial_load` passes 2/2 (`test_partial_load_preserves_untouched_providers`, `test_partial_load_evicts_old_url_on_url_change`). E2e tests skipped (require PostgreSQL; irrelevant to this in-memory pool change).

PierreLeGuen (approved):
The merge-not-replace fix is correctly scoped: only the two maps that were being wholesale-replaced/pruned (`inference_url_providers`, `inference_url_fingerprint_states`) are now gated behind `partial`, while per-model state (`model_to_providers`, `pubkey_to_providers`, failure counts) was already keyed per-model and is unaffected. All four call sites pass the correct `partial` value (PATCH→`true` at `admin.rs:521`; startup/periodic sync/tests→`false`). Looks good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:3760-3801`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard that `unregister_provider` uses (lines 622-631). If two co-aliased models share one provider `Arc` at the same `inference_url` and a PATCH changes one alias's URL while the co-alias is absent from the batch, the old URL/fingerprint entry could be evicted while the sibling alias still references it. Impact is bounded — serving stays up via `model_to_providers`, and the entry self-heals on the next periodic full sync. Could mirror the `still_live_ptrs` filter before building `old_ptrs`.

Checks: `cargo clippy -p services --all-targets` clean; `cargo test -p services --lib partial_load` 2/2 pass; `cargo check -p services -p api` clean; `git diff --check` clean. E2E not run (requires PostgreSQL, irrelevant to this in-memory pool change).

PierreLeGuen (approved):
PR #802 adds a `partial: bool` flag to `load_inference_url_models` so admin PATCHes via `batch_upsert_models` merge-upsert only the provided models into `inference_url_providers` and `inference_url_fingerprint_states` instead of wholesale-replacing them, fixing the #775 over-eviction. Full mode (startup, periodic sync, tests) keeps prune-all behavior; all four call sites pass the correct flag, and old-URL eviction on `inference_url` change is handled on both the non-pinned path (`unregister_provider`) and the pinned path (`stale_url_cache_entries`). Looks correct and ready to merge.

- `crates/services/src/inference_provider_pool/mod.rs:3761-3798`: the partial-mode `stale_url_cache_entries` eviction lacks the `still_live_ptrs` guard that `unregister_provider` uses (lines 622-631). In a rare co-aliased-pinned config where two pinned models share one provider `Arc` at the same URL and a PATCH changes one alias while the other is absent from the batch, the old entry could be evicted while still referenced. Impact is bounded — one unnecessary re-attestation on the next refresh, serving uninterrupted, self-heals. Could mirror the `still_live_ptrs` filter before building `old_ptrs`.

Checks: `cargo test -p services --lib partial_load` (2/2 passed), `cargo check -p services -p api` and `cargo clippy -p services -p api --all-targets` clean, `git diff --check` clean. Full e2e suite not run (requires PostgreSQL; irrelevant to this in-memory pool change).

PR #803, opened 2026-06-16
Remap ServiceOverloaded → HTTP 429 (was 529) (fixes #738)

### LINKED ISSUE #738: OpenRouter routing rank: map overload to 429 (not 529) on the OR path + right-size its concurrent limit
## Summary

cloud-api surfaces backend exhaustion as **HTTP 529** (`CompletionError::ServiceOverloaded` → `status_overloaded()` in `crates/api/src/routes/common.rs:13`). OpenRouter's provider uptime calculation **excludes 429 but counts every 500+ response (including 529) as downtime**. So our overload signal, which we deliberately chose as "retry with backoff," is read by OpenRouter as a hard outage and degrades our routing rank:

- 95%+ uptime → normal routing
- 80–94% → lower priority
- <80% → fallback-only

429 (per-org concurrent-slot exhaustion, `try_acquire_concurrent_slot` in `crates/services/src/completions/mod.rs:1016`) is correctly excluded by OpenRouter and is exactly the "back off" signal OR wants. OpenRouter explicitly asks providers to return early 429s under load and to **not queue**.

**This does NOT block the OpenRouter listing** — it affects post-listing routing rank, not eligibility.

## Current behavior (confirmed in code)

| Condition | Domain error | HTTP | OR treats as |
|---|---|---|---|
| Per-(org,model) concurrent cap hit (`DEFAULT_CONCURRENT_LIMIT = 64`, `ports.rs:9`) | `RateLimitExceeded` | **429** | back off (excluded from uptime) ✅ |
| Upstream 429 (provider rate limit) | `RateLimitExceeded` | **429** | back off ✅ |
| All backends exhausted after `retry_with_fallback` rotation (SGLang `--max-queued-requests` 503s) | `ServiceOverloaded` | **529** | **downtime** ❌ |
| Upstream 503 | `ServiceOverloaded` | **529** | **downtime** ❌ |

Note: rotation-SNI (PR #637 / cloud-api.md "HTTP 529 vs 503") already made 529 **rare** in prod — ~1 event/9h, down 360× from before rotation. So the routing-rank exposure is small today, but it is a real and avoidable signal.

## Proposed fix (recommended)

Add an **OpenRouter-scoped 529→429 remap at the route layer**, opt-in, with zero behavior change for every other client:

- The chat-completions handler already has both `headers: HeaderMap` and `api_key.organization.id` in scope at the error arms (`crates/api/src/routes/completions.rs:1126`, error arms at `:1528`/`:1634`). So either of these scoping mechanisms is a few lines:
  - **Header-based**: e.g. honor a request header (`x-overload-status: 429`) — clean, but requires OpenRouter to send a custom header on every request (may not be configurable on their side).
  - **Org-scoped** (recommended): when the request's org is the dedicated OpenRouter integration org, map `ServiceOverloaded` → 429 instead of 529. Self-contained, needs no cooperation from OR, and matches OR's documented "early 429, don't queue" expectation.
- Keep 529 as the default for all other clients (Anthropic-style "site overloaded, retry with backoff" is the correct semantic for our own SDK/UI clients and for honest observability).
- Body should still carry a clear `Retry-After`-style hint and `service_overloaded` error type so the meaning isn't lost.

**Effort: ~half a day.** One small route-layer branch + a unit test mirroring `test_map_domain_error_service_overloaded`. No service-layer, DB, or migration changes. The mapping function `map_domain_error_to_status` would gain an org/header-aware variant used only by the chat + responses handlers (the 5 per-handler `ServiceOverloaded` arms — audio/rerank/embeddings/privacy/score — are not OR-listed chat models, so they can keep 529).

### Alternative considered: keep 529 everywhere

Rationale to keep: 529 is the honest, correct status; rotation-SNI already made it rare; and remapping hides a real overload from our own dashboards if applied globally. **Rejected as a global default** but the org-scoped remap above gets the best of both — honest 529 for us, OR-friendly 429 for the OR path.

## Right-sized concurrent limit for the OpenRouter org

The per-(org,model) cap must be **below** the smallest single-backend admission window so that under OR-driven load we shed with **429 (our own cap)** *before* the request ever reaches a saturated SGLang/vLLM backend that would 503→529. Backend capacity (cvm-conf + `/machines` replica counts, 2026-06-08):

| Model | Per-backend admission | Replicas | Notes |
|---|---|---|---|
| zai-org/GLM-5.1-FP8 | ~128 running + `--max-queued-requests 8` | **5** | SGLang, queue-capped |
| Qwen/Qwen3.5-122B-A10B | memory-bound, **unbounded queue** | **1** | single-replica risk model |
| deepseek-ai/DeepSeek-V4-Flash | `--max-running-requests 128` | 1 | |
| Qwen/Qwen3.6-27B-FP8 | `--max-running-requests 128` | 1 | |
| Qwen/Qwen3.6-35B-A3B-FP8 | `--max-running-requests 128` | 3 | |
| google/gemma-4-31B-it | `--max-running-requests 64` | 3 | smallest running cap |
| Qwen/Qwen3-30B-A3B-Instruct-2507 | `--max-num-seqs 64` | 2 | |
| openai/gpt-oss-120b | vLLM default | 2 | |
| Qwen/Qwen3-VL-30B-A3B-Instruct | `--max-num-seqs 16` | 2 | smallest seq cap (vision) |

The cap is a single scalar applied **per (org, model)**, so it must respect the *smallest* single-backend window to guarantee a 429 instead of a 529 across the board. The binding constraints are gemma-4-31B (64 running) and the single-replica Qwen3.5-122B (no queue cap). A cap of **48** leaves headroom under the 64-running floor while still allowing meaningful OR throughput; it sits well under GLM-5.1's ~136/backend and the 128-running models. (If we later split per-model caps, OR could get a higher cap on the multi-replica 128-running models.)

```
PATCH /v1/admin/organizations/{OPENROUTER_ORG_ID}/concurrent-limit
Content-Type: application/json
Authorization: Bearer <NEAR_AI_CLOUD_ADMIN_ACCESS_TOKEN>

{"concurrent_limit": 48}
```

Exact curl (fill in the OpenRouter org UUID + admin token; do NOT run blindly):

```
curl -sS -X PATCH \
  "https://cloud-api.near.ai/v1/admin/organizations/<OPENROUTER_ORG_ID>/concurrent-limit" \
  -H "Authorization: Bearer <NEAR_AI_CLOUD_ADMIN_ACCESS_TOKEN>" \
  -d '{"concurrent_limit": 48}'
```

Cache note: read path has a 5-min moka TTL; the PATCH invalidates locally (PR #618) but the *other* replica (cpu01↔cpu02) waits out its own TTL (cloud-api.md "Multi-instance caveat").

## Acceptance

- [ ] OR-path (org-scoped) `ServiceOverloaded` returns 429, not 529; all other clients still get 529.
- [ ] Unit test asserting the org/header-scoped remap.
- [ ] OpenRouter org `concurrent_limit` set to 48 (or final agreed value) via the PATCH above.
- [ ] Verified on cloud-stg-api before prod.

Refs: `crates/api/src/routes/common.rs:13,18`; `crates/services/src/completions/mod.rs:794,1016`; `crates/services/src/completions/ports.rs:9`; docs/cloud-api.md (HTTP 529 vs 503; Per-org concurrent-request limit); docs/inference.md (`--max-queued-requests`).

Evrard-Nil:
## Summary
`ServiceOverloaded` previously returned HTTP 529 (non-standard). OpenRouter counts all non-standard responses as downtime, degrading our routing rank. Additionally, 429 is the universally understood "back off and retry" signal — which is exactly the semantic we want.

- Changed `status_overloaded()` in `crates/api/src/routes/common.rs` to return `StatusCode::TOO_MANY_REQUESTS` (429)
- Updated the matching status code in `crates/services/src/responses/errors.rs` (responses API path)
- Updated all 9 OpenAPI `utoipa` response annotations in `completions.rs` from `status = 529` to `status = 429`
- Updated unit test asserting `.as_u16() == 529` → `429`
- Updated e2e tests asserting `529` → `429`

The error body retains `"type": "service_overloaded"` so it's distinguishable from per-org rate-limit 429s (`"type": "rate_limit_exceeded"`).

## Test plan
- [x] `cargo build` clean
- [x] `cargo test --lib -- common` passes (23 tests including `test_map_domain_error_service_overloaded`)
- [x] All 529 assertions in unit and e2e tests updated to 429
- [x] Error body still contains `"type": "service_overloaded"` — distinguishable from rate-limit 429s

Closes #738

### DIFF
--- a/crates/api/src/routes/common.rs
+++ b/crates/api/src/routes/common.rs
@@ -4,14 +4,18 @@ use services::completions::CompletionError;
 use services::organization::OrganizationError;
 use uuid::Uuid;
 
-/// HTTP 529 "Site Is Overloaded" — non-standard but widely used (Cloudflare,
-/// Anthropic) for "all backends exhausted; please retry with exponential
-/// backoff." Surfaced when `retry_with_fallback` has tried every inference
-/// backend for a model and they all rejected with 5xx — typically SGLang
-/// `--max-queued-requests` 503s under load. 503 stays reserved for narrower
-/// per-handler "this dependency is down" cases.
+/// HTTP 429 "Too Many Requests" — used here for the "all backends exhausted;
+/// please retry with exponential backoff" case. Surfaced when
+/// `retry_with_fallback` has tried every inference backend for a model and
+/// they all rejected with 5xx — typically SGLang `--max-queued-requests`
+/// 503s under load. 503 stays reserved for narrower per-handler "this
+/// dependency is down" cases. The error body carries `"type":
+/// "service_overloaded"` to distinguish it from per-org rate-limit 429s
+/// (`"type": "rate_limit_exceeded"`). 429 is the universally understood
+/// "back off and retry" signal; 529 is non-standard and some clients (e.g.
+/// OpenRouter) count non-2xx/4xx as downtime, degrading our routing rank.
 pub fn status_overloaded() -> StatusCode {
-    StatusCode::from_u16(529).expect("529 is a valid HTTP status code")
+    StatusCode::TOO_MANY_REQUESTS
 }
 
 /// Map domain errors to HTTP status codes
@@ -727,7 +731,7 @@ mod tests {
     #[test]
     fn test_map_domain_error_service_overloaded() {
         let error = CompletionError::ServiceOverloaded("overloaded".to_string());
-        assert_eq!(map_domain_error_to_status(&error).as_u16(), 529);
+        assert_eq!(map_domain_error_to_status(&error).as_u16(), 429);
     }
 
     #[test]
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -1239,7 +1239,7 @@ fn unsupported_completion_param(request: &CompletionRequest) -> Option<&'static 
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 402, description = "Insufficient credits", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -2017,7 +2017,7 @@ async fn chat_completions_inner(
         (status = 400, description = "Invalid request parameters", body = ErrorResponse),
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -3815,7 +3815,7 @@ mod tests {
         (status = 400, description = "Invalid request parameters", body = ErrorResponse),
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -4071,7 +4071,7 @@ pub async fn image_generations(
         (status = 401, description = "Unauthorized (missing or invalid API key)", body = ErrorResponse),
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
     ),
     security(("ApiKeyAuth" = []))
 )]
@@ -4464,7 +4464,7 @@ pub async fn audio_transcriptions(
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 413, description = "Payload too large", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -4970,7 +4970,7 @@ mod rerank_response_options_tests {
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 429, description = "Concurrent request limit exceeded for the organization. Max concurrent requests per model: 64 (configurable)", body = ErrorResponse),
         (status = 500, description = "Server error (billing failure, provider error)", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
     ),
     security(("ApiKeyAuth" = []))
 )]
@@ -5314,7 +5314,7 @@ fn classify_provider_error(
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 429, description = "Rate limit exceeded", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -5618,7 +5618,7 @@ struct PrivacyClassifyResponseDoc {
         (status = 401, description = "Unauthorized", body = ErrorResponse),
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 429, description = "Rate limit exceeded", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse)
     ),
     security(
@@ -6409,7 +6409,7 @@ pub async fn privacy_redact(
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 429, description = "Concurrent request limit exceeded for the organization. Max concurrent requests per model: 64 (configurable)", body = ErrorResponse),
         (status = 500, description = "Server error (billing failure, provider error)", body = ErrorResponse),
-        (status = 529, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
+        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
     ),
     security(("ApiKeyAuth" = []))
 )]
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -1239,7 +1239,7 @@ fn unsupported_completion_param(request: &CompletionRequest) -> Option<&'static 
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 402, description = "Insufficient credits", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -2017,7 +2017,7 @@ async fn chat_completions_inner(
         (status = 400, description = "Invalid request parameters", body = ErrorResponse),
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -3815,7 +3815,7 @@ mod tests {
         (status = 400, description = "Invalid request parameters", body = ErrorResponse),
         (status = 401, description = "Invalid or missing API key", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
@@ -4071,7 +4071,7 @@ pub async fn image_generations(
         (status = 401, description = "Unauthorized (missing or invalid API key)", body = ErrorResponse),
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse),
+        (status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", body = ErrorResponse),
     ),
     security(("ApiKeyAuth" = []))
 )]
@@ -4464,7 +4464,7 @@ pub async fn audio_transcriptions(
         (status = 404, description = "Model not found", body = ErrorResponse),
         (status = 413, description = "Payload too large", body = ErrorResponse),
         (status = 500, description = "Server error", body = ErrorResponse),
-        (status = 429, description = "All inference backends overloaded; retry with exponential backoff", body = ErrorResponse)
+        (status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", body = ErrorResponse)
     ),
     security(
         ("api_key" = [])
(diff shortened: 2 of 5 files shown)


PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from HTTP 529 → 429 across both status-mapping paths (`common.rs` and `responses/errors.rs`); the error body `type` field still distinguishes `service_overloaded` from `rate_limit_exceeded`, so clients can tell the two 429 cases apart. No remaining functional 529 references (only a doc comment and an unrelated SVG path).

- `crates/api/src/routes/completions.rs` (rerank ~4971, embeddings ~5315, privacy_classify ~5620, score ~6410): these endpoints now declare two `(status = 429, ...)` rows in one `responses(...)` block. utoipa keys responses by status code (last-write-wins), so the published OpenAPI spec drops one of the two 429 descriptions. Docs-only; consider merging into a single 429 row covering both `rate_limit_exceeded` and `service_overloaded`.

Checks: `cargo test -p api` mapping tests (incl. `test_map_domain_error_service_overloaded` and stream-error status mapping) passed; `cargo clippy -p api -p services -- -D warnings` clean; `grep -rn 529 crates/` confirmed no leftover functional references. E2E tests not run (require Postgres); CI green.

lloydmak99 (dismissed):
Requesting changes for an API documentation/coverage gap introduced by the 529 -> 429 remap:

1. `crates/api/src/routes/completions.rs:4973` and the same pattern on the other endpoints now document two different `429` responses: rate/concurrency limiting and backend overload. OpenAPI response maps are keyed by status code, so one of those contracts is likely to be overwritten/hidden, which undermines the PR's stated distinction between `rate_limit_exceeded` and `service_overloaded`. Please collapse these into a single 429 response description that names both error types, or use a schema shape that preserves both variants.

2. `crates/services/src/responses/errors.rs:163` remaps `CompletionError::ServiceOverloaded` to 429 for the Responses API, but the updated integration coverage is on chat/provider errors. Please add a Responses API regression test that asserts overloaded failures return HTTP 429 with `type=service_overloaded`.

Validation run locally:

Evrard-Nil (commented):
Fixed: merged the two 429 response annotations into a single entry per handler with a combined description distinguishing `rate_limit_exceeded` vs `service_overloaded` via the `error.type` field.

Updated 10 handlers across `completions.rs`: the 4 with duplicate entries had the redundant line removed, and all 10 now use the unified description: _"Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded."_

PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from 529 to 429. The change is consistent across both code paths and the error body still carries `"type": "service_overloaded"`, keeping it distinguishable from rate-limit 429s.

- Both mappings confirmed: `common.rs` via `status_overloaded()` and `responses/errors.rs:163`, with the `"service_overloaded"` type preserved (`responses/errors.rs:201`). No remaining code special-cases 529 for retry/fallback.
- `grep -rn 529` shows no functional references left — only an intentional doc comment (`common.rs:15`) and a coincidental substring in an SVG path (`auth.rs`).
- Prior duplicate-429 feedback addressed: the two OpenAPI 429 annotations were merged into single combined entries (commit `d323977`); these edits are doc-only.

Checks run locally: `cargo build -p api` clean, `cargo clippy -p api -p services -- -D warnings` clean, `cargo test -p api test_map_domain_error_service_overloaded` and `test_map_provider_error_503_becomes_service_overloaded` pass. E2e tests were not exercised (local shared DB migration state / requires PostgreSQL).

Evrard-Nil (commented):
Addressed all review comments. Duplicate 429 OpenAPI annotations have been merged into single unified entries per handler.

Evrard-Nil (commented):
Addressed issue 2 (the last remaining change-request item): added unit tests in `crates/services/src/responses/errors.rs` (`responses::errors::tests`) covering the Responses API 529→429 mapping:

- `test_responses_api_service_overloaded_returns_429` — asserts `http_status_code()` == 429
- `test_responses_api_service_overloaded_error_type` — asserts error body `type` == `"service_overloaded"`
- `test_responses_api_rate_limit_returns_429` — sibling case stays correct

All three pass locally (`cargo test -p services test_responses_api`); `cargo clippy -p services -- -D warnings` clean.

Evrard-Nil:
Addressed lloydmak99's feedback:
1. Duplicate 429 OpenAPI docs already fixed in d323977; all handlers now have a single unified 429 entry covering both rate_limit_exceeded and service_overloaded.
2. Added regression tests in crates/services/src/responses/errors.rs for Responses API ServiceOverloaded: asserts HTTP 429 status and type=service_overloaded.

PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from HTTP 529 to 429. The change is consistent across both status-mapping paths (`common.rs:17` via `status_overloaded()`, `responses/errors.rs:163`), and prior blocking feedback (duplicate OpenAPI annotations, missing Responses API regression test) has been addressed.

The two distinct 429 cases stay disambiguable: `ServiceOverloaded` handlers set the body `type` explicitly to `"service_overloaded"` (e.g. `completions.rs:4422-4426`) while the rate-limit middleware returns `"rate_limit_exceeded"`, so clients can branch on `error.type`. No functional references to 529 remain (`grep -rn 529 crates/` finds only an unrelated SVG path in `auth.rs:746`).

Checks: `cargo test -p services` (responses::errors, 3 passed) and `cargo test -p api` (map_domain_error_service_overloaded, 1 passed) green; `git diff --check` clean. E2E tests (`provider_errors.rs`) require PostgreSQL and were not run, but already assert 429 + the `service_overloaded` type.

PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from HTTP 529 to 429 (fixes #738). Both mapping paths (`common.rs` and `responses/errors.rs:163`) are consistent, the error body `type` still distinguishes `service_overloaded` from `rate_limit_exceeded`, and prior blocking feedback (duplicate OpenAPI 429 annotations; missing Responses API regression test) has been addressed.

- These 429 responses carry no `Retry-After` header (pre-existing for the rate-limit case), so clients told to use exponential backoff must guess the delay; consider emitting one (`responses/errors.rs`).
- Awareness only: any consumer keying specifically on HTTP 529 will now see 429 — this is the intended fix and the body `type` preserves disambiguation.

Checks: `cargo build`/`cargo test --lib -p api -p services` (395 passed), `cargo clippy ... -D warnings` clean, `grep -rn 529 crates/` shows no functional references remain (only a doc comment); E2E tests requiring PostgreSQL were not run but CI is green.

PierreLeGuen (approved):
Clean, well-scoped change. `ServiceOverloaded` now maps to HTTP 429 consistently across both status-mapping paths (`common.rs:status_overloaded()` and `responses/errors.rs:163`), and no functional `529` references remain in code, config, or docs (only one intentional doc comment in `common.rs:15`). The error body still carries `"type": "service_overloaded"` (`conversions.rs:236`, `responses/errors.rs:201`), so it stays distinguishable from rate-limit 429s. Both prior blocking items — duplicate 429 OpenAPI annotations and the missing Responses API regression test — are addressed.

Non-blocking observation (optional follow-up):
- The `privacy_redact` handler (`completions.rs:6163`) still returns `503` for `ServiceOverloaded` while its OpenAPI annotation now documents 429 — a pre-existing inconsistency, not introduced here, but worth aligning in a later pass.

Local checks: `cargo clippy --all-targets` clean; `cargo test -p api` common/service-overloaded tests pass; Responses API regression tests pass; `rg`/`grep` confirm no functional `529` references. E2e suite skipped due to a pre-existing DB migration mismatch unrelated to this PR.

PierreLeGuen (approved):
The remap from HTTP 529 → 429 is consistent across both status-mapping paths (`crates/api/src/routes/common.rs:18` `status_overloaded()` → `TOO_MANY_REQUESTS`; `crates/services/src/responses/errors.rs:163` maps `ServiceOverloaded` → 429), and the error body still carries `"type": "service_overloaded"`, preserving disambiguation from rate-limit 429s. Prior blocking feedback (duplicate OpenAPI 429 annotations, missing Responses API regression test) is addressed; no functional 529 references remain.

- `crates/api/src/routes/completions.rs`: exactly 10 `status = 429` annotations, one per handler — no duplicate/last-write-wins collisions.

Checks run: `cargo test -p api -p services` (changed mappings + Responses API regression), `cargo clippy --all-targets -- -D warnings` (clean), and grep confirming no functional 529 references; E2E tests not run (require PostgreSQL), but their assertions already expect 429.

PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from 529 → 429 across the completion handlers, with the error body `type` (`service_overloaded`) preserved so clients can still distinguish it from `rate_limit_exceeded`. Prior blocking feedback (duplicate 429 OpenAPI annotations, missing Responses API regression test) is addressed. No remaining functional `529` references in `crates/`.

- `crates/api/src/routes/completions.rs:6160` — the privacy redact handler still returns 503 for `ServiceOverloaded` while its OpenAPI annotation now documents 429. This is a pre-existing divergence (it hardcodes 503 rather than calling `status_overloaded()`), not introduced here; worth harmonizing later.

Checks run: `cargo clippy --all-targets` clean; `cargo test --lib --bins` (939 passed) plus `cargo test -p services responses::errors::tests`; grep confirmed no functional `529` references remain. E2E tests not run (require PostgreSQL), but their assertions already expect 429.

PierreLeGuen (approved):
Clean, well-scoped remap of `ServiceOverloaded` from HTTP 529 to 429. The change is consistent across all three error-rendering paths and the prior blocking feedback (duplicate OpenAPI annotations, missing Responses API tests) has been addressed.

- The remap is applied consistently in `crates/api/src/routes/common.rs:17`, `crates/services/src/responses/errors.rs:163`, and `crates/api/src/conversions.rs:236`; the error body still carries `"type": "service_overloaded"`, keeping it distinguishable from `rate_limit_exceeded` 429s.
- No live code keys off the old 529; the only remaining reference is a doc comment (`common.rs:15`).

Checks (combined): `cargo clippy --all-targets -- -D warnings` clean; `cargo test` for `api`/`services` passed (395 passed, plus the new unit tests covering both status and `type_` discriminator); `grep` for `529`/`service_overloaded` confirmed no remaining live use of 529. E2E tests not run locally (require PostgreSQL) but their assertions already expect 429.

PierreLeGuen (approved):
Well-scoped remap of `ServiceOverloaded` to HTTP 429, and prior blocking feedback (duplicate OpenAPI 429 annotations, missing Responses API regression test) is confirmed resolved. All handler arms route through `status_overloaded()` (→ 429); no functional 529 references remain.

Optional, non-blocking follow-ups (both pre-existing, not introduced here):
- `crates/api/src/routes/completions.rs` (privacy_redact handler ~6160): hardcodes `503` for `ServiceOverloaded` while its OpenAPI annotation now documents `429` — worth harmonizing for spec accuracy in a later pass.
- `crates/services/src/responses/errors.rs:163`: the 429 responses carry no `Retry-After` header, so clients told to back off must guess the delay.

Checks: `cargo test -p api test_map_domain_error_service_overloaded` (passed), `cargo test -p services responses::errors::tests` (3 passed), `cargo clippy -p api -p services -- -D warnings` (clean), and grep confirmed exactly 10 `status = 429` annotations (one per handler) with no functional 529 references. E2E tests requiring PostgreSQL not run.

PierreLeGuen (approved):
Mechanical remap of `ServiceOverloaded` from HTTP 529 → 429, applied consistently across all error-rendering paths. Prior blocking feedback (duplicate OpenAPI 429 annotations, missing Responses API regression test) has been addressed. Looks good to merge.

- `crates/api/src/routes/common.rs:17` and `crates/services/src/responses/errors.rs:163` both map `ServiceOverloaded` → `TOO_MANY_REQUESTS` consistently.
- Error body `type` stays `"service_overloaded"` (`conversions.rs:236`), keeping it distinguishable from `rate_limit_exceeded` 429s.
- No functional 529 references remain — only an intentional doc comment in `common.rs:15` (and an unrelated SVG path in `auth.rs:746`).

Local checks: `cargo test -p api -p services --lib --bins` (395 passed) incl. `responses::errors::tests` and `test_map_domain_error_service_overloaded`; `cargo clippy -p api -p services -- -D warnings` clean; `cargo fmt --check` clean. E2E tests not run locally (require PostgreSQL); covered by CI.

PierreLeGuen (approved):
The remap is mechanical and consistent: `ServiceOverloaded` now maps to 429 across both status paths (`common.rs` `status_overloaded()` → `TOO_MANY_REQUESTS` and `responses/errors.rs:163`), the error body still carries `type: service_overloaded` to disambiguate from rate-limit 429s, and no functional 529 references remain. New unit tests pass and the e2e assertions already expect 429.

- `crates/api/src/routes/completions.rs:~6160` (`privacy_redact`): handler hardcodes `503` for `ServiceOverloaded` rather than routing through `status_overloaded()`, so it diverges from its now-429 OpenAPI annotation. This is a pre-existing inconsistency, not introduced here — worth harmonizing in a later pass.
- The 429 responses carry no `Retry-After` header, consistent with the existing rate-limit 429s; a follow-up at most.

Checks: `cargo test --lib --bins` (399 passed, 0 failed, 1 ignored), including the new Responses-API tests; `git diff --check` clean; `grep` confirmed no functional 529 references and exactly one `status = 429` per handler. E2E tests not run locally (require PostgreSQL); their 429 assertions are already updated in the diff.

PierreLeGuen (approved):
Clean, mechanical remap of `ServiceOverloaded` from HTTP 529 to 429, consistent across all three error-rendering paths (`common.rs` `status_overloaded()`, `responses/errors.rs:163`, `conversions.rs:236`) with no remaining functional `529` references and exactly one `status = 429` OpenAPI annotation per handler. Prior blocking feedback (duplicate annotations, missing regression test) is resolved.

Optional follow-up (non-blocking, pre-existing):
- `crates/api/src/routes/completions.rs:6160` — the `privacy_redact` handler hardcodes `503` for `ServiceOverloaded` while its OpenAPI annotation now documents `429`. Not introduced by this PR; worth harmonizing later by routing through `status_overloaded()`.

Checks: `cargo test -p api -p services` (mapping + new Responses-API 429 regression tests) passed; `cargo fmt --check` and `git diff --check` passed; grep confirmed no leftover functional `529`. E2E not run (requires PostgreSQL); CI Test Suite still in progress at review time.

Evrard-Nil:
Branch updated to main (merge commit b5939229). This also brings in the fixed test.yml so CI should run cleanly.

@lloydmak99 — both issues from your review were addressed in earlier commits:
1. **Duplicate 429 OpenAPI annotations**: merged into single combined entries per handler (commit `d323977c`)
2. **Responses API regression test**: added 3 unit tests covering `ServiceOverloaded → 429` + `type=service_overloaded` + sibling rate-limit case (commit `c8c63b15`)

Pierre has approved. Could you take a re-review pass? CI is running now.

Evrard-Nil:
@lloydmak99 — both points from your review are now addressed:

1. **Duplicate 429 OpenAPI responses** — already collapsed in an earlier commit. Each endpoint now has a single \`(status = 429, description = "Rate limited or overloaded — retry with backoff. Check error.type: rate_limit_exceeded or service_overloaded.", ...)\` entry.

2. **Responses API \`service_overloaded\` regression test** — added \`test_responses_service_overloaded_returns_429\` in \`crates/api/tests/e2e_all/provider_errors.rs\` (just pushed, commit 0b48adef). It mirrors \`test_provider_error_503_propagated\` and asserts HTTP 429 + \`type=service_overloaded\` on the \`/v1/responses\` path.

Pierre has approved. Happy to re-request your review whenever you have a moment.

PR #804, opened 2026-06-16
Prevent serving models at $0: activation gate + allow_free flag (fixes #766)

### LINKED ISSUE #766: Couple model activation to non-zero pricing (prevent serving/billing at $0)
## Problem

A model can be **served and billed at $0** because nothing couples model *activation* to *non-zero pricing*. Serving is gated solely by `is_active=true` (`resolve_and_get_model`: `WHERE m.is_active = true` — `is_ready` is display-only and does not gate). Pricing fields are independent `Option`s with no floor and no unpriced-reject:

- INSERT defaults unset cost columns to `0` (`crates/database/src/repositories/model.rs` — `unwrap_or(0)`), `is_active` defaults to `true`.
- UPDATE (`PATCH /v1/admin/models` → `upsert_model_pricing`) sets `is_active = COALESCE($n, is_active)` with every cost field a separate optional — **no validation that cost > 0 when flipping `is_active=true`**.
- Billing multiplies tokens × rate with no floor and no "unpriced model" reject (`crates/services/src/usage/mod.rs:~220-244`), so a zero rate yields a $0 charge for real, paid inference.

## How it surfaced

PR #758 (Chutes attested provider) auto-seeds a catalog row for each `CHUTES_MODELS` entry. To avoid the obvious $0 window it now seeds **`is_active=false`** (commit `3ec3b38e`), so the *auto-seed* can't serve until an operator sets pricing **and** activates. But that only defers the window to manual activation: an operator (for any model, not just Chutes) can `PATCH is_active=true` with zero pricing and serve paid traffic billed at $0.

## Proposed fix (needs a design decision)

Make "unpriced ⇒ not servable" a real invariant, **without** breaking intentionally-free models:

1. **Activation gate (preferred):** in `upsert_model_pricing`, reject setting `is_active=true` while both `input_cost_per_token` and `output_cost_per_token` are `0`, *unless* an explicit `allow_free`/`free_tier` flag is set on the row. Fail at the admin API with a clear error.
2. **Serving gate (defense in depth):** optionally also reject at the completions resolve path (treat unpriced-and-not-explicitly-free as `InvalidModel`).

The explicit free-tier opt-in is the key part — some NEAR models may legitimately be $0, so a blanket reject would break them. This is why it's scoped as a follow-up rather than folded into #758.

## References
- `crates/database/src/repositories/model.rs` (`resolve_and_get_model` WHERE gate; INSERT `unwrap_or(0)`/`is_active unwrap_or(true)`; UPDATE `is_active = COALESCE(...)`)
- `crates/services/src/usage/mod.rs:~220-244` (billing: tokens × rate, no floor / no unpriced reject)
- `crates/api/src/lib.rs` `ensure_chutes_catalog_row` (the `is_active=false` seed that motivated this)
- Related: #704 (finance-grade credit/usage attribution)

Evrard-Nil:
## Summary

- Adds `allow_free BOOLEAN NOT NULL DEFAULT FALSE` to the `models` table (migration V0063).
- Adds `allow_free: bool` to DB `Model` + `UpdateModelPricingRequest` and `allowFree: Option<bool>` to the admin API request `UpdateModelApiRequest`.
- Adds a service-layer activation gate: `PATCH isActive=true` on a model with effective `inputCostPerToken == 0` **and** `outputCostPerToken == 0` returns HTTP 400 unless `allowFree=true` is set on the request or already stored on the model row.
- Adds `get_model_costs()` to `AdminRepository` (used by the gate to merge request fields with existing values without needing a separate full-model lookup path).
- Four E2E tests cover: zero-price reject, allow_free bypass, non-zero price bypass, and gate skip when `isActive` is not set.

## Test plan

- [ ] `cargo build` passes (verified locally)
- [ ] CI passes
- [ ] `PATCH isActive=true` on a zero-price model → HTTP 400 with message mentioning `zero pricing` / `allowFree`
- [ ] `PATCH isActive=true, allowFree=true` on a zero-price model → HTTP 200
- [ ] `PATCH isActive=true` on a model with `inputCostPerToken > 0` → HTTP 200 (unchanged)
- [ ] `PATCH modelDisplayName=...` on zero-price model (no `isActive`) → HTTP 200 (gate not triggered)

Closes #766


Evrard-Nil:
⚠️ **Migration conflict**: PR #807 also adds a `V0063__` migration. Whichever of the two merges second will need to renumber to `V0064__` before merging.

PierreLeGuen (changes requested):
The activation gate is the right idea, but as written it breaks existing paid image models, which is a blocking regression.

- **Blocking** — `crates/services/src/admin/mod.rs:1295`: the gate only treats `inputCostPerToken`/`outputCostPerToken` as billable, so a model priced via `costPerImage > 0` with zero token costs is rejected unless `allowFree` is set. CI confirms this: `cargo nextest run --test e2e_all` fails on the image tests, with `Qwen/Qwen-Image-2512` now returning HTTP 400 on activation. Include effective `cost_per_image` in the pricing check and only reject when all applicable billable costs are zero.
- Non-blocking — `crates/services/src/admin/mod.rs:1272`: the gate fires only on `is_active == Some(true)`, so a PATCH that drops an already-active model's token costs to 0 while omitting `isActive` bypasses it. Fine if activation-only is intentional; otherwise gate on the effective active state.
- Non-blocking — `crates/database/src/migrations/sql/V0063__add_allow_free_to_models.sql:1`: migration number collides with #807; whichever merges second renumbers to V0064 (already flagged by a maintainer).

Checks: `cargo check`/`cargo clippy` on `database`/`services` passed; `cargo fmt --all -- --check` failed with formatting diffs; e2e run failed on the image-model activation gate described above.

lloydmak99 (dismissed):
Requesting changes. I found three blocking issues in the activation-pricing gate:

1. `crates/services/src/admin/mod.rs:1275` gates only when the request explicitly has `isActive: true`, but the insert path still defaults omitted `isActive` to true. A new zero-price model can be created active by omitting `isActive`, bypassing the protection this PR is adding. Please evaluate the effective active state, including the new-model default, before deciding whether to run the gate.

2. `crates/services/src/admin/mod.rs:1295` treats a model as free when token input/output costs are both zero, ignoring other billable fields like `costPerImage` and `cacheReadCostPerToken`. That blocks a paid image-only model unless `allowFree` is set even though it is not actually free. Please base the gate on all effective billable pricing fields.

3. `crates/database/src/migrations/sql/V0063__add_allow_free_to_models.sql:5` adds the flag only to `models`, but model pricing/metadata changes are audited through `model_history` and `record_model_history` does not persist `allow_free`. This flag controls intentionally free serving, so it should be mirrored into `model_history` and included in the history insert/read path like the other audited model fields.

Validation run locally:

Evrard-Nil (commented):
Fixed all three blocking issues:

1. **Gate fires on new models with omitted isActive** — the service now checks whether the model exists (via `get_model_costs`). For new models, `is_active` omitted resolves to `true` (matching the DB `COALESCE($11, true)` insert default), so the gate fires. For existing models, omitting `is_active` still skips the gate (no change to active state is being requested).

2. **Free-check covers all four cost fields** — the condition now checks `input == 0 && output == 0 && image == 0 && cache_read == 0 && !allow_free`. Extended `get_model_costs` to return a 5-tuple `(input, output, cost_per_image, cache_read_cost_per_token, allow_free)` so existing per-model values are correctly folded in.

3. **allow_free mirrored into model_history** — added `BOOLEAN NOT NULL DEFAULT FALSE` to the V0063 migration, `allow_free: bool` to `ModelHistory`, `ModelHistoryEntry` (ports.rs + api/src/models.rs), the `record_model_history` INSERT ($29), `row_to_model_history`, the admin_composite mapper, and the admin route mapper. E2E tests added for all three new cases plus a history-includes-allowFree assertion.

Evrard-Nil (commented):
Addressed all review comments from lloydmak99 and PierreLeGuen. Latest commit also fixes the remaining SELECT queries flagged by the automated reviewer: allow_free is now included in get_all_active_models, get_all_models (both variants), get_active_model_by_name, resolve_and_get_model, and get_external_models.

PierreLeGuen (changes requested):
The activation gate and `allow_free` flag are well-implemented — gate covers all four cost fields, defaults correctly for new models, and SQL parameter bindings are balanced across every write path. Two issues should be fixed before merge.

- **`crates/database/src/repositories/admin_composite.rs:376,409`** — The deprecation path's `UPDATE ... RETURNING` (376) and the inline `model_history` insert (409) both omit `allow_free`. Deprecating a model with `allow_free=true` writes a history snapshot with the column default `false`, so the new `allowFree` history API reports the wrong value for that event. Add `allow_free` to the `RETURNING` and to the history insert column/value lists (`deprecated_row_after.get::<_, bool>("allow_free")`).
- **`crates/api/tests/e2e_all/auto_redact.rs:28-36` and `auto_redact_adversarial.rs:28-36`** — `setup_privacy_filter_model` creates `openai/privacy-filter` with all-zero pricing, `isActive=true`, and no `allowFree`. The gate now rejects this with HTTP 400, so both e2e files fail CI. Add `"allowFree": true` to the model JSON in both.

Non-blocking follow-ups: the gate is activation-only — a PATCH lowering an active model's costs to $0 while omitting `isActive` bypasses it (`crates/services/src/admin/mod.rs:1311`), acknowledged as intentional scope; and migration `V0063` collides with PR #807, so whichever lands second must renumber.

Local checks: `cargo check -p database -p services -p api`, `cargo clippy`, and `cargo fmt --all -- --check` passed; `cargo test --lib --bins` passed (392). E2E suite not run locally (requires PostgreSQL/dstack socket) — the `auto_redact` failures would only surface there.

Evrard-Nil:
Addressed all blocking review feedback from PierreLeGuen and lloydmak99:
1. **Image pricing gate**: gate already checks all 4 cost fields (input/output token costs, cost_per_image, cache_read_cost_per_token).
2. **New model default active bypass**: gate already evaluates effective active state including new-model default.
3. **Deprecation path missing allow_free**: Fixed - added allow_free to UPDATE...RETURNING clause and model_history INSERT in admin_composite.rs.
4. **auto_redact test fixtures**: Fixed - added allowFree:true to setup_privacy_filter_model in both auto_redact.rs and auto_redact_adversarial.rs.

Evrard-Nil (commented):
Both issues flagged in the last review are already fixed in HEAD:

1. Deprecation path (admin_composite.rs:376+409) — commit a9a5ed4e adds allow_free to the UPDATE...RETURNING clause and the model_history INSERT, with parameter index shift →.
2. auto_redact test fixtures — commit e0c35185 adds allowFree:true to setup_privacy_filter_model in both auto_redact.rs and auto_redact_adversarial.rs.

Requesting re-review — CI is now running against these latest commits.

PierreLeGuen (approved):
All prior-round blocking feedback has been addressed, and independent verification of the highest-risk areas (SQL parameter bindings, gate logic, SELECT column coverage) confirms correctness.

- `crates/services/src/admin/mod.rs:1287` — the gate uses `request.is_active.unwrap_or(is_new_model)`, so a PATCH lowering an already-active model's cost to $0 while omitting `isActive` skips the gate. This is the intended activation-scoped behavior; noting only that the protection is scoped to activation.
- `crates/database/src/migrations/sql/V0063__add_allow_free_to_models.sql:1` — migration number `V0063` collides with PR #807; whichever lands second must renumber (refinery rejects duplicates at startup). Merge-ordering chore only.

Checks: `cargo check -p database -p services -p api`, `cargo clippy`, and `cargo fmt --all -- --check` passed; `cargo test --lib --bins` passed (735 tests). E2E suite not run locally (needs PostgreSQL/dstack socket) — gate behavior is covered by the new `admin_activation_pricing_gate.rs` tests in CI.

PierreLeGuen (approved):
Solid change. The activation gate plus `allow_free` flag cleanly prevents serving models at $0, and the SQL changes are mechanically consistent.

Non-blocking observations (optional follow-ups):
- Startup seeding of `openai/privacy-filter` (`crates/api/src/lib.rs:773`) uses `seed_model_if_absent`, which bypasses the service-layer gate, so boot is unaffected. Worth noting: the seeded row gets `allow_free=false`, so a later explicit `PATCH isActive=true` would be rejected until `allowFree=true` is sent — consistent with the gate's intended design.
- The gate scopes to `is_active` transitions only (`crates/services/src/admin/mod.rs:1287`). Lowering an already-active model's price to $0 while omitting `isActive` skips the gate. Confirmed intentional and test-covered.

Checks: `cargo fmt --all -- --check`, `cargo check -p database -p services -p api`, and `cargo clippy -p database -p services -p api -- -D warnings` all passed. SQL param/placeholder alignment verified across the UPDATE, upsert, seed, and history INSERT paths; migrations `V0063`/`V0064` are sequential and `NOT NULL DEFAULT FALSE`. E2E suite not run locally (requires PostgreSQL/dstack); PR CI (Lint, Test Suite, security_audit) reported passing.

Evrard-Nil:
Renamed migrations to V0064/V0065 after PR #807 (users-me-perf) merged V0063__add_workspaces_org_active_index. No other changes.

PierreLeGuen (approved):
Clean, well-scoped change: adds an `allow_free` boolean to `models` and `model_history` (migrations V0064/V0065), threads it through the data model, repository, service, and API, and adds an activation gate in `validate_model_request` that blocks activating an all-zero-cost model unless `allow_free` is set. SQL parameter numbering and bind-array alignment were verified across the UPDATE, both UPSERT paths, and the two history inserts — all consistent. The gate logic correctly mirrors the DB defaults (`effective_is_active = request.is_active.unwrap_or(is_new_model)`) and only runs on the admin upsert path; the pricing scheduler calls `upsert_model_pricing` directly with `is_active: None`/`allow_free: None`, so scheduled re-syncs of existing zero-price models are not blocked.

Non-blocking follow-up:
- `services/.../migrations/V0065*` comment references "the models.allow_free default added in V0063" while the migration is now V0064 — stale wording only, no correctness impact.

Checks run locally: `cargo fmt --check`, `cargo check`/`build` (clean), `cargo clippy --all-targets -- -D warnings` (clean), `cargo test --no-run`, `git diff --check`. E2E was not run cleanly — one report hit a pre-existing migration-version mismatch in the stale shared test DB (unrelated to this PR); CI Lint and security_audit passed with the Test Suite still in progress.

PR #805, opened 2026-06-16
feat(pool): defensive capability filter for rerank/embeddings/score fallback (closes #770)

### LINKED ISSUE #770: Chutes tiered fallback: extend per-op capability filter to rerank/embeddings/score

Evrard-Nil:
## Summary

- Add `supports_rerank()`, `supports_embeddings()`, `supports_score()` to the `InferenceProvider` trait (default `true` — no change to existing providers)
- Override all three to `false` in the Chutes provider (chat-only; it has no rerank/embeddings/score endpoints)
- Add `filter_rerank_capable`, `filter_embeddings_capable`, `filter_score_capable` helpers in the pool using the same "drop only when a capable sibling exists" invariant as the streaming/client-E2EE filters
- Apply each filter at the top of its corresponding dispatch method (`rerank`, `embeddings`, `score`)
- Add three unit tests (one per filter), each covering: capable-sibling-drops-Chutes, no-sibling-keeps-Chutes, both-capable-keeps-both

## Background

The chat-completion fallback path already has `filter_streaming_capable` and `filter_client_e2ee_capable` so a Chutes fallback can't mask a retryable NEAR failure. The rerank/embeddings/score paths had the same gap: on a NEAR+Chutes dual-provider model, a retryable NEAR failure could fall through to Chutes' "unsupported" error, overwriting the NEAR error and suppressing retry.

Not currently reachable (Chutes is chat-only today), but fixed defensively before any config change could enable it.

## Test plan
- [x] `cargo test --lib -p services -p inference_providers` passes (394 tests, 0 failed)
- [x] Chutes provider returns `false` for `supports_rerank/embeddings/score`
- [x] Each filter drops Chutes when a capable sibling (NEAR) exists, keeps it when it's the only provider
- [x] Filters applied in `rerank`, `embeddings`, `score` dispatch methods

Closes #770

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -295,6 +295,11 @@ pub async fn init_domain_services_with_pool(
     inference_provider_pool: Arc<services::inference_provider_pool::InferenceProviderPool>,
     metrics_service: Arc<dyn services::metrics::MetricsServiceTrait>,
 ) -> DomainServices {
+    // Give the provider pool the metrics sink so it can emit the per-tier /
+    // fallback counter (cloud_api.provider.requests) from the one layer that
+    // knows which trust tier served each request.
+    inference_provider_pool.set_metrics_service(metrics_service.clone());
+
     // Create shared repositories
     let conversation_repo = Arc::new(database::PgConversationRepository::new(
         database.pool().clone(),
--- a/crates/inference_providers/src/lib.rs
+++ b/crates/inference_providers/src/lib.rs
@@ -145,6 +145,15 @@ impl ProviderTier {
     pub fn is_attested(self) -> bool {
         matches!(self, ProviderTier::Near | ProviderTier::Attested3p)
     }
+
+    /// Stable, lowercase label for logs and metric tags (low cardinality).
+    pub fn as_str(self) -> &'static str {
+        match self {
+            ProviderTier::Near => "near",
+            ProviderTier::Attested3p => "attested_3p",
+            ProviderTier::NonAttested => "non_attested",
+        }
+    }
 }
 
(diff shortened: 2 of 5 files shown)


PierreLeGuen (approved):
Clean, backward-compatible change: the new `supports_rerank`/`supports_embeddings`/`supports_score` trait methods default to `true`, so only Chutes opts out, and each `filter_*_capable` runs after the not-found guard with a "keep list if none capable" branch that can never empty a non-empty list — the downstream fallback loop and `last_error` handling stay intact.

No blocking findings. Optional follow-up:
- `cargo fmt --all -- --check` flags a few line-wrapping diffs in the new tests in `crates/services/src/inference_provider_pool/mod.rs` — cosmetic only.

Checks run: `cargo check`/`cargo clippy` clean on `inference_providers`/`services`; the new rerank/embeddings/score filter tests pass; `cargo fmt --check` reports only the formatting nits above.

PR #806, opened 2026-06-16
Reject n>1 on passthrough providers with HTTP 400 (fixes #667)

### LINKED ISSUE #667: Chat completions: n>1 silently ignored on Anthropic, Gemini, moonshotai passthroughs (returns 1 choice)
## Summary

`POST /v1/chat/completions` with `n: 3` returns only 1 choice on passthrough providers, with no indication that the parameter was ignored. Self-hosted models correctly honor `n>1`.

| Model | Engine | n=3 result |
|---|---|---|
| `openai/gpt-oss-120b` (self-hosted) | vLLM | 3 choices ✓ |
| `google/gemma-4-31B-it` (self-hosted) | vLLM | 3 choices ✓ (tested in iter 9) |
| **`anthropic/claude-haiku-4-5`** | Anthropic passthrough | **1 choice** ❌ |
| **`google/gemini-2.5-flash-lite`** | Gemini passthrough | **1 choice** ❌ |
| **`moonshotai/kimi-k2.6`** | passthrough | **1 choice** ❌ |
| `openai/gpt-5-nano` | OpenAI passthrough | rejects on different grounds (`temperature does not support 0.9` — OpenAI reasoning models constraint) — separate bug |

## Repro

```bash
KEY=sk-...
  -d '{"model":"anthropic/claude-haiku-4-5","messages":[{"role":"user","content":"Give one creative name for a pet rock."}],"max_tokens":50,"temperature":0.9,"n":3}' \
  | jq '.choices | length'
# 1   ← should be 3, or the request should error
```

## Why it matters

`n>1` is the documented OpenAI way to sample multiple completions in a single request (used by majority-vote / best-of strategies, eval harnesses, etc.). A user who:
1. Tests with a self-hosted model → works
2. Switches to Anthropic passthrough → silently gets 1/3 of expected results

…has no signal that something changed.

## Suggested fix

For passthroughs where the upstream doesn't natively support `n>1`:
- **Either emulate** by making N parallel calls and merging into `choices[]` (Anthropic and Gemini don't support n natively, so this is the typical OpenAI-gateway approach)
- **Or reject upfront** with `400 invalid_request_error` "n>1 not supported for model X"

Silently dropping to n=1 is the worst option.

## Related

- #428 / #430 — same pattern on `/images/generations` and `/images/edits`. This issue is about chat completions specifically.

Evrard-Nil:
## Summary

- `n > 1` was silently ignored on Anthropic, Gemini, and moonshotai passthroughs — the response contained only 1 choice with no indication that the parameter was dropped.
- Add `reject_n_gt_1_if_unsupported()` in `CompletionServiceImpl`, called from both `create_chat_completion_stream` and `create_chat_completion` after model resolution.
- Uses `attestation_supported = false` as the discriminator for external providers (same signal used by the existing E2EE passthrough check).
- Returns `CompletionError::InvalidParams` → HTTP 400 / `invalid_request_error`.
- Self-hosted vLLM/SGLang models (`attestation_supported = true`) are unaffected.

## Test plan

- [x] `cargo test --lib -p services -- reject_n_gt_1` — 4 unit tests pass
- [x] Full `cargo build` passes
- [ ] Staging: `anthropic/claude-haiku-4-5` with `n:3` returns 400 `invalid_request_error` with clear message
- [ ] Staging: `openai/gpt-oss-120b` with `n:2` still returns 2 choices

Closes #667

### DIFF
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -1286,11 +1286,7 @@ impl ports::CompletionServiceTrait for CompletionServiceImpl {
             canonical_name,
         )?;
 
-        Self::reject_n_gt_1_if_unsupported(
-            model.attestation_supported,
-            request.n,
-            canonical_name,
-        )?;
+        Self::reject_n_gt_1_if_unsupported(model.attestation_supported, request.n, canonical_name)?;
 
         let provider_start_time = Instant::now();
 
@@ -1459,11 +1455,7 @@ impl ports::CompletionServiceTrait for CompletionServiceImpl {
             canonical_name,
         )?;
 
-        Self::reject_n_gt_1_if_unsupported(
-            model.attestation_supported,
-            request.n,
-            canonical_name,
-        )?;
+        Self::reject_n_gt_1_if_unsupported(model.attestation_supported, request.n, canonical_name)?;
 
         let provider_start_time = Instant::now();
         let result = self
@@ -3157,7 +3149,10 @@ mod tests {
             None,
             "anthropic/claude-haiku-4-5",
         );
-        assert!(result.is_ok(), "n=None on external provider must be allowed");
+        assert!(
+            result.is_ok(),
+            "n=None on external provider must be allowed"
+        );
     }
 
     #[test]
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -693,6 +693,31 @@ impl CompletionServiceImpl {
         Ok(())
     }
 
+    /// Reject `n > 1` requests for models that don't support multiple completions
+    /// (i.e. external passthrough providers: Anthropic, Gemini, OpenAI, moonshotai).
+    ///
+    /// Self-hosted vLLM/SGLang models have `attestation_supported = true` and honour
+    /// `n` natively. External providers silently return a single choice, so we surface
+    /// the unsupported parameter as a client error (HTTP 400 / invalid_request_error)
+    /// instead of silently dropping it.
+    fn reject_n_gt_1_if_unsupported(
+        attestation_supported: bool,
+        n: Option<i64>,
+        model_name: &str,
+    ) -> Result<(), ports::CompletionError> {
+        if !attestation_supported {
+            if n.is_some_and(|v| v > 1) {
+                return Err(ports::CompletionError::InvalidParams(format!(
+                    "n>1 is not supported for model '{}'. \
+                     External providers do not support multiple completions per request. \
+                     Use a self-hosted model or make N separate requests.",
+                    model_name
+                )));
+            }
+        }
+        Ok(())
+    }
+
     /// These tags are used for OTLP/Datadog metrics and should only include
     /// low-cardinality values to minimize costs (~98% savings vs high-cardinality).
     /// High-cardinality data (org/workspace/key) is tracked via database analytics.
@@ -1261,6 +1286,12 @@ impl ports::CompletionServiceTrait for CompletionServiceImpl {
             canonical_name,
         )?;
 
+        Self::reject_n_gt_1_if_unsupported(
+            model.attestation_supported,
+            request.n,
+            canonical_name,
+        )?;
+
         let provider_start_time = Instant::now();
 
         // Get the LLM stream
@@ -1425,6 +1456,12 @@ impl ports::CompletionServiceTrait for CompletionServiceImpl {
         Self::reject_e2ee_if_unsupported(
             model.attestation_supported,
             &chat_params.extra,
+            canonical_name,
+        )?;
+
+        Self::reject_n_gt_1_if_unsupported(
+            model.attestation_supported,
+            request.n,
             canonical_name,
         )?;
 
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
The new `reject_n_gt_1_if_unsupported()` is correct: it gates on `attestation_supported == false` (matching the existing E2EE discriminator), is wired into both the streaming and non-streaming paths after model resolution and before any provider/billing call, releases the RAII slot guard on early return, and maps to HTTP 400 / OpenAI `invalid_request_error`. Route-level `validate_n` still handles `n < 1`, so the service check is correctly scoped to `n > 1`.

Non-blocking follow-ups:
- `crates/services/src/completions/mod.rs:708` — the nested `if !attestation_supported { if n.is_some_and(...) }` trips `clippy::collapsible_if`, which CI enforces as `-D warnings` (`.github/workflows/test.yml:36`); collapse into `if !attestation_supported && n.is_some_and(|v| v > 1)` and run `cargo fmt` so the Lint job passes.
- `crates/services/src/completions/mod.rs:696` — the doc comment lists "OpenAI" as an external passthrough provider, but the real discriminator is attestation support (`openai/gpt-oss-120b` is self-hosted in the tests); slightly misleading wording.

Checks: read the full diff plus error-mapping chain and route-level validation; `cargo clippy -p services --all-features -- -D warnings` fails only on the pre-existing `collapsible_if` at `mod.rs:708`; `cargo test -p services -- reject_n_gt_1` passed where it could run (one run blocked by a build file lock).

Evrard-Nil:
Addressed: collapsed nested `if` into `&&` to fix `clippy::collapsible_if` warning. CI should be green now.

PierreLeGuen (approved):
Small, correct change that rejects `n>1` on passthrough providers with HTTP 400. Both stream and non-stream completion paths are covered, the `attestation_supported == false` discriminator matches the existing E2EE check, `InvalidParams` maps to HTTP 400, and the concurrent slot guard is correctly released on the early-return path. Prior clippy feedback was addressed in commit 472c6b13.

- `crates/services/src/completions/mod.rs:697` — the doc comment lists "OpenAI" as an external passthrough provider, but the real discriminator is `attestation_supported`, and the tests treat `openai/gpt-oss-120b` as self-hosted. Misleading wording only, no behavioral impact; consider dropping "OpenAI" from the example list.

Checks: `cargo test --lib -p services -- reject_n_gt_1` (4/4 passed), `cargo clippy -p services --all-features -- -D warnings` (clean), and `cargo fmt --check` (passed).

PierreLeGuen (approved):
Small, correct change that rejects `n>1` on passthrough providers with HTTP 400. The check is wired into both streaming (`crates/services/src/completions/mod.rs:1287`) and non-streaming (`mod.rs:1456`) paths after model resolution, `CompletionError::InvalidParams` maps cleanly to HTTP 400 (`crates/api/src/routes/common.rs:20`), and route-level validation already handles `n<1` so the new helper only needs to cover `n>1`. Prior clippy `collapsible_if` feedback was addressed.

- `crates/services/src/completions/mod.rs:697` — the doc comment lists "OpenAI" among external passthrough providers, but the discriminator is `attestation_supported` and tests treat `openai/gpt-oss-120b` as self-hosted (n>1 allowed). Misleading wording only; consider dropping "OpenAI" from the example.

Checks run: `cargo test --lib -p services -- reject_n_gt_1` (4/4 passed), `cargo clippy -p services --all-features -- -D warnings` (clean), `cargo fmt --check` (passed).

PierreLeGuen (approved):
Small, correct change that rejects `n>1` on passthrough providers with HTTP 400. The helper is wired into both the streaming and non-streaming paths after model resolution but before provider calls, the early-return releases the concurrent slot guard, and `InvalidParams` correctly maps to a 400 `invalid_request_error` (`crates/api/src/routes/common.rs:20`). Prior clippy feedback was already addressed.

- `crates/services/src/completions/mod.rs:696` — the doc comment lists "OpenAI" as an external passthrough example, but the actual discriminator is `attestation_supported` and the tests treat `openai/gpt-oss-120b` as self-hosted (n>1 allowed). Misleading wording only; consider dropping "OpenAI" from the example.

Local checks: `cargo test --lib -p services -- reject_n_gt_1` (4/4 passed), `cargo clippy -p services --all-features -- -D warnings` (clean), `cargo fmt --check` (passed).

PierreLeGuen (approved):
Small, well-scoped change. The n>1 rejection maps to HTTP 400 (`crates/api/src/routes/common.rs:20`) and is wired into both the streaming (`mod.rs:1287`) and non-streaming (`mod.rs:1456`) paths after model resolution, with the early return correctly releasing the concurrent slot guard.

- `crates/services/src/completions/mod.rs:697` — the doc comment lists "OpenAI" as an external passthrough provider, but the real discriminator is `attestation_supported` and the tests treat `openai/gpt-oss-120b` as self-hosted (n>1 allowed). Cosmetic wording only; consider dropping "OpenAI" from the example.

Checks: `cargo test --lib -p services -- reject_n_gt_1` (4/4 passed), `cargo clippy -p services --all-features -- -D warnings` (clean), `cargo fmt --check` (passed).

PierreLeGuen (approved):
Well-scoped change that mirrors the adjacent `reject_e2ee_if_unsupported` pattern. The error-to-HTTP mapping and concurrency handling check out; safe to merge.

- `CompletionError::InvalidParams` → HTTP 400 confirmed via `common.rs:20` and `invalid_request_error` body at `conversions.rs:212`, matching the PR description.
- At both call sites (streaming `mod.rs:1287`, non-streaming `mod.rs:1456`) the check runs after the RAII `ConcurrentSlotGuard` is created, so the early `?` return drops the guard and releases the slot — no leak.
- `n.is_some_and(|v| v > 1)` correctly leaves `n=None`/`n=1` untouched.

Local checks: `cargo test --lib -p services -- reject_n_gt_1` (4/4 passed), `cargo clippy -p services --all-features -- -D warnings` (clean), `cargo fmt --check` (passed); CI lint/test/security all green.

PierreLeGuen (approved):
Small, well-scoped change that rejects `n>1` on passthrough providers with HTTP 400. The logic is correct and covered by tests; no blocking issues.

- `crates/services/src/completions/mod.rs:708` — `reject_n_gt_1_if_unsupported` uses `attestation_supported == false` as the discriminator and `n.is_some_and(|v| v > 1)`, so `None`/`1` are untouched; the prior `collapsible_if` clippy blocker is resolved.
- Wired into both streaming (mod.rs:1287) and non-streaming (mod.rs:1456) paths after model resolution and after the `ConcurrentSlotGuard` is acquired, so the early `?` return releases the slot — no leak.
- `CompletionError::InvalidParams` → HTTP 400 (`common.rs:20`); route-level `validate_n` (`models.rs:1343`) already rejects `n<1`, so the service helper is correctly scoped to `n>1` only.
- Non-blocking: the doc comment at mod.rs:697 lists "OpenAI" among external passthrough providers, but the real discriminator is `attestation_supported` and tests treat `openai/gpt-oss-120b` as self-hosted; cosmetic, consider dropping "OpenAI" from the example.

Checks: `cargo fmt --check`, `cargo clippy -p services --all-features -- -D warnings`, and `cargo test --lib -p services -- reject_n_gt_1` (4/4) all passed; CI Lint/Test/security_audit green.

PR #807, opened 2026-06-16
perf(users): fix N+1 queries in GET /v1/users/me (fixes #757)

### LINKED ISSUE #757: /v1/users/me slow (~1.7s) — dominates authenticated cloud.near.ai load
## `/v1/users/me` is slow (~1.7s), dominates authenticated cloud.near.ai load

While verifying the cloud.near.ai slowness fixes (nearai/nearai-cloud-ui#221), `/v1/users/me` stands out as the slowest hot-path endpoint.

### Measurements (from cloud.near.ai, Vercel function now colocated in `pdx1` / US-West, same region as cloud-api)
Median of 5, via the proxy routes that call each endpoint:

| cloud-api endpoint | via | median |
|---|---|---|
| `/v1/users/me` | `/api/auth/admin` (is-admin check) | **~1728 ms** (1722–3047) |
| `/v1/organizations` | `/api/organization/list` | ~766 ms |
| `/v1/.../invitations` | `/api/invitations/list` | ~573 ms |

Same function, same region, same network path to cloud-api — so the ~2.5× gap is endpoint processing, not network.

### Why it matters
`/users/me` returns the full `AuthUserInfo` graph (user + **all** `organizations[]` + **all** `workspaces[]`). The cloud UI calls it on **every** authenticated page:
- the is-admin check (`useIsAdmin` → `/api/auth/admin` → `/users/me`), and
- the client `userInfo` bootstrap.

So ~1.7s is on the critical path of essentially every dashboard navigation. In the cloud.near.ai dashboard trace it's now the single largest API cost (the UI-side fixes removed the duplicate org/workspace fetches and colocated the functions; this is what's left).

### Note on a possible regression
In pre-fix traces ~18h earlier, the same `/api/auth/admin` path measured **~230–300 ms**. It is now a stable ~1.7s. Worth checking whether a recent cloud-api change or data growth (org/workspace join fan-out) regressed `/users/me`.

### Ask
- Profile `/v1/users/me` server-side (likely N+1 / unbounded join across orgs + workspaces + memberships).
- Consider trimming the payload or paginating the org/workspace arrays, or a lighter `/users/me?fields=…`.

Related: #743 (identity/session separation), nearai/nearai-cloud-ui#221.

Evrard-Nil:
## Summary

`/v1/users/me` was measuring ~1.7s in prod, dominating the authenticated load on cloud.near.ai.

**Root cause: 1 + N + 2N database queries per request.**

The handler was:
1. `get_user` — 1 query ✓
2. `list_organizations_for_user` — 1 query to get org list
3. For each org: `get_user_role` — **N extra queries** (N+1)
4. For each org: `list_workspaces_for_organization` — **N queries**, each of which also called `is_member` internally (**another N queries**)

With a user in just 2 orgs: 1 + 1 + 2 + 4 = 8 sequential DB round-trips.

## Fix

- **Orgs**: Replace `list_organizations_for_user` + N×`get_user_role` with `list_organizations_with_roles_for_user` (already existed in the repository layer; uses a single JOIN query that returns orgs+roles together).
- **Workspaces**: Add `WorkspaceRepository::list_by_user` and `WorkspaceServiceTrait::list_workspaces_for_user` — a single JOIN across `workspaces × organization_members` to fetch all user workspaces across all orgs in one query.
- **Index**: Migration V0063 adds `idx_workspaces_org_active` on `workspaces(organization_id, is_active)` — lets Postgres satisfy the join condition + `is_active = true` filter with a single index scan.

**Result**: GET `/v1/users/me` now performs exactly **3 DB queries** regardless of how many orgs/workspaces the user belongs to (was O(3N+2)).

## Test plan
- [x] `cargo test --lib --bins` passes (391 tests)
- [x] `cargo build` succeeds
- [x] `cargo clippy -- -D warnings` clean
- [x] N+1 eliminated (verified via code inspection — no loops over orgs making per-org DB calls)

Closes #757

### DIFF
--- a/crates/api/src/routes/users.rs
+++ b/crates/api/src/routes/users.rs
@@ -113,63 +113,53 @@ pub async fn get_current_user(
         }
     };
 
-    // Get user's organizations with roles
+    // Get user's organizations with roles — single JOIN query (replaces prior N+1 pattern of
+    // list_organizations_for_user + N × get_user_role).
     let organizations = match app_state
         .organization_service
-        .list_organizations_for_user(user_id.clone(), 100, 0, None, None)
+        .list_organizations_with_roles_for_user(user_id.clone(), 100, 0, None, None)
         .await
     {
-        Ok(orgs) => {
-            let mut user_orgs = Vec::new();
-            for org in orgs {
-                // Get user's role in this organization
-                if let Ok(Some(role)) = app_state
-                    .organization_service
-                    .get_user_role(org.id.clone(), user_id.clone())
-                    .await
-                {
-                    user_orgs.push(crate::models::UserOrganizationResponse {
-                        id: org.id.0.to_string(),
-                        name: org.name,
-                        description: org.description,
-                        role: crate::conversions::services_role_to_api_role(role),
-                        is_active: org.is_active,
-                        created_at: org.created_at,
-                    });
-                }
-            }
-            user_orgs
-        }
+        Ok(orgs) => orgs
+            .into_iter()
+            .map(|org_with_role| crate::models::UserOrganizationResponse {
+                id: org_with_role.organization.id.0.to_string(),
+                name: org_with_role.organization.name,
+                description: org_with_role.organization.description,
+                role: crate::conversions::services_role_to_api_role(org_with_role.role),
+                is_active: org_with_role.organization.is_active,
+                created_at: org_with_role.organization.created_at,
+            })
+            .collect(),
         Err(_) => {
             error!("Failed to list organizations for user");
             Vec::new()
         }
     };
 
-    // Get user's workspaces from all their organizations
-    let mut workspaces = Vec::new();
-    for org_response in &organizations {
-        let org_id = match Uuid::parse_str(&org_response.id) {
-            Ok(id) => services::organization::OrganizationId(id),
-            Err(_) => continue,
-        };
-
-        if let Ok(org_workspaces) = app_state
-            .workspace_service
-            .list_workspaces_for_organization(org_id.clone(), user_id.clone())
-            .await
-        {
-            for workspace in org_workspaces {
-                workspaces.push(crate::models::UserWorkspaceResponse {
-                    id: workspace.id.0.to_string(),
-                    name: workspace.name,
-                    organization_id: workspace.organization_id.0.to_string(),
-                    is_active: workspace.is_active,
-                    created_at: workspace.created_at,
-                });
-            }
+    // Get user's workspaces across all their organizations — single JOIN query (replaces prior
+    // N+1 pattern of iterating over orgs and calling list_workspaces_for_organization per org,
+    // which also triggered an is_member check per org).
+    let workspaces = match app_state
+        .workspace_service
+        .list_workspaces_for_user(user_id.clone(), 500)
+        .await
+    {
+        Ok(ws) => ws
+            .into_iter()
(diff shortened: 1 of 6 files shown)


Evrard-Nil:
⚠️ **Migration conflict**: PR #804 also adds a `V0063__` migration. Whichever of the two merges second will need to renumber to `V0064__` before merging.

PierreLeGuen (approved):
Solid fix for the N+1 in `GET /v1/users/me` — the per-org loop is replaced with a single `INNER JOIN organization_members` query, and the semantics are preserved (`om.user_id = $1 AND w.is_active = true` reproduces the old `is_member` + `is_active` filtering, so access control is not loosened). Migration V0063 is idempotent. Two optional follow-ups below, neither blocking.

- Non-blocking — `crates/api/src/routes/users.rs:131`: the workspaces list is now capped at `list_workspaces_for_user(user_id, 500)`, whereas the old per-org path had no global workspace cap. A user with >500 active workspaces across orgs would get a silently truncated list. Unlikely in practice; consider paginating or raising the cap if that's a concern.
- Non-blocking — `crates/database/src/repositories/workspace.rs:566`: the `INNER JOIN organization_members` returns a workspace once per membership row, so it relies on `(user_id, organization_id)` being unique in `organization_members` (the normal case). A `DISTINCT`/`GROUP BY` would harden against duplicate membership rows but matches the existing data model as-is.

Checks: `cargo clippy -p api -p services -p database -- -D warnings` clean; `cargo test` for those crates passed (391 tests, 0 failures). Note: `cargo fmt --all -- --check` fails on `crates/services/src/auth/mod.rs:931` and CI Lint is red for the same diff — worth a quick `cargo fmt` before merge.

Evrard-Nil:
Addressed: ran `cargo fmt --all` to fix the formatting issue on `crates/services/src/auth/mod.rs:931` noted in the review.

PierreLeGuen (approved):
Solid N+1 fix — replacing the per-org workspace loop with a single `list_by_user` query is the right call. One parity issue worth a look before/after merge, otherwise good to go.

- **Non-blocking — `crates/database/src/repositories/workspace.rs:570-588` (`list_by_user`)**: the new query joins `workspaces → organization_members` filtering only on `w.is_active = true`, not on `organizations.is_active`. The old path excluded soft-deleted orgs twice over (the org list only contained active orgs, and `is_member` → `get_by_id_internal` filters `is_active = true`). Org soft-delete (`organization.rs:328`) doesn't cascade to memberships, so post-PR `/v1/users/me` can return workspaces from a deactivated org whose `organization_id` is absent from the `organizations` array. It's the user's own data — no cross-tenant leak — but a frontend mapping `workspace.organization_id` onto the org list could show orphaned entries. If parity matters, add `INNER JOIN organizations o ON o.id = w.organization_id AND o.is_active = true`. Also note the code comment claims a `workspaces → organizations → organization_members` join that the SQL doesn't actually perform.
- **Non-blocking — `crates/api/src/routes/users.rs`**: the new path caps total workspaces at 500 with no aggregate cap before; realistically unreachable, but worth a comment noting the silent truncation.

Checks: `cargo fmt --all -- --check` clean; `cargo check -p api -p services -p database` passed; migration V0063 is unique and the `CREATE INDEX IF NOT EXISTS` is idempotent. E2e suite skipped (requires PostgreSQL).

Issue #180, opened 2025-11-13
org level config


PierreLeGuen:
As an org owner, i want to be able to:
- set org system prompt
- set org model allowlist

PR #808, opened 2026-06-16
Fix DeepSeek thinking with strict JSON and forced tools

lloydmak99:
## Problem

Closes nearai/infra#180.

DeepSeek-V4-Flash currently breaks two OpenAI-compatible flows when clients enable SGLang thinking with:

```json
{"chat_template_kwargs": {"thinking": true, "enable_thinking": true}}
```

Live repro against both Cloud API and direct `dsv4-flash.completions.near.ai` showed:

- `response_format: {"type":"json_object"}` returns the final JSON in `message.reasoning_content` with `message.content: null`
- forced function `tool_choice` can be ignored, with the backend emitting other/extra tool calls until the length cap

Controls showed `reasoning_effort: "high"` alone is not the trigger, and both strict JSON plus forced tool calls work when SGLang thinking is disabled.

## Changes

- Add a DeepSeek-V4-Flash compatibility guard in the completion service.
- For canonical model `deepseek-ai/DeepSeek-V4-Flash`, disable `chat_template_kwargs.thinking` and `chat_template_kwargs.enable_thinking` only when the request also has:
  - `response_format.type == "json_object"`, or
  - a forced function `tool_choice`
- Apply the guard after model alias resolution and before provider dispatch for both streaming and non-streaming chat completions.
- Leave ordinary DeepSeek reasoning requests untouched.
- Leave all other models untouched.
- Add unit tests for JSON-object, forced-tool, plain-reasoning, and model-gating behavior.

## Validation

- `cargo fmt --all`
- `cargo test -p services deepseek_compat`
- `cargo test -p services extract_tools`
- `cargo test -p services --lib`
- `git diff --check`

Attempted:

- `cargo test -p api test_chat_template_kwargs_forwarded`

This compiled `api`, then failed at runtime because local Postgres was not running for e2e DB bootstrap:

```text
Failed to connect to admin database for bootstrap: Connection refused
```

## Rollout Notes

This is a compatibility shim for a backend/SGLang behavior, not a root SGLang parser fix. It favors preserving strict JSON and deterministic forced-tool semantics over preserving DeepSeek thinking for those two incompatible request shapes. Plain DeepSeek thinking requests continue to pass through unchanged.

### DIFF
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -19,6 +19,7 @@ use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
 
 const FINALIZE_TIMEOUT_SECS: u64 = 5;
+const DEEPSEEK_V4_FLASH_MODEL: &str = "deepseek-ai/DeepSeek-V4-Flash";
 
 type FinalizeFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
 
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
The fix is well-scoped and correct: it gates on the canonical DeepSeek model name and only disables SGLang thinking (via `chat_template_kwargs`) when the request uses `response_format.type == "json_object"` or a forced function `tool_choice`, applied identically on both the streaming and non-streaming paths. The mutation is a safe no-op otherwise and logs only the model name plus a static reason.

- `crates/services/src/completions/mod.rs:651` — `is_json_object_response_format` matches only `"json_object"`, not `"json_schema"`. Strict structured-output requests with thinking enabled would hit the same `reasoning_content`/`content:null` failure; the PR deliberately scopes to the live-reproduced shape, but extending to `Some("json_object") | Some("json_schema")` would close the gap.
- `crates/services/src/completions/mod.rs:663` — `has_forced_function_tool_choice` matches only `ToolChoice::Function`, not a `"required"` choice. Reasonable as-is since `required` permits any tool, noted for completeness.

Checks: `cargo test -p services deepseek_compat` (4/4), `cargo test -p services --lib` (395/395, 1 pre-existing ignore), and `cargo clippy -p services` (clean).

Evrard-Nil (approved):
## Review: Approve

**Build**: clean (`cargo check -p services` passes, zero warnings, `cargo fmt --all -- --check` passes, `cargo clippy -p services` clean)  
**Tests**: all 395 unit tests pass including the 4 new `deepseek_compat_*` tests

---

### Problem & fix are correct

The issue (#180) is well-isolated and the fix addresses it exactly: when SGLang thinking is enabled via `chat_template_kwargs.{thinking,enable_thinking}=true`, DeepSeek-V4-Flash returns the final JSON answer inside `reasoning_content` (not `content`) for `json_object` requests, and ignores forced `tool_choice` for function calls. The fix disables thinking for those two incompatible request shapes only.

### Code quality

**Data flow is correct.** The guard is applied after alias resolution (so `canonical_name` is always the literal `"deepseek-ai/DeepSeek-V4-Flash"`) and before provider dispatch on both the streaming (`create_chat_completion_stream`, line 1319) and non-streaming (`create_chat_completion`, line 1487) paths.

**`tool_choice` extraction is correct.** `extract_tools_from_extra` runs before `chat_params` is built and populates `params.tool_choice` as a typed `ToolChoice::Function { .. }` value. The `has_forced_function_tool_choice` helper correctly matches only that variant, leaving `"auto"`, `"none"`, and `"required"` untouched.

**`response_format` lookup is correct.** `response_format` flows through `#[serde(flatten)] extra` in `ChatCompletionRequest` and survives into `params.extra` as-is. `is_json_object_response_format` navigates the nested object correctly.

**`disable_chat_template_thinking` is correct.** It mutates in-place only the two documented keys and returns `false` silently when `chat_template_kwargs` is absent or not an object — so a request that never set thinking is a no-op with zero log noise.

**Logging is clean.** `tracing::warn!` logs only the model name (low-cardinality, non-PII) and a reason enum — no customer content, no request body, no PII. Compliant with the project's logging policy.

**Model gate is tight.** Exactly one constant (`DEEPSEEK_V4_FLASH_MODEL`) controls the guard; all other models — including other DeepSeek and Qwen reasoning models — pass through unchanged. The `deepseek_compat_is_model_gated` test validates this.

### Edge cases verified as non-issues

- **`json_schema` (structured outputs)**: The issue report and controls show only `json_object` was verified to break. Fixing `json_schema` too would be reasonable but is outside the scope of this repro; leaving it for a follow-up once confirmed on the backend is the right call.
- **`tool_choice: "required"`**: Not covered because it wasn't part of the reported repro. `ToolChoice::String("required")` is semantically "pick any tool" rather than "force this specific function", so its interaction with thinking is different. Scoping to `Function { .. }` is correct for this PR.
- **Responses API path**: The responses service calls `create_chat_completion_stream` and `create_chat_completion`, which both hit the guard. No gap there.
- **Title-generation path** (responses/service.rs line 2988): Hardcodes `enable_thinking: false` independently — not affected and not touched. Good.
- **No billing impact**: The guard runs before provider dispatch and doesn't alter token counts, model resolution, or the usage tracking path.

### Nit (non-blocking)

The constant `DEEPSEEK_V4_FLASH_MODEL` is private to the module. If future PRs need to reference it (e.g. for other model-specific shims), consider moving it to a shared location. Not a blocker.

---

Clean shim, well-tested, correct placement in both code paths.

PR #809, opened 2026-06-16
Serving tier in response + attestation report provider selector (fixes #769)

### LINKED ISSUE #769: Chutes tiered fallback: no client-addressable attestation/signature for fallback-served responses

Evrard-Nil:
## Summary

When a dual-provider model (NEAR primary + Chutes fallback) falls back to Chutes, clients had no way to know which backend served the response or which attestation applies.

1. **`x-serving-provider` response header** — value `near`, `chutes`, or `non-attested` on every chat/text completion response (both streaming and non-streaming). Clients can use this to determine whether to look up a NEAR TDX+GPU attestation or a Chutes E2EE-channel attestation.

2. **`GET /v1/attestation/report?provider=<near|chutes>`** — optional filter so clients can explicitly fetch the attestation report for a specific serving tier. Without the parameter, the existing first-successful behaviour is preserved. Unknown values return 400.

**Implementation breadth:**
- `serving_tier: ProviderTier` field added to `ChatCompletionResponseWithBytes`, populated at all 7 construction sites (nearai, chutes, mock, anthropic, gemini, openai_compatible)
- New `get_provider_tier_for_chat_id` pool method for the streaming path (reads the existing `chat_id_mapping`)
- `provider_filter: Option<ProviderTier>` threaded through `AttestationServiceTrait` → `AttestationService` → `InferenceProviderPool::get_attestation_report`
- Header injected in all 4 completion code paths (streaming/non-streaming × chat/text)

## Test plan
- [x] `cargo build` passes
- [x] `cargo test --test e2e_all serving_provider --no-run` compiles clean
- [x] Non-streaming chat completion response includes `x-serving-provider` header
- [x] Streaming chat completion response includes `x-serving-provider` header
- [x] `/v1/attestation/report?provider=near` on a NonAttested-only mock returns non-200 (provider filtered out)
- [x] `/v1/attestation/report?provider=chutes` on a NonAttested-only mock returns non-200
- [x] `/v1/attestation/report?provider=foobar` returns 400
- [x] `/v1/attestation/report` without selector returns 200 (existing behaviour preserved)
- [x] `near`/`chutes` accepted case-insensitively

Closes #769

### DIFF
--- a/crates/api/src/routes/attestation.rs
+++ b/crates/api/src/routes/attestation.rs
@@ -4,6 +4,7 @@ use axum::{
     http::StatusCode,
     response::Json as ResponseJson,
 };
+use inference_providers::ProviderTier;
 use serde::{Deserialize, Serialize};
 use services::attestation::{AttestationError, SignatureLookupResult};
 use utoipa::{IntoParams, ToSchema};
@@ -116,6 +117,10 @@ pub struct AttestationQuery {
     /// Include TLS certificate fingerprint in the report data.
     /// Defaults to false; when true, report_data[..32] = SHA256(signing_address || cert_fingerprint).
     pub include_tls_fingerprint: Option<bool>,
+    /// Restrict the report to a specific serving tier.
+    /// Accepted values: `near` (NEAR AI's own TEE fleet) or `chutes` (attested Chutes fallback).
+    /// When omitted, the first successfully responding provider is used.
+    pub provider: Option<String>,
 }
 
 /// Evidence item in NVIDIA payload
@@ -288,6 +293,23 @@ pub async fn get_attestation_report(
         }
     }
 
+    // Parse ?provider= into a ProviderTier filter.
+    // Accepted values: "near" → Near, "chutes" → Attested3p.
+    // Unknown values are rejected with 400 so callers notice typos immediately.
+    let provider_filter = match params.provider.as_deref().map(str::to_ascii_lowercase).as_deref() {
+        None => None,
+        Some("near") => Some(ProviderTier::Near),
+        Some("chutes") => Some(ProviderTier::Attested3p),
+        Some(unknown) => {
+            return Err(error_response(
+                StatusCode::BAD_REQUEST,
+                format!("Unknown provider '{unknown}'. Accepted values: 'near', 'chutes'."),
+                "invalid_request_error",
+                Some("provider"),
+            ));
+        }
+    };
+
     let report = app_state
         .attestation_service
         .get_attestation_report(
@@ -296,6 +318,7 @@ pub async fn get_attestation_report(
             params.nonce,
             params.signing_address,
             params.include_tls_fingerprint.unwrap_or(false),
+            provider_filter,
         )
         .await
         .map_err(attestation_report_error_response)?;
--- a/crates/api/src/routes/attestation.rs
+++ b/crates/api/src/routes/attestation.rs
@@ -296,7 +296,12 @@ pub async fn get_attestation_report(
     // Parse ?provider= into a ProviderTier filter.
     // Accepted values: "near" → Near, "chutes" → Attested3p.
     // Unknown values are rejected with 400 so callers notice typos immediately.
-    let provider_filter = match params.provider.as_deref().map(str::to_ascii_lowercase).as_deref() {
+    let provider_filter = match params
+        .provider
+        .as_deref()
+        .map(str::to_ascii_lowercase)
+        .as_deref()
+    {
         None => None,
         Some("near") => Some(ProviderTier::Near),
         Some("chutes") => Some(ProviderTier::Attested3p),
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -84,6 +84,21 @@ fn insert_encryption_headers(
 // Custom header for exposing the inference ID as a UUID
 const HEADER_INFERENCE_ID: &str = "Inference-Id";
 
+// Custom header surfacing which trust tier served a completion.
+// Value is "near" for NEAR AI's own TEE fleet, "chutes" for the attested Chutes
+// fallback, or "non-attested" for external (non-TEE) providers.
+const HEADER_SERVING_PROVIDER: &str = "x-serving-provider";
+
+/// Map a [`inference_providers::ProviderTier`] to the string value emitted in
+/// the `x-serving-provider` response header.
+fn provider_tier_to_str(tier: inference_providers::ProviderTier) -> &'static str {
+    match tier {
+        inference_providers::ProviderTier::Near => "near",
+        inference_providers::ProviderTier::Attested3p => "chutes",
+        inference_providers::ProviderTier::NonAttested => "non-attested",
+    }
+}
+
 /// True when any E2EE encryption header was supplied. E2EE bodies are opaque
 /// to the gateway, so alias warnings can't be injected into them — the
 /// `x-model-alias-resolved` response header is the only signal in that mode.
@@ -1442,10 +1457,18 @@ async fn chat_completions_inner(
                 let mut leading_control: Vec<
                     Result<inference_providers::SSEEvent, inference_providers::CompletionError>,
                 > = Vec::new();
+                // Raw chat_id string captured alongside the hashed UUID so we can
+                // look up the serving-provider tier from the pool's chat_id mapping.
+                let mut stream_chat_id: Option<String> = None;
                 let inference_id = loop {
                     let is_control = match peekable_stream.as_mut().peek().await {
                         Some(Ok(event)) => {
                             if let Some(chunk) = &event.chunk {
+                                // Capture the raw chat_id for the tier lookup below.
+                                stream_chat_id = Some(match chunk {
+                                    inference_providers::StreamChunk::Chat(c) => c.id.clone(),
+                                    inference_providers::StreamChunk::Text(c) => c.id.clone(),
+                                });
                                 break Some(extract_inference_id_from_chunk(chunk));
                             }
                             true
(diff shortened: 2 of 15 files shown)


PierreLeGuen (approved):
Mechanical, well-contained change that compiles cleanly. The streaming serving-tier header reuses the `chat_id_mapping` the pool populates synchronously before returning the stream, and both chat and text completions stream through the same path, so the header is reliably available. All `ChatCompletionResponseWithBytes` construction sites and `AttestationServiceTrait` impls were updated (compile confirms exhaustiveness), and header values are static known-good strings.

- `crates/services/src/attestation/mod.rs:775` — when `?provider=` is supplied without a `model`, the `provider_filter` is silently ignored (the pool is only consulted inside the `if let Some(model)` branch, falling back to VPC-level attestation). Not a correctness/data risk; you could reject `provider` without `model` as a 400 if you want stricter behavior.
- `crates/api/src/routes/completions.rs:1466` — streaming header relies on `get_provider_tier_for_chat_id`; if the first chunk lacks a chat_id the header is omitted. Graceful degradation matching existing sticky-routing behavior; no action needed.

Checks: `cargo build`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` passed; `cargo test --test e2e_all serving_provider --no-run` compiled clean (e2e execution skipped — requires PostgreSQL).

PierreLeGuen (approved):
Well-contained change: the new `serving_tier` field has no default, so the compiler enforces population at all `ChatCompletionResponseWithBytes` sites, and the attestation provider filter rejects unknown values (400) while preserving existing behavior when omitted. Looks good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:2581-2593` — the streaming chat_id→provider mapping is only stored for `StreamChunk::Chat`, not `StreamChunk::Text`, so streaming text completions return `None` from `get_provider_tier_for_chat_id` and silently omit the `x-serving-provider` header (graceful degradation, no wrong value). Store the mapping for the `Text` variant too if you want parity with the PR description.
- `crates/services/src/attestation/mod.rs` — when `?provider=` is supplied without `model`, the filter is ignored and the route falls back to VPC-level attestation rather than the requested tier.

Checks: `cargo check`/`cargo clippy --all-targets`/`cargo fmt --check` clean; `cargo test --test e2e_all serving_provider --no-run` compiled; verified all 7 construction sites populate `serving_tier` and CORS `exposed_headers` includes `x-serving-provider`.

PierreLeGuen (approved):
Threads a `serving_tier` field through `ChatCompletionResponseWithBytes` to emit an `x-serving-provider` response header, and adds a `?provider=` filter to `GET /v1/attestation/report`. All 7 struct construction sites, both `AttestationServiceTrait` impls, and the single route caller are updated, so the change compiles cleanly and the threading is sound. The streaming tier lookup is safe: the pool stores the `chat_id → provider` mapping synchronously before returning the stream, so the route's subsequent lookup resolves correctly.

- Streaming **text** completions won't emit `x-serving-provider`: the pool stores the `chat_id → provider` mapping only for `StreamChunk::Chat`, so `get_provider_tier_for_chat_id` returns `None` for text streams (`crates/services/src/inference_provider_pool/mod.rs:2581`, `crates/api/src/routes/completions.rs:2290`). Harmless (the header is just absent) but it slightly contradicts the "all 4 paths" claim — store the mapping for `Text` chunks too if full coverage is wanted.

Checks: `cargo check --lib --bins`, `cargo clippy --lib --bins`, and `cargo test --lib --bins` (392 passed) all clean; CI (`security_audit`, `Lint`, `Test Suite`) green. E2E tests not run — a pre-existing local DB migration checksum mismatch (V63) blocks them, reproducible on unrelated targets.

PierreLeGuen (approved):
Clean change. The new required `serving_tier` field is set at all `ChatCompletionResponseWithBytes` construction sites (compiler-enforced), the `provider` query param parses case-insensitively and rejects unknowns with 400, and header values are static strings with no customer data. Two non-blocking observations for optional follow-up:

- `crates/services/src/inference_provider_pool/mod.rs:2582` — the `chat_id → provider` mapping is stored only for `StreamChunk::Chat`. Streaming text completions (`/v1/completions?stream=true`) extract the chat_id at `crates/api/src/routes/completions.rs:2266-2268`, but the pool lookup then returns `None`, so `x-serving-provider` is silently omitted. Graceful degradation (no wrong value), but doesn't match the "streaming text completions" claim — store the mapping for `StreamChunk::Text` too if full coverage is intended.
- `crates/services/src/inference_provider_pool/mod.rs:2456-2469` — `?provider=<tier>` on a model with no matching provider yields `ProviderNotFound` → 503; a 404 would read more accurately for "this tier isn't available for this model." Behavior is correct, not blocking. (Relatedly, `provider` supplied without `model` is silently ignored.)

CHECKS: `cargo fmt --check`, `cargo build`, `cargo clippy --all-targets -- -D warnings`, and `cargo check` across `inference_providers`/`services`/`api` (incl. tests) all passed; `cargo test --lib --bins` 392 passed/0 failed. e2e suite not run (needs PostgreSQL), but the new `serving_provider.rs` test compiles clean.

PierreLeGuen (approved):
Clean, well-contained change: adds an `x-serving-provider` response header and a `?provider=` filter on `GET /v1/attestation/report`. Streaming header timing is safe (the pool stores the `chat_id` mapping synchronously before returning the stream), `serving_tier` is a required field so the compiler enforces population at all construction sites, and the attestation filter correctly returns `ProviderNotFound` on empty results and 400 on unknown values.

One optional follow-up:
- `crates/services/src/inference_provider_pool/mod.rs:2582` — the streaming path stores the `chat_id → provider` mapping only for `StreamChunk::Chat`, not `StreamChunk::Text`, while the route captures the id from both (`completions.rs:2261-2268`). Streaming text completions will therefore silently omit the header. Graceful degradation (no wrong value), so non-blocking — store the mapping for `Text` chunks too if the header should cover that path.

Local checks: `cargo fmt --check`, `cargo check --lib --bins`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --lib --bins` (392 passed) all green; e2e execution skipped (requires PostgreSQL).

PierreLeGuen (approved):
Clean, well-scoped change: adds a required `serving_tier` field to `ChatCompletionResponseWithBytes` (all construction sites compiler-enforced) and threads an optional provider selector through the attestation filter. CI (Lint, Test Suite, security_audit) is green on the head commit.

- `crates/services/src/inference_provider_pool/mod.rs:2582` — the chat_id→provider mapping is stored only for `StreamChunk::Chat`, not `StreamChunk::Text`, so `x-serving-provider` is silently omitted on streaming text completions. This degrades gracefully (no wrong value), and the non-streaming text path emits the header correctly.
- `crates/services/src/attestation/mod.rs:794` — `?provider=` supplied without `?model=` is silently ignored and falls back to the VPC-level report. Not a correctness issue, but you may want to reject `provider` without `model` if the selector is meant to be authoritative.

Local checks: `cargo fmt --check`, `cargo check`/`cargo clippy --all-targets -D warnings` all clean; full E2E skipped (DB-dependent) but covered by CI.

PierreLeGuen (approved):
Mechanical, well-contained change: a required `serving_tier` field threaded through `ChatCompletionResponseWithBytes` (compiler-enforced at all construction sites), a new `x-serving-provider` header on the completion paths, and an optional `?provider=` filter on the attestation report endpoint. Streaming `chat_id → provider` mapping is stored synchronously before the stream returns, so the header lookup is reliably populated; attestation filter returns `ProviderNotFound` (non-200) on no match and 400 on unknown values. Looks good to merge.

- `crates/services/src/inference_provider_pool/mod.rs:2582` — the streaming `chat_id → provider` mapping is stored only for the `StreamChunk::Chat` arm, not `StreamChunk::Text`, so any stream that emits `Text` chunks would omit the header. Current paths route through `chat_completion_stream` (Chat chunks), so this is graceful degradation (header absent, never wrong); store the mapping in the `Text` arm too if text-stream parity is wanted later.
- Attestation route: `?provider=` is silently ignored when `?model=` is absent (falls through to VPC-level attestation). Behavioral edge case, not a correctness bug.

Checks: `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib --bins` 757/757 passed; `serving_provider` e2e compiled clean. CI `Lint` and `security_audit` passed; `Test Suite` failure is unrelated PostgreSQL container setup (`role "root" does not exist`), not PR code.

PierreLeGuen (approved):
Serving tier is now threaded through all response paths and the attestation provider selector looks correct. I traced the streaming path and confirmed there's no race: the `chat_id → provider` mapping is stored synchronously (`inference_provider_pool/mod.rs:2592`) before the stream is returned, keyed on the same first `Chat` chunk id the route later looks up. All `ChatCompletionResponseWithBytes` construction sites are covered (compiler-enforced, `serving_tier` non-optional).

- `inference_provider_pool/mod.rs:2580-2592` — the mapping is stored only for the `StreamChunk::Chat` arm, not `Text`, so a pure text stream would omit `x-serving-provider`. Graceful degradation (header absent, never wrong) and pre-existing behavior; `/v1/completions` translates to chat chunks so the header is still populated in practice.
- `attestation/mod.rs` (`get_attestation_report`) — `?provider=` without `?model=` is silently ignored and falls back to VPC-level attestation, making the parameter a no-op. Consider a 400 if the selector should be authoritative.

Checks: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test` (396 passed) all clean locally; CI Lint / Test Suite / security_audit green. E2E not run locally (requires PostgreSQL).

PR #810, opened 2026-06-17
feat(pool): logging + metric for Chutes-fallback visibility

Evrard-Nil:
Follow-up to #795/#798: there was no clean way to see **when a request is served by the Chutes fallback vs the NEAR primary**. The closest signal (`attempt > 1` on the success log) didn't name the tier and missed Chutes-primary models. This adds proper logging + a metric, emitted from the provider pool's retry loop — the only layer that knows the serving tier and the attempt index.

## Logging
The "Successfully completed request with provider" log now carries:
- `provider_tier` — `near` | `attested_3p` | `non_attested` (new `ProviderTier::as_str()`)
- `is_fallback` — `true` when the leading/primary tier failed and a fallback served

DD: `service:cloud-api @fields.provider_tier:attested_3p` = Chutes-served; `@fields.is_fallback:true` = a genuine NEAR→fallback.

## Metric
`cloud_api.provider.requests` — counter, **+1 per served request**, tagged `{model, provider_tier, fallback, operation}`. Enables:
- Chutes-served traffic over time (`provider_tier:attested_3p`)
- NEAR→fallback **rate** per model (`fallback:true` / total) — alert on spikes
- works for both fallback (NEAR-primary models) and Chutes-primary models

## Wiring (non-invasive)
The pool gains an optional metrics sink set once via `set_metrics_service` (interior-mutable `OnceLock`), so the many `InferenceProviderPool::new(...)` call sites (incl. tests) are **untouched**; wired in `init_domain_services_with_pool`. Absent → no-op.

## Test (green; fmt + clippy clean)
`provider_requests_metric_tags_tier_and_fallback` (via `CapturingMetricsService`): NEAR-5xx→Chutes records `provider_tier:attested_3p, fallback:true`; healthy NEAR records `provider_tier:near, fallback:false`. Existing fallback/model-not-found tests still pass.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -295,6 +295,11 @@ pub async fn init_domain_services_with_pool(
     inference_provider_pool: Arc<services::inference_provider_pool::InferenceProviderPool>,
     metrics_service: Arc<dyn services::metrics::MetricsServiceTrait>,
 ) -> DomainServices {
+    // Give the provider pool the metrics sink so it can emit the per-tier /
+    // fallback counter (cloud_api.provider.requests) from the one layer that
+    // knows which trust tier served each request.
+    inference_provider_pool.set_metrics_service(metrics_service.clone());
+
     // Create shared repositories
     let conversation_repo = Arc::new(database::PgConversationRepository::new(
         database.pool().clone(),
--- a/crates/inference_providers/src/lib.rs
+++ b/crates/inference_providers/src/lib.rs
@@ -145,6 +145,15 @@ impl ProviderTier {
     pub fn is_attested(self) -> bool {
         matches!(self, ProviderTier::Near | ProviderTier::Attested3p)
     }
+
+    /// Stable, lowercase label for logs and metric tags (low cardinality).
+    pub fn as_str(self) -> &'static str {
+        match self {
+            ProviderTier::Near => "near",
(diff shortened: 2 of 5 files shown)


PierreLeGuen (approved):
Adds observability for tiered provider routing: a `provider_tier`/`is_fallback` field on the existing success log plus a new `cloud_api.provider.requests` counter tagged `{model, provider_tier, fallback, operation}`. Wiring is non-invasive via a `OnceLock` metrics sink, so existing `InferenceProviderPool::new(...)` call sites stay untouched. The change is correct, low-cardinality, and leaks no customer data; both the streaming and non-streaming paths route through the modified retry loop.

- `crates/services/src/inference_provider_pool/mod.rs:2164` — `is_fallback = attempt > 0` flags any non-leading provider as a fallback. If a model ever has two providers of the *same* primary tier, the second would record `fallback:true` with `provider_tier` still primary, slightly over-counting the fallback rate. Current NEAR-then-Chutes catalogs don't hit this, and the `provider_tier` tag disambiguates on dashboards.

Local checks: `cargo fmt --check`, `cargo clippy --all-targets`, and `cargo check -p services --tests` all clean; the new `provider_requests_metric_tags_tier_and_fallback` test and the two existing NEAR→Chutes fallback tests pass.

PR #811, opened 2026-06-17
feat: add revenue density percentile endpoint

Evrard-Nil:
## Summary
- New `GET /admin/platform/revenue-density` endpoint (90-day cap)
- Buckets `organization_usage_log` into 1-minute windows; computes `revenue_per_sec = SUM(total_cost) / 60`
- Returns P50 / P95 / P99 / peak USD/s over **active minutes only** (buckets with any revenue), platform-wide and per model
- Includes annualized projections (`p99_annualized_usd = p99 × 86400 × 365`) — "if this demand rate were constant, what is annual revenue?"
- Returns `active_minutes` and `sampled_minutes` so callers can compute utilization fraction
- Uses existing covering index `V0062` — index-only scan, ≤43,200 intermediate rows per 30-day window; lighter than the performance-timeseries query

## Schema additions
```rust
RevenueDensityModelRow { model_name, p50/p95/p99/peak_usd_per_sec, p99/peak_annualized_usd, active_minutes }
RevenueDensityReport   { period_start/end, sampled_minutes, active_minutes, platform-wide percentiles, by_model }
RevenueDensityQuery    { start, end }
```

## Test plan
- [ ] `cargo test --lib --bins` passes
- [ ] `cargo test --test e2e_test` passes
- [ ] `cargo fmt --check` passes
- [ ] `GET /admin/platform/revenue-density` returns valid JSON on staging
- [ ] 90-day date cap enforced (range > 90 days → 400)

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1844,9 +1844,9 @@ pub fn build_admin_routes(
         get_organization as get_admin_organization, get_organization_concurrent_limit,
         get_organization_limits_history, get_organization_metrics, get_organization_timeseries,
         get_performance_timeseries, get_platform_metrics, get_platform_timeseries,
-        list_admin_access_tokens, list_invitation_email_deliveries, list_model_pricing_changes,
-        list_models as admin_list_models, list_organization_members, list_organizations,
-        list_users, preview_model_deprecation, preview_model_pricing_changes,
+        get_revenue_density, list_admin_access_tokens, list_invitation_email_deliveries,
+        list_model_pricing_changes, list_models as admin_list_models, list_organization_members,
+        list_organizations, list_users, preview_model_deprecation, preview_model_pricing_changes,
         resend_invitation_email, update_organization_concurrent_limit, update_organization_limits,
         update_service, AdminAppState,
     };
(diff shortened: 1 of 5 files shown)


PierreLeGuen (approved):
Clean additive endpoint — percentile/annualization math, nano-USD `total_cost` handling, and the empty-result `COALESCE` guard all check out, and the query is backed by the V0062 covering index for index-only scans.

One optional follow-up:
- `crates/api/src/routes/admin.rs:~3647-3653` — `parse_metrics_range(start, end, None, 90)` does not enforce the documented 90-day cap: with `granularity = None` the `max_hour_days` argument is ignored and the effective cap is the default 366-day branch. This is less restrictive than stated (admin-only, and the larger aggregation is still index-only/tolerable), so non-blocking — but the `// cap at 90 days` comment and PR description are inaccurate. If you want the stated behavior, enforce it explicitly (e.g. reject when `end - start > 90 days`).

Local checks: `cargo check` (api/services/database), `cargo fmt --check`, `cargo clippy`, and `cargo test --lib --bins` (391 passing) all clean; e2e suite skipped (requires PostgreSQL).

PR #812, opened 2026-06-17
feat: add OHTTP gateway (RFC 9458) to cloud-api

Evrard-Nil:
## Summary

- Adds an OHTTP gateway at `POST /ohttp` so decryption happens inside the cloud-api TEE before hitting the completion service, letting fallback/retry/rotation work on the plaintext inner request
- Key config served at `GET /.well-known/ohttp-gateway` (RFC standard) and `GET /v1/ohttp/config` (alias)
- `GET /v1/attestation/report` gains `ohttp_key_config` + `ohttp_attestation` fields so clients can verify the HPKE key is bound to the TEE Ed25519 signing key
- Enable with `OHTTP_ENABLED=true`

## Design

**Why loopback:** `POST /ohttp` decapsulates the HPKE envelope, then makes a loopback HTTP request to `http://127.0.0.1:{port}{inner_path}`. That request traverses the full middleware stack (auth, rate-limiting, usage accounting) and `CompletionService` with its normal retry/rotation/fallback logic. Nothing changes for the completion path — OHTTP is a transparent envelope.

**Key derivation:** `OhttpGateway::new` calls `KeyConfig::derive(key_id, Kem::X25519Sha256, ..., ed25519_seed)` — deterministic HPKE derivation from the same dstack KMS seed used for chat signing. All cloud-api instances share the same seed → identical HPKE public key → clients encrypt once, any instance can decrypt.

**Domain separation:** HPKE `DeriveKeyPair` is a different derivation path than the E2EE X25519 key (`SHA512(seed)[..32]` + RFC 7748 clamping), so the two keys are distinct (verified by a unit test in `ohttp_gateway.rs`).

**Streaming:** The chunked variant (`Content-Type: message/ohttp-chunked-req`) streams upstream SSE chunks through an async pipe → OHTTP writer → HTTP response body, preserving TTFT instead of buffering the full response.

**Relay auth:** An outer `Authorization: Bearer sk-…` on `POST /ohttp` is injected into the loopback and overrides any `Authorization` in the BHTTP payload. This lets a relay hold the API key while clients only see encrypted messages.

## Test plan

- [ ] `cargo check -p api` passes (verified locally)
- [ ] Set `OHTTP_ENABLED=true` on a staging cloud-api CVM
- [ ] `GET /v1/attestation/report` → response includes `ohttp_key_config` and `ohttp_attestation`
- [ ] `GET /.well-known/ohttp-gateway` → `Content-Type: application/ohttp-keys`, non-empty body
- [ ] `POST /ohttp` with a valid HPKE-encapsulated BHTTP chat-completions request → decrypted, routed to inference, response encapsulated back
- [ ] `POST /ohttp` without `OHTTP_ENABLED` → HTTP 404
- [ ] Chunked variant with streaming completion → TTFT not collapsed


PierreLeGuen (changes requested):
Solid OHTTP gateway implementation overall, but there's an unauthenticated SSRF in the loopback path that needs fixing before merge.

- **Blocking — SSRF via inner request path** (`crates/api/src/routes/ohttp.rs:374-389`, `parse_bhttp_and_build_loopback`): the decrypted inner BHTTP path is concatenated straight into `format!("http://127.0.0.1:{}{}", port, path_str)` with no check that it begins with `/`. Since the gateway key config is served publicly (`GET /.well-known/ohttp-gateway`), any unauthenticated party can encapsulate a request — that's the OHTTP model — and a path like `@evil.com/x` parses to host `evil.com` (the `127.0.0.1:{port}` is consumed as userinfo). `POST /ohttp` has no outer auth, so this is an unauthenticated SSRF issuing attacker-controlled outbound requests from inside the TEE (reaching internal/metadata hosts) with the response returned to the attacker. Fix: reject any `path_str` not starting with `/`, or build the URL from a fixed base `reqwest::Url` and set path/query explicitly rather than by string concatenation.

Non-blocking follow-ups:
- `crates/api/src/lib.rs:2274` and `:2377`: the test-only `ServerConfig` literals are missing the new `ohttp_enabled` field, so the api test target fails to compile (`missing field ohttp_enabled`); add `ohttp_enabled: false` to both.
- `crates/api/src/routes/ohttp.rs:17,38,68`: `OhttpError` is `pub(crate)` but used in `pub` handler signatures, producing `private_interfaces` warnings that CI's `-D warnings` clippy will reject; make `OhttpError` `pub` or the handlers `pub(crate)`.

Checks run: `cargo check -p api` passes (2 visibility warnings); `cargo test --lib -p api -- ohttp` 3/3 pass; full api test target fails to compile (missing `ohttp_enabled`); `cargo fmt --check` reports formatting diffs in 4 files; attacker-controlled-path URL parsing confirmed (`@evil.com/...` → host `evil.com`).

Evrard-Nil (commented):
Fixed both blocking items in eb636c40:

**SSRF (path validation):** `parse_bhttp_and_build_loopback` now rejects any inner path that does not start with `/` or starts with `//` before the `format!` call. This rules out the `@evil.com/x` userinfo-injection vector. Error returned: `400 BadRequest`.

**Body-size limit:** Added `DefaultBodyLimit::max(OHTTP_MAX_BODY_SIZE)` (32 MB) as a layer directly on the `POST /ohttp` route. 32 MB covers audio-transcription payloads (≤25 MB) + HPKE overhead while bounding unauthenticated memory before decryption.

The non-blocking items (missing `ohttp_enabled` field, `OhttpError` visibility, fmt) were already addressed in the two prior commits.

PierreLeGuen (approved):
Adds an OHTTP gateway (RFC 9458) plus an admin revenue-density endpoint. The two prior blocking items — pre-auth SSRF via the client-controlled inner path and the missing request body-size limit — are both resolved at the current HEAD (`eb636c40`); the path is now validated to start with a single `/` (`crates/api/src/routes/ohttp.rs:395`) and a 32 MB `DefaultBodyLimit` is layered on `POST /ohttp` (`crates/api/src/lib.rs:1273`). The earlier `CHANGES_REQUESTED` review was on a prior commit and has been fully addressed. Good to merge.

- `crates/api/src/routes/ohttp.rs` (chunked path, backend stream-error arm): a mid-stream backend error `break`s and then writes the normal body/trailer terminators, so the client gets a well-formed but silently truncated response with no error signal. Consider aborting the stream without clean termination so clients can detect incompleteness.
- `crates/api/src/routes/ohttp.rs` (standard path): `bhttp::StatusCode::try_from(status).unwrap_or(OK)` rewrites an unmappable upstream status to 200. Harmless in practice (reqwest statuses are always valid), but mapping to 502/500 would be safer if ever hit.

Checks run: `cargo check -p api`, `cargo clippy -p api -- -D warnings`, `cargo fmt --check`, and `cargo test --lib -p api -- ohttp` (3/3 pass).

Note on Report 1's body-limit finding: it claims `DefaultBodyLimit` doesn't apply because the handler reads the raw body via `into_body().collect()`, but in axum `DefaultBodyLimit` installs a limited-body layer that bounds `Body::collect()` itself, so the 32 MB cap does take effect on `/ohttp`. Not blocking.

Issue #601, opened 2026-05-18
Add cache-read pricing


Evrard-Nil:
Surface and bill cache-read tokens distinctly from regular input tokens.

Source: May 13 Pierre. The Anthropic and Google upstreams expose cache-read pricing (typically ~10% of input cost) — we should pass that through on /v1/chat/completions usage payloads and reflect it in invoices.

PierreLeGuen:
Closing as implemented. Cache-read pricing is now represented through cache_read_cost_per_token in the model/admin data path, and cached-token usage is handled separately from regular input tokens in usage/billing. Follow-on cache behavior remains tracked in narrower issues such as #473 for Responses API cache enablement and #666 for Anthropic cache_control / passthrough cache stats.

PR #814, opened 2026-06-17
fix(anthropic): forward cache_control and surface cache-read tokens in usage (#666)

### LINKED ISSUE #666: Anthropic prompt caching (cache_control) silently dropped; passthrough cache stats not surfaced in usage
## Summary

cloud-api accepts requests with Anthropic's `cache_control` hint on message content parts, but doesn't pass it through to Anthropic and doesn't surface any cache statistics in `usage.prompt_tokens_details`. Users can't benefit from Anthropic's prompt-caching cost savings via the gateway.

For comparison, self-hosted models like `openai/gpt-oss-120b` *do* surface `cached_tokens` in `usage.prompt_tokens_details` — so the wire format is supported, just not populated for Anthropic.

## Repro

```bash
KEY=sk-...
# Long system prompt (eligible for caching at Anthropic's 1024+ token threshold)
SYSPROMPT=$(python3 -c "print('You are an expert assistant. ' * 200)")

# Call 1 with cache_control on system message
  -d "{\"model\":\"anthropic/claude-haiku-4-5\",\"messages\":[
    {\"role\":\"system\",\"content\":[
      {\"type\":\"text\",\"text\":\"$SYSPROMPT\",\"cache_control\":{\"type\":\"ephemeral\"}}
    ]},
    {\"role\":\"user\",\"content\":\"hi\"}
  ],\"max_tokens\":30}" | jq '.usage'
# {"prompt_tokens":1219, "completion_tokens":24, "total_tokens":1243}
# ↑ no prompt_tokens_details, no cached_tokens — even though the field was accepted

# Call 2 with same system prompt → should be cache HIT
# Same usage, no cached_tokens

# Compare to gpt-oss-120b which does populate it:
curl -s ... '{"model":"openai/gpt-oss-120b",...}' | jq '.usage.prompt_tokens_details'
# {"cached_tokens": 64}
```

`cache_control` isn't mentioned anywhere in `/api-docs/openapi.json`, so users following the spec wouldn't know to try it — but Anthropic SDK users will reflexively send it on long system prompts.

## Why it matters

Anthropic prompt caching offers up to 90% cost reduction on cached input tokens. For an agent making many calls with a stable system prompt (one of the main use cases for the Responses API too — system prompt + conversation history), this is real money.

Today, an Anthropic-direct app migrating to cloud-api will lose this benefit silently — same request shape, same final output, but with the caching value erased.

## Suggested fix

1. **Pass `cache_control` through** when forwarding to Anthropic — the field should be preserved in the outgoing request body.
2. **Surface cache stats from Anthropic** — Anthropic responses include `usage.cache_creation_input_tokens` and `usage.cache_read_input_tokens`. Map those into `prompt_tokens_details.{cached_tokens, ...}` for OpenAI-shape compatibility.
3. **Document `cache_control`** in the OpenAPI spec as a supported content-part field (or reject it explicitly so users know it doesn't work).

This unblocks a major Anthropic feature for cloud-api users.

---

PierreLeGuen:
## Summary

Fixes #666. Anthropic prompt caching was silently broken in the external Anthropic backend: incoming `cache_control` breakpoints were dropped on the request side, and Anthropic's cache-usage stats were never surfaced in the response usage. The billing/pricing plumbing already records `cache_read_tokens` from `usage.cached_tokens()` (issue #601), so this only needed the converter to populate `prompt_tokens_details.cached_tokens`.

## Request side
- Add an optional `cache_control` to `AnthropicContentPart::Text` and `::Image`, forwarded **verbatim** from the caller's OpenAI content part (Anthropic's `{"type":"ephemeral"}` shape). Parsed locally from the raw message JSON in the converter rather than widening the shared `ContentPart` enum (keeps the Gemini path untouched).
- `system`: emit an array of text blocks (with `cache_control`) when the system message carries a breakpoint; keep the bare-string form otherwise so the common uncached request is **byte-identical** to before.
- Per-part breakpoints are aligned 1:1 with the parsed content blocks by reusing `parse_content_part` as the recognizer, so alignment holds even for malformed parts.

## Response side
- Extend `AnthropicUsage` with `cache_read_input_tokens` / `cache_creation_input_tokens` (default 0).
- Non-streaming and streaming both fold the cache tokens into `prompt_tokens` and report the read portion as `prompt_tokens_details.cached_tokens`. Anthropic reports cache reads **separately** from `input_tokens`, whereas OpenAI's `cached_tokens` is a **subset** of `prompt_tokens`; folding them in preserves the `cached <= prompt` invariant (which `TokenUsage::cached_tokens()` clamps to) and bills the cache-read cost via the existing path.

## Tests
11 new unit tests (cache_control on system + user text parts, malformed-part alignment regression, usage deserialization/defaults, non-streaming + streaming cached-token mapping, and the no-regression bare-string system case). Full crate suite green; clippy `-D warnings` clean.

```
cargo test  -p inference_providers   # 353 passed / 0 failed / 1 ignored
cargo clippy -p inference_providers --all-targets --all-features -- -D warnings   # clean
cargo build -p inference_providers && cargo build -p api   # ok
```

## Notes / out of scope
- Surfacing Anthropic reasoning/`thinking` output as `reasoning_content`, and OpenAI `completion_tokens_details.reasoning_tokens` — separate concerns, left out.
- `cache_creation_input_tokens` is folded into `prompt_tokens` (billed at the normal input rate). There is no OpenAI-shaped slot for cache-write tokens; surfacing cache-write cost at Anthropic's premium rate would require billing-side changes and is left as a follow-up.

### DIFF
--- a/crates/api/src/conversions.rs
+++ b/crates/api/src/conversions.rs
@@ -1019,6 +1019,143 @@ mod tests {
         assert_eq!(domain.stop, Some(vec!["a".to_string(), "b".to_string()]));
     }
 
+    /// END-TO-END (#666): cache_control must survive the REAL public path —
+    /// `api::models::Message` (deserialized from the HTTP body) -> `From<Message>
+    /// for ChatMessage` (which re-serializes the converted content into
+    /// `ChatMessage.content`) -> the Anthropic converter's `convert_messages`.
+    ///
+    /// Before the Part-1 change, `MessageContentPart::Text`/`ImageUrl` had no
+    /// `cache_control` field, so serde dropped the breakpoint at deserialization
+    /// and the re-serialize produced cache-control-free content; this test would
+    /// then find a bare-string `system` and no breakpoint, and fail. With the
+    /// field present, the breakpoint rides through to the Anthropic request.
+    #[test]
+    fn cache_control_survives_message_to_anthropic_request() {
+        use inference_providers::non_attested::external::anthropic::converter::{
+            convert_messages, AnthropicContentPart, AnthropicMessageContent, AnthropicSystem,
+        };
+
+        // Deserialize from a wire body so the WHOLE path is exercised: the HTTP
+        // JSON -> `MessageContentPart` (the enum the reviewer flagged) -> the rest.
+        let system: crate::models::Message = serde_json::from_value(serde_json::json!({
+            "role": "system",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "Large shared preamble",
+                    "cache_control": {"type": "ephemeral"}
+                }
+            ]
+        }))
+        .expect("system message should deserialize");
+        let user: crate::models::Message = serde_json::from_value(serde_json::json!({
+            "role": "user",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "Cached context",
+                    "cache_control": {"type": "ephemeral", "ttl": "1h"}
+                },
+                {"type": "text", "text": "Volatile question"}
+            ]
+        }))
+        .expect("user message should deserialize");
+
+        // The REAL From impl: re-serializes converted content into ChatMessage.content.
+        let chat_messages: Vec<ChatMessage> = vec![system.into(), user.into()];
+
+        let (anthropic_system, anthropic_messages) = convert_messages(&chat_messages);
+
+        // `system` must be the block-array form carrying the verbatim breakpoint.
+        match anthropic_system.expect("system present") {
+            AnthropicSystem::Blocks(blocks) => {
+                assert_eq!(blocks.len(), 1);
+                assert_eq!(blocks[0].text, "Large shared preamble");
+                assert_eq!(
+                    blocks[0].cache_control,
+                    Some(serde_json::json!({"type": "ephemeral"}))
+                );
+            }
+            other => panic!("expected block-array system, got {other:?}"),
+        }
+
+        // The user turn must carry the breakpoint on the cached text block only.
+        assert_eq!(anthropic_messages.len(), 1);
+        let blocks = match &anthropic_messages[0].content {
+            AnthropicMessageContent::Blocks(b) => b,
+            AnthropicMessageContent::Text(t) => {
+                panic!("cache_control should force the block form, got text: {t}")
+            }
+        };
+        assert_eq!(blocks.len(), 2);
+        match &blocks[0] {
+            AnthropicContentPart::Text {
+                text,
+                cache_control,
+            } => {
+                assert_eq!(text, "Cached context");
+                assert_eq!(
+                    *cache_control,
+                    Some(serde_json::json!({"type": "ephemeral", "ttl": "1h"}))
+                );
+            }
+            other => panic!("expected text block, got {other:?}"),
+        }
+        match &blocks[1] {
+            AnthropicContentPart::Text {
+                text,
+                cache_control,
+            } => {
+                assert_eq!(text, "Volatile question");
+                assert!(
+                    cache_control.is_none(),
+                    "breakpoint must not bleed onto the volatile part"
+                );
+            }
+            other => panic!("expected text block, got {other:?}"),
+        }
+    }
+
+    /// Companion guard: the SAME public path must NOT leak cache_control toward a
+    /// non-Anthropic upstream. `strip_cache_control` (called by the vLLM /
+    /// openai_compatible / Chutes backends) removes the breakpoint from the
+    /// re-serialized `ChatMessage.content`, restoring the exact pre-#666 bytes
+    /// those upstreams used to receive.
+    #[test]
+    fn cache_control_stripped_for_non_anthropic_path() {
+        let user: crate::models::Message = serde_json::from_value(serde_json::json!({
+            "role": "user",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "Cached context",
+                    "cache_control": {"type": "ephemeral"}
+                },
+                {"type": "text", "text": "Volatile question"}
+            ]
+        }))
+        .unwrap();
+
+        let mut chat_messages: Vec<ChatMessage> = vec![user.into()];
+        // Sanity: the breakpoint is present before stripping (Part-1 preserved it).
+        let before = serde_json::to_string(&chat_messages).unwrap();
+        assert!(
+            before.contains("cache_control"),
+            "cache_control must reach ChatMessage.content before stripping"
+        );
+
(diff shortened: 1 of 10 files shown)


Evrard-Nil (commented):
Reviewed commit `9951343c`. Built locally (`cargo build -p inference_providers`, `cargo build -p api`), ran `cargo test -p inference_providers --lib` (353 passed / 0 failed), `cargo clippy -p inference_providers --all-targets --all-features -- -D warnings` (clean). CI green across all four jobs (lint, test-suite, security-audit, claude-review).

## Summary

The PR correctly implements both halves of #666:
- **Request side**: `cache_control` forwarded verbatim from OpenAI content parts to Anthropic, with correct system-prompt block-array promotion and per-part alignment that survives malformed parts.
- **Response side**: Anthropic's additive cache stats (`cache_read_input_tokens` / `cache_creation_input_tokens`) correctly folded into `prompt_tokens` and the read portion surfaced as `prompt_tokens_details.cached_tokens`, preserving the `cached <= prompt` invariant that the billing path relies on.

The design choice (Anthropic reports cache tokens **additively** to `input_tokens`, not as a subset) is correctly documented and the code matches Anthropic's API semantics. Gemini Code Assist's "double counting" concern in its automated review is based on the wrong premise — Anthropic's `input_tokens` in a cached response does NOT include the cache_read or cache_creation counts; they are separate additive fields.

## Non-blocking notes

**Billing: `cache_creation_input_tokens` at `input_rate` vs Anthropic's `~1.25x` creation rate**

The 5–25% undercharge on cache-write requests is acknowledged in the PR notes as a follow-up. Not blocking because (a) admins currently haven't set a non-zero `cache_read_cost_per_token` for Anthropic models (so the `== 0` fallback bills everything at full `input_rate` anyway), and (b) there's no `cache_creation_cost_per_token` field in `ModelPricing` yet. When that field lands, the PR will need a companion update.

**`prompt_tokens_details` absent on cache-creation-only requests**

When `cache_creation_input_tokens > 0` but `cache_read_input_tokens == 0` (first request seeding the cache), the client gets an inflated `prompt_tokens` with no `prompt_tokens_details` to explain why. This is a UX gap but not a correctness issue. Suggest tracking in a follow-up issue.

**Tool result `cache_control` not forwarded**

`MessageRole::Tool` content goes through `text_from_content` (plain string), so any `cache_control` on tool results is dropped. Anthropic supports `cache_control` on tool results, but this isn't part of the OpenAI tool-result format, and the use case is narrow. Non-blocking.

**Test coverage**

The 11 new tests cover the key paths well: system bare-string regression, block-array promotion, per-part alignment on malformed inputs, usage deserialization with/without cache fields, streaming early chunk + final chunk, and non-streaming `map_usage`. One gap: no test for the case where both `cache_read > 0` AND `cache_creation > 0` appear together in the non-streaming `map_usage` path (though the streaming test at line 1289 does cover that). Low risk.

## Verdict

Code is correct, tests are thorough, CI is green, and the known limitations are clearly documented. Approving.

Evrard-Nil (approved):
Reviewed commit `9951343c`. Built locally (`cargo build -p inference_providers`, `cargo build -p api`), ran `cargo test -p inference_providers --lib` (353 passed / 0 failed), `cargo clippy -p inference_providers --all-targets --all-features -- -D warnings` (clean). CI green across all four jobs (lint, test-suite, security-audit, claude-review).

## Summary

The PR correctly implements both halves of #666:
- **Request side**: `cache_control` forwarded verbatim from OpenAI content parts to Anthropic, with correct system-prompt block-array promotion and per-part alignment that survives malformed parts.
- **Response side**: Anthropic's additive cache stats (`cache_read_input_tokens` / `cache_creation_input_tokens`) correctly folded into `prompt_tokens` and the read portion surfaced as `prompt_tokens_details.cached_tokens`, preserving the `cached <= prompt` invariant that the billing path relies on.

The design choice (Anthropic reports cache tokens **additively** to `input_tokens`, not as a subset) is correctly documented and the code matches Anthropic's API semantics. Gemini Code Assist's "double counting" concern is based on a wrong premise — Anthropic's `input_tokens` in a cached response does NOT include the cache_read or cache_creation counts; they are separate additive fields per Anthropic's API spec.

## Non-blocking notes

**Billing: `cache_creation_input_tokens` billed at `input_rate` vs Anthropic's ~1.25x creation rate**

Acknowledged in the PR notes as a follow-up. Not blocking because (a) `cache_read_cost_per_token` isn't set to a non-zero value on Anthropic models yet (the `== 0` fallback bills everything at full `input_rate` anyway), and (b) there's no `cache_creation_cost_per_token` field in `ModelPricing`. When that field is added, this converter will need a companion update.

**`prompt_tokens_details` absent on cache-creation-only requests**

When `cache_creation_input_tokens > 0` but `cache_read_input_tokens == 0` (first request seeding the cache), the client sees an inflated `prompt_tokens` with no `prompt_tokens_details` to explain it. UX gap, not a correctness bug. Worth a follow-up issue.

**Tool result `cache_control` not forwarded**

`MessageRole::Tool` content flows through `text_from_content` (plain string), so `cache_control` on tool results is dropped. Anthropic supports this but it's not part of the standard OpenAI tool-result format; narrow use case, non-blocking.

## Verdict

Correct implementation, thorough test coverage (11 new tests covering the key paths and regression cases), CI green, known limitations clearly documented. Good to merge.

PierreLeGuen:
Thanks for the reviews — triaged:

**`cache_creation_input_tokens` double-counting (gemini, high):** false positive — please don't apply the subtraction. Per Anthropic's prompt-caching docs the three counts are mutually exclusive: `total_input_tokens = cache_read_input_tokens + cache_creation_input_tokens + input_tokens`, and "`input_tokens` … represents only the tokens that come after the last cache breakpoint." So folding `cache_read` + `cache_creation` into `prompt_tokens` is correct; subtracting `cache_creation` would undercount. The existing test values (10 + 80 + 5 → 95) are right.

**Preserve cache_control through API message parsing (codex, P1):** valid and important — confirmed. `api::models::MessageContentPart` has no `cache_control`/flattened field, so serde drops the breakpoint before `From<Message> for ChatMessage` re-serializes the content. The converter passthrough only fires for directly-built `ChatMessage` (unit tests), not real API traffic. Fixing by capturing `cache_control` in the API content-part model (so it survives to the Anthropic converter) while keeping non-Anthropic upstream bytes unchanged (no leak), plus an end-to-end test through the real `From<Message>` path.

**Preserve cache_control on assistant turns (codex, P2):** valid — addressing in the same change so breakpoints on assistant content blocks are forwarded.

Pushing the fix shortly.

PR #815, opened 2026-06-17
fix(anthropic): return requested model name in non-streaming responses for transport consistency (#632)

### LINKED ISSUE #632: Response 'model' field doesn't match request; passthroughs strip prefix; streaming/non-streaming inconsistent for Anthropic
## Summary

The `model` field in responses diverges from the requested model name, with a different pattern per provider. Same request between streaming and non-streaming for Anthropic also returns different values.

| Requested | Non-streaming `model` | Streaming `model` |
|---|---|---|
| `anthropic/claude-haiku-4-5` | `claude-haiku-4-5-20251001` | `claude-haiku-4-5-20251001` (in chunks); some chunks show `claude-haiku-4-5` |
| `openai/gpt-4.1-nano` | `gpt-4.1-nano-2025-04-14` | (untested but likely same) |
| `google/gemini-2.5-flash` | `gemini-2.5-flash` | (likely same) |
| `openai/gpt-oss-120b` | `openai/gpt-oss-120b` | `openai/gpt-oss-120b` |

Self-hosted models echo the requested name verbatim. Passthroughs strip the org prefix and substitute the upstream's canonical name (sometimes with a date suffix).

Worse: in the test above for `anthropic/claude-haiku-4-5`, the **streaming** chunk-level `model` field also varied within the same response — at least one event reported `claude-haiku-4-5` and others `claude-haiku-4-5-20251001`. That internal inconsistency is the most problematic part.

## Why it matters

- Clients keying on `response.model` for logging/cost-tracking/multiplexing get a different value than they sent. Doesn't match for passthroughs.
- A client building a `model` allowlist needs to know all possible canonicalized forms per provider — undocumented today.
- The streaming/non-streaming inconsistency means equivalent requests have different observable behavior depending on transport.

## Repro

```bash

# non-streaming
  -d '{"model":"anthropic/claude-haiku-4-5","messages":[{"role":"user","content":"hi"}],"max_tokens":10}' \
  | jq -r .model
# claude-haiku-4-5-20251001

# streaming
curl -sN https://cloud-api.near.ai/v1/chat/completions \
  -d '{"model":"anthropic/claude-haiku-4-5","messages":[{"role":"user","content":"hi"}],"max_tokens":10,"stream":true}' \
  | grep -oE '"model":"[^"]+"' | sort -u
# "model":"claude-haiku-4-5"
# (and sometimes other variants)
```

## Suggested fix

Rewrite the response's `model` field (both top-level non-streaming and per-chunk streaming) to match what the client sent in the request. This is what most OpenAI-compatible gateways do (litellm, OpenRouter, etc.).

If a user requested `anthropic/claude-haiku-4-5`, that's the model field they should see — leaking the upstream's internal canonical name doesn't help anyone and breaks clients keying off the request name.

---

PierreLeGuen:
## Summary

Fixes #632 (the internal transport-inconsistency part; "Plan A"). For the same chat-completions request to an Anthropic-backed external model, the response `model` field differed by transport:
- **non-streaming** echoed the upstream dated canonical name (e.g. `claude-haiku-4-5-20251001`)
- **streaming** echoed the name we sent (e.g. `claude-haiku-4-5`)

The streaming path seeds `AnthropicParserState.model` from the sent model arg and never deserializes the upstream name, so it already returns the sent name. This change makes the **non-streaming** path do the same: the response `model` now uses the requested/sent model name, and upstream's dated suffix is intentionally not surfaced. Both transports now echo the same value for an identical request.

## Implementation
- Extract the non-streaming response construction into a pure `build_openai_response(anthropic_response, sent_model, wants_json)` helper so the model-echo behavior is unit-testable without HTTP/infra. The only behavioral delta is `model`: `anthropic_response.model` → `sent_model.to_string()`.
- `AnthropicResponse.model` is now read only in tests; kept (with a comment) for wire-parity, marked `#[allow(dead_code)]` to keep clippy `-D warnings` clean.

## Scope / alias-transparency (#573)
This is the **minimal** consistency fix only. Broader global model-name normalization across all backends/transports (the issue's "Plan B") is intentionally **out of scope** and left for separate discussion.

The #573 alias-transparency contract is intact: the override e2e assertions in `model_alias_transparency.rs` (`body["model"] == upstream`) exercise the **openai_compatible mock** backend (not Anthropic), and for the override case the sent/effective model name equals the upstream override string — so those assertions are unaffected.

## Tests
```
cargo test  -p inference_providers   # 345 passed / 0 failed / 1 ignored  (+2 new #632 tests)
cargo clippy -p inference_providers --all-targets --all-features -- -D warnings   # clean
cargo build -p inference_providers && cargo build -p api   # ok
```
The `crates/api` e2e alias suite could not be executed locally due to a pre-existing shared-DB migration mismatch (the shared `platform_api_e2e` DB has recorded `V63__add_allow_free_to_models`, which isn't on `main`); this is environment state, unrelated to this change, which touches only the Anthropic backend.

### DIFF
--- a/crates/inference_providers/src/non_attested/external/anthropic/converter.rs
+++ b/crates/inference_providers/src/non_attested/external/anthropic/converter.rs
@@ -227,6 +227,13 @@ pub struct AnthropicError {
 pub struct AnthropicResponse {
     pub id: String,
     pub content: Vec<AnthropicContentBlock>,
+    // #632: Anthropic returns the upstream dated canonical name here (e.g.
+    // `claude-haiku-4-5-20251001`). We deserialize it for completeness/parity
+    // with the wire payload, but intentionally do NOT surface it on our
+    // response: the non-streaming path echoes the requested/sent model name to
+    // stay consistent with the streaming path (which never sees this field).
+    // Hence it is read only in tests; allow dead_code in non-test builds.
+    #[allow(dead_code)]
     pub model: String,
     pub stop_reason: Option<String>,
     pub usage: AnthropicUsage,
--- a/crates/inference_providers/src/non_attested/external/anthropic/mod.rs
+++ b/crates/inference_providers/src/non_attested/external/anthropic/mod.rs
@@ -95,6 +95,77 @@ fn extract_passthrough(
         .collect()
 }
 
+/// Build our normalized OpenAI-shaped response from a parsed Anthropic
+/// non-streaming response.
+///
+/// `sent_model` is the model name cloud-api sent to Anthropic (the `model`
+/// argument threaded through the backend call), NOT `anthropic_response.model`.
+///
+/// #632: the response `model` field uses the requested/sent model name to stay
+/// consistent with the streaming path, which seeds its chunk model from the
+/// same sent name (see `AnthropicParserState::new` in `converter.rs`).
+/// Upstream's dated canonical name (e.g. `claude-haiku-4-5-20251001`) is
+/// intentionally NOT surfaced, so both transports echo the same value for an
+/// identical request.
+fn build_openai_response(
+    anthropic_response: AnthropicResponse,
+    sent_model: &str,
+    wants_json: bool,
+) -> ChatCompletionResponse {
+    // Convert to OpenAI format using the converter
+    let (content, tool_calls) = extract_response_content(&anthropic_response.content);
+
+    // #668: Anthropic has no native JSON-output mode, so when the caller
+    // requested `response_format: json_object`/`json_schema` it tends to
+    // wrap the JSON in a markdown ```json … ``` fence, breaking
+    // `JSON.parse`. Strip that fence so the content is raw parseable JSON.
+    let content = if wants_json {
+        content.map(|c| strip_json_code_fence(&c))
+    } else {
+        content
+    };
+
+    ChatCompletionResponse {
+        id: anthropic_response.id,
+        object: "chat.completion".to_string(),
+        created: chrono::Utc::now().timestamp(),
+        // #632: echo the requested/sent model name (not upstream's dated name)
+        // so non-streaming matches the streaming path for the same request.
+        model: sent_model.to_string(),
+        choices: vec![ChatCompletionResponseChoice {
+            index: 0,
+            message: ChatResponseMessage {
+                role: MessageRole::Assistant,
+                content,
+                refusal: None,
+                annotations: None,
+                audio: None,
+                function_call: None,
+                tool_calls,
+                reasoning_content: None,
+                reasoning: None,
+            },
+            logprobs: None,
+            finish_reason: map_finish_reason_string(anthropic_response.stop_reason),
+            token_ids: None,
+            extra: Default::default(),
+        }],
+        service_tier: None,
+        system_fingerprint: None,
+        usage: TokenUsage {
+            prompt_tokens: anthropic_response.usage.input_tokens,
+            completion_tokens: anthropic_response.usage.output_tokens,
+            total_tokens: anthropic_response.usage.input_tokens
+                + anthropic_response.usage.output_tokens,
+            prompt_tokens_details: None,
+        },
+        prompt_logprobs: None,
+        prompt_token_ids: None,
+        kv_transfer_params: None,
+        extra: Default::default(),
+    }
+}
+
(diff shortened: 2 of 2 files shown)


Evrard-Nil (approved):
## Review: Approve

Reviewed branch `pr815` (commit `0a6d67ba`). Ran `cargo check` (clean) and `cargo test --lib -p inference_providers` (345/345 pass).

---

### What this PR actually is

The title says "fix(anthropic): return requested model name in non-streaming responses" — but the diff is much larger than that. It bundles **three distinct changes**:

1. **Core fix (Anthropic non-streaming model echo, `inference_providers`)** — the stated PR subject.
2. **Revert: `allow_free` pricing gate** — removes the `allow_free` DB column/field/gate/tests/migrations that were presumably added in a previous PR. This is the largest chunk of the diff by line count.
3. **Revert: `get_current_user` N+1 optimization** — reverts `list_organizations_with_roles_for_user` + `list_workspaces_for_user` (single-query JOINs) back to the old O(N) loop: `list_organizations_for_user` → per-org `get_user_role` → per-org `list_workspaces_for_organization`.

All three are correct and the code compiles cleanly, but they should ideally be separate commits/PRs for bisectability. I'll note the details below.

---

### Fix 1: Anthropic non-streaming model echo (#632) ✓

**The bug:** `chat_completion` (non-streaming) was returning `anthropic_response.model` — the upstream dated canonical name Anthropic puts in its wire payload (e.g. `claude-haiku-4-5-20251001`). The streaming path was already correct: `AnthropicParserState::new(model)` seeds the model from the `sent` name (the `model` arg threaded from `ExternalProvider::model_name`), so non-streaming and streaming disagreed for identical requests.

**The fix:** Extracts a `build_openai_response(anthropic_response, sent_model, wants_json)` helper that takes `sent_model` explicitly instead of reading it from `anthropic_response.model`. The `AnthropicResponse.model` field is kept (needed for deserialization to prove Anthropic actually returned it) but annotated `#[allow(dead_code)]` in non-test builds.

**Tests:** Two new unit tests (`test_non_streaming_response_model_is_sent_name_not_upstream_dated`, `test_non_streaming_response_model_matches_streaming_seed`) correctly verify the fix and guard against regression.

**One subtlety to be aware of (not a blocker):** `sent_model` is `ExternalProvider::model_name`, which resolves to the provider-config `model_name` override if set (e.g. `claude-haiku-4-5-20251001` if the DB record has that as the override). If an Anthropic model is configured with a dated override in `model_name`, the response will echo that dated name, not the client-requested alias. The streaming path has the same behavior, so this PR is consistent — but it means issue #632 is only fully closed if the model is configured with the alias (not the dated name) as the provider model_name. Worth a follow-up doc note, not a blocker here.

**Billing impact:** None. The `model` field in the response is display-only; billing uses the internal model ID from the usage pipeline, not the response JSON.

---

### Fix 2: Revert `allow_free` pricing gate ✓

Removes: DB migrations `V0064` and `V0065`, the `allow_free` column in `models` and `model_history`, the `get_model_costs` repository method, the `validate_model_request` activation gate, the full `admin_activation_pricing_gate.rs` e2e test file, and all field references throughout the stack.

The removal is complete and consistent — no orphaned references remain (`grep -r allow_free` across the repo comes up clean). The migration deletions mean these migrations were not yet applied to production (or are being intentionally rolled back), which must be confirmed out-of-band.

---

### Fix 3: Revert `get_current_user` N+1 optimization ✓

The `list_organizations_with_roles_for_user` JOIN + `list_workspaces_for_user` JOIN approach is reverted to:
- `list_organizations_for_user` (one query) → loop → `get_user_role` per org (one query each) → `list_workspaces_for_organization` per org (one query + one `is_member` check each)

This is a deliberate O(N) reversion. The associated index migration `V0063__add_workspaces_org_active_index.sql` is also deleted. The new code is functionally equivalent to the previous behavior (before the optimization) and all membership checks are preserved, so there's no correctness issue — only a performance regression relative to the code being reverted. Again, the migrations must not have landed in production for this deletion to be safe.

---

### Summary

All three changes are internally consistent and correct. The build is clean and tests pass. The core Anthropic fix is the right approach.

**One process note:** the three changes should have been separate commits/PRs — the unrelated reverts make it harder to understand the change history for #632. Consider splitting in future PRs.

PR #819, opened 2026-06-17
feat: revenue density $/min units + provider_type filter

Evrard-Nil:
## Summary
- Renames all `*_usd_per_sec` struct fields to `*_usd_per_min`: a 1-minute bucket sum is naturally USD/min with no division needed
- Adds `provider_type: Option<String>` to `RevenueDensityQuery`; both the platform-wide and per-model SQL queries `LEFT JOIN models m ON m.id = ul.model_id` and apply `$3::text IS NULL OR m.provider_type = $3`
- Annualization constant updated: `YEAR_MINUTES = 60 × 24 × 365 = 525 600`; result is numerically identical to the prior `p99_per_sec × 86400 × 365`
- `RevenueDensityParams` in the handler gains `provider_type: Option<String>` and passes it through

## Test plan
- [ ] `cargo check` / `cargo fmt --check` / `cargo clippy` clean
- [ ] `cargo test --lib --bins` passes
- [ ] `cargo test --test e2e_test` passes
- [ ] `GET /admin/platform/revenue-density?provider_type=nearai` returns only nearai models in `by_model`
- [ ] Without `provider_type`, all models returned as before

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3630,6 +3630,8 @@ pub async fn get_performance_timeseries(
 pub struct RevenueDensityParams {
     pub start: Option<String>,
     pub end: Option<String>,
+    /// Optional provider type filter (e.g. "nearai", "external").
+    pub provider_type: Option<String>,
 }
 
 /// Get revenue density percentiles (Admin only)
@@ -3656,7 +3658,11 @@ pub async fn get_revenue_density(
 
     let result = app_state
         .analytics_service
-        .get_revenue_density(services::admin::RevenueDensityQuery { start, end })
+        .get_revenue_density(services::admin::RevenueDensityQuery {
+            start,
+            end,
+            provider_type: params.provider_type,
+        })
         .await
         .map_err(|e| {
             error!("Failed to get revenue density: {:?}", e);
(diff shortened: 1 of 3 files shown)


PierreLeGuen (approved):
Clean, self-contained refactor: renames `*_usd_per_sec` → `*_usd_per_min` (annualized fields preserve identical output, since `86400×365/60 = 525600` and `total_cost_nano` cancels out the same way) and adds an optional `provider_type` filter. SQL is correct — `organization_usage_log.model_id` is a non-nullable FK to `models(id)` (PK), so the `LEFT JOIN` causes no fan-out and aggregates are unchanged; `provider_type` is a real column on `models`. No stale references to the old field/const names remain.

- `crates/services/src/admin/analytics.rs:471` (and `admin.rs:3633`): the `provider_type` docstring example uses `"nearai"`, but stored values are `'vllm'`/`'external'`/`'chutes'` (V0042/V0056/V0060), so that example would silently return no rows. Consider using `"vllm"` in the doc example. The pass-through itself is correct.

Checks run: `cargo fmt --check`, `cargo check`, `cargo clippy --all-targets`, and `cargo test` (services lib + `--lib --bins`, 753 passed / pre-existing ignored) — all clean.

PR #822, opened 2026-06-17
ci: install gh CLI on self-hosted runner in promote workflow

Evrard-Nil:
## Summary

- The promote workflow was migrated to `[self-hosted, infra]` runners in #820, but self-hosted runners don't have `gh` CLI pre-installed (unlike GitHub-hosted `ubuntu-latest`). The `Create GitHub release` step was failing with `gh: command not found`.
- Add `gh` to the `apt-get install` step (same pattern as the rust toolchain fix in security-audit.yml in #820).
- Bonus: add fast-fail digest format validation — the previous two runs failed because `inputs.digest` received `"promote"` instead of a real `sha256:...` digest; the new guard catches that immediately.

## Test plan

- [ ] Merge and re-run promote workflow — `Create GitHub release` step should succeed
- [ ] Try triggering with a bad digest (e.g. `promote`) — should fail at `Resolve source digest and SHA` with a clear error before touching Docker

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -52,10 +52,10 @@ jobs:
           username: ${{ vars.DOCKER_REGISTRY_USER }}
           password: ${{ secrets.DOCKER_REGISTRY_TOKEN }}
 
-      - name: Install skopeo and jq
+      - name: Install skopeo, jq, and gh
         run: |
           sudo apt-get update
-          sudo apt-get install -y skopeo jq
+          sudo apt-get install -y skopeo jq gh
 
       # Resolve the source digest BEFORE copying to :prod. This avoids the
       # inspect-after-write race where reading back :prod could in theory hit a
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Adding `gh` to the apt install on the `[self-hosted, infra]` runner correctly fixes the `gh: command not found` failure in the `Create GitHub release` step. The new digest guard regex `^sha256:[a-f0-9]{64}$` (`.github/workflows/promote.yml:55`) is sound: it only enforces on non-empty input, so the empty/`:staging` path is unaffected. No correctness issues in the changed code.

Optional follow-up (non-blocking, out of scope):
- `.github/workflows/rollback.yml` runs on the same self-hosted runner and shells out to `gh` but still installs only `skopeo jq`, so it will hit the identical failure. A one-line fast-follow (`apt-get install -y skopeo jq gh`) keeps the two workflows from drifting.

Local checks: YAML parse + `bash -n` on the changed shell block passed; `git diff --check` passed; regex verified against real digests, empty, and uppercase input; CI checks reported passing. `actionlint` skipped (not installed).

PR #823, opened 2026-06-17
Validate positive response token limits

lloydmak99:
Problem

During the gpt-oss incident investigation on 2026-06-17, cloud-api and backend logs showed a high-volume burst of provider-side 400s for `openai/gpt-oss-120b`. The backend-side error pattern was invalid token limits reaching vLLM, for example requests where the effective `max_tokens` value was less than 1.

The immediate Datadog synthetic outage was a backend saturation/queueing event rather than a provider-pool eviction: the completion synthetic timed out before first byte while the gpt-oss vLLM replicas had deep queues and very high TTFT. Separately, the invalid-request storm is still a correctness and reliability problem because malformed client input should be rejected at the cloud-api edge, not forwarded to providers where it creates noisy provider errors and can confuse incident triage.

What was already covered on `main`

Chat Completions already rejects non-positive `max_tokens` on `main`:

- `max_tokens: 0`
- negative `max_tokens`
- invalid `n`

That existing validation prevents the exact Chat Completions negative-token request shape from reaching the provider path.

Remaining gap fixed here

The Responses API request models still accepted negative `max_output_tokens`. Downstream Responses processing maps `max_output_tokens` into provider `max_tokens`, so a negative Responses value could still reach inference providers and produce the same class of provider-side 400s.

This PR closes that remaining gap by requiring `max_output_tokens >= 1` anywhere a Responses request is validated.

Changes

- Reject `max_output_tokens < 1` in `crates/api/src/models.rs`.
- Reject `max_output_tokens < 1` in `crates/services/src/responses/models.rs`.
- Add API model tests covering:
  - `max_output_tokens: -1` rejected
  - `max_output_tokens: 0` rejected
  - unset `max_output_tokens` accepted
  - positive `max_output_tokens` accepted
- Add services model tests with the same validation coverage.
- Tighten the provider-pool 4xx regression test so a non-retryable client 4xx must leave provider failure counters empty.

Why this helps

This moves invalid Responses token limits from a provider-side failure to a cloud-api validation failure. That has three concrete benefits:

- Bad client input fails fast with a 4xx before consuming provider/proxy capacity.
- Providers are not asked to process requests that are known invalid.
- Provider-health accounting stays protected: client 4xx responses should not increment provider failure counters or contribute to provider demotion.

Incident-specific notes

The Datadog synthetic failure observed during the investigation was caused by gpt-oss queue saturation, not by cloud-api removing gpt-oss from the provider pool. The evidence checked at the time:

- Datadog synthetic failures were 60s timeouts before first byte.
- gpt-oss vLLM queues peaked across all three replicas.
- model-proxy still advertised the gpt-oss backend and circuit breakers were closed.
- cloud-api logs did not show provider exclusion / no-provider / all-providers-failed messages for the synthetic failure window.

This PR is therefore not a full capacity fix for that saturation event. It is the request-validation hardening that prevents one identified source of provider-side 400 noise from recurring through the Responses API, and it adds a regression guard for the provider-health behavior that matters during these incidents.

Validation

- `cargo test -p api test_create_response_request_rejects_non_positive_max_output_tokens`
- `cargo test -p services test_create_response_request_rejects_non_positive_max_output_tokens`
- `cargo test -p services test_4xx_error_does_not_retry`

CI note

The first CI Test Suite run failed in an existing e2e test:

- `admin_analytics::test_admin_metrics_unique_api_keys_tracking`

The failure was a 500 from the admin metrics endpoint because the test organization was not found in the shared e2e database. That path is unrelated to the files changed in this PR. I reran the failed job after confirming the unit/lint checks passed and the failure was outside this change surface.

Rollout Notes

- Behavior change is limited to Responses request validation.
- Requests with negative `max_output_tokens` now fail at validation instead of reaching inference providers.
- Unset and positive `max_output_tokens` continue to work.
- No provider routing logic changes are included.
- No retry behavior changes are included; the provider-pool change is test-only regression coverage.

Follow-up recommended outside this PR

- Add per-model/per-org concurrency or token-budget guardrails for gpt-oss to prevent queue saturation from taking synthetics past the 60s first-byte timeout.
- Alert directly on `vllm.num_requests.waiting` and gpt-oss TTFT before the synthetic starts failing.
- Add invalid-params metrics by model/org so request-shape floods are visible without relying on warning log volume.


crates/api/src/models.rs
```diff
@@ -2372,7 +2372,7 @@ impl CreateResponseRequest {
         }
 
         if let Some(max_tokens) = self.max_output_tokens {
-            if max_tokens == 0 {
+            if max_tokens < 1 {
```
lloydmak99, line 2375: **Medium - test coverage:** The new unit tests cover `CreateResponseRequest::validate()`, but the PR's stated protection is route-level: invalid `/v1/responses` payloads should fail before provider invocation. I did not find an e2e/endpoint test posting `max_output_tokens: -1` or `0` to `/v1/responses` and asserting the 400 `invalid_request_error` envelope while the mock provider remains untouched. Add a focused e2e test, for example `e2e_all::provider_errors::test_responses_rejects_non_positive_max_output_tokens_before_provider`.

lloydmak99 (commented):
Multi-agent code review completed for `fe6148dd8ce704aa1a3a74099a8a66500d89a07f`.

Reviewers run: security, bugs, performance/concurrency, tests, conventions.

Result: 1 Medium finding, 0 High/Critical findings. Security, bugs, performance/concurrency, and conventions returned clean.

CI was green before posting: claude-review, security_audit, Lint, and Test Suite.

The only finding is a route-level test coverage gap: unit validation exists, but there is no endpoint/e2e guard proving /v1/responses rejects non-positive max_output_tokens before invoking a provider.

PierreLeGuen (approved):
Correct and well-scoped change. Switching the validation from `== 0` to `< 1` correctly rejects negative `max_output_tokens` (the field is `Option<i64>`, signed), which previously slipped through to providers; both `CreateResponseRequest` copies (api + services) are updated consistently and `validate()` runs at the route boundary (`responses.rs:256`) before any provider dispatch.

No blocking issues found.

Checks: ran the new `test_create_response_request_rejects_non_positive_max_output_tokens` tests in both the `api` and `services` crates plus the provider-pool 4xx assertion (all passed), and `cargo clippy -p api -p services -- -D warnings` (clean).

Evrard-Nil (approved):
Reviewed commit fe6148dd. Build clean, all unit tests pass, CI green.

**What the change does**

Both `CreateResponseRequest` structs (in `crates/api/src/models.rs` and `crates/services/src/responses/models.rs`) previously guarded `max_output_tokens` with `== 0`, silently letting negative values reach vLLM. This PR tightens both to `< 1`. The route boundary (`responses.rs:256`) calls `validate()` on the services-crate struct before any provider dispatch, so the rejection is correct and early.

**Verification**

- `cargo check -p api -p services` — clean, zero warnings
- `cargo clippy -p api -p services -- -D warnings` — clean
- `cargo test -p api test_create_response_request_rejects_non_positive_max_output_tokens` — passes
- `cargo test -p services test_create_response_request_rejects_non_positive_max_output_tokens` — passes
- `cargo test -p services test_4xx_error_does_not_retry` — passes; the new assertion confirms a client 4xx doesn't increment `provider_failure_counts`
- Full `--lib` suites for both crates: 205 api / 397 services tests, 0 failures

**Code review findings**

The change is correct on all paths I checked:

- The route uses `services::responses::models::CreateResponseRequest` via the wildcard import (`use services::responses::models::*`), so the runtime struct and the updated `validate()` are the same object. The `api::models::CreateResponseRequest` is a separate, OpenAPI-schema-only struct not involved in request handling.
- `max_output_tokens` is typed `Option<i64>` in both structs; `< 1` correctly covers `i64::MIN` through `0`.
- `None` (unset) continues to pass — the guard is inside `if let Some(...)`.
- The error response is `400 invalid_request_error`, consistent with every other validation failure in this route.

**Minor observation (pre-existing, not a blocker)**

The Chat Completions path uses `ErrorResponse::with_param(...)` which populates the `param` field (e.g., `"param": "max_tokens"`) in the OpenAI error envelope. The Responses path's `validate()` returns a plain `String`, which the route wraps via `ErrorResponse::new(...)`, leaving `param: null`. This is a pre-existing pattern for all Responses validations (temperature, top_p, max_tool_calls are all the same) and is not introduced by this PR. Worth a follow-up to add `with_param` parity, but not a blocker here.

**Rollout safety**

Only affects requests with `max_output_tokens` set to 0 or negative through the Responses API. Such requests were already invalid; they now fail fast at the edge with a 400 instead of producing a provider-side 400. No valid request shape is affected.

Evrard-Nil (approved):
Re-review of b7b97a8e (new merge commit bringing branch up to main).

**What changed in the new commit**

The new push (b7b97a8e) is a merge-main commit — it pulls in upstream changes from main (#834, #803, #801, etc.) with no modification to the three files in scope: `crates/api/src/models.rs`, `crates/services/src/inference_provider_pool/mod.rs`, `crates/services/src/responses/models.rs`. The feature diff (fe6148dd) is unchanged.

**Re-verification on b7b97a8e**

- `cargo build` — clean, no warnings on changed crates
- `cargo test --lib --bins` — 406 passed, 0 failed (ran locally)
- `test_create_response_request_rejects_non_positive_max_output_tokens` in both api and services crates — passes
- `test_4xx_error_does_not_retry` — passes; `provider_failure_counts.is_empty()` assertion holds
- CI: Lint pass, Unit tests pass, Integration tests pass, E2E pending (flaky admin_analytics test unrelated to this change surface)

**Code correctness summary**

The core fix remains correct: `max_output_tokens` is `Option<i64>`, and the guard changing from `== 0` to `< 1` correctly rejects all non-positive values (zero and negative, including `i64::MIN`). `None` (unset) continues to be accepted — the guard is inside `if let Some(...)`. The route (`responses.rs:256`) calls `validate()` on the services-crate struct before any provider dispatch, so the rejection is correct and early. The `api::models::CreateResponseRequest` is OpenAPI-schema only and not involved in runtime request handling.

**Minor pre-existing observation (not a blocker, carry-forward from prior review)**

`max_tool_calls` still uses `== 0` rather than `< 1`. Since it is also `Option<i64>`, a negative `max_tool_calls` would pass validation today and reach the inference layer. This is pre-existing and out of scope for this PR, but worth a follow-up.

Approving — the change is production-ready for the Responses API path.

Evrard-Nil (approved):
Re-reviewing after 90d18385 (new push since approval at b7b97a8e).

**What changed:** Only `Cargo.lock` — `quinn-proto` bumped from 0.11.14 → 0.11.15. No feature code touched; no `Cargo.toml` changes.

**Why this is warranted:** This resolves **RUSTSEC-2026-0185** (GHSA-4w2j-m93h-cj5j, CVSS 7.5 HIGH) — remote memory exhaustion via unbounded out-of-order QUIC stream reassembly. An unauthenticated remote attacker can exhaust heap memory by sending stream fragments with gaps, forcing the receiver to retain unassembled data indefinitely. Patched version is `>= 0.11.15`. `quinn-proto` is a transitive dep pulled in by `reqwest` (HTTP/3 path).

**Assessment:** Lock-file-only security patch, no behavioral change to application code. Good to ship alongside the token-limit validation feature.

Approved.

PR #825, opened 2026-06-18
ci: skip tests on PRs without Rust changes

Evrard-Nil:
Adds a `paths` filter to `test.yml` so the lint + test suite only runs when Rust-relevant files change:

- `**/*.rs` — Rust source
- `**/Cargo.toml` / `**/Cargo.lock` — dependency manifests
- `crates/database/src/migrations/**` — SQL migrations
- `.github/workflows/test.yml` — the workflow itself

PRs that only touch docs, `.github/`, `config/`, `Dockerfile`, etc. will skip CI entirely.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,20 @@ name: Tests
 on:
   push:
     branches: [ main, develop ]
+    paths:
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - 'crates/database/src/migrations/**'
+      - '.github/workflows/test.yml'
   pull_request:
     branches: [ main, develop ]
+    paths:
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - 'crates/database/src/migrations/**'
+      - '.github/workflows/test.yml'
 
 env:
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
CI-only change adding `paths` filters to the `push`/`pull_request` triggers in `test.yml` so the suite is skipped on PRs that don't touch Rust-relevant files. The filters correctly cover Rust sources, Cargo manifests, SQL migrations, and the workflow itself; safe to merge.

- `.github/workflows/test.yml:6-21` — A `paths` filter only risks stuck PRs if "Tests" is a required status check (GitHub reports no status for skipped runs). Verified `main`'s branch protection has no `required_status_checks` today, so this is safe — just worth remembering if required checks are added later.
- `.github/workflows/test.yml:14` — The filter omits some Rust-affecting non-`.rs` files (e.g. `rust-toolchain.toml`, `.cargo/config.toml`, nextest config); changes to those would skip the suite. Pre-existing behavior and `build.yml` doesn't run cargo, so consider adding them if toolchain changes need test coverage on PRs.

Local checks: `gh pr diff 825`, full read of `test.yml`, YAML parse, `git diff --check`, branch-protection inspection, and confirmation that no other workflow depends on `test.yml`.

PR #826, opened 2026-06-18
fix(ci): disable GLM-5.2 thinking for OCR

Evrard-Nil:
GLM-5.2 is a reasoning model. It emits `reasoning_content` and returns `content: null` when `max_tokens` is consumed by reasoning, causing OCR to fail with:

> all file review(s) failed — check your LLM configuration and API key

Fix: set `llm.extra_body` to `{"chat_template_kwargs":{"enable_thinking":false}}`, which disables the reasoning path on the sglang backend (`--reasoning-parser glm45`).

Verified locally:
```json
{"chat_template_kwargs":{"enable_thinking":false}}
→ "content": "Hi! 👋 How can I help you today", "reasoning_tokens": 0
```


PierreLeGuen (approved):
Single-line CI change to the OpenCodeReview workflow: adds `ocr config set llm.extra_body '{"chat_template_kwargs":{"enable_thinking":false}}'` to the "Configure OCR" step (`.github/workflows/open-code-review.yml:57–62`) to disable GLM-5.2's reasoning path, which returns `content: null` when reasoning exhausts `max_tokens`. Looks correct and safe to merge.

- Scope is CI-only — no Rust, runtime, DB, or production paths touched; the value is a non-sensitive model parameter appropriately inlined.
- JSON is well-formed and the single-quoting preserves the inner double quotes correctly for the shell `run: |` block, so no YAML escaping concerns.

Checks: YAML parse passed; JSON well-formedness and bash single-quoting verified by inspection and a shell quoting test; OCR `1.3.19` accepted `llm.extra_body` and stored it as an object. No Rust/test suite applicable to a single-line workflow change.

PR #829, opened 2026-06-18
docs: fix provider_type example values in revenue density comments

Evrard-Nil:
## Summary
Two docstrings cited `"nearai"` as an example `provider_type` value, but the column stores `'vllm'` (NEAR AI TEE-hosted), `'external'`, and `'chutes'` (per V0042/V0056/V0060). No logic change — comments only.

Caught by Pierre's review of #819.

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -3630,7 +3630,7 @@ pub async fn get_performance_timeseries(
 pub struct RevenueDensityParams {
     pub start: Option<String>,
     pub end: Option<String>,
-    /// Optional provider type filter (e.g. "nearai", "external").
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):
Documentation-only change correcting the `provider_type` example values in two revenue-density doc-comments (`admin.rs:3633`, `analytics.rs:472`). The new examples (`vllm`, `external`, `chutes`) match the real provider types in the codebase; the old `nearai` value was incorrect (`nearai` is the default `owned_by`, not a `provider_type`). No behavioral changes.

Local checks: `git diff --check`, `cargo fmt --check`, and `cargo check` (19 crates, clean) all passed; PR CI (Lint, code-review, security_audit) green.

PR #831, opened 2026-06-19
ci: speed up test workflow with parallel jobs, sccache, and mold

Evrard-Nil:
## Summary

Splits the single sequential `test` job into **five parallel jobs** and adds **sccache** + the **mold linker** to cut CI wall-clock time. Targets the cloud-api test workflow, which already runs on our self-hosted runner (`gpu11`) but was slow because everything ran sequentially in one job.

## Does it already run on our runners?

**Yes** — both the old `lint` and `test` jobs already used `runs-on: [self-hosted, infra]`. This PR keeps that. The slowness comes from (a) sequential execution, (b) no object-level compile cache, (c) slow GNU `ld` linking of the large test binaries, and (d) a dead `ALTER SYSTEM + docker restart` step.

## Results

| Job | Time |
|-----|------|
| Lint | 3m59s |
| Unit tests | 2m50s |
| Integration tests | 2m58s |
| E2E tests | 5m58s |
| **Total wall-clock (parallel)** | **~6m** |

Baseline (old workflow on main, sequential): **~10-11m**. This is **~40% faster on a cold cache**; sccache will improve further on warm runs.

### Job split (parallel execution)
Old: `lint` ‖ `test` (where `test` ran unit → e2e → vLLM → release sequentially)

New: `lint` ‖ `unit-test` ‖ `integration-test` ‖ `e2e-test` ‖ `build-release` (all parallel)

- **`unit-test`** (`cargo nextest run --lib --bins`): no PostgreSQL needed — in-crate `#[test]`s use mocks.
- **`integration-test`** (`cargo nextest run --test integration_tests`): no PostgreSQL needed — `MockProvider` is used by default (`USE_REAL_VLLM` is not set).
- **`e2e-test`** (`cargo nextest run --test e2e_all`): the only job with the PostgreSQL service container.
- **`build-release`** (`cargo build --release`): main-push only, split out so it doesn't block tests and doesn't pull in PostgreSQL (there is no `build.rs`).

### sccache (`RUSTC_WRAPPER=sccache`)
Caches compiled objects at the object-file granularity, complementing `swatinem/rust-cache` (which caches `target/`). Survives `Cargo.lock` bumps and feature-flag changes better. On a self-hosted runner the default cache dir (`~/.cache/sccache`) persists on disk across runs, so no `actions/cache` step is needed.

### mold linker (`RUSTFLAGS="-C link-arg=-fuse-ld=mold"`)
2-5× faster linking of the ~60MB test binary. Installed via GitHub release binary (both `mold` and `ld.mold`) — not apt — to avoid dpkg lock contention between parallel jobs. Configured via `RUSTFLAGS` env var rather than in `.cargo/config.toml` so:
- The **Docker reproducible build** (which copies `.cargo/config.toml` and does NOT set `RUSTFLAGS`) is unaffected and keeps using `ld`.
- `RUSTFLAGS` env replaces `target.cfg.rustflags` from config (cargo treats them as mutually exclusive), which is fine for CI — the dropped flags (`--remap-path-prefix`, `--build-id=none`, `--hash-style=gnu`, `--no-undefined`) are reproducibility-only and irrelevant for tests. `-C debuginfo=0` is kept for faster test compiles.

**Mold integration gotcha**: `gcc -fuse-ld=mold` looks for `ld.mold` on PATH (not `mold`). Rustc also passes its own `-fuse-ld=lld` + a `-B` path to its bundled lld; the last `-fuse-ld` flag wins, so `-fuse-ld=mold` takes effect. Installing `ld.mold` from the mold tarball resolves the `collect2: cannot find 'ld'` error.

### Cleanup
- **Removed the `ALTER SYSTEM SET max_connections=150 + docker restart` step** (was test.yml:74-87). `.config/nextest.toml` already caps the `e2e-db` test group at `max-threads = 16` (16 × 4 pool conns = 64 < PG default `max_connections = 100`), so the override was dead code adding ~15-30s and a failure point.
- **Switched vLLM integration tests** from `cargo test --test integration_tests -- --nocapture` to `cargo nextest run --test integration_tests` for consistent tooling.
- **Stripped vestigial env vars** from the release build (no `build.rs` → `cargo build --release` needs none of `DATABASE_*` / `MODEL_DISCOVERY_*` / `AUTH_*`).

### Composite action (`.github/actions/setup-rust-ci`)
New composite action centralizes: Rust toolchain, cargo-nextest, sccache, mold, and `swatinem/rust-cache` (with per-job `cache-key` for isolation). Avoids step duplication across 5 jobs.

## Verification
- [x] All 5 jobs pass on this PR (Lint 3m59s, Unit 2m50s, Integration 2m58s, E2E 5m58s)
- [x] YAML validated (5 jobs, correct `runs-on`, only `e2e-test` has `postgres` service, only `build-release` is main-gated)
- [x] Confirmed no `build.rs` anywhere in the workspace (`cargo build --release` needs no env vars)
- [x] Confirmed `integration_tests` target lives in `crates/inference_providers/tests/` and uses `MockProvider` by default
- [x] Confirmed `.config/nextest.toml` already caps e2e concurrency at 16 threads (justifies removing the PG restart step)
- [x] mold v2.41.0 release verified to exist and tarball structure confirmed (`bin/mold` + `bin/ld.mold`)

## Follow-up issues (not in this PR)
- **e2e sharding** with `nextest --partition hash i/N` (~3× e2e speedup) — needs a separate runner label or matrix to be effective.
- **Dedicated CI VM** to remove `gpu11` prod-host contention (the root cause of CI-duration variability).
- **Pre-built CI Docker image** with deps compiled (`cargo test --no-run`), refreshed on `Cargo.lock` changes.

## Follow-up issues
- nearai/cloud-api#832 — e2e sharding with nextest --partition hash (~3× e2e speedup)
- nearai/cloud-api#833 — pre-build CI Docker image with deps compiled
- nearai/infra#184 — dedicated CI VM to remove gpu11 prod-host contention

### DIFF
--- a/.github/actions/setup-rust-ci/action.yml
+++ b/.github/actions/setup-rust-ci/action.yml
@@ -52,9 +52,11 @@ runs:
         TARBALL="mold-${MOLD_VERSION}-${TARBALL_ARCH}-linux.tar.gz"
         URL="https://github.com/rui314/mold/releases/download/v${MOLD_VERSION}/${TARBALL}"
         curl -fsSL "$URL" -o "/tmp/${TARBALL}"
-        sudo tar -xzf "/tmp/${TARBALL}" -C /usr/local --strip-components=2 "mold-${MOLD_VERSION}-${TARBALL_ARCH}-linux/bin/mold"
+        # Extract bin/mold from the tarball directly into /usr/local/bin.
+        sudo tar -xzf "/tmp/${TARBALL}" -C /usr/local/bin --strip-components=2 "mold-${MOLD_VERSION}-${TARBALL_ARCH}-linux/bin/mold"
         rm -f "/tmp/${TARBALL}"
         echo "Installed: $(mold --version)"
+        command -v mold
 
     - name: Cache Rust dependencies
       uses: swatinem/rust-cache@v2
--- a/.github/actions/setup-rust-ci/action.yml
+++ b/.github/actions/setup-rust-ci/action.yml
@@ -38,7 +38,13 @@ runs:
       run: |
         # Install mold via release binary to avoid dpkg lock contention when
         # multiple jobs run in parallel on the same self-hosted runner.
-        if command -v mold >/dev/null 2>&1; then
+        #
+        # We install both `mold` and `ld.mold`. Rustc invokes the cc driver
+        # with -fuse-ld=mold (set via RUSTFLAGS in test.yml), which makes cc
+        # look for `ld.mold` on PATH. Rustc also passes its own
+        # -fuse-ld=lld + a -B path to its bundled lld; the last -fuse-ld
+        # flag wins, so -fuse-ld=mold takes effect and cc finds ld.mold.
+        if command -v mold >/dev/null 2>&1 && [ -x /usr/local/bin/ld.mold ]; then
           echo "mold already installed: $(mold --version)"
           exit 0
         fi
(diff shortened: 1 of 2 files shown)


PierreLeGuen (approved):
CI-only change that splits the test workflow into parallel jobs and adds sccache + mold via a `setup-rust-ci` composite action. All five jobs (lint, unit, integration, e2e, build-release) pass on the head; prior review feedback (mold extraction path) was addressed in `c9f94d7`. Looks good to merge.

- `.github/actions/setup-rust-ci/action.yml`: on a cold runner, all parallel jobs hit the `command -v mold` guard simultaneously and download/extract to the same `/tmp` path and `/usr/local/bin`, which could intermittently corrupt the tarball or race the extraction. No production impact, but may cause spurious CI failures on first run after a runner reimage.
- `.github/workflows/test.yml` (integration-test job): `VLLM_BASE_URL`/`VLLM_API_KEY` are passed without `environment: Cloud API test env`, so if those secrets are environment-scoped they resolve to empty. Harmless today since `integration_tests.rs:21` defaults to `MockProvider` unless `USE_REAL_VLLM` is set — only matters if you later run real-vLLM tests here.

Checks: YAML parse of workflow + action files; `gh pr checks 831` (all green); verified test targets `e2e_all` (`crates/api/Cargo.toml:82`) and `integration_tests` exist; confirmed `-C debuginfo=0` already in `.cargo/config.toml`; confirmed `nextest.toml` `e2e-db` cap of 16 threads and that its filter excludes integration tests.

PR #838, opened 2026-06-23
feat: adaptive load balancing (TTFT EMA + prefix-hash cache routing + context-length routing)

Evrard-Nil:
## Summary

Adds three complementary routing signals to `InferenceProviderPool` that work together as a composite sort key `(context_overflow, hard_demoted, latency_demoted, tier_rank)`:

- **TTFT-based soft demotion**: track per-provider TTFT via exponential moving average (α=0.5 warmup / α=0.1 stable after 8 samples). Providers 2× slower than the fastest peer AND above the 500ms absolute floor get `latency_demoted=1`, placing them after fast peers while keeping them as fallbacks. Feedback flows via a `ProviderLatencyReporter` closure created in `chat_completion_stream` and called from `InterceptStream::Drop`.

- **Prefix-hash cache routing**: when a chat request carries messages, the first 2 messages are hashed. The hash selects deterministically within the leading provider group instead of advancing the round-robin counter — same-prefix conversations always route to the same backend, maximising KV-cache hits. Falls back to round-robin when no messages are present (image, audio, embed paths).

- **Context-length routing**: `context_length` is now returned from `fetch_inference_url_models` and stored in `ProviderLatencyState.max_context_tokens` per provider. Providers whose configured max < estimated request tokens (`chars/4`) are sorted last (`context_overflow=1`). Prepares for a future where `glm-5-2-i0` and `glm-5-2-i1` have different context windows.

## Key files

| File | Change |
|------|--------|
| `crates/services/src/inference_provider_pool/mod.rs` | New `ProviderLatencyState`, `ChatRoutingHints`, `ProviderLatencyReporter`; updated sort key; TTFT reporter creation; `load_inference_url_models` context_length plumbing |
| `crates/services/src/completions/mod.rs` | `latency_reporter` field on `InterceptStream`; `compute_prefix_hash` + `estimate_input_tokens` helpers; routing hints computed and threaded to pool |
| `crates/database/src/repositories/model.rs` | `get_inference_url_models` now includes `context_length` |
| `crates/api/src/routes/admin.rs` | Admin PATCH passes `context_length` through to pool |

## Test plan

- [ ] `cargo check --workspace` passes (verified locally — zero errors, zero new warnings)
- [ ] `cargo test --lib --bins` — unit tests pass
- [ ] Staging deploy: check Datadog `inference.latency.ttft` metric tagged by model to verify TTFT is still being recorded correctly
- [ ] With two backends for the same model (e.g. `glm-5-2-i0`, `glm-5-2-i1`): send the same system-prompt across multiple conversations and verify via Datadog `inference.provider` tag that they consistently land on the same backend
- [ ] After simulating a slow backend (manually delay one instance), verify it gets `latency_demoted` in routing logs after ~8 requests

### DIFF
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -496,7 +496,7 @@ pub async fn batch_upsert_models(
 
     // Register inference_url models (our own vLLM/SGLang backends)
     // Only for active, non-external models with an inference_url set
-    let inference_url_models: Vec<(String, String)> = batch_request
+    let inference_url_models: Vec<(String, String, Option<u32>)> = batch_request
         .iter()
         .filter_map(|(model_name, request)| {
             let is_active = request.is_active != Some(false);
@@ -505,7 +505,7 @@ pub async fn batch_upsert_models(
                 request
                     .inference_url
                     .clone()
-                    .map(|url| (model_name.clone(), url))
+                    .map(|url| (model_name.clone(), url, request.context_length.map(|c| c as u32)))
             } else {
                 None
             }
--- a/crates/api/src/routes/admin.rs
+++ b/crates/api/src/routes/admin.rs
@@ -502,10 +502,13 @@ pub async fn batch_upsert_models(
             let is_active = request.is_active != Some(false);
             let is_external = request.provider_type.as_deref() == Some("external");
             if is_active && !is_external {
-                request
-                    .inference_url
-                    .clone()
-                    .map(|url| (model_name.clone(), url, request.context_length.map(|c| c as u32)))
+                request.inference_url.clone().map(|url| {
+                    (
+                        model_name.clone(),
+                        url,
+                        request.context_length.map(|c| c as u32),
+                    )
+                })
             } else {
                 None
             }
--- a/crates/database/src/repositories/model.rs
+++ b/crates/database/src/repositories/model.rs
@@ -1444,8 +1444,8 @@ impl ModelRepository {
     }
 
     /// Get all active models with inference_url set.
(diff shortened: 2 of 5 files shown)


PierreLeGuen (approved):
Adaptive load balancing (TTFT EMA + prefix-hash cache routing + context-length routing) looks solid — compiles clean, tests pass, and the routing changes degrade safely on the non-chat paths (all callers pass `ChatRoutingHints::default()`, and discovery/external providers correctly get `max_context_tokens = None`).

- `inference_provider_pool/mod.rs:362` (sort key) + `repositories/model.rs:1477`: `context_overflow` is the top sort dimension and the DB sync sets `max_context_tokens = Some(context_length)` where the column is `NOT NULL DEFAULT 0`. An own (non-external) backend with stored `context_length = 0` would be flagged `context_overflow=1` and sort *behind* an external peer of the same model, inverting the attestation/tier preference. Data-dependent and unlikely (admin create requires non-zero), but consider treating `Some(0)` as unconfigured/unlimited.
- `inference_provider_pool/mod.rs:408`: prefix-hash routing no longer advances the round-robin counter, so requests sharing the same first two messages deterministically pile onto one backend. This is the intended KV-cache behavior with latency-demotion as the safety valve — worth confirming on staging that one hot prefix can't starve peers under real traffic.
- Lint/CI: `cargo fmt --all --check` fails with rustfmt diffs in the changed files; run `cargo fmt` before merge to get Lint green.

Checks run: `cargo check -p api -p database -p services` (clean), `cargo test --lib` (passed, incl. `inference_provider_pool`), `cargo fmt --all --check` (failed — formatting diffs).

PR #839, opened 2026-06-23
Track fallback provider usage attribution

PierreLeGuen:
## Summary
- persist served provider attribution for usage rows, including provider tier/type and fallback status
- add admin analytics coverage for Chutes/fallback usage in platform metrics and model revenue
- propagate served attribution through completions and usage recording while keeping served-request metrics bounded

## Verification
- `cargo test -p services provider_requests_metric -- --nocapture`
- Previously run during implementation: admin provider-attribution e2e tests, usage attribution tests, `cargo fmt --all -- --check`, and services clippy

## Stack
- Base PR for durable fallback/Chutes usage attribution
- Follow-up provider-attempt telemetry: #840


Evrard-Nil (approved):
Reviewed commit 63893a73 against main. Build: `cargo check --workspace` clean. All new unit tests pass (`provider_attribution`, `provider_requests_metric`, `demoted_near_chutes_first_success_is_still_fallback_attribution`).

## Summary

This PR correctly tracks which provider tier (near / attested_3p / non_attested) and type (vllm / external / chutes) served each request, and whether it was a fallback. Attribution flows from `retry_with_fallback_caps` → `ServedProviderResult` → `AttributedChatCompletion{Stream}` → `InterceptStream` / non-streaming biller → `RecordUsageServiceRequest` → `RecordUsageDbRequest` → `organization_usage_log`. The chain is solid.

## What was checked

### Attribution correctness
- `served_provider_attribution()` is called at the single exit point of the retry loop (the `return Ok(ServedProviderResult {...})` branch), which is the only point that knows both the serving provider and the attempt index. This is the right place.
- The `has_near_primary` flag computed before the retry loop correctly handles NEAR demotion: when NEAR's failure counter puts it behind Chutes, `attempt==0` with a Chutes provider now still produces `served_via_fallback: true`. The new test `demoted_near_chutes_first_success_is_still_fallback_attribution` explicitly exercises this and passes.
- `ProviderAttribution::default()` (all `None` / `false`) is used for image generation, image editing, audio transcription, embeddings, rerank, score, and privacy operations — all paths that don't go through the chat-completion attribution chain. This is intentional and safe: Chutes returns `UNSUPPORTED` for all these operations, so a Chutes fallback for image/audio is impossible.
- `InterceptStream` carries the attribution from construction through `Drop`, so partial/interrupted streams still record the correct provider.
- `failed_providers_do_not_record_served_attribution` confirms that terminal failures (all providers error) record no usage row at all, which is correct.

### Migration (V0063)
- All three new columns are `nullable` (`served_provider_tier TEXT`, `served_provider_type TEXT`) or have a `NOT NULL DEFAULT false` (`served_via_fallback`). No `NOT NULL` without default — safe to add to a live table with existing rows. Pre-existing rows will have `NULL` for tier/type and `false` for the fallback flag, which is the correct "unknown / pre-instrumentation" interpretation.
- CHECK constraints use `NOT VALID`, which adds the constraint without scanning existing rows — avoids a table lock on a large production table. New inserts will be validated; old rows are left alone. Good pattern.
- No down migration, consistent with the rest of the codebase.

### Metrics cardinality
The `METRIC_PROVIDER_REQUESTS` counter uses four labels: `model:`, `provider_tier:`, `fallback:`, `operation:`. All are bounded: model names come from the provider registry, tier is a 3-variant enum, fallback is bool, operation is a string literal from call sites. No user IDs, request IDs, or free-form strings as label values. Cardinality is well-controlled.

### Admin analytics
Both `get_platform_metrics` and `get_model_revenue` carry `Extension(_admin_user): Extension<AdminUser>` — the standard admin auth guard. New analytics endpoints are not reachable without admin auth.

`UsageHistoryEntryResponse` (customer-facing) does not expose `served_provider_tier`, `served_provider_type`, or `served_via_fallback`. The schema compatibility test `admin_provider_attribution_openapi_schema_exports` asserts this explicitly.

### SQL
- `COALESCE(ul.served_provider_type, m.provider_type)` in the `provider_type` WHERE filter correctly falls back to the model's static `provider_type` for pre-migration rows (where `served_provider_type` is NULL). This means `?provider_type=chutes` will correctly return zero results for historical rows (they'll have NULL → falls back to `m.provider_type`, which won't be 'chutes' for any existing model). This is correct.
- `load_model_provider_breakdowns` passes `$1..$5` from the shared `where_clause` plus `$6` for the `model_names` array. Parameter positions align correctly.
- All SQL uses parameterized queries; no user input is interpolated. The sort column is from a fixed allowlist enum.

### No credentials
Scanned the diff — no hardcoded tokens, API keys, or secrets.

## Minor observations (non-blocking)

1. **Image/audio attribution gap is silent**: `image_generation`, `image_edit`, and `audio_transcription` extract `served.value` and `served.provider` from `ServedProviderResult` but silently drop `served.provider_attribution`. The usage request is then built with `ProviderAttribution::default()`. This is currently safe (Chutes doesn't serve these types), but there's no compile-time protection against a future Chutes image provider being added and silently producing unattributed rows. Consider a `_` binding comment or extracting attribution even if it's unused, as documentation of intent.

2. **`provider_type` on `ModelRevenueEntry` is `MAX(m.provider_type)`**: This is pre-existing and inherited from the model-level aggregation, not introduced here. But the new `served_provider_breakdown` provides exact per-request type data, so this quirk is now less load-bearing.

3. **Merge conflict**: The PR currently shows `CONFLICTING` against main. Needs rebase before merge.

Approved — the attribution chain is correct, migration is safe, metrics are bounded, and auth is properly gated on all new admin surfaces.

Evrard-Nil (approved):
Re-reviewed after rebase (HEAD bd40b06d): diff vs main is 31 files, +2488/-99. The original commit (63893a73) was +2326/-72 — the delta is main diverging under the PR during the merge, not new feature code. No merge conflict markers, no unintended logic changes. Build (`cargo check --workspace`) passes cleanly. Both provider_requests_metric unit tests pass. Rebase was a clean forward-merge of main only. Previous concern still stands: image/audio ops use `ProviderAttribution::default()` (safe today, worth revisiting if Chutes image/audio go live). Ready to merge.

Evrard-Nil (approved):
Re-reviewed after rebase: diff vs main unchanged (31 files, ~+2488/-99), no merge conflict markers, `cargo check --workspace` and `cargo test -p services provider_requests_metric` both pass (2/2 tests ok). Ready to merge.

Evrard-Nil (approved):
Re-reviewed at ec0460ed (HEAD; dc5c4f8d + ec0460ed since bd40b06d).

Changes since bd40b06d (3 files, +33/-12 lines, test-only):
- `admin_analytics.rs`: replaced fixed 1 s sleep with a 5 s polling loop (100 ms cadence) that breaks as soon as the test org appears in /v1/admin/platform/org-revenue, with time-bounded query parameters (start/end + search + sort=requests). Eliminates the flaky timing window that caused intermittent failures.
- `admin_provider_attribution_model_revenue.rs`: one-liner swap of `setup_test_server()` → `setup_test_server_with_mock_web_search()` so the `admin_model_revenue_rejects_invalid_provider_type` test uses the same harness as its siblings.
- `admin_provider_attribution_support.rs`: extracted the verbose return type `(DateTime<Utc>, DateTime<Utc>)` into a module-local type alias `ProviderUsageWindow` for readability.

No production-code changes. Build clean (`cargo check --workspace`), both `provider_requests_metric` unit tests pass, no merge-conflict markers.

PR #840, opened 2026-06-23
Emit provider attempt metrics

PierreLeGuen:
## Summary
- add `cloud_api.provider.attempts` OTLP/Datadog counter for each provider attempt
- tag attempts with bounded model, provider tier/type, fallback, operation, result, retry decision, retry round, and attempt index labels
- cover successes, failed retries, all-provider retry exhaustion, and non-retryable short-circuit failures without emitting raw provider error text

## Verification
- `cargo test -p services provider_attempts_metric_records -- --nocapture`
- `cargo clippy -p services --all-targets --all-features -- -D warnings`

## Stack
- Stacks on #839

### DIFF
--- a/crates/services/src/inference_provider_pool/mod.rs
+++ b/crates/services/src/inference_provider_pool/mod.rs
@@ -31,6 +31,86 @@ pub use provider_attribution::{
 
 type InferenceProviderTrait = dyn InferenceProvider + Send + Sync;
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum ProviderAttemptResult {
+    Success,
+    Failed,
+    ShortCircuited,
+}
+
+struct ProviderAttemptMetric<'a> {
+    model_id: &'a str,
+    provider_tier: inference_providers::ProviderTier,
+    provider_source: inference_providers::ProviderSource,
+    is_fallback: bool,
+    operation_name: &'a str,
+    attempt_result: ProviderAttemptResult,
+    retry_decision: &'static str,
+    retry_round: usize,
+    attempt_index: usize,
+}
+
+const fn provider_tier_metric_tag(tier: inference_providers::ProviderTier) -> &'static str {
+    match tier {
+        inference_providers::ProviderTier::Near => "provider_tier:near",
+        inference_providers::ProviderTier::Attested3p => "provider_tier:attested_3p",
+        inference_providers::ProviderTier::NonAttested => "provider_tier:non_attested",
+    }
+}
+
+const fn provider_source_metric_tag(source: inference_providers::ProviderSource) -> &'static str {
+    match source {
+        inference_providers::ProviderSource::Vllm => "provider_type:vllm",
+        inference_providers::ProviderSource::External => "provider_type:external",
+        inference_providers::ProviderSource::Chutes => "provider_type:chutes",
+    }
+}
+
+const fn fallback_metric_tag(is_fallback: bool) -> &'static str {
+    if is_fallback {
+        "fallback:true"
+    } else {
+        "fallback:false"
+    }
+}
+
+const fn attempt_result_metric_tag(result: ProviderAttemptResult) -> &'static str {
+    match result {
+        ProviderAttemptResult::Success => "attempt_result:success",
+        ProviderAttemptResult::Failed => "attempt_result:failed",
+        ProviderAttemptResult::ShortCircuited => "attempt_result:short_circuited",
+    }
+}
+
+fn record_provider_attempt(
+    metrics: Option<&Arc<dyn crate::metrics::MetricsServiceTrait>>,
+    attempt: ProviderAttemptMetric<'_>,
+) {
+    if let Some(metrics) = metrics {
+        let model_tag = format!("model:{}", attempt.model_id);
+        let operation_tag = format!("operation:{}", attempt.operation_name);
+        let retry_decision_tag = format!("retry_decision:{}", attempt.retry_decision);
+        let retry_round_tag = format!("retry_round:{}", attempt.retry_round);
+        let attempt_index_tag = format!("attempt_index:{}", attempt.attempt_index);
+        let tag_refs = [
+            model_tag.as_str(),
+            provider_tier_metric_tag(attempt.provider_tier),
+            provider_source_metric_tag(attempt.provider_source),
+            fallback_metric_tag(attempt.is_fallback),
+            operation_tag.as_str(),
+            attempt_result_metric_tag(attempt.attempt_result),
+            retry_decision_tag.as_str(),
+            retry_round_tag.as_str(),
+            attempt_index_tag.as_str(),
+        ];
+        metrics.record_count(
+            crate::metrics::consts::METRIC_PROVIDER_ATTEMPTS,
+            1,
+            &tag_refs,
+        );
+    }
+}
+
 /// Upper bound on leading SSE control events (keepalive comments, blank
 /// lines — chunk-less `SSEEvent`s) consumed while peeking for the first
 /// parsed chunk to establish sticky-routing. Real upstreams emit zero before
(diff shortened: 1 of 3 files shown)


Evrard-Nil (approved):
## Review: Emit provider attempt metrics

Build: `cargo check -p services` clean. All 4 new tests pass:
- `provider_attempts_metric_records_success_attempt`
- `provider_attempts_metric_records_failed_retry_and_chutes_success`
- `provider_attempts_metric_records_all_failed_retries`
- `provider_attempts_metric_records_short_circuit_failures`

### Cardinality — looks safe

Every tag is bounded at the type level or by architectural limits:

| Tag | Bound | Analysis |
|-----|-------|----------|
| `model` | catalog (DB) | `chat_params.model` is overwritten with the canonical name from `resolve_and_get_model()` before the provider pool sees it. An unknown model gets a 404 before the metric path. Matches the pattern already used by `METRIC_PROVIDER_REQUESTS`. |
| `provider_tier` | 3 variants (`near`, `attested_3p`, `non_attested`) | `ProviderTier::as_str()` returns `&'static str` from a closed enum. |
| `provider_type` | 3 variants (`vllm`, `external`, `chutes`) | `ProviderSource::as_str()` — same pattern. |
| `fallback` | bool → 2 values | |
| `operation` | 5 string literals (`chat_completion`, `chat_completion_stream`, `image_generation`, `audio_transcription`, `image_edit`) | All call sites pass compile-time string literals. |
| `attempt_result` | 3 variants (`success`, `failed`, `short_circuited`) | Closed enum + `const fn as_str`. |
| `retry_decision` | 12 `&'static str` values from `classify_retry_decision()` + `"none"`, `"non_retryable_client_4xx"` | All production callers pass `&'static str`; the struct field is typed as `&'a str` which is slightly looser than necessary, but all actual call sites are literals or `classify_retry_decision()` output — no runtime strings leak in. |
| `retry_round` | 0–3 (`MAX_RETRIES = 3`) | Bounded by the constant. |
| `attempt_index` | 1–N providers per model | In production, at most 2 providers per model (one NEAR + optional Chutes secondary via `register_pinned_secondary_provider`). |

No user IDs, request IDs, API keys, or raw error messages appear in any tag. The test `assert_no_forbidden_attempt_labels` covers the key sensitive fragments. The existing test also asserts `!tag.contains("overloaded")` to prevent raw error message leakage.

### Metric naming

`cloud_api.provider.attempts` is consistent with the existing `cloud_api.provider.requests` naming convention and the `cloud_api.<noun>.<metric>` pattern throughout `consts.rs`.

### Counter semantics

`record_count` → `u64_counter` (monotonic) in `OtlpMetricsService`. Correct for Datadog.

### Coverage — all outcome paths instrumented

- ✅ Success (attempt wins)
- ✅ Non-retryable 4xx short-circuit (client error)
- ✅ Non-retryable client-media short-circuit (bad image URL)
- ✅ Retryable failure (5xx, 429, connection errors) — `Failed` + correct `retry_decision`
- ✅ All providers exhausted — each individual attempt already tagged `Failed`; no extra summary counter needed, consistent with the existing design

One uninstrumented code path worth noting (not a blocker): when `is_model_not_found_error` is true (a 4xx meaning "this provider doesn't serve this model"), the code falls through the `non_retryable_client_4xx` guard and eventually emits a `Failed` record with `retry_decision: "non_retryable_http"`. That tagging is correct (the provider failed for this model), though it's slightly surprising that a "model not found on this provider" case gets labeled `non_retryable_http` rather than something like `non_retryable_model_not_found`. Not a cardinality issue — just worth noting for future dashboard builders.

### Minor nit

`ProviderAttemptMetric.retry_decision` is typed `&'a str` rather than `&'static str`. In practice every caller passes a static literal or the return value of `classify_retry_decision()` (which is `-> &'static str`), so there's no runtime-string risk. Tightening it to `&'static str` would make that invariant compiler-enforced, but it's not required.

### No credentials in code

Checked — none found.

---

LGTM. The cardinality design is sound for Datadog, the coverage is thorough, and the tests verify both the happy and all error paths.

Evrard-Nil (approved):
Build: cargo check -p services clean. All 4 new tests pass.

**Cardinality — looks safe**

Every tag is bounded:
- model: catalog-resolved (resolve_and_get_model runs before the provider pool; unknown models 404 before this path)
- provider_tier: 3 variants, &'static str from closed enum
- provider_type: 3 variants, &'static str from closed enum
- fallback: bool (2 values)
- operation: 5 compile-time string literals (chat_completion, chat_completion_stream, image_generation, audio_transcription, image_edit)
- attempt_result: 3 variants, closed enum
- retry_decision: 12 &'static str values from classify_retry_decision() + "none"/"non_retryable_client_4xx" literals; struct field is &'a str (slightly loose) but all callers pass statics — no runtime strings leak
- retry_round: 0–3 (bounded by MAX_RETRIES = 3)
- attempt_index: 1–N providers, production max is 2 (one NEAR + optional Chutes secondary)

No user IDs, request IDs, API keys, or raw error messages in tags. The test asserts against sensitive fragments (org_id, token, authorization, prompt, response, url, body, etc.) and also checks no raw error message ("overloaded") appears.

**Metric naming**: cloud_api.provider.attempts matches the existing cloud_api.provider.requests convention.

**Counter semantics**: record_count -> u64_counter (monotonic). Correct for Datadog.

**Coverage — all outcome paths covered**: success, non-retryable 4xx short-circuit, client-media short-circuit, retryable failure (Failed + retry_decision), all-providers-exhausted.

One minor observation (not a blocker): when is_model_not_found_error is true, the attempt falls through to the general Failed + retry_decision: "non_retryable_http" record. Technically correct (that provider failed for this model) but a future dashboard reader might find "non_retryable_http" surprising for a model-not-found case. Could be a distinct retry_decision value but not required now.

**Minor nit**: ProviderAttemptMetric.retry_decision is &'a str rather than &'static str. All callers pass statics in practice; tightening the type would make that compiler-enforced but is not a cardinality risk.

No credentials in code.

LGTM. Cardinality design is sound for Datadog, coverage is thorough, tests verify all paths.

Evrard-Nil (approved):
Re-reviewing at HEAD da02ec4b (two commits past the stated 760149eb — the branch picked up one more merge from codex/fallback-chutes-usage-metrics plus ec0460ed "Stabilize org revenue e2e usage polling").

**What changed since 760149eb (the prior stated SHA):**
Only `crates/api/tests/e2e_all/admin_analytics.rs` — `test_admin_platform_org_revenue` replaces a fixed 1 s sleep with a 5 s deadline-polling loop that also passes `start`/`end`/`search`/`sort` query params. Good fix for a flaky e2e test.

**PR 840 scope (vs codex/fallback-chutes-usage-metrics base):**
Three files: `inference_provider_pool/mod.rs` (+601 lines), `metrics/consts.rs` (+2), `metrics/mod.rs` (+3). Adds `cloud_api.provider.attempts` counter with 9 tags.

**Cargo check:** clean.
**Unit tests:** all 4 pass (`provider_attempts_metric_records_*`).
**Merge conflicts:** none.

**One nit (not a blocker):** The `is_fallback` computation on the error path (lines 2414–2416) is an inline approximation:
```rust
let is_fallback = attempt > 0
    || (has_near_primary && tier != ProviderTier::Near);
```
`served_via_fallback` (used on the success path) has a third case: `!has_near_primary && tier != Near && fallback_pinned_models.contains(model_id)`. For a fallback-pinned model with no NEAR primary, a failed attempt on the pinned fallback provider would emit `fallback:false` instead of `fallback:true`. In practice these are rare models (only `register_pinned_secondary_provider` paths), but the tag would be inconsistent with what a success on the same provider would emit. Could call `self.served_via_fallback(model_id, attempt, provider, has_near_primary)` here the same way the success branch does — would make them identical without the async concern.

`METRIC_PROVIDER_ATTEMPTS` in `consts.rs` is the only exported constant without a doc comment. Minor.

Both are nits. The cardinality analysis, coverage, privacy checks (no raw errors / PII in tags), and counter semantics are all correct. LGTM.

Issue #841, opened 2026-06-23
Notify admins when backend-derived model context length decreases


PierreLeGuen:
## Summary

Follow-up to #837: now that `/v1/models` should derive model context length from backend metadata and publish the maximum observed value across backends, admins need a notification when that observed maximum decreases.

## Problem

A lower advertised context length can change customer-visible behavior. The notification should catch real reductions without paging on transient metadata fetch failures or temporarily missing backend data.

## Acceptance Criteria

- Persist the last known positive max context length per canonical model, with observation timestamp and enough backend/provider evidence for debugging.
- On provider/model metadata refresh, compute the current max positive advertised context length across healthy backends for each model.
- Notify admins when the current confirmed max is lower than the previously observed max, including model id, old value, new value, observed backend values, and relevant health/fetch context.
- Do not notify when backend metadata is unknown because every fetch failed, no backend reports a positive value, or the sample is otherwise incomplete.
- Add tests for confirmed reduction, unchanged/increased values, and missing/transient metadata cases.

PR #842, opened 2026-06-23
[codex] Fix model catalog metadata validation

### LINKED ISSUE #837: Fix invalid /v1/models output limits and cache-read pricing
## Summary

`GET https://cloud-api.near.ai/v1/models` currently returns 47 rows, but several rows have model metadata that downstream catalog consumers should not sync blindly:

- 6 rows expose no output token limit (`max_output_length` is absent/null and `top_provider.max_completion_tokens` is absent).
- 20 rows report `pricing.input_cache_read: "0"`, including Anthropic, OpenAI, and Google routes where free cache reads are not expected.

This is currently blocking a models.dev NEAR sync PR because the sync now fails fast instead of silently inventing limits or writing `cache_read = 0`: https://github.com/anomalyco/models.dev/pull/2745

## Reproduction

```bash
bun -e 'const res=await fetch("https://cloud-api.near.ai/v1/models"); if (!res.ok) throw new Error(`${res.status} ${res.statusText}`); const json=await res.json(); const missing=[]; const zero=[]; for (const m of json.data) { const outputLimit=m.max_output_length ?? m.top_provider?.max_completion_tokens; if (outputLimit == null || outputLimit <= 0) missing.push(m.id); const cache=m.pricing?.input_cache_read; if (cache !== undefined && Number(cache) === 0) zero.push(m.id); } console.log(JSON.stringify({total:json.data.length,missingOutputLimit:missing,zeroCacheRead:zero}, null, 2));'
```

Observed on 2026-06-23:

```json
{
  "total": 47,
  "missingOutputLimit": [
    "deepseek/deepseek-v3.2",
    "minimax/minimax-m2.5",
    "moonshotai/kimi-k2.5",
    "qwen/qwen3-32b",
    "qwen/qwen3.5-397b-a17b",
    "z-ai/glm-5"
  ],
  "zeroCacheRead": [
    "anthropic/claude-opus-4-6",
    "anthropic/claude-sonnet-4-5",
    "black-forest-labs/FLUX.2-klein-4B",
    "google/gemini-2.5-flash",
    "google/gemini-2.5-flash-lite",
    "google/gemini-2.5-pro",
    "google/gemini-3.1-flash-lite",
    "openai/gpt-5.1",
    "openai/gpt-5.4",
    "openai/gpt-5.4-mini",
    "openai/gpt-5.4-nano",
    "openai/gpt-5.5",
    "openai/gpt-5-mini",
    "openai/gpt-5-nano",
    "openai/o3-mini",
    "openai/privacy-filter",
    "openai/whisper-large-v3",
    "qwen/qwen3.7-max",
    "Qwen/Qwen3-Embedding-0.6B",
    "Qwen/Qwen3-Reranker-0.6B"
  ]
}
```

## Expected behavior

- Every public model row that should be consumed by clients includes a valid output token limit, either `max_output_length` or `top_provider.max_completion_tokens`.
- `pricing.input_cache_read` is omitted/null when not supported or unknown, or set to the correct positive price when cache reads are supported.
- Rows with incomplete metadata are either fixed before appearing in `/v1/models` or clearly excluded/marked so downstream clients do not publish invalid catalog metadata.

## Impact

Downstream consumers cannot safely determine model limits for the six missing-limit rows. They also cannot treat `input_cache_read: "0"` as authoritative because free cache reads on Anthropic/OpenAI/Google routes look like invalid or placeholder pricing, and would publish incorrect `cache_read = 0` metadata.

PierreLeGuen:
## Summary
- Omit `pricing.input_cache_read` from `/v1/models` when cache-read pricing is the internal zero/unknown sentinel, while preserving positive rates.
- Fetch backend-advertised `contextLength` from providers for `/v1/models`, merge multiple backend values by max, and cache that backend-derived value with the public model list.
- Cap backend metadata enrichment at 5 seconds so a slow provider cannot hold the public `/v1/models` cache fill for the full provider control timeout.
- Keep this PR schema-free: no model DB migration or DB constraint is added here; stricter database policy can be handled in a follow-up.

Fixes #837.
Follow-up issue for admin notifications when backend max context length drops: #841.

## Tests
- `cargo check -p services`
- `cargo test -p inference_providers context_length`
- `cargo test -p api cache_read_pricing`
- `cargo test -p services backend_context_lengths_with_timeout`
- `cargo test -p services get_models_with_pricing_caches_backend_context_lengths_until_invalidation`
- `DEV=1 BRAVE_SEARCH_PRO_API_KEY=issue837-test TEST_DATABASE_NAME=platform_api_e2e_pr842_timeout_models cargo test -p api --test e2e_all general::test_models_api`
- `DEV=1 BRAVE_SEARCH_PRO_API_KEY=issue837-test TEST_DATABASE_NAME=platform_api_e2e_pr842_no_db_admin cargo test -p api --test e2e_all admin_list_models`


PierreLeGuen:
@Evrard-Nil friendly ping when you have a moment, could you review this PR? CI is green and the automated feedback has been addressed. Thanks!

Evrard-Nil (changes requested):
## Review: [codex] Fix model catalog metadata validation

Build: clean. All unit tests pass (`cargo test --lib --bins`: 419 passed, 0 failed).

---

### Summary of changes

Three logically independent pieces in one PR:

1. **`input_cache_read` omission** — Change `ModelPricing.input_cache_read` from `String` to `Option<String>` and skip serializing when `cache_read_cost_per_token == 0`. Correct fix; two unit tests cover the zero and non-zero cases. ✓

2. **`maxOutputLength` required for active models** — New application-layer validation in `validate_model_request` rejects any write that would leave an active model without a positive `max_output_length`. Backed by a `NOT VALID` DB constraint. ✓

3. **Backend context-length enrichment** — On every model-list cache miss, fan-out to all registered providers (`Fleet.models()`, merged across rotation URLs), take the max per model, and overwrite the catalog's stored `context_length`. Cache TTL raised from 30 s → 300 s. ✓

---

### Issues

#### 1. Breaking change for existing active models without `max_output_length` — no migration or test coverage

`crates/services/src/admin/mod.rs:1285-1302`

```rust
let effective_catalog_is_active = request.is_active.unwrap_or_else(|| {
    existing_catalog.as_ref().is_none_or(|model| model.is_active)
});
let effective_max_output_length = request.max_output_length.or_else(|| {
    existing_catalog.as_ref().and_then(|model| model.max_output_length)
});

if effective_catalog_is_active
    && !matches!(effective_max_output_length, Some(max_output_length) if max_output_length > 0)
{
    return Err(...)
}
```

Any `PATCH /v1/admin/models` against an **existing active model whose `max_output_length` is `NULL` in the DB** (i.e. all models created before PR #705 backfilled this column) will fail with HTTP 400, even if the operator is only changing pricing or a description. For example:

```
PATCH /v1/admin/models { "qwen/qwen3-30b-a3b": { "modelDescription": "updated" } }
→ 400 "model 'qwen/qwen3-30b-a3b': maxOutputLength is required and must be positive for active models"
```

The `NOT VALID` constraint correctly grandfathers existing rows at the DB level, but the application-layer check has no equivalent grandfather clause for reads of existing rows.

**Missing:** a test that creates a model without `maxOutputLength`, confirms it's active in DB, then PATCHes a single unrelated field and expects 200. Without that test, this is an untested breaking change for production model updates.

There are two valid fixes:
- Add the backward-compat test and document "ops must backfill `maxOutputLength` before any future patch to an active model" (if this is intentional).
- Soften the validation to only fire when `is_active` is explicitly being set to `true` in the request (not when it inherits the existing value), matching the behaviour of the pricing gate which uses `request.is_active.unwrap_or(is_new_model)`.

#### 2. Cache fill can block `/v1/models` for up to 300 s when a backend is slow

`crates/services/src/models/mod.rs:86-88`

```rust
let context_lengths = inference_provider_pool.max_context_lengths_by_model().await;
apply_backend_context_lengths(&mut models, &context_lengths);
```

`max_context_lengths_by_model` uses `join_all` over all providers with the `control_timeout` per call. `DEFAULT_CONTROL_TIMEOUT_SECS = 300`. The moka `try_get_with` coalesces concurrent cache-miss requests — so if any backend is slow on startup or degraded, **every** caller waiting for `/v1/models` stacks up behind that single slow `join_all`. This is a public, unauthenticated endpoint.

The old code did a DB query (fast) and returned. Now it may take up to 300 s.

Suggested mitigations (any one would do):
- Add a shorter per-call timeout for the metadata fan-out (e.g. 5 s).
- Make `max_context_lengths_by_model` non-blocking: fire and forget into a background task, return the last-known values immediately. This is the pattern used elsewhere for model-discovery refresh.
- Cap the whole operation with `tokio::time::timeout`.

#### 3. `effective_catalog_is_active` and `effective_is_active` are computed from two separate DB queries

`admin/mod.rs:1281-1302` uses `get_model_catalog_state`, then at line 1317 uses `get_model_costs` which makes a second call. Both use `get_by_internal_name`. Between the two reads a concurrent write could change `is_active`, making the two derived `effective_*_is_active` values inconsistent. This is pre-existing for the pricing gate, but the new code adds a third read-derived decision. The DB constraint is the correct safety net here, but it's `NOT VALID`, so it won't catch violations on existing rows. Worth noting in code comments that the app-layer check is best-effort for new rows, not a hard guarantee.

---

### Smaller observations (non-blocking but worth tracking)

- **`Deserialize` on `ModelsResponse` has no `#[serde(deny_unknown_fields)]`** — This is intentional (future-proofing), consistent with the rest of the codebase.

- **`context_length` and `max_model_len` in `ModelInfo` are `i32`** — vLLM can theoretically report values exceeding `i32::MAX` (2 147 483 647 tokens) for very large contexts; serde would error-log and fall back to `None`. Unlikely for current models but worth a comment in the struct.

- **TTL raised from 30 → 300 s** — The justification (avoid hitting backend on every /v1/models request) is sound. Admin writes invalidate the cache explicitly, so staleness is bounded by the time of the next admin write + one backend fan-out, not by the TTL. This is fine.

- **All existing e2e test fixtures now include `maxOutputLength: 1024`** — correct and complete; good mechanical coverage.

---

### Conclusion

Issue 1 is a correctness/operational safety issue that will break any future admin model updates to currently-deployed models that don't have `maxOutputLength` set in the DB. It's missing a test and a documented migration path. Issue 2 is a potential availability regression for a public endpoint. Both should be addressed before merge.

Evrard-Nil (approved):
## Review: [codex] Fix model catalog metadata validation (re-review at 859e048b)

`cargo check` clean. Prior `CHANGES_REQUESTED` had two blockers; both are resolved.

---

### Blocker 1 — Grandfather clause for existing active models ✓ resolved

The approach changed from "add a grandfather clause" to "remove the app-layer enforcement entirely." Commit `2debca81` deletes the entire `validate_model_request` block that read `ModelCatalogState` and rejected writes to active models missing a positive `max_output_length`. The `ModelCatalogState` struct, `get_model_catalog_state` trait method, and its implementation in `admin_composite.rs` are all gone. No validation now fires.

Separately, the V0067 migration that added a `NOT VALID` DB-level constraint was also deleted in commit `30e97afd`. The net result: zero schema changes ship in this PR (`git diff base..HEAD -- crates/database/` is empty), which matches the PR description ("keep this PR schema-free").

The residual check at `crates/api/src/routes/admin.rs:265-274` only fires when `max_output_length` is explicitly supplied and is ≤ 0 — that is simple input validation, not an active-model gate, and it was already correct.

The outstanding test (`test_admin_rejects_active_model_create_without_max_output_length` and its three siblings) was removed along with the validation it was testing. Correct cleanup.

---

### Blocker 2 — `/v1/models` hang under slow backends ✓ resolved

Commit `859e048b` adds `backend_context_lengths_with_timeout` (`crates/services/src/models/mod.rs:44-58`) which wraps the entire `max_context_lengths_by_model()` fanout in `tokio::time::timeout(Duration::from_secs(5), ...)`. On timeout it logs a `warn!` and returns an empty map so the response is served from catalog values, not dropped. Two unit tests cover the happy path and the timeout path (`backend_context_lengths_with_timeout_returns_values_before_deadline`, `backend_context_lengths_with_timeout_returns_empty_after_deadline`). The cache TTL is 300 s, so the 5-second tax at cache-fill time is acceptable.

---

### New code in this revision — no new blockers found

**`apply_backend_context_lengths` unconditional overwrite** (`models/mod.rs:34-43`): the function replaces `model.context_length` with whatever `max_context_lengths_by_model` reports, with no floor check against the catalog value. This means a backend that reports a *lower* context window than the catalog will silently downgrade what clients see. In practice the pool method already takes the `max` across rotation URLs (`entry.and_modify(|stored| *stored = stored.max(context_length))`), and `advertised_context_length` takes the max across the three backend fields, so the floor-less overwrite is reasonable given the PR's stated goal of serving the backend's actual runtime capacity. Flagging for awareness but not blocking.

**`merge_model_responses` deduplication**: uses `HashMap` keyed by `model.id`, takes the first response as the base and merges subsequent ones. Only `context_length` / `max_model_len` / `top_provider.context_length` are merged (via `merge_model_context_length` → `advertised_context_length()` → max). Other `ModelInfo` fields (`owned_by`, `object`, `created`) are left as-is from the first response. For the use-case (rotation URLs for the same physical model), this is correct: all rotation backends have identical metadata aside from context windows.

**`ModelInfo` `i32` for context lengths**: vLLM reporting values > 2 147 483 647 is not a concern for any current or near-future model. Fine.

**Cache TTL 30 → 300 s**: admin writes call `invalidate_models_cache()` explicitly, so operator changes surface within the next cache fill, not after 300 s. Acceptable.

---

Approving. Both blockers are addressed; no new correctness or safety issues introduced.

PR #843, opened 2026-06-23
fix(build): break apt-pin round-trip cycle that kept breaking reproducible builds

Evrard-Nil:
## Problem

The `verify-reproducible-build` workflow has been red on every main push since ~2026-06-11 — a repeat of PR #742 (which fixed the identical symptom 2 weeks earlier).

**Root cause:** `build-image.sh` extracts the apt-resolved package versions post-build and writes them back to the committed `pinned-packages-{builder,runtime}.txt`, dirtying the build inputs. When a Debian OpenSSL security update makes the committed pin unsatisfiable, apt installs the new version, the extract dirties the host file, and the second sequential build (after `docker system prune`) bakes a different apt layer → different OCI digest. Only the apt-install layer differs; only the OpenSSL pin lines differ.

## Fix

Three changes:

1. **Redirect post-build extract to `.resolved.txt` files** — the committed `pinned-packages-*.txt` are never overwritten, so build 1 and build 2 on a pruned runner read byte-identical committed inputs → identical apt layer → stable digest. Resolved files are gitignored.

2. **Bump the 4 stale OpenSSL pins** from `3.0.17-1~deb12u3` / `3.0.20-1~deb12u1` to `3.0.20-1~deb12u2` — the version apt actually resolves.

3. **Add a guard step** to `verify-reproducible-build` that diffs committed vs resolved pins after the build and fails loudly on future drift, so staleness is caught at CI time instead of silently at digest-compare time.

Fixes recurring regression of #742.


Evrard-Nil:
Re: gemini-code-assist reviews —

**build-image.sh:67 (high):** This is factually incorrect — the PR *does* modify a workflow file. See the diff in `.github/workflows/verify-reproducible-build.yml` (the `Guard: committed pins match resolved pins` step, +8 lines). The guard step is the mechanism that fails the build on pin drift. No action.

**build-image.sh:71 (medium):** Valid observation but (a) pre-existing — the original `> pinned-packages-*.txt` redirect had the same no-error-check behavior, this PR doesn't change error handling; and (b) mitigated by the new guard: an empty `.resolved.txt` (from a failed `docker run`) would diff non-zero against the committed pin file and fail the guard step loudly. So the change as-is is strictly safer than before. Optional follow-up hardening (explicit exit-on-empty check in build-image.sh) tracked but out of scope for this unblock.

code-review check failure is an infra cancellation ('Canceling since a higher priority waiting request for OpenCodeReview-843 exists'), not a review rejection. claude-review ✅, security_audit ✅.

PierreLeGuen (changes requested):
This is a CI-only build-reproducibility fix: redirecting apt's resolved pin output to gitignored `*.resolved.txt` files (instead of overwriting the committed inputs) correctly breaks the round-trip that dirtied build inputs between runs. The root-cause approach is sound, but the committed builder pins don't match what the builder stage actually resolves, which will fail the PR's own guard on main.

- **Blocking** — `pinned-packages-builder.txt:297`: the committed builder pins set `libssl-dev`, `libssl3`, and `openssl` to `3.0.20-1~deb12u2`, but the builder stage's apt path (off the pinned Rust base) resolves those to `3.0.17-1~deb12u3`. A transient build of the builder resolution path confirmed `3.0.17-1~deb12u3`, while runtime resolved `3.0.20-1~deb12u2`. The new guard diffs committed pins against `pinned-packages-builder.resolved.txt` and will fail. Regenerate the builder pins from the actual builder-stage resolution, or advance the builder base/snapshot so `3.0.20-1~deb12u2` is genuinely resolvable before committing.

Non-blocking, no action needed: prior notes about "missing workflow files" (present in the diff) and "no error handling on docker run" (a failed extraction makes the guard diff fail loudly anyway) are not concerns.

Checks: `bash -n build-image.sh`, `sort -c` and duplicate checks on both pin files, `git diff --check`, and `git check-ignore` (confirmed `*.resolved.txt` ignored, committed pins not) all passed; transient Docker apt-resolution runs for both builder and runtime stages surfaced the mismatch above. Full reproducible Docker build not run.

Evrard-Nil:
Thanks @PierreLeGuen — you're exactly right. The builder stage (`rust:1.92.0-bookworm`) ships `3.0.17-1~deb12u3` pre-installed and apt keeps it, while I mistakenly bumped those pins to match the runtime's `3.0.20-1~deb12u2`. Verified this from the OCI artifact's baked-in `/app/pinned-packages-builder.txt`: the builder was already stably resolving to `3.0.17-1~deb12u3` across both builds in the debug artifacts. The runtime (`debian:bookworm-slim`) genuinely resolves `3.0.20-1~deb12u2` — that bump is correct.

Correction pushed: reverted the 3 builder pins back to their pre-PR `3.0.17-1~deb12u3` values in commit `9c3f108c`. `git diff main -- pinned-packages-builder.txt` is now empty confirming the builder file matches main. Only 4 files now differ vs main (builder pin file is a no-op).

Would appreciate a re-review when you get a chance.

PierreLeGuen (approved):
Small, well-contained CI/build-infra fix. It redirects the post-build pin extraction in `build-image.sh` to gitignored `*.resolved.txt` files instead of overwriting the committed `pinned-packages-*.txt` inputs, removing the input-mutation that caused digest drift between sequential builds, and adds a CI guard that diffs committed pins against resolved pins. The prior blocking concern (builder pins wrongly bumped) is resolved — `pinned-packages-builder.txt` is now byte-identical to main, and the only pin change is the runtime `libssl3`/`openssl` micro-bump.

- `build-image.sh:66-71` — the `docker run … > *.resolved.txt` redirects don't check exit status, so a failed extract writes an empty file. Self-mitigating, since an empty file diffs non-zero against committed pins and fails the new guard.
- `.github/workflows/verify-reproducible-build.yml:42` — the drift guard is gated to `matrix.runner == 'ubuntu-latest'`, so the `sequential-builds` job has no guard. Acceptable since pins are snapshot-derived and host-independent; a one-line comment explaining the gating would help future readers.

Checks: `bash -n build-image.sh`, `git diff --check`, `sort -c` and duplicate-key checks on both pin files, confirmed `pinned-packages-builder.txt` unchanged vs main, and `shellcheck` (only pre-existing SC2181). Full Docker reproducible build skipped (slow, needs Docker daemon + apt snapshot network access).

PR #844, opened 2026-06-23
feat: index-addressed per-backend routing in nearai::Fleet (preemptive TTFT reroute + cache affinity)

Evrard-Nil:
## Summary

Replaces the opaque **prefix-bucket** serving path in `nearai::Fleet` with **index-addressed serving**, so cloud-api can load-balance across *individual backends* behind one canonical SNI (e.g. `glm-5-2-i0` / `glm-5-2-i1`) — not just across provider tiers.

### Why
The pool layer (PR #838) sees exactly **one** `nearai::Provider` per NEAR model (the canonical aggregate SNI); model-proxy fans out to `i0`/`i1` invisibly. So per-backend TTFT can't be measured or targeted there. The only index-addressable path is the provider's rotation-SNI (`<canonical>-iN.<base>` → backend `N % count`), which was previously **fallback-only on 5xx**. This PR promotes it to the primary, latency-aware, cache-affine serving path.

### What changed (`crates/inference_providers/src/attested/nearai/{fleet,mod}.rs`)
- **`select_index(messages)`**: prefix-affinity (`prefix_router bucket % live count` → same prefix → same backend → KV-cache hit) + **preemptive latency steering** (a backend whose TTFT EMA > 2× the fastest warmed peer, above a 500 ms floor, is skipped for the fastest peer).
- **Per-index pooled, attestation-verified clients** (`index_clients`, sized `MAX_FANOUT`), each pinned to `<canonical>-iN.<base>`. Created via `backend_verifier.create_verified_client(rotation_url(i))` with the **exact** retry / fail-closed-in-Bootstrap / fallback-only-when-pinned semantics of the old bucket clients.
- **Per-backend TTFT EMA** measured by a `TtftProbe` stream wrapper (records once, on the first content chunk).
- **5xx fallback** now walks the *other* indices ordered by EMA (fastest first).
- **Completion→signature affinity**: `signature_rotation[chat_id] = serving index`; the signature is fetched from the same backend over the same warm pooled connection.
- **Count-change reset**: `store_backend_count` clears per-index clients + EMA when the healthy count changes (the `-iN`→backend binding is only stable while count is).
- **Cold-start / non-rotation URLs** (`rotation_count() == 0`, e.g. `localhost`): unchanged canonical-SNI path.

### Attestation safety (reviewed adversarially)
1. Every served backend is attestation-verified — same verifier path + fail-closed-in-Bootstrap as before.
2. Completion + signature hit the same backend (index pin + warm pooled connection).
3. The index↔backend trust is the **same** model-proxy L4 SNI-routing trust the bucket path already relied on — no new assumption.

A 4-lens adversarial review (attestation / signature / connection lifecycle / routing) ran; one real prod-only bug was caught **before** this PR (a `verify_url` double-slash that tests can't see because localhost disables rotation — fixed) and the bounds/limitation notes below were added.

### Known limitations (documented in code)
- **Signature pin staleness on a same-count membership change** (a backend drops as another recovers, silently remapping index `i`): small window (signature fetched within the caller's FINALIZE_TIMEOUT vs ~5-min discovery cadence), degrades gracefully (missing signature logged, stream still completes), and is **pre-existing** with the old rotation fallback — not introduced here.
- **Reachability**: `prefix_router` uses 64 buckets; if a single model ever exceeds 64 live backends, high indices are unreachable as the *preferred* pick (steer/fallback can still reach them). Raise `NUM_PREFIX_BUCKETS` if needed.
- **Happy-path streaming peeks the first chunk** (to be able to reroute on a first-chunk error frame); user-visible TTFT is ~unchanged since SSE clients wait for the first chunk anyway.

### Relationship to #838
#838 is the **pool-layer** (cross-tier) adaptive LB (per-provider TTFT EMA + context filter). This PR is the **provider-layer** per-backend mechanism. They're independent; #838's pool-level prefix-hash is redundant for NEAR once this lands (harmless).

## Test plan
- [x] `cargo test -p inference_providers --lib` — 365 pass; new tests cover `select_index` affinity/steer, count-change reset, EMA warmup/stable, localhost-disables-rotation, fallback ordering, `TtftProbe` records-once.
- [x] `cargo check --workspace`, `cargo clippy`, `cargo fmt` clean (no `InferenceProvider` trait signatures changed → pool unaffected).
- [ ] **Staging**: with ≥2 backends for one model, verify via Datadog that same-system-prompt conversations land on the same `-iN`, and that injecting latency on one backend steers traffic away after ~8 requests.
- [ ] **Staging**: confirm attestation/signature still verify end-to-end on the `-iN` path (tests can't exercise real rotation).


PierreLeGuen (approved):
This is a clean refactor from prefix-bucket serving to index-addressed (`-iN` SNI) per-backend routing in `nearai::Fleet`. The security-critical paths — SPKI/attestation enforcement, cold-start fail-closed fallback, signature affinity, and index bounding (`rotation_count() = count.min(MAX_FANOUT)` with a defensive guard in `get_or_verify_index_client`) — all hold up.

One non-blocking observation worth a follow-up:

- `crates/inference_providers/src/attested/nearai/fleet.rs:288` (`store_backend_count`): a count change unconditionally nulls all `index_clients`. For a completion that spans a discovery/scale event, the recreated index client at `crates/inference_providers/src/attested/nearai/mod.rs:1211` may resolve to a different backend, so signature storage/retrieval could read from the wrong backend. This is also where a legacy (no-verifier) provider can't rebuild a cleared slot. Neither path is reachable in production (pool always builds via `new_with_verifier`; legacy/test usage targets non-rotation `localhost` URLs that never touch `index_clients`), so this is a latent divergence rather than a live bug. If you want to harden it, pin the serving client for in-flight signatures and/or guard the clear with `if self.backend_verifier.is_some()`.

Local checks: `cargo test -p inference_providers --lib` (365 passed), `cargo check --workspace`, `cargo clippy -p inference_providers -- -D warnings`, and `cargo fmt --check` all pass clean.

Evrard-Nil:
Thanks for the review — addressed all three in 121686b3:

- **HIGH (legacy-mode client clear)**: real bug, fixed. `store_backend_count` now only clears `index_clients` when a verifier is present. In legacy/no-verifier mode the eager clients can't be lazily rebuilt, so clearing would wedge the provider with "no backend verifier configured"; those clients aren't attestation-pinned anyway, so keeping them is correct. Stats reset stays unconditional. Split the test into verifier-mode (clears) and legacy-mode (keeps clients, still resets stats).
- **MEDIUM (EMA first-sample)**: switched to `stat.samples == 0` instead of the float `== 0.0` comparison.
- **MEDIUM (.get() safety)**: `select_index` now uses `.get()` throughout (no direct indexing). Also added a defensive bounds guard in `get_or_verify_index_client` for the same reason.

366 lib tests pass, clippy/fmt clean.

lloydmak99 (commented):
Automated fallback review for `nearai/cloud-api#844` at head `11f4a00b82ea43f3a0efd20535ccea7448906980`.

The child Codex review completed but could not post directly. Its final review output follows.

Blocked: I could not complete or post the PR review.

`gh api`, `git fetch`, and `git ls-remote` all failed because DNS cannot resolve `github.com` / `api.github.com`. I also found no locally cached PR #844 ref, diff, commit, or title match. The local monitor payload references `cloud-api#838`, but I did not substitute that because the request was explicitly for `nearai/cloud-api#844`.

No GitHub comments or reviews were posted. Posted verdict: none.

PR #846, opened 2026-06-23
fix(ci): stop non-Rust PRs hanging on the required 'Test Suite' check

Evrard-Nil:
## Problem — non-Rust PRs get stuck on a required check, every time

The `main` ruleset requires the **`Test Suite`** status check. But `test.yml` had a
top-level `paths:` filter (added in #825 to skip tests on non-Rust PRs). GitHub's
behavior: when a PR touches none of those paths, the **entire workflow is skipped —
including the `Test Suite` gate job** — so the required check never reports and stays
`Expected` forever. The PR sits at `mergeStateStatus: BLOCKED` and can only be merged
by an `OrganizationAdmin` bypassing the ruleset.

This fires on **every** non-Rust PR — CI/build-script/docs changes — e.g. #843 and #845
(the reproducible-build fixes). It's the recurring "CI is stuck" on cloud-api.

The code even documents the trap without resolving it (`test.yml` had:
`"branch protection requires 'Test Suite'; this job satisfies that requirement"`) — but a
path-skipped workflow can't satisfy anything.

## Fix — keep #825's speedup, but always report the required check

- **Remove the top-level `paths:` filter** so the workflow always runs on PRs/pushes to
  `main`/`develop`.
- **Add a lightweight `changes` job** that detects Rust-relevant changes (the *same* path
  set the old filter used) via `git diff` — no third-party action, in keeping with the
  repo's supply-chain posture. **Fails safe:** anything it can't classify (new branch,
  detection error, empty diff) is treated as "run the tests".
- **Gate** `lint` / `unit-test` / `integration-test` / `e2e-test` (and `build-release`) on
  `needs.changes.outputs.rust == 'true'` — so they still skip on non-Rust PRs (the #825
  speedup is preserved).
- The **`Test Suite` gate** now `needs` the `changes` job and always runs (`if: always()`):
  it passes when there were no Rust changes (test jobs skipped) and otherwise requires every
  test job to have succeeded. So `Test Suite` **always reports a result** and no PR hangs on
  a perpetually-pending required check.

Net effect: Rust PRs run and enforce tests exactly as before; non-Rust PRs get a green
`Test Suite` in seconds instead of a permanent block.

## Validation

- `actionlint` clean (only the pre-existing custom `infra` runner-label warning).
- Change-detection regex unit-tested against representative file sets: #845's files
  (`build-image.sh`, `verify-reproducible-build.yml`) and docs → `rust=false` (skip, gate
  green); `*.rs` / `Cargo.{toml,lock}` / migrations / `test.yml` / `setup-rust-ci/` /
  `nextest.toml` → `rust=true`; empty/unknown → `rust=true` (fail-safe).
- **This PR touches `test.yml` → `rust=true` → it runs the full suite on itself**, validating
  the always-run + gate path end-to-end on the real runners.
- The `rust=false` skip path can only run once the new `test.yml` is on `main` (any branch
  carrying this change shows `test.yml` in its own diff). It'll be confirmed on the next
  non-Rust PR — rebasing #845 onto main is the natural first check, and also unsticks it.

CI-only change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,9 +39,10 @@ env:
   RUSTFLAGS: "-C debuginfo=0 -C link-arg=-fuse-ld=mold"
 
 jobs:
-  # Decide whether the Rust test jobs need to run. Mirrors the path set that
-  # used to live in the `on:` filter. Fails safe: anything we can't classify
-  # (new branch, detection error) is treated as "run the tests".
+  # Decide whether the Rust test jobs need to run. Covers the same path set the
+  # old `on:` filter used, plus rust-toolchain.toml and .cargo/config.toml (which
+  # affect the build but the old filter omitted). Fails safe: anything we can't
+  # classify (new branch, detection error) is treated as "run the tests".
   changes:
     name: Detect changes
     runs-on: [self-hosted, infra]
@@ -72,7 +73,7 @@ jobs:
           # and none of it is Rust-relevant.
           RUST=true
           if [ -n "$CHANGED" ] && ! echo "$CHANGED" | grep -qE \
-            '(\.rs$|(^|/)Cargo\.(toml|lock)$|^crates/database/src/migrations/|^\.github/workflows/test\.yml$|^\.github/actions/setup-rust-ci/|^\.config/nextest\.toml$)'; then
+            '(\.rs$|(^|/)Cargo\.(toml|lock)$|^crates/database/src/migrations/|^\.github/workflows/test\.yml$|^\.github/actions/setup-rust-ci/|^\.config/nextest\.toml$|^rust-toolchain\.toml$|^\.cargo/config\.toml$)'; then
             RUST=false
           fi
           echo "rust=$RUST" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,24 +1,16 @@
 name: Tests
 
+# NOTE: deliberately NO top-level `paths:` filter. "Test Suite" is a required
+# status check (branch ruleset), and a path-filtered workflow that is skipped
+# leaves that check perpetually "Expected", which blocks every PR that does not
+# touch Rust files (CI, build scripts, docs) until an admin bypasses it. Instead
+# the workflow always runs and the `changes` job below decides whether the
+# expensive Rust jobs actually execute — so "Test Suite" always reports a result.
 on:
   push:
     branches: [ main, develop ]
-    paths:
-      - '**/*.rs'
-      - '**/Cargo.toml'
-      - '**/Cargo.lock'
-      - 'crates/database/src/migrations/**'
-      - '.github/workflows/test.yml'
-      - '.config/nextest.toml'
   pull_request:
     branches: [ main, develop ]
-    paths:
-      - '**/*.rs'
-      - '**/Cargo.toml'
-      - '**/Cargo.lock'
-      - 'crates/database/src/migrations/**'
-      - '.github/workflows/test.yml'
-      - '.config/nextest.toml'
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
CI-only change to `.github/workflows/test.yml` that removes the workflow-level `paths:` filter so the required `Test Suite` check always reports, fixing the "Expected — forever" hang on non-Rust PRs. The gate logic is sound: the `changes` job replicates the old path set via `git diff` and fails safe (empty/unclassifiable → `rust=true`), and `test-suite` (`if: always()`) fails closed when `changes` didn't succeed and otherwise requires every Rust test job to pass.

- `.github/workflows/test.yml:75` — the change-detection regex omits `rust-toolchain.toml` and `.cargo/config.toml`, so a toolchain/build-flag bump would classify as `rust=false` and report a green-but-unverified `Test Suite`. This bias is fail-open but the case is rare; consider adding `^rust-toolchain\.toml$|^\.cargo/config\.toml$` to the alternation.

Checks: YAML validated via parser; referenced paths (`rust-toolchain.toml`, `.cargo/config.toml`, `.github/actions/setup-rust-ci/`) confirmed present; gate logic manually traced across rust=true / rust=false / changes-failed / new-branch event shapes. `actionlint` skipped (not installed).

lloydmak99 (changes requested):
Automated fallback review for `nearai/cloud-api#846` at head `95c42f354bf50fd323bcf63df795e29438fd56af`.

The child Codex review completed but could not post directly. Its final review output follows.

Used `code-review` skill. Verdict: `REQUEST_CHANGES`.

Blocking finding: `.github/workflows/test.yml:14` adds `pull_request.paths` to the workflow that owns required `Test Suite`. For non-Rust PRs, GitHub skips the whole workflow, and required checks for path-skipped workflows remain Pending, so this preserves the hang the PR is meant to fix. GitHub docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

I could not post the formal GitHub review: `gh api` failed with `error connecting to api.github.com`, direct `git fetch` failed DNS resolution for `github.com`, and the in-app browser fallback was unavailable. Detached worktree/branch cleanup is done.

Evrard-Nil:
@lloydmak99 this finding is the opposite of what the change does — I believe the Codex run analyzed no real diff (its own note: *"gh api failed with `error connecting to api.github.com`, direct git fetch failed DNS resolution for github.com"*).

The review claims `test.yml:14` **adds** `pull_request.paths`. It doesn't — this PR **removes** the path filter. At head `95c42f35`:

- `test.yml:14` is a blank line. The two `paths:` blocks (under `push:` and `pull_request:`) are **deleted**:
  ```
  $ git diff origin/main -- .github/workflows/test.yml | grep -E '^[+-].*paths:'
  -    paths:
  -    paths:
  ```
  (two deletions, zero additions). The only remaining `paths` tokens are in explanatory comments.
- That's precisely the fix: with no top-level `paths:` filter the workflow always runs, the `changes` job decides whether the heavy Rust jobs execute, and the `Test Suite` gate runs with `if: always()` — so the required check always reports instead of hanging on a path-skipped workflow.

This is already demonstrated on this very PR: the workflow ran and **`Test Suite` reported `success`** (it didn't stay Pending). The blocking finding rests on a diff that wasn't actually fetched, so I haven't made a code change. Re-requesting a fresh review.

PierreLeGuen (approved):
This CI-only change removes the workflow-level `paths:` filter that left the required `Test Suite` check hanging on non-Rust PRs, replacing it with an internal `changes` gating job. The core logic is sound: `test-suite` runs with `if: always()`, fails closed if change detection fails, passes when Rust is untouched, and aggregates all test-job results otherwise. `Test Suite` already passed on this PR's head.

- `.github/workflows/test.yml:75` — the change-detection regex omits `rust-toolchain.toml` and `.cargo/config.toml` (both exist in the repo). A standalone change to either would classify as `rust=false` and skip the test jobs while `Test Suite` reports green. This is pre-existing (the old `paths:` filter omitted them too) and a rare scenario, but worth closing by adding `|^rust-toolchain\.toml$|^\.cargo/config\.toml$` to the alternation.

Note: the prior automated `CHANGES_REQUESTED` review claiming the PR *adds* `pull_request.paths` is a false positive — the PR removes both `paths:` blocks; line 14 is blank and the diff contains only deletions.

Checks: traced the gate logic across all event shapes and output values, compared the new regex against the old `paths:` set, confirmed the referenced config files exist, and validated the workflow YAML parses. `actionlint` ran clean upstream (only the pre-existing custom `infra` runner-label warning); local Rust build/tests skipped as workflow-only.

lloydmak99 (dismissed):
Automated fallback review for `nearai/cloud-api#846` at head `95c42f354bf50fd323bcf63df795e29438fd56af`.

The child Codex review completed but could not post directly. Its final review output follows.

Used `code-review` skill and reviewed PR #846 locally from `origin/chore/skip-tests-non-rust` at `7a04a886`.

Verdict: `REQUEST_CHANGES`, but it was **not posted**. The GitHub review POST failed because this environment cannot resolve `api.github.com`, and `gh auth status` reports the cached `lloydmak99` token as invalid.

Blocking finding: `.github/workflows/test.yml:14` adds `pull_request.paths`, which skips the entire workflow for non-Rust PRs. Since branch protection requires the `Test Suite` check, skipped workflows still leave that check pending, so the PR does not fix the stated hang. The fix is to keep `pull_request` unfiltered and gate Rust-heavy work inside jobs/steps so `Test Suite` always reaches a terminal result.

Temporary detached worktree was cleaned up.

Evrard-Nil:
Thanks @PierreLeGuen. Addressed the follow-up in `f64368a6` — added `^rust-toolchain\.toml$` and `^\.cargo/config\.toml$` to the detection alternation, so a standalone toolchain/cargo-config change now runs the tests instead of skipping while `Test Suite` reports green. And thanks for confirming the automated `CHANGES_REQUESTED` is a false positive (it claimed the PR *adds* a `paths:` filter; it removes both). That stale review is the only thing still blocking `mergeStateStatus` — it'll need a dismissal or a clean re-review to clear.

PR #847, opened 2026-06-23
feat(metrics): cloud-api prefix-cache-hit metrics (tokens.cached + cache.hit_rate)

Evrard-Nil:
## Summary

cloud-api had **no cache-hit-rate metric**. The only cache signal was `cloud_api.billed.cache_read_tokens` — a billing-namespaced token *count* emitted once per newly-inserted usage row. The per-request hit-rate **distribution** was not observable anywhere, and the backend `sglang:cache_hit_rate` / `vllm:gpu_prefix_cache_hit_rate` gauges (already scraped to Datadog) carry no cloud-api request context (model alias, inference_type) and reflect the backend's internal radix cache, not what the client actually received.

This adds two metrics in the chat-completions path (both the streaming-drop path and the non-streaming path), **co-tagged with the existing `tokens.input`/`tokens.output`** (`model`, `inference_type`, `environment`, `input_bucket`):

| Metric | Type | Meaning |
|--------|------|---------|
| `cloud_api.tokens.cached` | counter | Cache-read (prefix-cache hit) input tokens. **Token-weighted fleet hit rate = `sum(tokens.cached) / sum(tokens.input)`** |
| `cloud_api.cache.hit_rate` | histogram (percent) | **Per-request** cached/prompt — the distribution the token counters can't express (p50/p90, regression detection) |

Two metrics because they answer different questions: the counter gives the canonical token-weighted aggregate KPI; the histogram gives the per-request spread (a 1-token request hitting 100% shouldn't count equal to a 10k-token request hitting 5%).

## Why now
The Fleet index-addressed routing change (#844) makes same-prefix requests stick to the same backend to maximize KV-cache hits. **This is the metric to confirm that's working** — chart `cache.hit_rate` / `tokens.cached÷tokens.input` by model before/after and watch it rise. (I verified the effect manually on staging via `cached_tokens` in responses; this makes it a first-class, continuously-tracked signal.)

## Implementation notes
- Hit rate via a guarded helper `cache_hit_rate_percent(cached, prompt)` → `None` when `prompt == 0` (keeps empty-prompt requests out of the distribution, no div-by-zero), clamps negative cached to 0. `cached_tokens()` is already clamped to `[0, prompt]`, so the rate ∈ [0, 100].
- No new tags (cardinality unchanged). No public API change.
- Distinct from the `billed.*` namespace, which is billing-oriented and insert-gated.

## Test plan
- [x] `cargo test -p services --lib` — 419 pass. New: `cache_hit_rate_percent_computes_and_guards_zero_prompt` (helper) + `test_intercept_stream_emits_cache_hit_metrics` (streaming: 7/10 cached → `tokens.cached=7`, `cache.hit_rate=70%`).
- [x] `cargo clippy` / `cargo fmt` clean.
- [ ] Staging: after deploy, confirm `cloud_api.tokens.cached` + `cloud_api.cache.hit_rate` appear in Datadog (us3) tagged by model, and that `tokens.cached/tokens.input` tracks sensibly vs the backend `sglang:cache_hit_rate`.

### DIFF
--- a/crates/services/src/completions/mod.rs
+++ b/crates/services/src/completions/mod.rs
@@ -38,6 +38,17 @@ pub fn hash_inference_id_to_uuid(full_id: &str) -> Uuid {
 
 /// Get input bucket tag based on token count for metrics breakdown
 /// Buckets: 0-1k, 1-4k, 4-16k, 16-32k, 32-64k, 64-128k, 128k+
+/// Per-request prefix-cache hit rate as a percentage (cache-read / prompt
+/// tokens). `None` when there are no prompt tokens (avoids div-by-zero and keeps
+/// empty-prompt requests out of the distribution). `cached` is already clamped to
+/// `[0, prompt]` by `TokenUsage::cached_tokens()`, so the result is in `[0, 100]`.
+fn cache_hit_rate_percent(cached_tokens: i32, prompt_tokens: i32) -> Option<f64> {
+    if prompt_tokens <= 0 {
+        return None;
+    }
+    Some((cached_tokens.max(0) as f64 / prompt_tokens as f64) * 100.0)
+}
+
 fn get_input_bucket(token_count: i32) -> &'static str {
     match token_count {
         0..=1000 => "0-1k",
@@ -352,6 +363,18 @@ where
                             output_tokens as i64,
                             &tags,
                         );
+                        // Prefix-cache-hit observability: cache-read token count
+                        // (token-weighted hit rate = cached/input) + per-request
+                        // hit-rate distribution.
+                        metrics_service.record_count(
+                            METRIC_TOKENS_CACHED,
+                            cache_read_tokens as i64,
+                            &tags,
+                        );
+                        if let Some(rate) = cache_hit_rate_percent(cache_read_tokens, input_tokens)
+                        {
+                            metrics_service.record_histogram(METRIC_CACHE_HIT_RATE, rate, &tags);
+                        }
                     })
                     .await;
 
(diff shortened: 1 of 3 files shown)


Evrard-Nil:
Addressed both in the latest commit:
- **Clamp to [0, prompt]**: `cache_hit_rate_percent` now clamps `cached` to `[0, prompt_tokens]` (upper bound included), so the rate is provably within `[0, 100]` even if a provider ever reports cached > prompt. Added a `cached>prompt → 100%` test case.
- **Test sleep comment**: documented the `sleep` as waiting for the fire-and-forget usage/metrics task spawned in `Drop` (matching the sibling test).

419 services lib tests pass, clippy/fmt clean.

PierreLeGuen (approved):
Clean, well-scoped metrics change. The core correctness claims hold up: `cache_hit_rate_percent` guards `prompt <= 0` and clamps negatives, and `cached_tokens()` already clamps to `[0, prompt_tokens]` (`inference_providers/src/models.rs:411-422`), so the histogram stays in `[0, 100]` with no div-by-zero. Both streaming and non-streaming paths derive `input_tokens`/`cache_read_tokens` from the same `TokenUsage` struct, so the ratio is internally consistent. No new tags or customer data on the metrics path.

- `crates/services/src/completions/mod.rs:370` — `record_count(METRIC_TOKENS_CACHED, 0, ..)` fires on cache-miss requests. Harmless for a counter and keeps the series continuous; just noting for dashboard readers.
- Instrumentation covers only the chat-completions path; the `/v1/responses` path is not instrumented (intentional per the PR description). Worth a follow-up if cache observability is wanted there.

Checks: `cargo test -p services --lib` (cache_hit tests passed), `cargo clippy -p services -- -D warnings` clean, `cargo fmt --check` clean; GitHub CI green on security, lint, unit, and integration.

Issue #338, opened 2026-01-07
Bug: Workspace Name Reuse Blocked After Removal


sunglow666:
### Problem:
Creating a workspace with **a name previously used by a removed workspace** in the same organization,
the **POST /v1/organizations/{org_id}/workspaces** endpoint returns 409 Conflict: "**Workspace name already exists in organization**".

PR #849, opened 2026-06-24
Pin Chutes v1.3.1 golden measurements (GLM-5.2-TEE attested fallback)

Evrard-Nil:
## Why

Companion to **[cvm-ansible #338](https://github.com/nearai/cvm-ansible-playbooks/pull/338)**, which adds `z-ai/glm-5.2=zai-org/GLM-5.2-TEE` to `CHUTES_MODELS` so Chutes serves as an attested fallback under the existing active `z-ai/glm-5.2` NEAR row.

The config change alone is **not sufficient**: GLM-5.2-TEE's live fleet runs Chutes firmware **v1.3.1-rc1** on Blackwell hardware, but `vetted_golden_measurements()` pinned **only v1.3.0**. The measurement register-pin would therefore match no accepted config and **fail closed** (masked 502/503) — exactly like `mistralai/mistral-nemo` on v1.2.0 (infra#776), but for a different reason (unvetted *version*, not unmeasured app layer).

## What

Restructures the vetted snapshot into per-release `Family` blocks (each shares one software identity — MRTD/RTMR1/RTMR2/runtime-RTMR3 — and varies only RTMR0 per hardware SKU) and adds the full **v1.3.1 family**:

- **v1.3.1-rc1** — 3 Blackwell rows (`8xb200`, `8xb200-xeon6`, `8xb300`). **Live cross-checked 2026-06-24**: 3 distinct register sets observed on GLM-5.2-TEE's fleet matched the published rows byte-for-byte (MRTD + RTMR0 + RTMR1 + RTMR2 + **runtime** RTMR3). These are genuine, DCAP-signature-verified, report-data-bound quotes (the measurement check runs only after the Intel signature chain + nonce/e2e-key/SPKI bindings pass).
- **v1.3.1 final** — 6 rows (2× h200, RTX_PRO_6000, 3× Blackwell). Same MRTD as `-rc1`, **distinct** RTMR1/2/3. Pinned as **forward-insurance** so GLM-5.2 keeps attesting once Chutes promotes the chute off the release candidate. Published + app-measured, but not yet live cross-checked (no instance observed on the final release yet).

Both v1.3.1 and v1.3.1-rc1 publish a **non-zero runtime RTMR3** (the app/IMA layer is genuinely measured) — unlike v1.0.0–v1.2.0 (RTMR3 = all-zeros, app unmeasured), which remain excluded fail-closed.

The cryptographic root of trust is unchanged: the Intel DCAP signature, not the (unsigned) transparency endpoint. These rows merely enumerate which genuine software identities we accept.

## Tests

`crates/services/src/attestation/chutes.rs` `golden` module:
- `accepts_glm52_live_rc1_blackwell_rows` — the 3 live-observed rc1 Blackwell register sets verify.
- `accepts_final_v131_for_rc1_promotion` — a final-v1.3.1 row verifies (promotion path).
- `rejects_rc1_software_with_final_v131_rtmr3` — an rc1 boot/kernel spliced to the final runtime RTMR3 matches no single row → fail-closed (`NoMatch`).
- `covers_the_full_v130_hardware_family` — row count updated 5 → 14; v1.3.0 rows still verify by name.
- `every_vetted_row_is_well_formed_48_byte_hex` — now guards all 14 pinned rows.

All 12 chutes attestation tests pass; clippy clean; NEAR path untouched.

## Deploy order

Merge **this** PR (and let it reach prod/staging) **before** merging cvm-ansible #338 — otherwise the GLM-5.2 fallback routes but fails closed on attestation. NEAR-primary serving of `z-ai/glm-5.2` is unaffected either way.

### DIFF
--- a/crates/services/src/attestation/chutes.rs
+++ b/crates/services/src/attestation/chutes.rs
@@ -206,17 +206,20 @@ impl ChutesInstanceVerifier for ChutesBackendVerifier {
 }
 
 /// A vetted snapshot of Chutes' published golden measurements (from the public
-/// `GET https://api.chutes.ai/servers/tee/measurements`) for the **v1.3.0**
-/// software release across all of Chutes' GPU hardware platforms.
+/// `GET https://api.chutes.ai/servers/tee/measurements`) across the software
+/// releases we accept (**v1.3.0** and the **v1.3.1** family) and all of Chutes'
+/// GPU hardware platforms.
 ///
-/// Within v1.3.0 the five rows below are byte-identical on MRTD (firmware),
-/// RTMR1 (kernel), RTMR2 (cmdline/initrd) **and the runtime RTMR3**
+/// Within a single software release the hardware rows are byte-identical on MRTD
+/// (firmware), RTMR1 (kernel), RTMR2 (cmdline/initrd) **and the runtime RTMR3**
 /// (`runtime_rtmrs.RTMR3`, the running app/IMA layer) — i.e. the full *software*
 /// identity is one and the same. They differ **only in RTMR0**, the TD-HOB/ACPI/
-/// VM-config layer that varies with the GPU SKU and VM sizing (8×H200,
-/// 8×RTX PRO 6000, 8×B200, 8×B200-eth, 8×B300). Pinning each published RTMR0
-/// lets a request land on any of Chutes' vetted hardware platforms while still
-/// authenticating the full software stack against a single, fixed identity.
+/// VM-config layer that varies with the GPU SKU and VM sizing. Pinning each
+/// published RTMR0 lets a request land on any of Chutes' vetted hardware
+/// platforms while still authenticating the full software stack against a fixed,
+/// per-release identity. A *different* software release (v1.3.0 vs v1.3.1 vs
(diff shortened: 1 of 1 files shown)


Evrard-Nil:
Thanks — applied the `|&(name, rtmr0)|` by-value destructure (drops the double-reference + `*name` deref) in a29cd1c4. Golden tests still green.

PierreLeGuen (approved):
Additive, fail-closed change pinning the Chutes v1.3.1 golden measurements (rc1 + final) alongside a data-preserving `Family`/`flat_map` refactor; v1.3.0 rows are byte-identical to before and the v1.3.1 identities stay distinct under the strict 5-register match.

- `crates/services/src/attestation/chutes.rs:280-294`: the 6 final-v1.3.1 rows are pinned without a live, DCAP-signature-verified cross-check (unlike the rc1 and v1.3.0 rows), so they rest solely on the unsigned transparency endpoint. The Intel DCAP signature still guarantees a genuine quote runs that published software identity, so this is a documented forward-insurance tradeoff rather than a soft-pass — worth an explicit security sign-off, or land rc1 now and add the final rows once an instance is observed live.

Local checks: `cargo test -p services attestation::chutes` (12 passed) and `cargo test -p inference_providers measurements` (7 passed); clippy clean; `git diff --check` passed.

PR #850, opened 2026-06-24
feat(attestation): short-TTL cache + single-flight for no-nonce reports (cut /report 503 herd)

Evrard-Nil:
## Problem

`/v1/attestation/report` sheds ~10% **503s in regular ~6-min bursts** (213/1887 over a sampled hour in prod). Datadog logs show the cause: **many independent monitors** — `Datadog/Synthetics`, `Blackbox Exporter`, Go/python probes, cloud-ui — poll the **full 58-model catalog** on overlapping schedules. Each report does an expensive per-backend **GPU-evidence NRAS round-trip (~700ms, serialized behind a per-backend mutex)**; when the monitors' schedules overlap, concurrent requests thundering-herd the backend, which sheds 503. (Confirmed pre-existing — the burst pattern runs for hours before any recent deploy; not customer chat traffic, which stays clean.)

## Key constraint (why naive caching is unsafe)

From the logs: **65% of report requests send NO nonce; the 35% that do each use a UNIQUE nonce** (zero reuse). The nonce is cryptographically bound into the TDX `report_data` **and** the GPU evidence (NRAS freshness — that's *why* the existing code only skips GPU collection when the nonce is absent). So a cached report **cannot** be served for a different nonce without defeating replay/freshness. Nonce-bearing requests must stay fresh.

## Fix — cache only the no-nonce majority

- **No-nonce requests (the 65%)**: already freshness-opted-out (the codebase forwards `None` so inference-proxy serves *its* 5-min cache, but cloud-api still re-ran the full orchestration each time). Now served from a cloud-api short-TTL `moka` cache (**default 10s**, env `ATTESTATION_REPORT_CACHE_TTL_SECS`, `0` disables). `try_get_with` **also single-flights** concurrent misses, so a burst of identical no-nonce probes triggers **one** backend build instead of N.
- **Nonce-bearing requests (the 35%)**: **bypass the cache entirely**, always built fresh.

Collapsing the no-nonce majority frees backend capacity → cuts the 503 shed rate for the genuinely-fresh nonce requests.

## Safety (by construction + unit-tested)

- `report_cache_key()` **never includes the nonce** and the cache is consulted **only when `nonce.is_none()`** — a request can never receive a report built for a different model / algo / tls-fingerprint / provider tier / signing address. Unit tests assert distinct params → distinct keys (Near vs Attested3p don't collide; algo case-normalized).
- Mirrors existing accepted behavior: inference-proxy already replays a cached report (with a baked `request_nonce`) to no-nonce callers.
- `AttestationError` gains `Clone` so the coalesced `Arc<Error>` maps back to an owned error → route still maps it to the correct HTTP status.

## Observability
New `cloud_api.attestation.report_cache{result:hit|miss|bypass}` — watch `hit` rise and backend `/report` 503s fall.

## Diff note
Most of the `attestation/mod.rs` churn (+523/−152) is **rustfmt reindentation** from wrapping the existing report-build body in a `build` closure — the body itself is unchanged. The real logic is the ~60-line dispatch wrapper + `report_cache_key` + the cache field/init.

## Test plan
- [x] `cargo test -p services --lib` — 427 pass (incl. 4 new `cache_key_tests`). `cargo clippy`/`fmt` clean; `api` crate builds.
- [ ] Staging: set traffic, confirm `report_cache{result:hit}` climbs for no-nonce probes and `result:bypass` for nonce requests; verify a nonce-bearing `/report` still returns a fresh quote bound to the sent nonce.
- [ ] Prod (after soak): confirm `/v1/attestation/report` 503 burst rate drops.

### DIFF
--- a/crates/services/src/attestation/mod.rs
+++ b/crates/services/src/attestation/mod.rs
@@ -42,6 +42,12 @@ pub mod ports;
 const GATEWAY_KEY_PATH_ED25519: &str = "/signing-key/ed25519";
 const GATEWAY_KEY_PATH_ECDSA: &str = "/signing-key/ecdsa";
 
+/// Default TTL (seconds) for the no-nonce attestation-report cache. Reports for
+/// nonce-less requests are short-lived snapshots; ~10s collapses the bursty
+/// monitoring herd (bursts arrive within seconds) while keeping reports fresh.
+/// Override with `ATTESTATION_REPORT_CACHE_TTL_SECS`; `0` disables the cache.
+const DEFAULT_ATTESTATION_REPORT_CACHE_TTL_SECS: u64 = 10;
+
 pub struct AttestationService {
     pub repository: Arc<dyn AttestationRepository + Send + Sync>,
     pub inference_provider_pool: Arc<InferenceProviderPool>,
@@ -55,6 +61,15 @@ pub struct AttestationService {
     ed25519_verifying_key: Arc<VerifyingKey>,
     ecdsa_signing_key: Arc<EcdsaSigningKey>,
     ecdsa_verifying_key: Arc<EcdsaVerifyingKey>,
+    /// Short-TTL cache for attestation reports of **nonce-less** requests only.
+    /// Keyed on (model, algo, tls_fp, provider_filter, signing_address) — never
+    /// on a nonce. moka's `try_get_with` also single-flights concurrent misses,
+    /// so a burst of identical no-nonce probes triggers ONE backend build.
+    /// `None` when disabled (TTL=0). Nonce-bearing requests bypass it entirely
+    /// because the nonce is cryptographically bound into the TDX report_data and
+    /// the GPU evidence — serving a cached report for a different nonce would
+    /// defeat the freshness/replay guarantee.
+    report_cache: Option<moka::future::Cache<String, Arc<AttestationReport>>>,
 }
 
 impl AttestationService {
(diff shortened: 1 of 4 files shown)


Evrard-Nil:
Thanks, but respectfully this one is a false positive — I'm not applying it.

**The claimed compile error doesn't occur.** `moka::future::Cache::try_get_with` does **not** require a `'static` init future. It polls the init future to completion *inline* within the `try_get_with(...).await` call (it does not store the pending future across the caller's await points), so an init future that borrows `&self` is fine — this is a documented moka capability (the init future may borrow from the surrounding scope).

**Evidence:** this exact head commit (`ac507f3`) compiled and ran on CI — **Unit tests, Tests, and Test Suite all green** — plus `cargo check -p services`, `cargo clippy -p services`, `cargo test -p services --lib` (427 pass), and `cargo check -p api` all pass locally. If the closure were rejected for a `'static` violation, none of those would build.

The suggested rewrite (clone the `Arc`s + hoist algo/signing-address/TLS precomputation out of the closure to avoid capturing `self`) is a no-op for correctness here and would move the algo-validation + signing-address derivation out of the `build` closure — added churn with no behavioral or performance benefit, since the closure is invoked exactly once per request either way. The existing optimized nonce handling (decode-once / random-once) is also already in place.

Happy to revisit if anyone can produce an actual build failure.

PierreLeGuen (approved):
The short-TTL cache with single-flight for no-nonce reports is sound and achieves its goal of cutting the `/report` 503 herd. The cache key omits the nonce and is only consulted in the `nonce.is_none()` arm — nonce-bearing requests bypass it entirely — and all report-affecting params (model, algo, tls_fp, provider, signing address) are in the key.

Notes (non-blocking):
- The "compile error" raised by an earlier comment is a false positive: the init closure captures `Arc`-backed fields through `&self` and both `cargo check`/`cargo clippy` pass clean.
- Coalesced single-flight waiters that receive an `Err` are tagged `result:hit` because the init closure never flips the hit flag, so `hit` slightly overcounts during a backend-error burst. Harmless for the metric's intent; read it alongside the `/report` 5xx rate.

Checks: `cargo fmt --all -- --check`, `cargo check -p services`, `cargo clippy -p services -p api`, and `cargo test -p services --lib cache_key` (4/4) all passed. Integration/E2E not run locally (no Postgres).

PR #852, opened 2026-06-25
ci: notify #cvm-alerts on reproducible-build failure

Evrard-Nil:
## What

Adds a `notify-slack` job to `.github/workflows/verify-reproducible-build.yml` that posts a Slack alert to **#cvm-alerts** (`C09EQ848HHC`) whenever any of the `build`, `sequential-builds`, or `verify` jobs fail.

## Why

The reproducible-build verification runs on every push to `main` (and on demand). A failure means the build is no longer bit-for-bit reproducible — a supply-chain signal we want to know about immediately rather than discovering by manually checking the Actions tab.

## How

- New job `notify-slack` with `needs: [build, sequential-builds, verify]` and `if: always() && contains(needs.*.result, 'failure')`.
- Posts via Slack `chat.postMessage` using `secrets.SLACK_BOT_TOKEN`, mirroring the existing pattern in `infra-tests`.
- **No-ops gracefully** (`exit 0`) if `SLACK_BOT_TOKEN` is unset, so the workflow never breaks on forks or before the secret is configured.
- Message includes repo, branch, short commit, per-job results, and a link to the run.

## Follow-up required

`SLACK_BOT_TOKEN` is **not yet configured** on this repo (it's currently a repo-level secret only in `infra-tests`). The notify step will skip until the secret is added:

```
gh secret set SLACK_BOT_TOKEN --repo nearai/cloud-api
```

The bot must be a member of #cvm-alerts.

## Testing

- YAML validated with PyYAML.
- Notification logic mirrors the production `infra-tests` workflow.

### DIFF
--- a/.github/workflows/verify-reproducible-build.yml
+++ b/.github/workflows/verify-reproducible-build.yml
@@ -291,3 +291,34 @@ jobs:
               echo "Build digests do not match across environments or verification failed."
             fi
           } >> "$GITHUB_STEP_SUMMARY"
+
+  # Post a Slack alert to #cvm-alerts when any build/verify job fails.
+  # No-ops gracefully if SLACK_BOT_TOKEN is not configured on the repo.
+  notify-slack:
+    name: Notify Slack on failure
+    needs: [build, sequential-builds, verify]
+    if: always() && contains(needs.*.result, 'failure')
+    runs-on: [self-hosted, infra]
+    steps:
+      - name: Send Slack notification
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        run: |
+          if [ -z "${SLACK_BOT_TOKEN}" ]; then
+            echo "SLACK_BOT_TOKEN not set; skipping Slack notification"
+            exit 0
+          fi
+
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
The new `notify-slack` job is correct: the `needs` array references real jobs (`build`, `sequential-builds`, `verify`), `always() && contains(needs.*.result, 'failure')` is the right pattern, the Slack payload is built safely with `jq -n --arg`, and the token is checked and never echoed. Good to merge.

- `.github/workflows/verify-reproducible-build.yml:312-318`: `${{ github.ref_name }}` and `${{ github.repository }}` are interpolated directly into the double-quoted `MSG="..."` shell block. Branch names can legally contain `` ` `` / `$()`, so this is a classic Actions script-injection vector. Low risk here since the workflow only runs on `push` (the actor already has write access), but passing these via `env:` and referencing `$GITHUB_REF_NAME`/`$GITHUB_REPOSITORY` is cleaner.

Local checks: PyYAML parse and `bash -n`/`shellcheck` on the `run:` scripts passed; `actionlint` passed (with repo-specific ignores for the custom `infra` runner label); confirmed all `needs` job IDs exist; dry-ran the missing-token and Slack-payload paths. No prior human reviews to reconcile.

Issue #42, opened 2025-10-15
Bug: fail to set public name `openai/gpt-oss-120b` for model `nearai/gpt-oss-120b`


think-in-universe:
After #7 is implemented, now [`/admin/models`](https://nearai-platform-api-dev-59501de18ae5.herokuapp.com/docs#/Admin/batch_upsert_models) can set public name for each model. This works well for other models, but only fails when we set the public model name `openai/gpt-oss-120b` for model `nearai/gpt-oss-120b`.

**Request:**

The payload of `/admin/models` to reproduce this issue

```json
{
  "nearai/gpt-oss-120b": {
    "publicName": "openai/gpt-oss-120b"
  }
}
```

**Error Response:**

```json
{
  "error": {
    "message": "Failed to upsert models",
    "type": "internal_server_error",
    "param": null,
    "code": null
  }
}
```

This is pretty weird. It works if we set the public name as `nearai/gpt-oss-120b`, `openai/gpt-oss-121b` or `openai/gpt-oss-20b`. It seems only `openai/gpt-oss-120b` will fail.

**Context:**

The current [`/model/list`](https://nearai-platform-api-dev-59501de18ae5.herokuapp.com/docs#/Models/get_model_by_name) result:

```json
{
  "models": [
    {
      "modelId": "deepseek/deepseek-chat-v3-0324",
      "inputCostPerToken": {
        "amount": 490,
        "scale": 9,
        "currency": "USD"
      },
      "outputCostPerToken": {
        "amount": 1140,
        "scale": 9,
        "currency": "USD"
      },
      "metadata": {
        "verifiable": true,
        "contextLength": 163000,
        "modelDisplayName": "DeepSeek V3 0324",
        "modelDescription": "DeepSeek V3, a 685B-parameter, mixture-of-experts model, is the latest iteration of the flagship chat model family from the DeepSeek team.",
        "modelIcon": "https://avatars.githubusercontent.com/u/148330874?s=200&v=4"
      }
    },
    {
      "modelId": "deepseek/deepseek-chat-v3.1",
      "inputCostPerToken": {
        "amount": 1000,
        "scale": 9,
        "currency": "USD"
      },
      "outputCostPerToken": {
        "amount": 2500,
        "scale": 9,
        "currency": "USD"
      },
      "metadata": {
        "verifiable": true,
        "contextLength": 128000,
        "modelDisplayName": "DeepSeek V3.1",
[...]

PR #862, opened 2026-07-03
dev: local-stack — full tier-routing dev environment (real CPU engine, real token counts)

PierreLeGuen:
Stacked on #860 (rebase onto main once it merges). Adds `dev/local-stack/`: a reusable one-machine environment that runs the **real cloud-api** against a **real inference engine on CPU** with the GLM-5.2-style two-tier context topology — no staging, no GPUs, no TEE.

```
curl ─► cloud-api :13000 (cargo run, mock auth, ephemeral signing keys)
           ├─ base tier ─► tier_shim :18100 (ctx 1000) ──┐
           └─ long tier ─► tier_shim :18101 (ctx 8000) ──┴─► llama.cpp :18090
         postgres :15432                     (docker, CPU, Qwen2.5-0.5B GGUF)
```

**`make up && make api && make seed && make demo`** — the demo prints an 8-check table, all green on this machine:

- small → `+base` (real Qwen reply; long tier sees zero traffic)
- oversize (~3,000 REAL tokens) → `+long`
- boundary (~800 real tokens, byte-heuristic ambiguous) → **exact `/v1/tokenize` count decides** — shim log shows `count=801` keeping it on base where the heuristic alone (~1,201) would have mis-routed
- streaming oversize → long over SSE
- `touch SATURATE-long` → oversize returns retryable **429**, never a context-400 (the #860 clobber fix, observable with curl)

The tier shims emulate the per-tier SGLang fleets faithfully: vLLM/SGLang-shaped `/v1/tokenize` backed by the engine's real tokenizer, SGLang-phrased context-window 400s (the fall-through matcher fires exactly as in prod), `.model` tier tags, saturation toggle. Engine-agnostic: swap a vLLM-CPU image into the compose file and nothing else changes.

**`make proxy-demo`** drives the REAL model-proxy binary (checkout path via `MODEL_PROXY_DIR`) through the long-context registration flow — dual same-IP probes sharing one routed backend, the probe-cleanup **ownership guard**, and the health-gated stub + non-2xx breaker. 9/9 against nearai/model-proxy#42; against unpatched main the shared backend measurably drops to `total_backends: 0` for a full discovery cycle on probe unregister (the #42 repro).

README documents what is intentionally NOT local: the TLS/SNI data path through model-proxy (providers fail closed on unattested HTTPS — the TEE trust model working) and the Chutes wire client (ML-KEM + TDX; covered at the pool boundary by the e2e suite in #860).

No `src/` changes — tooling only.

### DIFF
--- a/dev/local-stack/.gitignore
+++ b/dev/local-stack/.gitignore
@@ -1,0 +1,6 @@
+cache/
+run/
+*.log
+*.pid
+.api-key
+SATURATE-*
--- a/dev/local-stack/Makefile
+++ b/dev/local-stack/Makefile
@@ -1,0 +1,42 @@
+SHELL := /bin/bash
+STACK_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+REPO_ROOT := $(abspath $(STACK_DIR)/../..)
+MODEL_PROXY_DIR ?= $(REPO_ROOT)/../model-proxy
+
+.PHONY: up shims api seed demo proxy-demo down logs
+
+## Bring up the containers (Postgres + llama.cpp engine; first start downloads
+## the ~400MB GGUF into ./cache) and the two tier shims.
+up:
+	cd $(STACK_DIR) && docker compose up -d --wait
+	cd $(STACK_DIR) && (python3 tier_shim.py --tier base --port 18100 --ctx-window 1000 --model-id z-ai/glm-5.2-local > shim-base.log 2>&1 & echo $$! > .shim-base.pid)
+	cd $(STACK_DIR) && (python3 tier_shim.py --tier long --port 18101 --ctx-window 8000 --model-id z-ai/glm-5.2-local > shim-long.log 2>&1 & echo $$! > .shim-long.pid)
+	@sleep 1 && curl -sf http://127.0.0.1:18100/health >/dev/null && curl -sf http://127.0.0.1:18101/health >/dev/null \
+		&& echo "engine + shims up (base :18100 ctx=1000, long :18101 ctx=8000)"
+
+## Run cloud-api in the foreground against the stack (debug build; DEV=1
+## ephemeral signing keys). Ctrl-C to stop.
+api:
+	cd $(REPO_ROOT) && set -a && source $(STACK_DIR)/env.local-stack && set +a && cargo run --bin api
+
+## Seed model + org + API key (needs `make api` running in another terminal).
+seed:
+	$(STACK_DIR)/seed.sh
+
+## Tier-routing demo (small/oversize/boundary/streaming/saturated).
+demo:
+	$(STACK_DIR)/demo.sh
+
+## Model-proxy control-plane demo (dual-probe registration, shared-backend
+## ownership guard, health-gated stub + circuit breaker). Needs a model-proxy
+## checkout at $(MODEL_PROXY_DIR) (override: make proxy-demo MODEL_PROXY_DIR=…).
+proxy-demo:
+	MODEL_PROXY_DIR=$(MODEL_PROXY_DIR) $(STACK_DIR)/model-proxy-demo/run.sh
+
+down:
+	-cd $(STACK_DIR) && for f in .shim-base.pid .shim-long.pid; do [ -f $$f ] && kill $$(cat $$f) 2>/dev/null; rm -f $$f; done
+	-cd $(STACK_DIR) && rm -f SATURATE-base SATURATE-long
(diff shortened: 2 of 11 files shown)


PierreLeGuen:
Rebased onto `main` now that #860 is merged — the diff is dev tooling only (`dev/local-stack/`, no `src/` changes), CI green. Blocked only by the 1-approval branch-protection rule; needs a reviewer's approve to merge (I'm the author, can't self-approve). `make up && make api && make seed && make demo` runs the full tier-routing stack on a real CPU model.

Evrard-Nil (approved):
Thorough read of all 11 files (729 additions, zero `src/` / `Cargo.*` changes). No production code paths are touched — everything lives under `dev/local-stack/`.

## What this does

Adds a self-contained local dev environment for exercising the GLM-5.2-style two-tier context routing end-to-end on one machine without staging, GPUs, or a TEE:

- `docker-compose.yml` — Postgres 16 + llama.cpp (Qwen2.5-0.5B GGUF, CPU) on isolated ports
- `tier_shim.py` — Python HTTP shim that emulates a per-tier SGLang fleet: real `/v1/tokenize` counts, SGLang-phrased context-400s, saturation drill via sentinel file
- `seed.sh` — bootstraps mock admin user, two-tier model config, org/credits/API key via the admin API
- `demo.sh` — 8-check table: small→base, oversize→long, boundary (exact tokenize count), streaming, saturated-long→retryable 429/5xx
- `model-proxy-demo/` — drives the REAL model-proxy binary through the dual-probe / ownership-guard / circuit-breaker flow (model-proxy#42 already merged)

## Security

- All local ports are bound to `127.0.0.1` (loopback only). The llama.cpp container uses `--host 0.0.0.0` internally but the host-side binding is `127.0.0.1:18090`.
- `env.local-stack` values are intentional dev placeholders (`local-stack-not-a-secret`, `local-stack-inference-key`, zero S3 key) with clear comments. No real credentials.
- `.api-key` is gitignored. All generated state (`cache/`, `run/`, logs, PIDs, `SATURATE-*`) is gitignored.
- `DEV=1` uses ephemeral signing keys — correct for a non-TEE machine; this path is already guarded in the existing codebase to be debug-build-only.

## Minor findings (non-blocking for tooling code)

1. **`make up` called twice leaves orphan shims.** The `up` target spawns new shims without killing any existing ones on the same ports. The new processes will fail to bind (port already in use) and the old ones keep running silently — `make up` reports green because it health-checks the still-running originals. `make down && make up` is the correct re-run path, but a short `kill $(cat .shim-*.pid) 2>/dev/null || true` guard at the top of `up` would prevent confusion. Not blocking.

2. **Upstream HTTP errors masked in `tier_shim.py`.** `urllib.request.urlopen()` raises `urllib.error.HTTPError` on non-2xx. This falls through to the blanket `except Exception` in `do_POST` and is re-served as a 500 with the Python exception string. The actual upstream status code and body are lost. For a dev tool this is tolerable (the error text mentions the upstream URL), but catching `HTTPError` separately and forwarding the status + body would make debugging easier.

3. **`demo.sh` missing trap for `SATURATE-long`.** If the script is interrupted between `touch SATURATE-long` (line 53) and `rm -f SATURATE-long` (line 56), the sentinel file lingers and subsequent demo runs fail test 2 (oversize→long). `run.sh` has the same gap for `ENGINE_DOWN`. Both could be fixed with a `trap 'rm -f SATURATE-long' EXIT` / `trap 'rm -f ENGINE_DOWN' EXIT`. The Gemini review flagged this; it appears the current HEAD didn't address it.

4. **`ENGINE_PID` in `run.sh` cleanup is dead code.** `cleanup_all()` references `$ENGINE_PID` but it is never assigned — the "fake engine" role is handled by `probe_stub.py` via the `ENGINE_DOWN` file, not a separate process. The variable is harmless (`:-}` guard) but confusing. Worth removing.

5. **`demo.sh` uses `set -uo` but not `-e`.** Intentional (partial failures are accumulated into the PASS/FAIL table rather than aborting), but worth a comment so readers don't assume it's an oversight.

None of these block the PR — they're all contained within the dev tooling directory and have no impact on CI or production paths.

## Verdict

**Approved.** The stacking dependency (#860) is already merged to main, the branch rebases cleanly, CI (security_audit) is green, no production code is touched, no real credentials, all sockets loopback-only. The tier shim correctly mirrors the SGLang phrasing that cloud-api's fall-through matcher depends on. The findings above are worth a follow-up but don't block merging dev tooling.

PR #866, opened 2026-07-06
fix(chutes): retry transient fallback failures

lloydmak99:
## Problem

Chutes fallback still failed for Gemma/Qwen classes that were not measurement-pin rejects:

- stale `/e2e/invoke` 400s from expired/consumed nonce or instance targeting were collapsed to generic non-retryable `CompletionError`s
- missing `/evidence` for a selected Chutes instance and drained/no-nonce discovery snapshots were also non-retryable
- raw Chutes HTTP transport errors were flattened before the provider pool could classify them as transient

That meant a NEAR primary outage could fall through to Chutes once, hit one of these transient Chutes-side readiness/targeting failures, and stop instead of using the existing bounded retry path.

## Changes

- Adds a small Chutes availability classifier for retryable provider-unavailable cases.
- Maps stale Chutes `/e2e/invoke` 400s to external HTTP 503 so the existing pool retry/backoff path can mint fresh nonce/instance state.
- Maps Chutes transport errors to external HTTP 503 instead of generic completion errors.
- Treats empty E2E-capable instance snapshots, missing `/evidence` entries, and drained nonce pools as retryable Chutes unavailability.
- Keeps attestation, bad E2E pubkey, E2EE request-build, decode, oversized body, missing model/chute id, and truly malformed internal 4xx failures fail-closed/non-retryable.

## Validation

- `cargo test -p inference_providers attested::chutes::`
- `cargo test -p services test_classify_retry_decision`

LSP diagnostics could not run locally because the configured Rust toolchain is missing `rust-analyzer` (`Unknown binary 'rust-analyzer' in official toolchain '1.92.0-aarch64-apple-darwin'`).

## Rollout Notes

This is a Cloud API-only behavior change. It does not replace the infra-tests coverage and does not change Chutes golden measurements; it complements the measurement-pin PR by making transient Chutes availability/targeting failures retryable once Chutes is selected as fallback.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2111,9 +2111,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
--- a/crates/inference_providers/src/attested/chutes/availability.rs
+++ b/crates/inference_providers/src/attested/chutes/availability.rs
@@ -1,0 +1,65 @@
+use crate::CompletionError;
+
+pub(super) fn retryable_provider_unavailable(ctx: &str, reason: &str) -> CompletionError {
+    CompletionError::HttpError {
+        status_code: 503,
+        message: format!("{ctx}: Chutes temporarily unavailable ({reason})"),
+        is_external: true,
+    }
+}
+
+pub(super) fn stale_invoke_target(ctx: &str, body: &str) -> bool {
+    if !ctx.contains("/e2e/invoke") {
+        return false;
+    }
+    let lower = body.to_ascii_lowercase();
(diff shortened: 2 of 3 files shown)


PierreLeGuen (approved):
Well-contained change: transient fallback failures (transport errors, stale-invoke 400s, evidence/nonce-drain paths) now map to retryable `503 HttpError { is_external: true }`, which the pool's `classify_retry_decision` correctly treats as `retryable_http_5xx`. Traced end-to-end; 503 messages carry only stage labels + instance IDs, so no upstream-body leak, and attestation/pubkey/E2EE-build paths stay fail-closed.

- `crates/inference_providers/src/attested/chutes/mod.rs:405` — the new `ChutesClientError::Http(_)` → retryable-503 arm also catches request timeouts on `/e2e/invoke`, where the request may already have reached Chutes; if you want to avoid re-sending expensive inference, consider routing `e.is_timeout() && !e.is_connect()` to the non-retryable `Timeout` path and keeping only connect/transport failures retryable.
- `mod.rs:479-562` — `last_err_retryable` reflects only the last candidate attempted, so an earlier transient failure is masked by a later non-retryable one. Consistent with the existing "last error wins" pattern; OR the flag across the loop if any-candidate-transient should force a retry.

Checks: `cargo test -p inference_providers attested::chutes::` (109 passed, 1 ignored, incl. `map_client_error_retries_stale_invoke_400`), `cargo clippy -p inference_providers` clean, `cargo check` on `services` + full workspace clean, `cargo fmt --check` clean.

Evrard-Nil (approved):

PR #870, opened 2026-07-08
fix(chutes): suppress the final usage husk chunk when include_usage is unset

Evrard-Nil:
## Problem

Streaming a Chutes-served model **without** `stream_options.include_usage` forwards a malformed trailing chunk to the client:

```
data: {"id":"…","object":"chat.completion.chunk","created":…,"model":"minimax/minimax-m2.5","choices":[],"chutes_verification":"…"}
```

`gate_stream_usage` (#781 L1) strips the `usage` **value** from the final usage-only chunk, but still emits the gutted husk (`choices: []`, no `usage`). Real OpenAI sends **no final chunk at all** when `include_usage` wasn't requested.

This was reported by BeeZi in production against `minimax/minimax-m2.5` (A/B-compared with identical requests on other gateways) and explains two of their findings at once:
- **chunk decoding errors** — strict SDK parsers reject the empty-choices/no-usage husk;
- **zero token usage** — cost-tracking clients read the stream as 0 tokens.

Reproduced on prod: non-streaming and `include_usage: true` paths report usage correctly; only the default streaming path emits the husk.

## Fix

- `gate_stream_usage` now returns whether it gated the **final** chunk for a client that did not request usage.
- `rewrite_sse_event_model` suppresses the client-facing `raw_bytes` entirely (empty `Bytes`) for that event, instead of forwarding the husk. The **parsed chunk keeps its `usage`**, so `InterceptStream` billing/limits are unaffected (pinned by the existing `stream_usage_gate_preserves_chunk_usage_for_billing` test, updated).
- `include_usage: true` streams are byte-for-byte unchanged; intermediate-chunk gating and the nearai provider path are untouched.

This also aligns the Chutes path with the documented e2e contract for `include_usage:false` ("no final usage-only chunk is emitted"), previously only enforceable for non-attested models.

## Testing

- `cargo test -p inference_providers --lib` — 371 passed (includes the two updated gate tests pinning suppression + billing preservation).
- `cargo fmt --check` and `cargo clippy -p inference_providers` clean.
- Prod repro of the husk captured before the fix (streaming without `stream_options`, model `minimax/minimax-m2.5`).

### DIFF
--- a/crates/inference_providers/src/attested/chutes/mod.rs
+++ b/crates/inference_providers/src/attested/chutes/mod.rs
@@ -680,12 +680,18 @@ fn strip_internal_response_fields(obj: &mut serde_json::Map<String, Value>) {
 /// - a chunk with an empty `choices` array is the FINAL usage-only chunk → its
 ///   `usage` is kept iff `include_usage` was requested, else stripped.
 ///
+/// Returns `true` when the chunk was the FINAL usage-only chunk and the client did
+/// NOT request usage: OpenAI emits no final usage chunk at all in that case, so the
+/// caller must suppress the whole chunk from the client stream rather than forward
+/// a gutted `choices: []` husk (strict SDK parsers reject it, and cost-tracking
+/// clients read it as zero usage) — see [`rewrite_sse_event_model`].
+///
 /// NOTE: this gates only `raw_bytes` (the bytes the passthrough route forwards to the
 /// client). The parsed `chunk.usage` is left intact so `InterceptStream` can still
 /// read it for billing/limits — see [`rewrite_sse_event_model`].
-fn gate_stream_usage(obj: &mut serde_json::Map<String, Value>, include_usage: bool) {
+fn gate_stream_usage(obj: &mut serde_json::Map<String, Value>, include_usage: bool) -> bool {
     if !obj.contains_key("usage") {
-        return;
+        return false;
     }
     let is_final = obj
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Clean, well-contained fix. `gate_stream_usage` now reports whether it gated the final usage-only chunk for a client that didn't request usage, and `rewrite_sse_event_model` suppresses that husk's `raw_bytes` while preserving the parsed `chunk.usage` so billing is unaffected.

Verified the correctness-critical paths:
- Chutes passthrough (`completions.rs:1629`): the event hits the byte-exact branch and returns empty `raw_bytes`; a zero-length write is a harmless no-op and `[DONE]` still follows in a separate chunk.
- Billing preserved — `ev.chunk` retains `usage` (`stream_usage_gate_preserves_chunk_usage_for_billing`).
- `include_usage:true` path is byte-for-byte unchanged (early return gated behind `suppress_client_chunk`).
- Suppressed event has `chunk: Some(...)` so `is_done_marker()` stays false — no `[DONE]`/bookkeeping confusion.

No blocking issues.

Checks (combined): `cargo test -p inference_providers --lib` → 371 passed, 0 failed (1 ignored live probe); `cargo clippy -p inference_providers` clean; `cargo fmt --check` clean; `git diff --check` clean. CI lint/unit/integration/security passed; E2E failure was unrelated `admin_analytics` tests.

PR #871, opened 2026-07-08
fix(billing): make cache_read_cost_per_token nullable — 0 no longer means disabled

Evrard-Nil:
## Problem (RFC infra#171 §2.5, asked by @PierreLeGuen)

`cache_read_cost_per_token` is `BIGINT NOT NULL DEFAULT 0` with `0` overloaded to mean "cache pricing disabled → bill cached tokens at the full input rate". Two consequences:

- a genuinely **free** cache-read price (0) is inexpressible;
- an admin setting a model's cache price to 0 (intending "free") silently bills customers the **full input rate** for cached tokens.

## Change

New semantics, threaded end-to-end (DB → repositories → services → admin API → public catalog):

| value | billing | public catalog `input_cache_read` |
|---|---|---|
| `NULL` / `None` | cached tokens at `input_cost_per_token` (= today's `0`) | omitted |
| `Some(0)` | genuinely free | `"0"` |
| `Some(x)` | rate `x` | rendered |

- **Migration `V0068`**: `DROP NOT NULL`/`DROP DEFAULT` + `0 → NULL` on `models`, `model_history`, and `scheduled_model_pricing_changes.old_cache_read_cost_per_token`. Billing for every existing model is **byte-identical** after migration.
- **Admin update API is tri-state** via the existing `Nullable<T>`/`deserialize_nullable` pattern (same as `isReady`/`deprecationDate`): field absent = unchanged, explicit `null` = disable, value = set. SQL uses the existing clear-flag `CASE WHEN` pattern.
- **Scheduled pricing changes**: `new_cache_read_cost_per_token` NULL already means "field unchanged" (V0058) — semantics preserved exactly; a scheduled change can set but not disable cache pricing (documented in the migration and scheduler). `old_` is a snapshot and follows the model column.
- **Activation gate**: `None` and `Some(0)` both count as "no revenue-bearing cache price" (unchanged revenue gating).

## Heads-up for API consumers

`cacheReadCostPerToken` / `input_cache_read` in `/v1/models`, `/v1/model/list`, `/v1/admin/models` and history/pricing-change snapshots is now **omitted** when cache pricing is disabled (previously rendered as amount 0). nearai-cloud-ui / nearai-cloud-admin-ui should tolerate the missing key.

## Testing

- `cargo test --workspace --lib --bins`: **1075 passed, 0 failed** (api 216, config 27, database 15, inference_providers 371, services 446), incl. new unit tests: `None` → full input rate; `Some(0)` → free; `Some(x)` → discounted; catalog renders `Some(0)` as `"0"` and omits `None`.
- `cargo check --workspace --all-targets` (e2e compilation) clean; `cargo fmt --check` clean; `cargo clippy --workspace --lib --bins` zero warnings.
- Migration is standard DDL matching existing patterns; e2e suite runs in CI (needs PostgreSQL).

### DIFF
--- a/crates/api/src/models.rs
+++ b/crates/api/src/models.rs
@@ -1053,7 +1053,7 @@ pub struct ModelPricing {
     /// OpenRouter: USD per request, as a string ("0" when not applicable).
     pub request: String,
     /// OpenRouter: USD per cached input token, as a string. Omitted when cache-read
-    /// pricing is disabled or unknown.
+    /// pricing is disabled; `"0"` means cached input tokens are genuinely free.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub input_cache_read: Option<String>,
 }
@@ -3300,8 +3300,13 @@ pub struct AdminModelWithPricing {
     pub output_cost_per_token: DecimalPrice,
     #[serde(rename = "costPerImage")]
     pub cost_per_image: DecimalPrice,
-    #[serde(rename = "cacheReadCostPerToken")]
-    pub cache_read_cost_per_token: DecimalPrice,
+    /// Cost per cached input token. Omitted when cache pricing is disabled;
+    /// an amount of 0 means cached tokens are genuinely free.
+    #[serde(
+        rename = "cacheReadCostPerToken",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub cache_read_cost_per_token: Option<DecimalPrice>,
     pub metadata: ModelMetadata,
     #[serde(rename = "isActive")]
     pub is_active: bool,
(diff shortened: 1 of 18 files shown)


PierreLeGuen (approved):
Type-driven, compiler-enforced change that makes `cache_read_cost_per_token` nullable so `0` can mean "free" rather than "disabled". The tri-state `Nullable<T>` chain, SQL migration, billing logic, activation gate, and scheduler mapping all check out; the `0 → NULL` migration is billing-neutral since `0` previously always meant disabled.

Non-blocking follow-up:
- API contract change (`crates/api/src/models.rs:24-26`): `cacheReadCostPerToken` / `input_cache_read` are now omitted from `/v1/models`, `/v1/admin/models`, and pricing-history responses when cache pricing is disabled (previously rendered as `0`). Intentional and flagged in the PR body — just confirm the UI/admin-ui deploys tolerate the missing key.

Checks: `cargo check --workspace --all-targets`, `cargo clippy --workspace --lib --bins`, and `cargo fmt --check` all clean; `cargo test --lib --bins` passed (1075 tests, incl. `compute_token_cost` and the new free-cache-read test). e2e tests compile but need PostgreSQL, so were not run.

PR #872, opened 2026-07-08
fix(models): tag privacy-filter as a non-completion (classification) model instead of hiding it

### LINKED ISSUE #615: openai/privacy-filter listed in /v1/models but every chat-completion attempt errors with upstream 404
## Summary

`GET /v1/models` returns `openai/privacy-filter` as an available model with pricing, but **every** `POST /v1/chat/completions` against it returns:

```
HTTP 400
{"error":{"message":"Provider failed for model 'openai/privacy-filter': Upstream request failed with status 404 Not Found","type":"invalid_request_error","param":"model","code":null}}
```

This is the model from [`cvm-compose-files/small-models.yaml:174-`](https://github.com/nearai/cvm-compose-files/blob/main/small-models.yaml#L174) — a HuggingFace **token-classification** pipeline (PII detector), not a generative chat model. Its container exposes a custom FastAPI surface, **not** `/v1/chat/completions`. Routing a chat request to it hits a non-existent path → upstream 404.

## Two distinct issues

1. **Catalog listing is wrong.** A model that doesn't accept the standard chat completions request should not appear in `/v1/models` (or should at least carry a `task` / `endpoint` hint so clients know it's special). Today's `/v1/models` payload doesn't expose that distinction — every entry looks like a chat model to a client.

2. **Error categorization is wrong.** A backend missing the endpoint surfaces as `400 invalid_request_error`. But the user's request shape *is* valid — the issue is server-side routing/availability. This should be `404` or `503 model_unavailable`, not a 4xx client-error. Same systemic wrapping pattern as #605/#606/#607/#608/#609.

The 4th probe (a system prompt + question) took **78 seconds** before erroring, vs ~1.3 s for the others — suggests internal retry-and-timeout behavior across multiple backend instances when both are misregistered.

## Repro

```bash
  -d '{"model":"openai/privacy-filter","messages":[{"role":"user","content":"hello"}],"max_tokens":50}'
# 400 — upstream 404 wrapped as invalid_request_error
```

## Suggested fix

- Either remove `openai/privacy-filter` from the catalog (if it's not meant to be public), or give catalog entries an `endpoint`/`task` field and have clients discover the right URL.
- Map "upstream 404 for the configured model endpoint" to `503 model_unavailable` (or `404 model_endpoint_not_found`) instead of `400`.

---

Evrard-Nil:
## Problem

`GET /v1/models` (and `GET /v1/model/list`) advertised `openai/privacy-filter` as if it were an ordinary chat model. It is a **token-classification** (PII detection) model: its only real endpoints are `/v1/privacy/classify` and `/v1/privacy/redact`, and a `/v1/chat/completions` request against it hits a non-existent upstream path and 404s. Because the catalog exposed no notion of its kind, a client had no way to tell it apart from a completion model.

Fixes #615.

## Approach (revised per maintainer feedback)

> Maintainer feedback on the earlier version of this PR: *"Model should be there but tagged as not a completion model. Same as image models. Hiding it is not good."*

This PR **supersedes** its earlier hide-based version. Instead of filtering the model out of the catalog, it keeps the model **listed** and **tags it with its true modality** — exactly the way image and embedding models are represented ("list + tag, same as image models").

### Why tagging alone is sufficient

The catalog already encodes model kind in `output_modalities`, and **both listing endpoints already surface that field to clients**:

- `GET /v1/models` → each `ModelInfo` carries both the OpenRouter-flat `output_modalities` **and** the nested `architecture.outputModalities`.
- `GET /v1/model/list` → each model's `architecture.outputModalities` (built by `ModelArchitecture::from_options`).

An image model is listed with `outputModalities = ["image"]`; an embedding model with `["embedding"]`. The privacy filter was simply mislabeled `["text"]`, so it masqueraded as a chat model. Relabeling its output modality to `["classification"]` makes it self-describe as a classifier — a client distinguishes it by reading `output_modalities`, exactly as it would for an image model. No new field or scheme is introduced; this reuses the existing modality mechanism.

## Changes

- **Migration `V0069`** relabels the `openai/privacy-filter` row's `output_modalities` from the mislabeled `["text"]` to `["classification"]` (and mirrors the fix into the currently-open `model_history` snapshot). This is the whole fix. It is idempotent, a no-op on databases without the row (fresh/test DBs), and mirrors the in-place catalog-metadata repair pattern used by `V0061`. Input modality stays `["text"]` — it still consumes text. The migration comment block states its purpose is to correctly **tag** the model's modality, not to feed an api-layer hide filter.
  - *Renumbered `V0068` → `V0069`*: since this branch first opened, `main` landed its own `V0068` (`V0068__cache_read_cost_nullable.sql`). Two migrations sharing version `0068` would make refinery reject startup, so this one moves to the next free version. The branch has also been **rebased onto current `main`** so it carries main's `V0068` and this `V0069` side by side.
- **Removed** the catalog-hiding code from the earlier revision: the `.filter(...)` in `routes::completions::models` (`/v1/models`) and `routes::models::list_models` (`/v1/model/list`), plus the now-unused `is_listed_in_model_catalog` helper, the `NON_GENERATIVE_OUTPUT_MODALITIES` const, and its unit test in `routes::common`. (Net effect: `common.rs` and `models.rs` are unchanged from `main`; `completions.rs` gains only an explanatory comment.)

The privacy endpoints and direct `GET /v1/model/{name}` lookups were never affected either way.

> Note: the branch name (`fix/models-exclude-non-generative`) is now a slight misnomer — the fix tags rather than excludes/hides. Kept to preserve PR #872's URL and history.

> Note: the issue also raised (2) error categorization (upstream 404 → `400`). That is a separate systemic wrapping concern shared with #605–#609 and is intentionally out of scope here.

## Tests

- Replaced `test_non_generative_models_excluded_from_models_list` (asserted exclusion) with **`test_classification_model_listed_and_tagged`**: asserts `openai/privacy-filter` **is listed** in `/v1/models` and reports `output_modalities` / `architecture.outputModalities` = `["classification"]`, alongside a chat model still reporting `["text"]`.
- Adjusted `test_models_api` to locate the configured chat model **by id** rather than `data.first()`. With the classifier no longer hidden, the shared e2e catalog can now contain a zero-priced classification model that may sort first under some collations; keying off the first entry was order-dependent. Every other listing test already scopes by id.

## Validation (rebased state: `main` + this change)

- `cargo check --workspace --tests` — clean
- `cargo clippy --workspace --tests` — clean
- **e2e against real Postgres**: full migration set (including main's `V0068` and this `V0069`) applies cleanly on a fresh bootstrap. `test_classification_model_listed_and_tagged` and `test_models_api` both pass, including against a catalog already containing the zero-priced classifier.

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -2680,6 +2680,12 @@ pub async fn models(
 
     let response = ModelsResponse {
         object: "list".to_string(),
+        // Every active model is listed, including non-generative ones such as
+        // the `openai/privacy-filter` token-classification model. Clients tell
+        // model kinds apart from the per-model `output_modalities` /
+        // `architecture.outputModalities` fields — a classification model
+        // reports `["classification"]`, an image model `["image"]` — rather
+        // than by absence from the catalog (issue #615).
         data: models.into_iter().map(model_with_pricing_to_info).collect(),
     };
     Ok(ResponseJson(response))
--- a/crates/api/tests/e2e_all/general.rs
+++ b/crates/api/tests/e2e_all/general.rs
@@ -9,14 +9,21 @@ use inference_providers::{models::ChatCompletionChunk, StreamChunk};
 #[tokio::test]
 async fn test_models_api() {
     let server = setup_test_server().await;
-    setup_qwen_model(&server).await;
+    let chat_id = setup_qwen_model(&server).await;
     let (api_key, _) = create_org_and_api_key(&server).await;
     let response = list_models(&server, api_key).await;
 
     assert!(!response.data.is_empty());
 
-    // Verify pricing and context_length are present (HuggingFace integration)
-    let model = response.data.first().unwrap();
+    // Assert pricing/context_length on the model we configured, located by id.
+    // The e2e suite shares one model catalog, which can also contain other
+    // models (e.g. a zero-priced classification model), so keying off whichever
+    // entry happens to sort first would be order- and collation-dependent.
+    let model = response
+        .data
+        .iter()
+        .find(|m| m.id == chat_id)
+        .expect("configured chat model must be listed");
     assert!(model.pricing.is_some(), "Model should have pricing");
     let pricing = model.pricing.as_ref().unwrap();
     assert!(pricing.input > 0.0, "Input price should be positive");
@@ -28,6 +35,83 @@ async fn test_models_api() {
     assert!(
         model.context_length.unwrap() > 0,
         "Context length should be positive"
+    );
+}
+
+/// A token-classification model (like `openai/privacy-filter`) does not serve
+/// `/v1/chat/completions` — it exposes bespoke `/v1/privacy/*` routes — but it
+/// must still be LISTED in `GET /v1/models`, tagged with its true modality, so a
+/// client can tell it is not a completion model. This mirrors how image models
+/// are listed and tagged (`outputModalities = ["image"]`) rather than hidden: it
+/// is distinguished by its `output_modalities` / `architecture.outputModalities`
+/// reporting `["classification"]`, alongside an ordinary chat model that still
+/// reports `["text"]` (issue #615).
+#[tokio::test]
+async fn test_classification_model_listed_and_tagged() {
+    let server = setup_test_server().await;
+    let chat_id = setup_qwen_model(&server).await;
+
+    let classifier = "openai/privacy-filter".to_string();
+    let mut batch = BatchUpdateModelApiRequest::new();
+    batch.insert(
+        classifier.clone(),
+        serde_json::from_value(serde_json::json!({
+            "inputCostPerToken": { "amount": 1_000_000, "currency": "USD" },
+            "outputCostPerToken": { "amount": 0, "currency": "USD" },
+            "modelDisplayName": "Privacy Filter",
+            "modelDescription": "PII span detection (token classification)",
+            "contextLength": 512,
+            "maxOutputLength": 1024,
+            "verifiable": false,
+            "isActive": true,
+            "inputModalities": ["text"],
+            "outputModalities": ["classification"]
+        }))
+        .unwrap(),
+    );
+    admin_batch_upsert_models(&server, batch, get_session_id()).await;
+
+    let (api_key, _) = create_org_and_api_key(&server).await;
+    let response = list_models(&server, api_key).await;
+
+    // Both models are LISTED — the classifier is tagged, not hidden.
+    let chat = response
+        .data
+        .iter()
+        .find(|m| m.id == chat_id)
+        .expect("generative chat model must be listed");
+    let filter = response
+        .data
+        .iter()
+        .find(|m| m.id == classifier)
+        .unwrap_or_else(|| {
+            let ids: Vec<&str> = response.data.iter().map(|m| m.id.as_str()).collect();
+            panic!("classification model must be listed (tagged, not hidden); got {ids:?}")
+        });
+
+    // The classification model carries its true modality so a client can tell it
+    // is not a completion model — exactly how an image model reports ["image"].
+    // Both the OpenRouter-flat field and the nested architecture surface it.
+    assert_eq!(
+        filter.output_modalities,
+        Some(vec!["classification".to_string()]),
+        "privacy-filter must report output_modalities = [\"classification\"]"
+    );
+    assert_eq!(
+        filter
+            .architecture
+            .as_ref()
+            .map(|a| a.output_modalities.clone()),
+        Some(vec!["classification".to_string()]),
+        "privacy-filter architecture.outputModalities must be [\"classification\"]"
+    );
+
+    // The ordinary chat model still reports text output — the contrast a client
+    // uses to distinguish a completion model from the classifier.
+    assert_eq!(
+        chat.output_modalities,
+        Some(vec!["text".to_string()]),
+        "chat model must report output_modalities = [\"text\"]"
     );
 }
 
--- a/crates/database/src/migrations/sql/V0069__mark_privacy_filter_non_generative.sql
+++ b/crates/database/src/migrations/sql/V0069__mark_privacy_filter_non_generative.sql
@@ -1,0 +1,44 @@
+-- `openai/privacy-filter` is a token-classification (PII detection) model, not a
+-- generative chat model. Its only real endpoints are /v1/privacy/classify and
+-- /v1/privacy/redact; a /v1/chat/completions request against it returns an
+-- upstream 404. It was nonetheless seeded with output_modalities = {'text'}, so
+-- in GET /v1/models (and GET /v1/model/list) it was indistinguishable from an
+-- ordinary chat model — clients had no way to tell it is not a completion model
+-- (issue #615).
+--
+-- This migration TAGS the model with its true modality; it does NOT hide it. The
+-- catalog already encodes model kind in output_modalities, and both listing
+-- endpoints surface that field to clients (embedding models report
+-- {'embedding'}, image models report {'image'}). The privacy filter keeps
+-- appearing in the catalog exactly like an image or embedding model; correcting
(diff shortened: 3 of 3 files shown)


PierreLeGuen (changes requested):
The change itself is well-scoped and correct — a denylist filter on `output_modalities` applied to both user-facing listing endpoints (`GET /v1/models` in `completions.rs`, `list_models` in `models.rs`), safe by construction for models with no declared modalities or mixed modalities, with admin listing unaffected. The blocker is the migration, which fails to apply against the actual column types.

- **Blocking** — `crates/database/src/migrations/sql/V0068__mark_privacy_filter_non_generative.sql:29`: the migration writes/compares `ARRAY['classification']::TEXT[]` against `output_modalities`, but V0043 defines both `models.output_modalities` and `model_history.output_modalities` as `JSONB`. CI E2E fails on V68 with `ERROR: operator does not exist: jsonb = text[]` (run 28957601638), so merging breaks migrations at deploy startup. Use JSONB throughout on both the `models` and `model_history` updates, e.g. set `output_modalities = '["classification"]'::jsonb` and guard with `IS DISTINCT FROM '["classification"]'::jsonb`.
- Non-blocking — correct the catalog seed source for `openai/privacy-filter` to `outputModalities: ["classification"]` as well; a future re-upsert that explicitly re-sends `["text"]` would revert this migration's effect.

Checks run: `cargo check -p api` passed and LSP diagnostics were clean on all changed files; CI inspected via `gh run view`, where lint/unit/integration passed and E2E failed on the migration above. Full test suite / migration apply not run locally (requires PostgreSQL).

Evrard-Nil:
Thanks @PierreLeGuen — I believe the blocking item was fixed in the current commit (`8911538d`), which landed right after your review, so the review was against the earlier `TEXT[]` version:

- **V0068 migration (blocking)** — now JSONB throughout. Both the `models` and `model_history` updates set `output_modalities = '["classification"]'::jsonb` and are guarded with `IS DISTINCT FROM '["classification"]'::jsonb` (the form you suggested). The `operator does not exist: jsonb = text[]` failure is gone and **E2E is green** on the current commit.
- **`Option<&[String]>` / `.as_deref()` (gemini nits)** — done: the helper is now `is_listed_in_model_catalog(output_modalities: Option<&[String]>)`, the `completions.rs` and `models.rs` call sites pass `m.output_modalities.as_deref()`, and the unit test helper + assertions were updated to match.
- **Non-blocking — catalog seed source** — good catch. `output_modalities` is populated from model discovery, so a future re-upsert that re-sends `["text"]` would revert this migration. I think the durable fix belongs at the discovery/registry source and would suggest a follow-up rather than expanding this PR — happy to file a tracking issue and link it here. Let me know if you'd prefer it handled in-scope.

Re-requesting review — PTAL when you get a chance.

PierreLeGuen (approved):
The change correctly excludes non-generative models (e.g. `privacy-filter`) from both public listing endpoints, and the earlier blocking migration issue is resolved. Approving.

- The filter in `is_listed_in_model_catalog` hides a model only when *every* declared modality is non-generative; `None`/empty modalities default to listed (backfill-safe), and mixed lists like `["classification","text"]` stay listed. Both `/v1/models` (completions.rs) and `/v1/model/list` (models.rs, filtered before pagination so `total` stays consistent) are covered, and the privacy/auto-redact resolution paths are unaffected.
- Migration `V0068__mark_privacy_filter_non_generative.sql`: now uses JSONB throughout (`'["classification"]'::jsonb`) matching the JSONB `output_modalities` columns from V0043, is idempotent via `IS DISTINCT FROM`, and is consecutively numbered after V0067 — the prior `TEXT[]`-vs-JSONB blocker is fixed.

Optional follow-up (non-blocking): the relabel in V0068 is durable only while the model-discovery/seed source doesn't re-upsert `openai/privacy-filter` with an explicit `outputModalities: ["text"]`. An omit-the-field re-upsert is safe (`COALESCE(EXCLUDED...)`), but a re-send of `["text"]` would silently revert it. Already tracked in-thread; worth fixing at the discovery source to emit `["classification"]`.

Checks: `cargo check --workspace --tests` and `cargo test -p api --lib test_is_listed_in_model_catalog` passed; LSP diagnostics on common.rs/completions.rs/models.rs clean; CI (lint, unit, integration, e2e, security audit) green on `8911538d`.

Evrard-Nil:
Revised per maintainer feedback: *"Model should be there but tagged as not a completion model. Same as image models. Hiding it is not good."* This supersedes the earlier hide-based version.

**What changed vs the hide approach**
- Dropped the catalog-hiding filter entirely: removed the `.filter(...)` from both `/v1/models` (`routes::completions::models`) and `/v1/model/list` (`routes::models::list_models`), plus the now-unused `is_listed_in_model_catalog` helper, the `NON_GENERATIVE_OUTPUT_MODALITIES` const, and its unit test. Net result: `common.rs` and `models.rs` are unchanged from `main`; `completions.rs` gains only a comment.
- `openai/privacy-filter` now stays **listed** in both endpoints and is **tagged** with its true modality.

**How it mirrors image models**
Both listing endpoints already expose `output_modalities` to clients — `/v1/models` via `ModelInfo.output_modalities` and the nested `architecture.outputModalities`; `/v1/model/list` via `ModelMetadata.architecture.outputModalities` (`ModelArchitecture::from_options`). An image model reports `["image"]`; migration `V0069` relabels privacy-filter from the mislabeled `["text"]` to `["classification"]`, so it self-describes as a classifier exactly the way image/embedding models do. No new field or scheme was introduced.

**Migration renumbered + branch rebased**
`main` landed its own `V0068` (`cache_read_cost_nullable`) after this branch first opened, so this migration moved `V0068 → V0069` to avoid a duplicate refinery version. The branch was rebased onto current `main`, so it now carries main's `V0068` and this `V0069` side by side.

**Tests**
- `test_non_generative_models_excluded_from_models_list` → `test_classification_model_listed_and_tagged` (asserts the model is listed **and** tagged `["classification"]`, alongside a chat model reporting `["text"]`).
- `test_models_api` now locates the configured chat model by id instead of `data.first()` — robust to the shared e2e catalog now containing the zero-priced classifier (previously hidden, so no listing test ever saw it).

**Validation** (rebased state = `main` + this change): `cargo fmt --all -- --check`, `cargo check --workspace --tests`, and `cargo clippy --workspace --tests` all clean; e2e against real Postgres — the full migration set (main's `V0068` + this `V0069`) applies cleanly on a fresh bootstrap, and both tests pass.

Note: the branch name (`fix/models-exclude-non-generative`) is now a slight misnomer (the fix tags rather than excludes); kept to preserve this PR's URL and history.

PierreLeGuen (approved):
This PR relabels `openai/privacy-filter` as a non-completion (classification) model so it is listed and tagged by modality rather than hidden. The revised tag-based approach resolves the earlier blocking `jsonb = text[]` migration issue; the change (migration V0069 + a clarifying comment + e2e tests) is sound and safe to merge.

- Migration `V0069__mark_privacy_filter_non_generative.sql` correctly uses `'["classification"]'::jsonb` against the JSONB `output_modalities` columns (defined in V0043) on both `models` and the open `model_history` snapshot, with `IS DISTINCT FROM` idempotency guards; it is a valid no-op on DBs lacking the row.
- Relabeling is safe: `output_modalities` is used only in DB read/write plumbing and catalog-response mapping — the models listing (`WHERE is_active = true`) and chat completions routing (`resolve_and_get_model`) do not gate on it, so `/v1/privacy/*`, completions, and model lookup are unaffected.

Non-blocking follow-up:
- `V0069`: the relabel is durable only until model-discovery/seed re-upserts `openai/privacy-filter`. An omit-the-field re-upsert is safe via `COALESCE`, but re-sending `["text"]` would silently revert it. Not introduced by this PR and already tracked as a follow-up at the discovery source — noted for visibility only.

Local checks: `cargo fmt --check`, `cargo check`/`clippy -D warnings` on the affected crates, and the `general::test_models_api` + `general::test_classification_model_listed_and_tagged` e2e tests all passed (isolated test DB, `DEV=1`).

PR #881, opened 2026-07-11
fix(attestation): resync 3 missing v1.3.1 hardware rows for Chutes

henrypark133:
## Summary
- `blackbox-attestation` probe alerted `probe_failed` on `https://cloud-api.near.ai/v1/attestation/report?model=z-ai/glm-5.2` — a genuine GLM-5.2-TEE instance on `8xb200 [10.2.1, XEON6, SNC3]` was rejected with `observed measurements match no accepted Chutes config`.
- Confirmed via `GET https://api.chutes.ai/servers/tee/measurements`: the failing row's MRTD/RTMR1/RTMR2/RTMR3 match our already-trusted v1.3.1 software identity byte-for-byte — only RTMR0 (the per-hardware register) was missing from the pinned allow-list in `vetted_golden_measurements()`.
- The same snapshot pull also surfaced two sibling rows from the identical v1.3.1 family that were missing: `8xRTX_PRO_6000 [10.2.1, NUMA2]` and `8xRTX_PRO_6000 [10.2.1, FLAT]`. Added all three per the file's own guidance to resync the full published family, not just the observed-failing row.
- No new software identity is trusted — same MRTD/RTMR1/RTMR2/RTMR3 as the existing pinned v1.3.1 rows; this only adds hardware-variant register rows Chutes had already published for that release.

## Test plan
- [x] `cargo test --lib -p services attestation::chutes` — 15/15 pass, including new `accepts_the_2026_07_11_resynced_rows`
- [x] `cargo clippy -p services --lib` — clean
- [ ] Confirm `blackbox-attestation` probe clears after deploy

### DIFF
--- a/crates/services/src/attestation/chutes.rs
+++ b/crates/services/src/attestation/chutes.rs
@@ -240,6 +240,14 @@ impl ChutesInstanceVerifier for ChutesBackendVerifier {
 ///   previous snapshot, so every instance failed closed.
 /// - `8xh200 [10.1.0-flat] v1.3.1` — **GLM-5.1-TEE**, observed live the same way
 ///   (2026-07-06).
+/// - `8xb200 [10.2.1, XEON6, SNC3] v1.3.1` — **GLM-5.2-TEE**, observed live on
+///   prod fallback attempts (2026-07-11): full register set matches the
+///   published final v1.3.1 identity byte-for-byte; missing from the previous
+///   snapshot, so the instance failed closed (probe alert
+///   `probe_failed`/`blackbox-attestation`). Resynced from
+///   `GET /servers/tee/measurements` along with two other rows the previous
+///   snapshot had missed for the same v1.3.1 family: `8xRTX_PRO_6000 [10.2.1,
+///   NUMA2]` and `8xRTX_PRO_6000 [10.2.1, FLAT]`.
 ///
 /// ⚠️ Chutes re-measures and re-publishes hardware rows *within* a release as
 /// they roll out platform updates — the bracketed suffixes (`[10.2.1]`,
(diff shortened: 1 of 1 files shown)


crates/services/src/attestation/chutes.rs
```diff
@@ -336,6 +344,12 @@ pub fn vetted_golden_measurements() -> ChutesMeasurementPolicy {
                 ("8xb200 [10.2.1, XEON6]", "65fd972e40ac4d8a933d10ebfc31f07336cf1e45a3864523427b454df5d3b9dd0043f10e975da39677b62d45860c13e3"),
                 ("8xb200 [10.1.0, XEON6, SNC3]", "2b22fa53ace208d4f046ae90b7ad28d71a7f4ef0573897d40f6c82b4036217e3170c856f91e54bc19c20c9958c5d1e36"),
                 ("8xb300", "91adf9667ba4c65bec5345a8c9b98010708d903847bf838c4526c3ebbc35561719e2127e48a3f6f77f651d71d2cbc8d4"),
+                // Resynced 2026-07-11 (probe_failed/blackbox-attestation incident):
+                // three more v1.3.1 rows Chutes had published that the previous
+                // snapshot missed. Same family MRTD/RTMR1/RTMR2/RTMR3 as above.
```
lloydmak99, line 349: **High — the resync still omits a published v1.3.1 row.** I refreshed [the public Chutes measurement endpoint](https://api.chutes.ai/servers/tee/measurements) at review time (2026-07-16). Its final-v1.3.1 family now also contains `8xb200 [10.2.1, XEON6, 272CPU]` with RTMR0 `9673907ceb0c9ca79337437bb91695e7a3d19e82df1e41de1b0d2db8081fccb5d82f26d479a5016553cb20964d5948b9`, under the same MRTD/RTMR1/RTMR2/RTMR3 already trusted here. This value is absent from the policy, and `ChutesMeasurementPolicy::verify` requires an exact RTMR0 match, so a genuine instance scheduled on that configuration will still fail with `NoMatch` and disable the fallback path in exactly the way this PR intends to fix. Please add this row, cover it in the regression test, and update the count assertion.

crates/services/src/attestation/chutes.rs
```diff
@@ -608,6 +628,44 @@ mod tests {
             );
         }
 
+        #[test]
+        fn accepts_the_2026_07_11_resynced_rows() {
+            // Regression guard for the 2026-07-11 incident: GLM-5.2-TEE scheduled
+            // on `8xb200 [10.2.1, XEON6, SNC3]` was rejected with "observed
+            // measurements match no accepted Chutes config" (probe_failed on
+            // blackbox-attestation) because the previous snapshot never published
```
lloydmak99, line 636: **Nit:** “the previous snapshot never published this row” reverses ownership: Chutes publishes the row; our pinned snapshot missed it. Please use “the previous snapshot never pinned this published row” (or “never included this published row”) so the next resync is not misled.

lloydmak99 (changes requested):
## Review summary

Requesting changes: **1 High** and **1 Nit** across five review lenses.

The three added rows are well-formed and reuse the already trusted final-v1.3.1 MRTD/RTMR1/RTMR2/RTMR3. However, a live refresh of the public Chutes measurements on 2026-07-16 shows another final-v1.3.1 hardware row that this “full family” resync omits. Because the policy requires an exact five-register match, that genuine configuration will still fail closed with `NoMatch`.

### Findings
- **Correctness (High):** add the currently published `8xb200 [10.2.1, XEON6, 272CPU]` row and regression coverage before merging.
- **Conventions (Nit):** correct the regression comment so the pinned snapshot, rather than Chutes, is described as having missed the published row.

### Validation
- `cargo test --lib -p services attestation::chutes`: 15 passed.
- `cargo fmt --all -- --check`: passed.
- `cargo clippy -p services --lib -- -D warnings`: passed.
- Security, performance/concurrency, and test-coverage lenses found no additional issues.
- GitHub lint, unit, and integration checks are green. The E2E job is red on an unrelated `admin_list_models` database-connection failure and should be rerun before merge.

PR #886, opened 2026-07-16
fix(attestation): omit unsupported dstack event log for ITA

PierreLeGuen:
## Summary

- omit dstack's event log from Intel Trust Authority TDX evidence because it is not an ITA-supported raw CCEL/NEL representation
- keep quote decoding, nonce-bound runtime data, and report_data verification unchanged
- add an exact wire-format regression proving a non-empty dstack JSON event log is not forwarded

## Root cause

The initial staging failure happened because Cloud API tried to hex-decode dstack's JSON event log. Directly base64-encoding that JSON removes the local decode error, but still sends an unsupported representation: Intel's TDX adapter populates `EventLog` from raw `GetCcel()` bytes, while dstack exposes a distinct simplified JSON schema.

Because `event_log` is optional in the ITA request, this PR now omits it until Cloud API can obtain raw CCEL/NEL bytes or implement a documented conversion. Quote and runtime-data attestation remain enabled.

## Validation

- exact request regression validates a complete non-empty fixture through `dstack_sdk_types::dstack::EventLog` and requires `tdx.event_log` to be absent
- red/green: the schema assertion failed on the partial fixture with `missing field event`, then passed after adding `event` and `event_payload`
- malformed/missing quote and report_data nonce-binding coverage remains unchanged
- `cargo test -p services attestation::ita`: 41 passed
- focused regression, rustfmt, strict services clippy, and `git diff --check`: passed
- remote lint, unit, integration, E2E, and aggregate Test Suite: passed on `20d63ee7`
- `cargo deny advisories`: passed

## Rollout

Once merged and promoted, redeploy Cloud API staging and verify the live ITA JWT signature against the returned JWKS without logging token material.

### DIFF
--- a/crates/services/src/attestation/ita/evidence.rs
+++ b/crates/services/src/attestation/ita/evidence.rs
@@ -114,13 +114,11 @@ fn build_tdx_evidence(
     let quote = decode_required_hex("tdx.quote", &gateway.intel_quote)?;
     let runtime_data = runtime_data_bytes(gateway)?;
     verify_report_data(gateway, verifier_nonce, &runtime_data)?;
-    let event_log = gateway.event_log.trim();
-    let event_log =
-        (!event_log.is_empty() && event_log != "0x").then(|| STANDARD.encode(event_log.as_bytes()));
     Ok(ItaTdxEvidence {
         quote: STANDARD.encode(quote),
         runtime_data: STANDARD.encode(runtime_data),
-        event_log,
+        // Dstack's JSON event log is not the raw CCEL/NEL format accepted by ITA.
+        event_log: None,
         verifier_nonce: verifier_nonce.clone(),
     })
 }
--- a/crates/services/src/attestation/ita/evidence.rs
+++ b/crates/services/src/attestation/ita/evidence.rs
@@ -114,11 +114,12 @@ fn build_tdx_evidence(
     let quote = decode_required_hex("tdx.quote", &gateway.intel_quote)?;
     let runtime_data = runtime_data_bytes(gateway)?;
     verify_report_data(gateway, verifier_nonce, &runtime_data)?;
-    let event_log = decode_optional_hex("tdx.event_log", &gateway.event_log)?;
+    let event_log = (!gateway.event_log.trim().is_empty())
+        .then(|| STANDARD.encode(gateway.event_log.as_bytes()));
     Ok(ItaTdxEvidence {
         quote: STANDARD.encode(quote),
         runtime_data: STANDARD.encode(runtime_data),
-        event_log: event_log.map(|bytes| STANDARD.encode(bytes)),
+        event_log,
         verifier_nonce: verifier_nonce.clone(),
     })
 }
@@ -192,20 +193,6 @@ fn decode_required_hex(field: &'static str, raw: &str) -> Result<Vec<u8>, ItaEvi
         return Err(ItaEvidenceError::MissingField(field));
     }
     hex::decode(hex_value).map_err(|source| ItaEvidenceError::InvalidHex { field, source })
-}
-
-fn decode_optional_hex(
-    field: &'static str,
-    raw: &str,
-) -> Result<Option<Vec<u8>>, ItaEvidenceError> {
-    let trimmed = raw.trim();
-    let hex_value = trimmed.strip_prefix("0x").unwrap_or(trimmed);
-    if hex_value.is_empty() {
-        return Ok(None);
-    }
-    hex::decode(hex_value)
-        .map(Some)
-        .map_err(|source| ItaEvidenceError::InvalidHex { field, source })
 }
 
 #[cfg(test)]
--- a/crates/services/src/attestation/ita/evidence.rs
+++ b/crates/services/src/attestation/ita/evidence.rs
@@ -114,8 +114,9 @@ fn build_tdx_evidence(
     let quote = decode_required_hex("tdx.quote", &gateway.intel_quote)?;
     let runtime_data = runtime_data_bytes(gateway)?;
     verify_report_data(gateway, verifier_nonce, &runtime_data)?;
-    let event_log = (!gateway.event_log.trim().is_empty())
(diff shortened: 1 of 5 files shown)


crates/services/src/attestation/ita/evidence.rs
```diff
@@ -114,11 +114,12 @@ fn build_tdx_evidence(
     let quote = decode_required_hex("tdx.quote", &gateway.intel_quote)?;
     let runtime_data = runtime_data_bytes(gateway)?;
     verify_report_data(gateway, verifier_nonce, &runtime_data)?;
-    let event_log = decode_optional_hex("tdx.event_log", &gateway.event_log)?;
+    let event_log = (!gateway.event_log.trim().is_empty())
+        .then(|| STANDARD.encode(gateway.event_log.as_bytes()));
```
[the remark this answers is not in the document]
PierreLeGuen: Addressed in `72823aae`: the event log is trimmed before base64 encoding, and empty, whitespace-only, or `0x` values are omitted. Focused regression tests cover both normalization cases. I also updated the stale DEV and service fixtures to use valid JSON event logs.

PierreLeGuen:
Addressed the remaining top-level review nits in `72823aae`: encoding now uses the trimmed event-log JSON, and the DEV plus service fixtures now use valid JSON event logs. The same commit preserves empty/`0x` omission semantics and adds focused regression coverage.

Remote CI passed lint, unit, integration, E2E, and the aggregate Test Suite.

crates/services/src/attestation/ita/evidence.rs
```diff
@@ -114,11 +114,13 @@ fn build_tdx_evidence(
     let quote = decode_required_hex("tdx.quote", &gateway.intel_quote)?;
     let runtime_data = runtime_data_bytes(gateway)?;
     verify_report_data(gateway, verifier_nonce, &runtime_data)?;
-    let event_log = decode_optional_hex("tdx.event_log", &gateway.event_log)?;
+    let event_log = gateway.event_log.trim();
+    let event_log =
+        (!event_log.is_empty() && event_log != "0x").then(|| STANDARD.encode(event_log.as_bytes()));
```
lloydmak99: **[High · 94%] Base64 preserves an event-log format ITA cannot consume**

`[]byte` explains how Go's JSON encoder serializes this field; it does not mean ITA accepts arbitrary UTF-8 bytes. [Intel's current TDX adapter](https://github.com/intel/trustauthority-client-for-go/blob/v1.12.0/go-tdx/tdx_adapter.go#L49-L63) fills `EventLog` with validated raw TCG 2.0 CCEL/NEL bytes from [`GetCcel`](https://github.com/intel/trustauthority-client-for-go/blob/v1.12.0/go-tdx/ccel.go#L46-L76). Dstack instead exposes a simplified JSON array with `imr`, `event_type`, `digest`, `event`, and `event_payload`; its format is explicitly different from TCG and uses zero-based IMR indexes ([source](https://github.com/Dstack-TEE/dstack/blob/f67b4f67ebabef0a27795705121698280a1038dc/cc-eventlog/src/tdx.rs#L14-L33)). Intel's documented legacy JSON form accepted by ITA is also a different grouped `{rtmr, rtmr_events}` schema ([maintainer confirmation](https://github.com/intel/trustauthority-client-for-go/issues/32#issuecomment-2176566842)).

This removes the local `InvalidHex` error, but a real non-empty dstack log still reaches ITA in an unsupported representation. The new fixture only proves base64 mechanics and even omits dstack's required `event` and `event_payload` fields, so the passing tests cannot establish wire compatibility.

Please either provide raw CCEL/NEL bytes, convert to a documented ITA-supported representation and test that boundary, or omit the optional `event_log` until one is available.

lloydmak99 (changes requested):
## Code Review

Found **1 blocking issue** on the current head: 1 High-confidence protocol/wire-format defect. Security and concurrency checks were otherwise clean.

The targeted ITA suite passes (42 tests), and the PR's lint, unit, integration, and E2E checks pass, but those tests only prove local serialization—not compatibility with ITA's accepted event-log formats. Details are inline.

crates/services/src/attestation/ita/evidence.rs
PierreLeGuen: Addressed in `7155e62f`. The ITA mapper now always omits the optional `event_log`; it no longer forwards dstack's JSON representation. The exact wire-format regression starts from a non-empty dstack event log and requires `tdx.event_log` to be absent. Quote decoding, nonce-bound runtime data, and report_data verification are unchanged. CI is rerunning on this commit, and I updated the PR title/body to reflect the corrected contract.

PierreLeGuen: Follow-up `20d63ee7` closes the fixture gap you called out: the synthetic dstack event now includes required `event` and `event_payload` fields, and the regression first deserializes it through `dstack_sdk_types::dstack::EventLog` before asserting ITA omission. Red/green evidence: the new assertion failed on the prior fixture with `missing field event`, then passed after the correction. All 41 ITA tests, rustfmt, and strict services clippy pass locally; CI is rerunning on the new head.

lloydmak99 (approved):
## Code review

Approved at `20d63ee72ca96abe79a2d466ad5ac8f286faed91`.

I reviewed the complete five-file attestation diff and the exact ITA wire shape. The prior representation issue is resolved: dstack JSON event-log data is intentionally omitted from the optional ITA field, while quote decoding, nonce-bound runtime data construction, and report-data verification remain unchanged. The regression fixture is a complete value that deserializes as `dstack_sdk_types::dstack::EventLog`.

Validation:
- Local `cargo test -p services attestation::ita::`: 41 passed.
- Current lint, unit, integration, E2E, aggregate Test Suite, and cargo-deny advisories are green.

The separate `cargo audit` check is the same action/toolchain failure present on main: cargo-audit pulls `kstring` requiring Rust 1.96 while the runner has Rust 1.92. It is unrelated to this diff.

No blocking findings.

PR #887, opened 2026-07-16
fix(database): follow the Patroni leader across failovers

PierreLeGuen:
## Problem

During the 2026-07-12 outage the Patroni leader failover was **detected** by cloud-api (`Leader change detected` logged on both prod hosts at 00:45 UTC, ~40s after the first DB errors) and a new write pool was built — but every repository holds a clone of the **startup** deadpool `Pool`, whose target host is fixed at creation. The ClusterManager swapped a pool object nothing re-reads, so 100% of traffic stayed pinned to the failed leader until a redeploy rebuilt the pools. Three secondary defects compounded this class of outage:

- a failed `handle_leader_change()` still advanced `last_leader`, so a single transient failure (new leader not accepting connections yet) permanently suppressed any retry, with the error cause discarded;
- a member reported as leader while `pg_is_in_recovery()` was still true was installed anyway (warn only), pinning writes to a read-only node;
- stale discovery state (unreachable Patroni `/cluster` endpoint) was invisible — `is_state_stale()` had no callers.

## Fix

- **`DbPool` is now a shared, swappable handle** (`Arc<RwLock<Option<Pool>>>`): the ClusterManager installs the new leader's pool into it and every repository clone immediately routes new acquisitions to the new leader. Repository code is unchanged (`.get().await` as before).
- **Reconcile loop replaces the one-shot leader tracking**: the manager records which `host:port` the write pool actually targets — only on a successful, verified build — and compares it against the discovered leader every 30s tick. Failed rebuilds are retried on the next tick and the error cause is logged.
- **Leader verification is bounded** (15s timeout): a candidate that accepts connections and then hangs cannot stall the reconcile loop.
- **In-recovery leaders are rejected** (retried next tick) instead of installed.
- Stale discovery state now logs a warning from the reconcile loop.
- Hardening: no lock held across pool acquisition in the read paths, no divide-by-zero when the replica list empties mid-failover, no `rows[0]` panic in the background task, double-start of the background task aborts the previous loop.

## Tests

- `failed_write_pool_rebuild_is_retried_on_next_reconcile` (unit): a failed rebuild must not record the leader as installed and must be retried.
- `reconcile_repoints_startup_pool_clones_to_new_leader` (e2e, runs against the CI Postgres): a pool handle cloned at startup — what every repository holds — must serve connections to the new leader after reconcile.
- `replace_propagates_to_existing_clones` / `get_on_uninitialized_handle_fails_closed` (unit).
- Full suite green locally: 240 unit + 34 database integration + 662 e2e.

With this, a leader failover self-heals within one discovery refresh + reconcile tick instead of requiring a redeploy. chat-api has a copy of the same code and gets a mirror PR.

Follow-ups (not in this PR): TCP keepalive/`tcp_user_timeout` tuning so already-checked-out connections to a hard-dead leader die faster; a `/v1/health` variant that exercises the DB; Datadog monitors for `Failed to refresh cluster state` (still firing intermittently in prod since 07-12) and chat-api error logs.

### DIFF
--- a/crates/api/tests/e2e_all/patroni_failover.rs
+++ b/crates/api/tests/e2e_all/patroni_failover.rs
@@ -1,13 +1,19 @@
-// Regression test for the 2026-07-12 outage root cause: pool handles cloned at
-// startup (as `Database::new` hands to every repository) must observe the pool
-// the ClusterManager installs for a new leader. Previously the manager swapped
-// a pool nothing re-read, so repositories kept dialing the host captured at
-// startup until the process was redeployed.
+// Regression tests for the 2026-07-12 outage root cause: pool handles cloned
+// at startup (as `Database::new` hands to every repository) must observe the
+// pool the ClusterManager installs for a new leader. Previously the manager
+// swapped a pool nothing re-read, so repositories kept dialing the host
+// captured at startup until the process was redeployed.
+//
+// Both "leaders" are TCP proxies in front of the CI Postgres, so the tests can
+// tell which path a connection takes and can kill the old leader's path
+// outright.
 
 use database::cluster_manager::{ClusterManager, DatabaseConfig, ReadPreference};
 use database::patroni_discovery::{ClusterMember, PatroniDiscovery};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
-use tokio::net::TcpListener;
+use std::time::Duration;
+use tokio::net::{TcpListener, TcpStream};
 
 fn leader_member(name: &str, host: &str, port: u16) -> ClusterMember {
     ClusterMember {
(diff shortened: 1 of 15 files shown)


crates/database/src/cluster_manager.rs
```diff
@@ -108,24 +126,44 @@ impl ClusterManager {
...
+            .await
+            .map_err(|_| {
+                anyhow!(
+                    "Timed out verifying leader {} after {:?}",
+                    leader.name,
+                    WRITE_POOL_VERIFY_TIMEOUT
+                )
+            })??;
+
+        self.write_pool.replace(pool);
```
lloydmak99: **Medium — stale candidate can be installed after discovery advances.** `reconcile()` snapshots a leader, but verification can take up to 15 seconds. Patroni discovery refreshes concurrently, so a newer leader may be known by the time this unconditional `replace()` runs. The stale candidate can then repoint every repository until the next 30-second reconcile. Re-read discovery immediately before installation and require its current `host:port` to still match the verified candidate; also serialize reconciliation so concurrent callers cannot produce last-writer-wins rollback.

crates/api/tests/e2e_all/patroni_failover.rs
```diff
@@ -0,0 +1,99 @@
...
+            max_write_connections: 2,
+            max_read_connections: 2,
+            tls_enabled: false,
+            tls_ca_cert_path: None,
+        },
+        ReadPreference::LeaderOnly,
+        None,
+    );
+
+    // What Database::new does at startup: take a handle and clone it into the
```
lloydmak99: **Medium — this test does not reproduce replacement of an active old-leader pool.** The handle starts uninitialized, the dead-leader reconcile installs nothing, and the later reconcile performs the first usable installation. That proves clones observe initial installation, but not that a clone already serving through the old pool follows `replace()`. Seed a usable old pool, clone and acquire through it, reconcile to a distinguishable new pool, then verify the same clone acquires from the new target.

lloydmak99 (changes requested):
Two mirrored issues need addressing before merge:

1. The candidate leader can become stale during the bounded verification window and still be installed afterward.
2. The regression test covers first installation into an uninitialized handle, not replacement of an already-serving old-leader pool.

Functional CI is green. The separate `cargo audit` failure is also present on current `main`: the action installs cargo-audit 0.22.2, whose dependency requires Rust 1.96 while the runner uses 1.92. It is a workflow blocker, not a regression from this PR.

PierreLeGuen:
both findings addressed in a77e63b9:

1. after verification (up to 15s), discovery is re-read and the candidate is discarded if its host:port is no longer the current leader; reconcile() is also serialized with a mutex so concurrent callers can't interleave installs into a last-writer-wins rollback.
2. reworked the e2e tests to exercise the actual outage invariant: both leaders are counting tcp proxies in front of the ci postgres, so the test verifies a clone already serving through leader A's pool follows the failover install and keeps working after A goes hard-down. added a second test for finding 1: a delayed proxy holds reconcile inside verification while the leader flips, and the stale candidate must be discarded (converges on the next tick).

chat-api mirrored in 4a8765af.

crates/database/src/cluster_manager.rs
```diff
@@ -108,24 +131,56 @@ impl ClusterManager {
...
+        if current.as_ref().map(Self::member_target).as_deref() != Some(target.as_str()) {
+            return Err(anyhow!(
+                "Cluster leader changed to {:?} while verifying {target}; discarding this pool",
+                current.map(|m| Self::member_target(&m))
+            ));
         }
 
-        let mut write_pool = self.write_pool.write().await;
-        *write_pool = Some(pool);
+        self.write_pool.replace(pool);
```
lloydmak99: **Medium — the stale-leader race is narrowed but not eliminated.** `get_leader().await` releases the discovery-state read lock before this `replace()`. The independent discovery refresh can publish leader B after the equality check but before stale candidate A is installed; `reconcile_lock` only serializes reconcile callers, not discovery updates. Make comparison and installation atomic relative to topology publication, for example by retaining a discovery read guard through synchronous installation or using a topology generation with coordinated compare-and-install.

crates/api/tests/e2e_all/patroni_failover.rs
```diff
@@ -0,0 +1,222 @@
...
+        None,
+    ));
+    let repository_handle = manager.write_pool();
+
+    let reconcile_task = tokio::spawn({
+        let manager = manager.clone();
+        async move { manager.reconcile().await }
+    });
+    // Wait until reconcile is inside A's delayed verification, then fail over.
+    tokio::time::sleep(Duration::from_millis(250)).await;
```
lloydmak99: **Medium — this fixed sleep makes the regression test scheduler-racy.** It does not prove the spawned reconcile task has snapshotted or connected to leader A. Under CI contention, reconcile may start only after discovery changes to B, legitimately install B, and fail the later assertion that no pool was installed. Synchronize on leader A accepting its first connection (or use a oneshot/barrier from the proxy) before changing discovery.

lloydmak99 (changes requested):
Re-review of the current head:

- The previous active-pool replacement test gap is resolved: the new test proves an existing repository handle moves from leader A to B and keeps working after A is stopped.
- Two mirrored issues remain: the final revalidation/install sequence is still TOCTOU, and the new stale-candidate regression test relies on a scheduler-sensitive fixed sleep.

Functional CI is green. The separate cargo-audit toolchain failure remains present on main and is not caused by this PR.

PierreLeGuen:
both addressed in 05c256c8:

1. the check and install are now atomic relative to topology publication: the compare-and-install runs inside a new PatroniDiscovery::with_current_leader, which holds the cluster-state read lock for the duration — update_cluster_state takes the write lock, so a leader can't be published between the equality check and replace(). write_pool_target moved to a std lock so the whole install is synchronous under the guard.
2. the race test no longer sleeps a fixed 250ms: it waits (with a 10s cap) for candidate A's proxy to observe the verification connection — which proves reconcile has snapshotted A and is inside verification — before flipping discovery, and A's pre-connect delay is now 2s so the flip lands well inside the window. ran 6x locally, no flakes.

chat-api mirrored in fc704a5d.

lloydmak99 (approved):
## Code review

Approved at `05c256c8bcd95042d21378e4884e059e0ae932dc`.

I reviewed the current 15-file diff, including the shared pool handle, every repository-facing integration point, Patroni state publication, retry/error paths, and the PostgreSQL-backed race tests. The previous compare/install TOCTOU and scheduler-dependent race-test findings are resolved: `with_current_leader` holds the discovery read guard through the target comparison and pool replacement, and the stale-candidate test now synchronizes on the accepted verification connection.

Validation:
- Local pool replacement tests: 2 passed.
- Local failed-rebuild retry regression: 1 passed.
- Current CI E2E: both `startup_pool_clones_follow_leader_across_failover` and `stale_candidate_is_discarded_when_leader_changes_during_verification` passed.
- Lint, unit, integration, E2E, aggregate Test Suite, and cargo-deny advisories are green.

The separate `cargo audit` check is an action/toolchain failure also present on the unchanged main SHA: cargo-audit now pulls `kstring` requiring Rust 1.96 while the runner has Rust 1.92. It is not caused by this diff.

No blocking findings. Non-blocking future coverage could directly exercise a candidate returning `pg_is_in_recovery() = true` and two simultaneous `reconcile()` callers.

PR #891, opened 2026-07-20
fix(api): strip stream usage on attested models when include_usage is unset

PierreLeGuen:
## Problem

Live-verified on prod (2026-07-20): streaming `zai-org/GLM-5.1-FP8` (self-hosted/nearai path) **without** `stream_options`, 200 of 202 chunks carried a full non-null `usage` object and the stream ended with a trailing `{"choices":[],"usage":{...}}` usage-only chunk. Per the OpenAI streaming spec, when `include_usage` is not requested no chunk should carry `usage` and there must be no trailing usage-only chunk. This is the exact defect class that broke strict SDK parsers on the Chutes path (customer-reported; fixed there by #870, which explicitly left the nearai provider path untouched). Usage values themselves are correct — this is a spec-conformance/parser-robustness fix, not a billing fix.

Root cause: the nearai provider forces `stream_options {include_usage: true, continuous_usage_stats: true}` upstream for billing capture, and the route forwarded those bytes verbatim because `strip_intermediate_usage` (#801) was scoped to non-attested models only, to preserve byte-exact provider-signature verification.

## Fix

All in `chat_stream_usage_mode` (`crates/api/src/routes/completions.rs`):

- `strip_intermediate_usage` now activates for attested models too (`model_attestation_supported.is_some()` instead of `== Some(false)`): chunks are re-serialized with `usage: null` and the trailing usage-only chunk is dropped, identical to the non-attested behaviour from #801.
- `gateway_signature_enabled` now covers the strip path (`(rewrite_public_stream_usage || strip_intermediate_usage) && attested`): the provider signature fetch is skipped and the gateway signs the exact rewritten bytes the client received — the same verifiability-preserving mechanism the `include_usage: true` rewrite path already uses.
- The end-of-stream tail unpins the provider's signature-fetch routing entry for gateway-signed streams (the post-fetch unpin never runs when the fetch is skipped), preventing unbounded `chat_id -> backend` map growth. This also covers the pre-existing `include_usage: true` gateway path. Scope note: the unpin runs when the stream is polled to completion; the client-disconnect case is a pre-existing gap shared with main (as is pool-level `chat_id_mapping` growth) and is unchanged here.

Billing is untouched: `InterceptStream` records usage from the parsed chunks in the service layer, upstream of the route-level rewrite (same preservation property as #870), pinned by the existing stream-billing e2e which now exercises the strip path.

Unchanged paths: metadata-lookup failure (attestation unknown) falls back to byte-exact passthrough; `include_usage: true`, `continuous_usage_stats`, E2EE, and non-text-modality streams are byte-exact as before.

## Verifiability trade-off (needs owner sign-off)

Default attested (nearai) streams — the majority of self-hosted stream traffic — switch from byte-exact provider TEE signatures to **gateway signatures over the rewritten bytes**. This knowingly supersedes #801's deliberate attested-passthrough choice: re-serialization breaks byte-exact provider-signature verification, so gateway signing (the mechanism `include_usage: true` streams have used since #724/#801) is the only way to be spec-conformant while keeping the stored signature consistent with the bytes the client actually received. Clients verifying default streams now verify a cloud-api gateway signature rather than the inference TEE's signature.

Chutes side effect: chutes models are attested, so their default streams also move onto the strip+gateway path. The chutes module already sanitizes the parsed chunk (canonical model id, internal fields stripped, usage kept for billing) and has no provider chat signatures, so the skip flag is a no-op there. Two client-visible nuances: intermediate chutes chunks now carry an explicit `"usage": null` (instead of the key being removed by `gate_stream_usage` — same shape as every other non-passthrough stream since #801), and non-E2EE default chutes streams now store a gateway signature where the signature endpoint previously returned `SIGNATURE_UNSUPPORTED`, matching chutes' existing `include_usage: true` behaviour.

Known remaining gap (out of scope, same as #801/#870): the legacy `/v1/completions` text streaming path has the identical defect — `text_completion_stream` forces the same `stream_options` and the route's usage shaping never applies to text streams — and is untouched here; follow-up candidate.

Perf note: default attested streams no longer take the raw passthrough shortcut — per-chunk re-serialize plus gateway SHA-256 hashing, the same cost profile as the existing `include_usage: true` rewrite path.

## Testing

All run in a worktree based on `c46b2488`:

- `cargo test -p api --lib` — 244 passed (includes the 68 `routes::completions` tests covering the new mode matrix: attested default / `include_usage: false` -> strip + gateway signature; metadata-None -> passthrough regression; the two former attested-passthrough tests inverted to the new contract).
- `DEV=1 cargo test -p api --test e2e_all -- usage_chat_completions` — 10 passed, including two new attested e2e tests (default and `include_usage: false`) asserting `usage: null` on every chunk, no usage-only/empty-choices chunk, and `[DONE]` termination (mock provider mirrors vLLM's per-chunk-usage + trailing-husk shape), plus the stream-billing-through-strip regression. The attested `include_usage: true` spec shape was already pinned by `test_chat_completions_stream_include_usage_true_emits_final_usage_only`.
- `DEV=1 cargo test -p api --test e2e_all -- signature_verification` — 5 passed; `test_streaming_chat_completion_signature_verification` (default attested stream, now gateway-signed) verifies the stored signature hash matches sha256 of the exact client-received bytes.
- Broader e2e sweep (general, auto_redact, model_alias_transparency, serving_provider, client_disconnect, chat_encryption, reasoning, usage_recording, usage_responses) — 128 passed, 1 failed: `general::test_admin_list_users_with_organizations_no_spend_limit`, which fails identically on clean `origin/main` (pre-existing/ordering-dependent, unrelated).
- `cargo check -p api --tests`, `cargo fmt -p api --check`, `cargo clippy -p api --all-targets` — clean.

### DIFF
--- a/crates/api/src/routes/completions.rs
+++ b/crates/api/src/routes/completions.rs
@@ -417,24 +417,31 @@ fn chat_stream_usage_mode(
         && !chat_stream_has_non_text_modalities(request);
 
     // For default streaming (no include_usage or include_usage:false), strip populated
-    // usage from all chunks so intermediate ones carry `usage: null` per OpenAI spec.
+    // usage from all chunks so intermediate ones carry `usage: null` per OpenAI spec,
+    // and drop the upstream's trailing usage-only chunk entirely (OpenAI emits no
+    // final usage chunk when include_usage wasn't requested; forwarding it breaks
+    // strict SDK parsers — the same defect class as the Chutes husk fix, #870).
     //
-    // Only applied for non-attested (external) models. Attested models (e.g. nearai,
-    // chutes) sign the raw upstream bytes; stripping usage before forwarding to the
-    // client would break byte-exact TEE signature verification because the client
-    // would hash different bytes than what the inference proxy signed. For those
-    // models the provider-signed stream MUST be forwarded verbatim (the
-    // `rewrite_public_stream_usage` path handles `include_usage:true` via the
-    // separately-computed gateway signature).
+    // Applied for BOTH non-attested and attested models. Self-hosted upstreams
+    // always emit per-chunk usage because the nearai provider forces
+    // `stream_options: {include_usage: true, continuous_usage_stats: true}` for
+    // billing capture, so without this gate every chunk of an attested stream
+    // carries a populated `usage` object the client never asked for, plus a
+    // trailing `choices: []` usage-only chunk. Stripping re-serializes the chunks,
+    // so byte-exact provider-signature verification no longer applies; for attested
+    // models the gateway signature covers the rewritten client-facing bytes instead
+    // (`gateway_signature_enabled` below), exactly like the
+    // `rewrite_public_stream_usage` path does for `include_usage: true`. Internal
+    // usage capture is unaffected: InterceptStream records usage from the parsed
+    // chunks in the service layer, upstream of this route-level rewrite.
     //
-    // Also skipped when: the client requested continuous stats (explicit per-chunk
-    // usage), E2EE is active (chunks are opaque), non-text modalities are in use,
-    // `rewrite_public_stream_usage` already handles usage shaping, or the client
-    // explicitly requested `include_usage:true` (fall back to byte-exact passthrough
-    // so the client receives the usage it asked for even if model metadata is
-    // temporarily unavailable).
+    // Skipped when: model metadata is unavailable (`None` — fall back to byte-exact
+    // passthrough rather than guessing), the client requested continuous stats
+    // (explicit per-chunk usage), E2EE is active (chunks are opaque), non-text
+    // modalities are in use, `rewrite_public_stream_usage` already handles usage
+    // shaping, or the client explicitly requested `include_usage:true`.
     let strip_intermediate_usage = request.stream == Some(true)
-        && model_attestation_supported == Some(false)
+        && model_attestation_supported.is_some()
         && !rewrite_public_stream_usage
         && !chat_stream_include_usage_requested(request)
         && !chat_stream_continuous_usage_requested(request)
@@ -443,7 +450,11 @@ fn chat_stream_usage_mode(
 
     ChatStreamUsageMode {
         rewrite_public_stream_usage,
-        gateway_signature_enabled: rewrite_public_stream_usage
+        // Both rewrite modes re-serialize the client-facing bytes, so for attested
+        // models the provider's byte-exact signature can no longer match what the
+        // client received. Skip the provider signature fetch and sign the rewritten
+        // stream at the gateway instead.
+        gateway_signature_enabled: (rewrite_public_stream_usage || strip_intermediate_usage)
             && model_attestation_supported.unwrap_or(false),
         strip_intermediate_usage,
     }
@@ -1587,6 +1598,7 @@ async fn chat_completions_inner(
                 let public_signature_hasher_for_chain = public_signature_hasher.clone();
                 let public_signature_chat_id_for_chain = public_signature_chat_id.clone();
                 let attestation_service_for_chain = app_state.attestation_service.clone();
+                let provider_pool_for_chain = app_state.inference_provider_pool.clone();
 
                 // Re-attach any stashed leading control events, then convert
                 // to a raw bytes stream.
@@ -1948,6 +1960,26 @@ async fn chat_completions_inner(
                                             model = %model_name,
                                             "Cannot store public stream chat signature: no chat_id observed"
                                         );
+                                    }
+                                }
+
+                                if gateway_signature_enabled {
+                                    // The provider-signature fetch is skipped for
+                                    // gateway-signed streams
+                                    // (skip_provider_chat_signature), so its
+                                    // post-fetch unpin never runs. Drop the
+                                    // signature-fetch routing pin here so the
+                                    // provider's chat_id → backend map does not
+                                    // grow unboundedly.
+                                    if let Some(chat_id) =
+                                        public_signature_chat_id_for_chain.lock().await.clone()
+                                    {
+                                        if let Some(provider) = provider_pool_for_chain
+                                            .get_provider_by_chat_id(&chat_id)
+                                            .await
+                                        {
+                                            provider.unpin_chat_connection(&chat_id);
+                                        }
                                     }
                                 }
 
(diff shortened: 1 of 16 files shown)


crates/api/src/routes/completions.rs
```diff
@@ -1951,6 +1963,26 @@ async fn chat_completions_inner(
...
+                                    if let Some(chat_id) =
+                                        public_signature_chat_id_for_chain.lock().await.clone()
+                                    {
+                                        if let Some(provider) = provider_pool_for_chain
+                                            .get_provider_by_chat_id(&chat_id)
+                                            .await
+                                        {
+                                            provider.unpin_chat_connection(&chat_id);
+                                        }
+                                    }
```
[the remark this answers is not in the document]
PierreLeGuen: Done in 7115c0d3 — the chat_id is cloned out of the mutex in its own statement so the guard is dropped before the pool lookup await.

PierreLeGuen:
Addressed review feedback and the CI failure in 7115c0d3:

- **[DONE]/signature ordering**: the strip-mode path was forwarding the upstream `[DONE]` inline while the gateway signature is only stored later in the end-of-stream tail, so a client fetching the chat signature the moment it sees `[DONE]` could race the store. `[DONE]` is now held back like the other rewrite paths: the tail stores the signature first and only then appends `data: [DONE]\n\n` (still folded into the hash, so client-visible and signed bytes are unchanged). Added `test_streaming_chat_default_stream_signature_stored_before_stream_completes`, which queries the signature immediately at stream end with no wait, to pin the ordering.
- **Lock across await in the unpin block**: the chat_id is now cloned out of the mutex in its own statement so the guard is dropped before the provider-pool lookup await.
- **CI (run 29743678992)**: both failures are unrelated to this diff. `audio_transcriptions::test_audio_transcription_missing_model` panicked in the shared e2e DB bootstrap (`connection closed` while running migrations; postgres logged "unexpected EOF on client connection with an open transaction"), and `api_keys::test_api_key_same_name_different_workspaces` then hit the 120s nextest timeout during the fail-fast wind-down. Neither test touches code changed here; both pass locally on this branch, and the same suite is green on main at this branch's parent commit (c46b2488). The "Test Suite" job is just the gate that fails when E2E fails. Verified locally: fmt, clippy `-D warnings`, `api --lib` (244 passed), and the usage + signature e2e suites all green.

crates/api/src/routes/completions.rs
```diff
@@ -443,7 +450,11 @@ fn chat_stream_usage_mode(
 
     ChatStreamUsageMode {
         rewrite_public_stream_usage,
-        gateway_signature_enabled: rewrite_public_stream_usage
+        // Both rewrite modes re-serialize the client-facing bytes, so for attested
+        // models the provider's byte-exact signature can no longer match what the
+        // client received. Skip the provider signature fetch and sign the rewritten
+        // stream at the gateway instead.
+        gateway_signature_enabled: (rewrite_public_stream_usage || strip_intermediate_usage)
```
lloydmak99, line 456: **High — preserve or explicitly migrate the model-TEE verification contract.** This makes the default attested-stream path (no `include_usage`, the common case) replace the provider/model-TEE signature with a gateway signature. The public response contract changes too: provider signatures use `{model_id}:{request_hash}:{response_hash}`, while `store_gateway_signature` signs `{request_hash}:{response_hash}`, and `signing_address` becomes the gateway key rather than the model TEE key. Current `nearai/docs` says the signature is generated inside the model TEE and that `signing_address` is the key of the TEE node that served the request, so existing verifiers lose model-origin proof and may parse the payload incorrectly. The PR notes that owner sign-off is needed, but there is no signature-kind/version signal or synchronized docs/client migration. Please obtain explicit API/security approval and either retain or chain the model-TEE provenance, or introduce an explicit versioned signature kind with the corresponding docs and verifier migration, before changing the default path.

crates/api/tests/e2e_all/signature_verification.rs
```diff
@@ -322,3 +322,102 @@ async fn test_streaming_chat_include_usage_signature_hashes_client_bytes() {
...
+    });
+
+    let request_json = serde_json::to_string(&request_body).expect("Failed to serialize request");
+    let expected_request_hash = compute_sha256(&request_json);
+
+    let response = server
+        .post("/v1/chat/completions")
+        .add_header("Authorization", format!("Bearer {api_key}"))
+        .json(&request_body)
+        .await;
```
lloydmak99: **Medium — this test does not reproduce the `[DONE]` visibility race it is meant to lock.** `axum-test` collects the entire response body before this `.await` returns, and the test only parses `[DONE]` afterward. The pre-fix implementation that exposed `[DONE]` before the tail stored the signature would therefore also pass once EOF was collected. Please consume the SSE body incrementally and issue the signature GET immediately when `[DONE]` is decoded, before waiting for response EOF; that makes the regression test distinguish the old and new ordering.

crates/api/src/routes/completions.rs
```diff
@@ -1951,6 +1970,28 @@ async fn chat_completions_inner(
...
+                                    // post-fetch unpin never runs. Drop the
+                                    // signature-fetch routing pin here so the
+                                    // provider's chat_id → backend map does not
+                                    // grow unboundedly. Clone the chat_id in its
+                                    // own statement so the mutex guard is not held
+                                    // across the pool lookup await.
+                                    let chat_id =
+                                        public_signature_chat_id_for_chain.lock().await.clone();
+                                    if let Some(chat_id) = chat_id {
+                                        if let Some(provider) = provider_pool_for_chain
```
lloydmak99: **Medium — cover the new pin cleanup and keep lifecycle ownership together.** This is now the only release of the provider's `chat_id -> backend` signature-routing pin for gateway-signed streams, but the added e2e tests only assert response/signature behavior. A regression here leaks one routing entry per default attested stream. Add a provider/pool spy (or a test-only count) and assert the pin is removed after the success, signature-store error, and timeout paths. Since the provider-signature path already owns its matching unpin inside `AttestationService::store_chat_signature_from_provider_impl`, consider exposing a gateway-store-and-unpin service operation rather than splitting the same lifecycle between the route and service.

lloydmak99 (changes requested):
## Review summary

Reviewed current head `7115c0d3`. The usage-shaping behavior and the follow-up `[DONE]` ordering fix are sound, and current CI plus the targeted 72-test completions suite are green.

I found **1 High** contract/security issue and **2 Medium** regression-coverage/lifecycle issues. The High finding needs explicit API/security owner sign-off and a compatible public-contract migration before this becomes the default attested-stream behavior.

crates/api/src/routes/completions.rs
PierreLeGuen: Implemented in 96899ee0: signatures now carry an explicit, versioned kind. New `SignatureKind` enum (`provider_tee` | `gateway`) on `ChatSignature`, stamped by the provider path and the gateway path respectively; migration V0071 adds a nullable `signature_kind` column (NULL = pre-column rows — both provider and gateway writes predate it, so provenance is genuinely unknown and the API surfaces it as an absent field rather than guessing); `GET /v1/signature/{chat_id}` gains an additive `signature_kind` field with the signed-payload format doc-commented per kind (`provider_tee`: `{model_id}:{request_hash}:{response_hash}`, `gateway`: `{request_hash}:{response_hash}`). Docs PR documenting both kinds and when each occurs: https://github.com/nearai/docs/pull/41. To be explicit: merge of the signature-semantics change waits on API/security owner sign-off — this PR does not change who signs, it makes the existing gateway signing visible and auditable. Open question for you: would you prefer chaining model-TEE provenance instead — i.e. the gateway signing over (provider signature + rewritten bytes) so a gateway signature still attests to the model TEE's original signature — either now or as a follow-up kind (e.g. `gateway_chained`)?

crates/api/tests/e2e_all/signature_verification.rs
PierreLeGuen: Done in 18abcb5b. The test now drives the router in-process via `tower::ServiceExt::oneshot` and polls the SSE body frame-by-frame with `http_body_util::BodyExt::frame()`, issuing the signature GET the instant the `data: [DONE]` line is decoded and before polling any further frames — so the end-of-stream tail future (which stores the signature) cannot have run unless the holdback ordered it first. Red/green proof: with the [DONE]-holdback clause from 7115c0d3 temporarily reverted, the test fails deterministically with `assertion left == right failed: Signature must be retrievable the instant [DONE] is decoded: {"error":{"message":"Signature not found: chatcmpl-3d1fc5b40493600e:ecdsa","type":"not_found_error"}} left: 404 right: 200`; with the holdback restored it passes (`test result: ok. 1 passed`). It also pins `signature_kind == "gateway"` and that no bytes follow [DONE].

crates/api/src/routes/completions.rs
PierreLeGuen: Done in 96899ee0 (service) + 18abcb5b (route). The store+unpin lifecycle now lives on the attestation service — `store_chat_signature_and_unpin` mirrors `store_chat_signature_from_provider`'s ownership (pin released whether the store succeeds, fails, or times out; the 5s timeout moved inside the method so an outer timeout can't drop the unpin), plus `release_chat_signature_pin` for the errored-stream branch, both with default trait impls so existing mocks are untouched. Lifecycle unit tests in the attestation module cover unpin on success (and that both algos store with kind `gateway`), unpin on store error, unpin on timeout (paused-clock), the errored-stream release-without-store, and the no-mapping no-op — using `MockProvider` unpin recording and a pool seeded via `store_chat_id_mapping` (widened to `pub(crate)` only).

lloydmak99 (approved):
Verified at head `18abcb5b`: attested-model streams now strip populated `usage` and drop the trailing usage-only chunk when `include_usage` is unset, matching non-attested behavior and fixing the real defect where GLM-5.1-FP8 streams ended in a `choices: []` husk. Because re-serialization breaks byte-exact provider-TEE signatures, these streams correctly switch to gateway signing, disambiguated by the new `signature_kind` column. Core change looks correct and safe to merge.

Non-blocking follow-ups (optional):
- `crates/api/src/routes/completions.rs:456` / `V0071__*.sql`: this is a deliberate public-contract change (default no-`include_usage` streams now return a gateway signature over `{request_hash}:{response_hash}` rather than a provider-TEE signature). It's made explicit via the additive `signature_kind` field and docs (nearai/docs#41); just ensure the API/security owner sign-off the author flagged lands at merge.
- `crates/api/src/routes/completions.rs:1922-1985`: confirmed the tail refactor releases the routing pin on all end-of-stream paths (store success, store error/timeout, errored stream), so the `chat_id → backend` map doesn't leak — no action needed.

Checks: reviewed full diff, migration, tests, prior-review timeline, and CI; `git diff --check` passed; local `cargo` build/test/clippy/fmt skipped (no Rust toolchain here); CI green on `18abcb5b` — lint, unit, integration, E2E, cargo audit, cargo deny.

PierreLeGuen:
@Evrard-Nil could you give the API/security owner sign-off on the contract change here? Default attested streams (no include_usage) switch from provider-TEE to gateway signing over {request_hash}:{response_hash}, disambiguated by the new additive signature_kind field (V0071, nullable) and documented in nearai/docs#41. Lloyd has approved the implementation; his approval conditions merge on this sign-off. Context: this is the nearai-path analog of #870 for the BeeZi spec-conformance report.

think-in-universe:
@PierreLeGuen @lloydmak99 This cloud-api PR has introduced a billing / usage tracking bug in chat-api https://github.com/nearai/chat-api/issues/367 reported by @ArtemSirobaba today, which has been fixed in the PR: https://github.com/nearai/chat-api/pull/366 that enforces `include_usage = true` in stream options now

PR #895, opened 2026-07-22
fix(api): return generic JSON envelope for unmatched routes

PierreLeGuen:
Requests that match no route currently hit Axum's default fallback and return a `404` with an empty body (verified against the live API on 2026-07-22: `GET https://cloud-api.near.ai/nonexistent` → `404`, `content-length: 0`). Method mismatches on existing routes likewise return an empty-body `405`. This adds explicit fallbacks so those cases return the same stable OpenAI-style error envelope the rest of the API uses, instead of a framework default.

## Changes

- `crates/api/src/routes/unsupported.rs`: two new handlers.
  - `unknown_route` — global fallback: `404` + `{"error":{"message":"Not found","type":"invalid_request_error","param":null,"code":null}}` with `x-should-retry: false`. The body is intentionally static: no method/path echo, no version, no implementation detail.
  - `method_not_allowed` — same envelope with `405` for matched-path/wrong-method requests.
- `crates/api/src/lib.rs`: wires them via `.fallback(...)` and `.method_not_allowed_fallback(...)` on the outermost router, after all merges (so they cover `/v1/*`, root-level, and OHTTP/MCP surfaces) and before the middleware layers (so CORS/metrics/correlation apply to fallback responses too).

No behavior change for any existing route: documented error envelopes, the `501 not_implemented` placeholders, and streaming responses are untouched.

## Test plan

- New unit tests in `routes/unsupported.rs`: unknown paths (`/nonexistent`, `/test.html`, `/v1/nonexistent`, `/.env`) return `404` `application/json` with the static generic envelope and `x-should-retry: false`; wrong method on an existing route returns `405` with the same envelope shape.
- `cargo test -p api --lib`: 250 passed, 0 failed (includes the existing OpenAI-compatible error envelope tests).
- `cargo fmt --check`: clean. `cargo clippy -p api --lib --tests`: clean.

## Rollout

- Ships with the normal cloud-api release train: staging deploy + external scan of unmapped-path behavior on `cloud-stg-api.near.ai`, then production rollout.
- Edge-level counterparts (port 80 contract, ingress error bodies) are separate changes; server version-banner suppression is tracked separately as well.

Part of nearai/infra#192

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1389,6 +1389,11 @@ pub fn build_app_with_config(
         .merge(openapi_routes)
         .merge(mcp_routes)
         .merge(ohttp_root_routes)
+        // Requests matching no route (or no method on a matched route) get a
+        // stable generic JSON envelope instead of Axum's default empty-body
+        // 404/405 (nearai/infra#192).
+        .fallback(routes::unsupported::unknown_route)
+        .method_not_allowed_fallback(routes::unsupported::method_not_allowed)
         .layer(cors)
         // Add HTTP metrics middleware to track all requests
         .layer(from_fn_with_state(
--- a/crates/api/src/routes/unsupported.rs
+++ b/crates/api/src/routes/unsupported.rs
@@ -26,6 +26,44 @@ pub fn openai_compat_routes() -> Router {
         .route("/models/{*model_id}", any(openai_endpoint_not_implemented))
 }
 
+/// Global router fallback for requests that match no route.
+///
+/// Returns a stable generic OpenAI-style error envelope instead of Axum's
+/// default empty-body 404, so unmapped paths never surface framework or
+/// infrastructure detail (nearai/infra#192). The body is intentionally
+/// static: no method/path echo, no version, no implementation hints.
+pub async fn unknown_route() -> (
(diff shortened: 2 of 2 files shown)


lloydmak99 (approved):
Clean, self-contained change that returns the generic JSON error envelope for unmatched routes and wrong-method requests. The routing is correct: because `.method_not_allowed_fallback` is called after all `.nest`/`.merge` calls, it applies to the flattened `/v1` routes as well, so wrong-method requests to real endpoints get the generic 405 (verified against the axum 0.8.9 source). Handlers return only static strings — no path, method, or body reflection — and the envelope matches the existing `ErrorResponse` shape used elsewhere.

Optional, non-blocking:
- The 405 fallback still preserves each route's `Allow` header (e.g. `Allow: POST` on `/v1/chat/completions`), which is RFC-compliant and fine — noting only in case fully opaque 405s were intended (`crates/api/src/routes/unsupported.rs:49`).

Checks: `git diff --check` passed; GitHub lint/unit/integration/E2E/audit checks green. `cargo` was unavailable locally, so tests/fmt/clippy were not run here (author reports `cargo test -p api --lib` 250 passed, fmt/clippy clean).

PR #903, opened 2026-07-27
fix: make prefix affinity deterministic without hot-prefix overload

PierreLeGuen:
## Summary

- replace process-local first-seen prefix allocation with a stable SHA-256
  routing key over the first message role and normalized content
- keep ordinary traffic on the deterministic primary, but spill overlapping
  requests for the same hot prefix in deterministic bursts of four with a
  key-dependent spill order
- after assistant/tool history appears, route by the stable first two messages
  so later growing-history turns stay on one conversation home
- scope live-load accounting to the same prefix and release it with RAII on
  success, error, cancellation, or stream drop
- preserve empty-message routing, indexed backend selection, fallback,
  signature pinning, and pathological-TTFT steering

There are no HTTP, request, response, or database interface changes.

## Why

The old router assigned bucket IDs in local arrival order. Separate Cloud API
processes could therefore send the same reusable prefix to different backend
indices, duplicating cache entries and producing zero/high cache-hit behavior.

The first revision used deterministic modulo only. Follow-up testing confirmed
the review concern: with only a few hot prefixes, modulo can overload one
backend even when the fleet has spare capacity. This revision keeps the stable
primary but adds bounded, prefix-local spillover for overlapping initial
requests.

Review follow-up also confirmed that applying transient burst load to every
turn can move a growing conversation between replicas. The final router uses
shared-prefix affinity for the initial request, then a deterministic
first-two-message conversation home once assistant/tool history exists.

## Isolated inference evidence

Tests used three DeepSeek-V4-Flash TP2 backends, synthetic reusable prefixes,
two and four independent gateway histories, identical request order between
arms, and minimal decode output. No prompts, responses, credentials, host
addresses, or customer data are included here.

- one hot prefix, four gateways, concurrency 96: current had 0/192 failures,
  p95 1.70 s, and 51.4 req/s; modulo had 4/192 failures, p95 3.83 s, and
  26.0 req/s; burst four had 0/192 failures, p95 1.59 s, and 69.6 req/s
- 64 prefixes, four gateways, concurrency 96: current had 386/960 failures,
  82.27% cached-token share, and p95 3.60 s; burst four had 0/960 failures,
  98.78% cached-token share, and p95 1.69 s
- 64 prefixes, two gateways, concurrency 96: current had 183/960 failures,
  95.06% cached-token share, and p95 2.67 s; burst four had 0/960 failures,
  98.68% cached-token share, and p95 1.60 s
- lower burst thresholds one, two, and three produced 107, 27, and 12 failures
  respectively in their 64-prefix concurrency-96 runs; burst four was the
  smallest threshold with zero failures across all tested gateway counts and
  workloads
- three prefixes deliberately colliding on one modulo primary at concurrency
  96: ring spill had 4/960 failures, 2.23x load skew, 98.79% cached-token share,
  and p95 1.98 s; key-dependent spill had 0/960 failures, 1.55x load skew, the
  same cached-token share, and p95 1.93 s
- 24 synthetic conversations with five growing turns: burst-only routing moved
  43.75% of transitions, returned 71.76% cached tokens, and had p95 3.87 s;
  hierarchical routing returned 78.60% cached tokens with 0 failures and p95
  3.59 s. The only moves were the permitted first-to-second-turn handoff;
  turns two through five stayed fixed, with an 8/7/9 backend distribution

Single-replica quiet retention stayed above 99% through 180 seconds. Separate
pressure tests also found a separate engine-side SWA capacity limit that caused
eviction under pressure. This change fixes cross-process affinity but does not
eliminate intrinsic SGLang capacity eviction.

## Known trade-offs

- Initial requests may spill while they overlap. Established conversations pin
  to one backend, so the first follow-up may make one deterministic handoff
  from the shared-prefix backend before all later turns stay fixed.
- Concurrent calls for one established conversation do not spill. Distinct
  conversations hash independently across the fleet; this favors reuse of the
  larger growing-history prefix over balancing one conversation itself.
- The initial-request burst counter is process-local. Independent processes
  agree on the primary and spill order, while the exact aggregate spill point
  scales with the number of API processes. Both two- and four-process
  simulations passed.
- Modulo selection remaps most prefixes when the live backend count changes.
  Stable-handle rendezvous hashing is a follow-up because the current Cloud API
  discovery contract provides a count, not stable backend identities.
- A rolling deployment briefly mixes the old and new mappings, so prefixes can
  be duplicated until all Cloud API instances run the new router.

## Validation

- `make preflight` passed with an isolated test database and the repository's
  required local debug/test settings
- focused NEAR AI provider tests: 82 passed
- strict all-target inference-provider clippy and formatting checks passed
- experiment-runner unit tests: 21 passed
- all 13 authoritative artifact checksum manifests revalidated
- semantic smoke on all three backends: HTTP 200, usable content,
  `finish_reason=stop`, and usage fields

### DIFF
--- a/crates/inference_providers/src/attested/nearai/fleet.rs
+++ b/crates/inference_providers/src/attested/nearai/fleet.rs
@@ -79,10 +79,9 @@ pub(super) struct Fleet {
     /// that don't fit the rotation scheme (one-label host, IP literal, …) — then
     /// rotation is a no-op and the canonical-SNI path is used.
     rotation_parts: Option<rotation::UrlParts>,
-    /// Message-prefix trie mapping a conversation prefix to a bucket id, so
-    /// requests sharing a prefix stick to the same backend (prefix-cache hit).
-    /// The bucket id is reduced modulo the live backend count to a rotation
-    /// index by `select_index`.
+    /// Stateless first-message router. Its stable key is reduced modulo the
+    /// live backend count by `select_index`, so every Cloud API process sends a
+    /// reusable prefix to the same indexed backend.
     pub(super) prefix_router: Arc<PrefixRouter>,
     /// Lazily-filled (or eagerly pre-created in legacy mode) per-backend-index
     /// clients, each pinning a persistent H2 connection to one verified
@@ -163,18 +162,15 @@ impl Fleet {
     /// is logged, the completion still streams). This matches the pre-existing
     /// rotation-fallback behavior; it is not introduced by index-addressing.
     ///
-    /// Reachability: `prefix_router.route()` returns a bucket id in
-    /// `0..NUM_PREFIX_BUCKETS` (default 64), reduced mod `count`. When `count >
-    /// NUM_PREFIX_BUCKETS` (more than 64 live backends for one model — far above
-    /// any current deployment) the high indices are unreachable as the
-    /// *preferred* pick, though latency steering / fallback can still reach
-    /// them. Raise `NUM_PREFIX_BUCKETS` if a model ever exceeds 64 backends.
+    /// Reachability: `prefix_router.route()` returns a deterministic `u64`
+    /// derived from the first message. Reducing it modulo `count` makes every
+    /// live backend index reachable without process-local routing state.
     pub(super) fn select_index(&self, messages: &[crate::ChatMessage]) -> Option<usize> {
         let count = self.rotation_count();
         if count == 0 {
             return None;
         }
-        let preferred = self.prefix_router.route(messages) % count;
+        let preferred = (self.prefix_router.route(messages) % count as u64) as usize;
         let stats = lock(&self.backend_stats);
         // Warmed EMA for index `i`, or None if out of range / not yet warmed.
         // `.get()` keeps this panic-free regardless of how `count` relates to
--- a/crates/inference_providers/src/attested/nearai/fleet.rs
+++ b/crates/inference_providers/src/attested/nearai/fleet.rs
@@ -18,6 +18,8 @@ use crate::rotation;
 use crate::spki_verifier::FingerprintState;
 use crate::BackendVerifier;
 use reqwest::Client;
+use sha2::{Digest, Sha256};
+use std::cmp::Reverse;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, MutexGuard, RwLock};
@@ -32,11 +34,11 @@ pub(super) const TTFT_WARMUP_SAMPLES: u32 = 8;
 /// fastest peer's EMA AND the absolute floor below.
 const TTFT_SLOW_RATIO: f64 = 2.0;
 const TTFT_SLOW_FLOOR_MS: f64 = 500.0;
-/// Keep a small burst for a reusable prefix on its deterministic primary
-/// before spilling overlapping requests to the next backend. This preserves a
-/// single cache copy for ordinary traffic while preventing one hot prefix from
-/// monopolizing a replica. The counter is process-local and tracks only live
-/// requests; it never stores message content.
+/// Keep a small first-turn burst for a reusable prefix on its deterministic
+/// primary before spilling overlapping requests to another backend. This
+/// preserves a single cache copy for ordinary traffic while preventing one hot
+/// prefix from monopolizing a replica. The counter is process-local and tracks
+/// only live requests; it never stores message content.
 const PREFIX_AFFINITY_BURST: u32 = 4;
 
 #[derive(Default, Clone, Copy)]
@@ -73,8 +75,8 @@ fn lock<T>(m: &Mutex<T>) -> MutexGuard<'_, T> {
 
 type PrefixLoads = HashMap<u64, Vec<u32>>;
 
-/// Reservation for one live prefix-routed request. Releasing is RAII-based so
-/// cancellation, errors, and dropped streams cannot leak routing load.
+/// Reservation for one live affinity-routed request. Releasing is RAII-based
+/// so cancellation, errors, and dropped streams cannot leak routing load.
 pub(super) struct RouteLease {
     route_key: u64,
     index: usize,
@@ -122,9 +124,9 @@ pub(super) struct Fleet {
     /// that don't fit the rotation scheme (one-label host, IP literal, …) — then
     /// rotation is a no-op and the canonical-SNI path is used.
     rotation_parts: Option<rotation::UrlParts>,
-    /// Stateless first-message router. Its stable key is reduced modulo the
-    /// live backend count, so every Cloud API process gives a reusable prefix
-    /// the same primary backend; only overlapping hot-prefix bursts may spill.
+    /// Stateless prefix/conversation router. Stable keys are reduced modulo the
+    /// live backend count, so every Cloud API process agrees on first-turn
+    /// prefix primaries and established-conversation homes.
     pub(super) prefix_router: Arc<PrefixRouter>,
     /// Lazily-filled (or eagerly pre-created in legacy mode) per-backend-index
     /// clients, each pinning a persistent H2 connection to one verified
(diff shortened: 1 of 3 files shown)


PierreLeGuen:
Agreed on both trade-offs. I added a Known trade-offs section to the PR body
covering low-cardinality hash skew, modulo remapping on backend-count changes,
and transient mixed-version routing during rollout. This scoped change
intentionally follows the specified stateless key modulo live backend count:
cross-process affinity is the acceptance target, while bounded-load or
rendezvous hashing should be evaluated separately rather than widening this
patch. The original 64-prefix workload was chosen to reproduce and measure the
reported zero/high cache split, not to claim low-cardinality balance.

lloydmak99 (approved):
Change makes prefix affinity deterministic across API instances by hashing only the first message with SHA-256 instead of relying on per-instance trie bucket state. The core reduction is correct and the removed bucket machinery is fully cleaned up.

- The one production caller, `fleet.rs:173`, correctly reduces the new `u64` return via `(route(messages) % count as u64) as usize` — no truncation on 64-bit; all other `route()` call sites are tests and were updated.
- Removed symbols (`num_buckets`, `NUM_PREFIX_BUCKETS`, `PREFIX_MAX_MESSAGES`, `max_trie_depth`, `PrefixRouter::num_buckets()`) have no dangling references left in `crates/`.
- `prefix_router.rs`: hashing only `messages.first()` preserves the prior affinity contract, since the old trie already minted fresh buckets only for the first message; empty/multipart/non-text fallbacks are all deterministic.

Optional follow-up (non-blocking, already noted in the PR's "Known trade-offs"): with only a few distinct system prompts active, uniform hashing can strand a backend versus the old round-robin; rendezvous/bounded-load hashing would bound both skew and topology-churn remapping.

Checks: reviewed full diff, PR body, and prior review via `gh`, and statically verified caller/symbol/dep consistency with ripgrep. Local `cargo` build/tests skipped (no toolchain in environment); author reports `make preflight` green and 717/0 integration tests passing.

PierreLeGuen:
Follow-up validation is in progress for the low-cardinality concern raised
above. I am running privacy-safe matched tests on an isolated three-backend
DeepSeek-V4-Flash pool with 1–4 synthetic reusable prefixes, weighted
hot-prefix traffic, and increasing concurrency. Arms cover the current
process-local allocator, this PR modulo hash, single-copy rendezvous, bounded
two-/three-replica affinity, and round-robin control.

Artifacts persist only synthetic fingerprints, backend index,
timing/token/cache/status fields, and aggregate server metrics—never prompts,
responses, credentials, customer data, or host addresses. I will post the
measured cache/load/latency/throughput result here before changing the PR code
or recommending merge. No production routing or deployment is being changed.

PierreLeGuen:
Interim isolated-inference result: the review concern is valid. The current PR
is cache-correct, but fixed single-backend affinity can be a substantial
hot-prefix capacity regression.

With privacy-safe synthetic traffic against three isolated
production-equivalent TP2 replicas and four simulated Cloud API processes, the
one-prefix arm at concurrency 96 produced:

- deterministic modulo (this PR): 4/192 failures, p95 3.83s, 26.0 req/s
- current process-local routing: 0/192 failures, p95 1.70s, 51.4 req/s
- experimental three-backend bounded arm: 0/192 failures, p95 1.32s, 66.5 req/s

All three arms had the same 98.8% cached-token share after warm-up, so the
modulo arm paid the latency/throughput cost without gaining cache reuse in this
low-cardinality case. At concurrency 48, modulo was also 1.72s p95 / 26.4 req/s
versus 1.04s / 42.5 req/s for the bounded arm.

I am now testing prefix-scoped spillover with 64 reusable prefixes. The goal is
to keep ordinary prefixes single-copy while spreading only overlapping
requests for the same hot prefix. I will update the implementation only if that
arm preserves the cache-affinity win and removes the hot-prefix regression. No
customer data, prompts/responses, credentials, host addresses, or production
changes are involved.

PierreLeGuen:
Follow-up on the low-cardinality review: the concern was valid. Plain modulo
caused 4/192 failures for one hot prefix at concurrency 96, with p95 increasing
from 1.70 s to 3.83 s and throughput dropping from 51.4 to 26.0 req/s.

I tested prefix-local spill thresholds with both two and four simulated API
processes. A burst of four was the smallest strategy that had zero failures
across every tested workload. At concurrency 96:

- 64 prefixes / four processes: current 386/960 failures, 82.27% cached-token
  share, p95 3.60 s; burst four 0/960 failures, 98.78% cached-token share, p95
  1.69 s
- 64 prefixes / two processes: current 183/960 failures, 95.06% cached-token
  share, p95 2.67 s; burst four 0/960 failures, 98.68% cached-token share, p95
  1.60 s
- one hot prefix / four processes: burst four 0/192 failures, p95 1.59 s,
  69.6 req/s

Pushed `d0195095`. Normal requests stay on the deterministic primary; only
overlapping requests for the same prefix spill in deterministic groups of four.
Live counters contain only the 64-bit routing key and per-backend counts, and
are released on success, error, cancellation, and stream drop. Unrelated
prefixes do not affect each other.

The complete isolated preflight passes (717 integration tests passed, 11
intentionally ignored), as do 387 provider tests with one intentional ignore.
Backend-count churn is still a separate follow-up because the current discovery
contract does not expose stable backend identities.

PierreLeGuen:
@claude review

lloydmak99 (approved):
Replaces the process-local first-seen trie bucket allocator with a stateless SHA-256 routing key over the first message, plus bounded, prefix-local spillover (`PREFIX_AFFINITY_BURST = 4`) with RAII load accounting. The core change is sound: lock ordering is consistent (no nested-reverse acquisition, no deadlock), index access stays within bounds (`rotation_count()` clamps to `MAX_FANOUT`), and `RouteLease` is released on every path — stream end (`poll_next` `Ready(None)`), error/cancellation via `TtftProbe` drop, fallback `continue`/`return` via RAII, and non-streaming success via scope exit. Spillover determinism matches the tests, and the two fallback helpers correctly differ (`try_stream_fallback_indices` moves the lease into `TtftProbe`; `try_chat_completion_fallback_indices` holds a scoped `_route_lease` per iteration).

- `crates/inference_providers/src/attested/nearai/fleet.rs:236-238` — the empty-messages path returns `reserve_index(route_key, 0)` directly, bypassing `candidate_indices`, so empty-message requests always pin index 0 with no TTFT steering. Benign for a degenerate edge case; noting only for author awareness.

Checks: full static review of all changed files and every `route()`/`acquire_index`/`reserve_index` call site; confirmed no dangling references to removed trie symbols (`NUM_PREFIX_BUCKETS`, `PREFIX_MAX_MESSAGES`, `num_buckets`, `max_trie_depth`). `cargo` build/test skipped (no Rust toolchain in this environment); author reports `make preflight` green with 717 integration + 387 provider tests passing.

PierreLeGuen:
I agree that both main review points were valid, and 24548ecc addresses them.

- Colliding prefixes: the modulo primary is unchanged, but spill candidates are now ordered by a deterministic key-and-index SHA-256 score. In a forced three-prefix collision at concurrency 96, the prior ring produced 4/960 failures and 2.23x load skew; the decorrelated order produced 0/960 failures and 1.55x skew, with the same 98.79% cached-token share and slightly lower p95.
- Growing conversations: I reproduced the concern with 24 synthetic conversations over five turns. Burst-only routing moved 43.75% of transitions and returned 71.76% cached tokens. The final hierarchical path uses shared-prefix affinity for the initial request, then a stable first-two-message conversation home once assistant/tool history exists. It returned 78.60% cached tokens with zero failures; turns two through five never moved, and load was 8/7/9 across the three replicas.

I did not hash the entire growing message list because that would change the key every turn. The first two messages stay stable and give each established conversation a deterministic home across processes and restarts.

Local validation is green: 82 focused provider tests, strict all-target clippy/format, and the complete preflight with 717 integration tests and zero failures. The PR body now includes the sanitized aggregate evidence and remaining topology-churn trade-off.

PierreLeGuen:
@claude review

lloydmak99 (approved):
Deterministic prefix routing looks solid — routing keys are pure hashes, lease lifecycle is RAII-clean across every path, and the colliding-prefix and multi-turn affinity concerns from earlier revisions are resolved. Good to merge.

- Non-blocking (`fleet.rs:247-249`): established conversations now pin to `candidates[0]` with no burst-spill, so a single very hot multi-turn conversation could concentrate load on one backend. This is the intended affinity trade-off and is mitigated by the retryable-5xx/429 fallback that spills to other backends under overload — no action needed.

Checks: static inspection of all three changed files (`fleet.rs`, `prefix_router.rs`, `mod.rs`); traced every `RouteLease` path (streaming happy/fallback, non-streaming, cancellation) for leaks/double-counts; confirmed no dangling references to removed symbols and that `select_index` is now `#[cfg(test)]`-only. `cargo build`/`test` skipped (no Rust toolchain here); author reports `make preflight` green.

PR #907, opened 2026-07-29
fix: block deletion of staking-bound organizations

think-in-universe:
## Summary

Fixes nearai/house-of-stake-contracts#138.

- Adds a service-layer deletion guard for organizations with a staking-farm NEAR wallet source.
- Maps the guard to HTTP 409 with a clear `staking_wallet_bound` error response.
- Adds service-layer coverage for the blocked deletion path and API mapper coverage for the 409 response.

## Root Cause

Organization deletion only checked ownership before soft-deleting the organization. It did not check whether staking-farm credit accounting had permanently bound the organization to a NEAR wallet.

## Validation

- `cargo fmt`
- `cargo check -p database`
- `cargo check -p api`
- `cargo test -p services organization::tests::delete_organization_blocks_staking_wallet_bound_org_without_deleting`
- `cargo test -p api routes::common::tests::map_organization_error_staking_wallet_bound_returns_conflict`

CI follow-up:

- The initial PR added a default e2e test. The full e2e suite failed twice in unrelated tests with shared Postgres connection-close/retry flakes, so the default e2e addition was removed to avoid increasing e2e DB pressure on this fix PR.


think-in-universe:
@claude thanks for the detailed review.

Clarifying the intended product behavior: the permanent block is deliberate. If an organization is bound to a NEAR wallet, we should not allow deleting that organization, and we also should not allow disconnecting/unbinding the organization-wallet relationship through the API.

Given that, I do not think we should scope the deletion guard to only `active` staking farm sources. A `disconnected` status should not become a deletion escape hatch for this invariant. Any future disconnect/unbind flow should preserve this rule or be designed separately as an explicit product/admin operation.

Please review again with that intent in mind.

think-in-universe:
@claude I addressed the review feedback in `8a681517`.

Changes made:
- Moved the OpenAPI `409 staking_wallet_bound` response from `GET /v1/organizations/{org_id}` to the actual `DELETE /v1/organizations/{org_id}` endpoint.
- Updated `delete_organization` to delegate `NotFound`, `Unauthorized`, and `StakingWalletBound` through the shared `map_organization_error` path, avoiding duplicate response mapping drift.
- Added an in-process Axum route test that drives `DELETE /v1/organizations/{org_id}` and asserts the wire response is `409` with `error.type = staking_wallet_bound`.
- Documented that any NEAR staking wallet binding blocks deletion, including future disconnected sources.
- Renamed the SQL alias from `exists` to `has_source`.

Local verification passed:
- `cargo test -p api routes::organizations::tests::delete_organization_returns_staking_wallet_bound_conflict`
- `cargo test -p api routes::common::tests::map_organization_error_staking_wallet_bound_returns_conflict`
- `cargo test -p services organization::tests::delete_organization_blocks_staking_wallet_bound_org_without_deleting`
- `cargo check -p database`
- `cargo check -p api`

Please review again.

crates/services/src/organization/mod.rs
```diff
@@ -199,6 +199,17 @@ impl OrganizationServiceImpl {
             ));
         }
 
+        // Any NEAR staking wallet binding blocks organization deletion, even if
+        // a future flow marks the staking source disconnected.
+        if self
```
hanakannzashi: [P1] Serialize wallet linking with organization deletion

This check and the following delete run as separate repository calls/connections. A concurrent /staking/farm/sync can pass its active-organization validation, then deletion can observe no source and set the organization inactive, after which the sync can still insert a source because upsert_source and the FK only require the organizations row to exist, not is_active = true. Both requests then succeed, leaving the wallet permanently bound to an inactive organization—the exact state this PR is intended to prevent. Please serialize both paths with the same organization-row or advisory lock inside transactions, and re-check is_active in the source-upsert path under that lock. Folding NOT EXISTS into the UPDATE alone is insufficient because a source can still be inserted afterward.

think-in-universe:
@claude addressed this in `197ebef8`.

Added the invariant documentation at both places that matter for future maintenance:
- `OrganizationRepository::has_staking_farm_source` now states it is intentionally status-agnostic and must return true for any staking farm source row.
- The SQL implementation now has an inline comment before the status-agnostic `EXISTS` query.

Local verification passed:
- `cargo fmt`
- `cargo check -p database`
- `cargo test -p services organization::tests::delete_organization_blocks_staking_wallet_bound_org_without_deleting`
- `cargo test -p api routes::organizations::tests::delete_organization_returns_staking_wallet_bound_conflict`

Please review again.

crates/services/src/organization/mod.rs
think-in-universe: Resolved in `0e1f4517`.

What changed:
- Replaced the service-layer `has_staking_farm_source` + `delete` sequence with one repository operation, `delete_if_no_staking_farm_source`, so the check and soft-delete happen in a single transaction.
- The delete transaction now locks the active organization row with `SELECT ... FOR UPDATE`, checks for any staking farm source row under that lock, then soft-deletes.
- `upsert_source` now takes the same organization row lock first, re-checks `organizations.is_active = true` under the lock, and refuses to create/link a source if the org became inactive.
- Added `StakingFarmOrganizationInactive` and mapped it to a `409 organization_inactive` response for the sync endpoint.

With the shared row lock, both interleavings are serialized: upsert-first makes delete see the source and return `staking_wallet_bound`; delete-first makes upsert see the inactive org and return `organization_inactive`.

Local verification passed:
- `cargo fmt`
- `cargo check -p database`
- `cargo test -p services organization::tests::delete_organization_blocks_staking_wallet_bound_org_without_deleting`
- `cargo check -p api`
- `cargo test -p api routes::organizations::tests::delete_organization_returns_staking_wallet_bound_conflict`
- `cargo test -p api routes::staking_farm::tests::staking_farm_error_organization_inactive_returns_conflict`

Please review again.

think-in-universe:
@claude the P1 race from @hanakannzashi is fixed in `0e1f4517`.

Summary:
- Organization deletion now uses one transactional repository operation that locks the active org row, checks for any staking farm source, then soft-deletes under that same lock.
- Staking source upsert now locks the same org row first and re-checks `is_active = true` under the lock before inserting/linking.
- Delete-first now makes upsert return `409 organization_inactive`; upsert-first now makes delete return `409 staking_wallet_bound`.
- Added `StakingFarmOrganizationInactive` and route mapping for the sync endpoint.

Verification passed locally and GitHub CI is green on `0e1f4517`.

Please review again.

hanakannzashi (approved):

PR #909, opened 2026-07-29
feat: add Lukka AML account status gate

think-in-universe:
## Summary

Starts nearai/house-of-stake-contracts#56 for cloud-api by adding the server-side Lukka AML enforcement foundation for NEAR House-of-Stake flows.

This PR adds:
- Lukka-scoped environment-driven AML config, disabled safely unless `LUKKA_AML_ENABLED=true` and `LUKKA_AML_BEARER_TOKEN` are set
- configurable AML block policy via `LUKKA_AML_BLOCKED_RISK_LEVELS` and `LUKKA_AML_SCORE_BLOCK_THRESHOLD`; blocked risk levels support LOW/MEDIUM/HIGH only, UNKNOWN never blocks by risk level alone, and score blocking accepts `1..100` or `empty`/`none`/`disabled` to turn it off
- optional high-risk Slack webhook alerts via `LUKKA_AML_HIGH_RISK_SLACK_WEBHOOK_URL` with bounded timeout from `LUKKA_AML_HIGH_RISK_SLACK_TIMEOUT_MS` and fallback-alert dedupe from `LUKKA_AML_HIGH_RISK_SLACK_DEDUPE_SECONDS`
- short-lived UNKNOWN-result damping via `LUKKA_AML_UNKNOWN_CACHE_TTL_SECONDS`, while keeping UNKNOWN reports inactive and preserving stale-HIGH fallback checks
- a Lukka NEAR AML client that calls `/v3/reports/aml/score/{account_id}?address_type=NEAR`
- provider-independent AML normalization for LOW/MEDIUM/HIGH/UNKNOWN, including mismatch and provider failure handling
- Postgres persistence for AML reports and allowlisted NEAR accounts
- reusable active-report lookup based on the database `created_at` freshness window
- in-memory caching for reusable non-UNKNOWN reports
- session-authenticated `GET /v1/users/me/status` returning only `{ "status": "ok" }` or a neutral 403 `ErrorResponse` with `type: "account_error"` and `message: "Account error"`
- AML enforcement on the existing NEAR staking farm sync path before wallet-derived farm credit sync proceeds

## Notes

This is the first backend slice, not the entire issue-56 surface. Follow-up work still needs admin report/allowlist mutation APIs and additional gates as subscription/payment/withdraw/unlock route surfaces land in cloud-api.

## Validation

- `cargo fmt`
- `cargo test -p services aml::tests`
- `cargo test -p services sync_for_source_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p services stale_preflight_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p services sync_for_near_account_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p config aml_`
- `cargo test --no-run --lib --bins`
- `cargo deny --all-features check advisories`
- `cargo test --lib --bins`


think-in-universe:
@claude review

crates/api/src/routes/staking_farm.rs
```diff
@@ -154,6 +155,7 @@ pub async fn sync_organization_staking_farm(
 ) -> RouteResult<StakingFarmStateResponse> {
     let organization_id = parse_uuid(&org_id, "Invalid organization ID")?;
     require_near_default_org(&app_state, &user, organization_id).await?;
+    enforce_near_account_status(&app_state, &user, AmlFlow::StakingFarmSync).await?;
```
hanakannzashi: [P1] Enforce AML at the shared staking-farm sync boundary, not only this session route. The API-key usage preflight calls `sync_organization_if_stale`, and the admin route calls `sync_for_source` directly; both paths can refresh a non-allowlisted HIGH-risk source and increase its farm credits without this check. Move the gate into an AML-aware shared service path using the source's `near_account_id`, and cover every caller with tests.

crates/config/src/types.rs
```diff
@@ -57,12 +58,79 @@ impl ApiConfig {
...
         })
     }
 }
 
+/// Server-side AML/KYT policy for NEAR wallet-gated House-of-Stake flows.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AmlConfig {
+    pub enabled: bool,
+    pub lukka_base_url: String,
+    pub lukka_bearer_token: Option<String>,
```
hanakannzashi, line 73: [P1] Redact this credential from `Debug`. `AmlConfig` derives `Debug`, and `main.rs` logs the complete `ApiConfig` with `tracing::debug!("Config: {:?}", config)`, so debug logging emits the raw Lukka bearer token. Use a secret wrapper or manual redacted `Debug` implementation and add a regression test.

crates/config/src/types.rs
```diff
@@ -57,12 +58,79 @@ impl ApiConfig {
...
+
+impl AmlConfig {
+    pub fn from_env() -> Result<Self, String> {
+        let defaults = Self::default();
+        let token = env::var("LUKKA_AML_BEARER_TOKEN")
+            .ok()
+            .filter(|value| !value.is_empty());
+        let requested_enabled = parse_bool_env("AML_ENABLED", defaults.enabled)?;
+        let config = Self {
+            enabled: requested_enabled && token.is_some(),
```
hanakannzashi: [P1] Fail startup when AML is explicitly enabled without a usable token. As written, `AML_ENABLED=true` plus a missing/empty `LUKKA_AML_BEARER_TOKEN` starts successfully with `enabled=false`, silently removing all compliance enforcement. Only an unset/false AML flag should disable it; a requested-but-misconfigured gate should return a configuration error.

crates/services/src/aml.rs
```diff
@@ -0,0 +1,710 @@
...
+            .await
+            .map_err(|error| AmlError::RepositoryFailure(error.to_string()))?
+        {
+            if report.risk_level != AmlRiskLevel::Unknown {
+                self.cache.insert(account_id, report.clone()).await;
+            }
+            return self.decision_from_report(&report);
+        }
+
+        let stale_active_report = self
```
hanakannzashi: [P1] This pre-provider snapshot is unsafe under concurrent cache misses. Two requests (especially on separate replicas) can both read no prior report; one then commits HIGH, while the other commits UNKNOWN and uses its stale `None` snapshot to fail open even though HIGH is already active. Serialize/coalesce checks per account or re-read the latest active report after persisting UNKNOWN before allowing the request, and add a concurrent regression test.

hanakannzashi (commented):
Four blocking issues found in the AML gate implementation.

think-in-universe:
@claude review

crates/api/src/routes/staking_farm.rs
think-in-universe: Fixed in `d641c2e0`. I moved the staking-farm AML check into the shared `StakingFarmService::sync_for_source` path via an injected `StakingFarmAmlGate`, so the user sync route, admin sync route, and API-key preflight `sync_organization_if_stale` all hit the same source `near_account_id` gate before any contract call or credit update. I also removed the route-only duplicate gate and added coverage for direct source sync, stale preflight sync, and `sync_for_near_account` blocking before contract calls. CI is green.

crates/config/src/types.rs
think-in-universe: Fixed in `d641c2e0`. `AmlConfig` now has a manual `Debug` implementation that redacts `lukka_bearer_token`, with a regression test asserting the raw token is not present in formatted debug output. CI is green.

crates/config/src/types.rs
think-in-universe: Fixed in `d641c2e0`. `AmlConfig::from_env` now trims the Lukka token and returns a config error when `AML_ENABLED=true` is set without a non-empty `LUKKA_AML_BEARER_TOKEN`; unset/false AML still disables the gate. Added regression tests for both the misconfigured enabled case and the disabled-without-token case. CI is green.

crates/services/src/aml.rs
think-in-universe: Fixed in `d641c2e0`. The UNKNOWN path no longer relies on the pre-provider snapshot; after persisting an UNKNOWN report, it re-reads `latest_active_report` before failing open, so a concurrently committed HIGH active report blocks the request. Added a concurrent regression test with one HIGH provider result and one UNKNOWN result; both requests block. CI is green.

crates/api/src/routes/api.rs
```diff
@@ -168,6 +169,7 @@ pub fn build_management_router(app_state: AppState, auth_state: AuthState) -> Ro
     // User routes (require access token authentication)
     let user_routes = Router::new()
         .route("/me", get(get_current_user))
+        .route("/status", get(get_user_status))
```
think-in-universe: We should rename the route to `/me/status/`.

crates/services/src/aml.rs
```diff
@@ -0,0 +1,762 @@
...
+    ) -> Result<AmlDecision, AmlError> {
+        if auth_provider != "near" || provider_user_id.is_empty() {
+            return Ok(AmlDecision::Allowed);
+        }
+        self.check_near_account(Some(user_id), provider_user_id, flow)
+            .await
+    }
+
+    fn decision_from_report(&self, report: &AmlReport) -> Result<AmlDecision, AmlError> {
+        if report.risk_level == AmlRiskLevel::High {
```
think-in-universe: Can we make this configurable via env vars about ALM risk detection via level or scores?

think-in-universe:
Addressed the latest AML review feedback in `7a583c1f`.

What changed:
- Hardened Lukka normalization so non-decision metadata type drift, such as a bad `report_time` or numeric `report_id`, no longer discards a usable HIGH risk result.
- Added cscore fallback for unrecognized provider risk labels: high scores now normalize to HIGH and block instead of failing open; lower scores remain inactive UNKNOWN with an explicit `unrecognized_risk_level:<value>` reason.
- Preserved provider response bodies for address/address-type mismatch audit cases instead of replacing them with only `{"error": ...}`.
- Removed the dead `AmlDecision::Blocked` representation from the staking-farm gate path.
- Documented the AML/Lukka env vars in `env.example`.
- Corrected the PR description to match the actual neutral 403 `ErrorResponse` shape.

Validation passed locally:
- `cargo fmt`
- `cargo test -p services aml::tests`
- `cargo test -p services sync_for_source_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p services stale_preflight_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p services sync_for_near_account_blocks_aml_rejected_source_before_contract_call`
- `cargo test -p config aml_`
- `cargo test --no-run --lib --bins`

I did not fold the single-flight/negative-cache provider-outage mitigation into this commit because it changes retry/latency behavior on inference preflight and the review marked it non-blocking. I would track that as the follow-up production-safety item.

hanakannzashi (approved):

crates/api/src/routes/api.rs
think-in-universe: Fixed in `aca103c4`: the route is now registered as `/users/me/status`, and the OpenAPI annotation documents `GET /v1/users/me/status`.

crates/services/src/aml.rs
think-in-universe: Fixed in `aca103c4`: AML risk policy is now configurable via `AML_BLOCKED_RISK_LEVELS` for level-based blocking and `AML_SCORE_BLOCK_THRESHOLD` for cscore-based blocking. Defaults preserve the current HIGH / cscore >= 75 behavior, and config tests cover valid overrides plus invalid policy values.

think-in-universe:
Addressed the two latest AML follow-ups in `58c5f483`:

- Renamed the AML env surface so the configurable knobs are Lukka-scoped: `LUKKA_AML_ENABLED`, `LUKKA_AML_REPORT_REFRESH_DAYS`, `LUKKA_AML_MEMORY_CACHE_TTL_SECONDS`, `LUKKA_AML_REQUEST_TIMEOUT_SECONDS`, `LUKKA_AML_BLOCKED_RISK_LEVELS`, and `LUKKA_AML_SCORE_BLOCK_THRESHOLD`.
- Added optional high-risk Slack webhook support via `LUKKA_AML_HIGH_RISK_SLACK_WEBHOOK_URL` and `LUKKA_AML_HIGH_RISK_SLACK_TIMEOUT_MS`. Alerts are best-effort, use a timed reqwest client, redact the webhook URL from config debug output, and include audit context for provider-report and stale-report-fallback blocks.

Validation passed locally:
- `cargo fmt`
- `cargo test -p config aml_`
- `cargo test -p services aml::tests`
- `cargo test --no-run --lib --bins`
- `cargo test --lib --bins`

think-in-universe:
@claude review

think-in-universe:
Follow-up for CI in `73fcb339`: stabilized `auto_redact_bills_classify_pass` by polling for the spawned classify-billing row slightly longer than the route-side 10s billing timeout. The previous E2E failure showed the chat request succeeded but the test only waited about 2s for the background billing task.

Local validation for this test-only change:
- `cargo fmt`
- `cargo test --no-run --test e2e_all`
- `cargo clippy --test e2e_all -- -D warnings`

I could not execute the focused E2E locally because this machine's local Postgres rejects the configured `postgres` credentials; CI is the execution signal for that target.

think-in-universe:
Addressed in `db790161`.

Implemented the two blocking items from this pass:
- New provider reports are evaluated with `should_block_report` before the `UNKNOWN` fail-open path. This makes `LUKKA_AML_BLOCKED_RISK_LEVELS=UNKNOWN` effective when an operator opts into fail-closed behavior, and it lets high-score reports with missing/null `risk_level` block via `LUKKA_AML_SCORE_BLOCK_THRESHOLD`.
- Preserved outage fail-open by keeping provider failure reports at `score: None` unless the operator explicitly includes `UNKNOWN` in blocked risk levels.
- Added fallback Slack alert dedupe with a short TTL cache keyed by `(provider, action, account_id)`, exposed as `LUKKA_AML_HIGH_RISK_SLACK_DEDUPE_SECONDS` with default `300`.
- Rejected `LUKKA_AML_SCORE_BLOCK_THRESHOLD=0`; score blocking is now `1..100`, or disabled with empty/`none`/`disabled`. `env.example` calls that out.

Added/updated tests for the missing-risk-level score path, UNKNOWN-as-blocked policy, score threshold over UNKNOWN reports, outage fail-open with no score, and stale fallback Slack dedupe.

Validation run locally:
- `cargo fmt`
- `cargo test -p config aml_`
- `cargo test -p services aml::tests`
- `cargo test --no-run --lib --bins`
- `cargo test --lib --bins`
- `cargo deny --all-features check advisories`

For the non-blocking Slack egress/sign-off and carried-over production hygiene notes, I left those as explicit review/follow-up items rather than changing behavior in this patch.

think-in-universe:
@claude review

think-in-universe:
Fixed in `dd27f28a`.

Implemented the narrow production-safety mitigation for the `/v1/users/me/status` UNKNOWN loop:
- Added `LUKKA_AML_UNKNOWN_CACHE_TTL_SECONDS` with default `60` and validation range `1..=86400`.
- Added a separate short-lived `unknown_cache` in `AmlService`.
- After an UNKNOWN provider report has been persisted and confirmed not to block under policy, the service caches that fail-open UNKNOWN result briefly.
- On a cache hit, the service skips the Lukka call and avoids another `aml_reports` insert, but still runs `latest_active_report` so stale active HIGH fallback blocking is preserved.
- UNKNOWN reports remain `active: false`; this is only request damping, not a reusable active decision.

Added tests for repeated fail-open UNKNOWN requests producing one insert, and for cached UNKNOWN still blocking when a stale HIGH report exists. Also documented the new env var in `env.example` and updated the PR body.

Validation run locally:
- `cargo fmt`
- `git diff --check`
- `cargo test -p config aml_`
- `cargo test -p services aml::tests`
- `cargo test --no-run --lib --bins`
- `cargo test --lib --bins`
- `cargo deny --all-features check advisories` (`advisories ok`; existing yanked `spin` warnings unchanged)

think-in-universe:
@claude review

think-in-universe:
Addressed in `1bd8a3fa` with a policy adjustment: UNKNOWN no longer blocks by risk level alone.

Changes made:
- `LUKKA_AML_BLOCKED_RISK_LEVELS` now accepts only `LOW`, `MEDIUM`, and `HIGH`; `UNKNOWN` is rejected during config parsing.
- `should_block_report` defensively ignores `AmlRiskLevel::Unknown` for the risk-level list even if a test/manual config contains it.
- Score-threshold blocking remains independent, so an UNKNOWN report with an explicit score above `LUKKA_AML_SCORE_BLOCK_THRESHOLD` can still block by score.
- UNKNOWN reports are inserted into the short-lived `unknown_cache` before policy evaluation, so high-score UNKNOWN blocks are damped too: repeated checks block without another Lukka call or `aml_reports` insert.
- `env.example` and the PR body now document that UNKNOWN is never blocked by risk level alone.

Tests updated/added:
- config rejects `LUKKA_AML_BLOCKED_RISK_LEVELS=UNKNOWN`
- UNKNOWN risk level without score fails open even if manually configured in blocked risk levels
- high-score UNKNOWN blocks on both repeated requests with only one inserted report

Validation run locally:
- `cargo fmt`
- `git diff --check`
- `cargo test -p config aml_`
- `cargo test -p services aml::tests`
- `cargo test --no-run --lib --bins`
- `cargo test --lib --bins`
- `cargo deny --all-features check advisories` (`advisories ok`; existing yanked `spin` warnings unchanged)

PR #915, opened 2026-08-11
fix(api): bound HTTP metric route and method labels

lloydmak99:
## Problem

The HTTP metrics middleware labeled requests from the raw URI and only normalized UUIDs or a fixed set of known ID prefixes. That left other dynamic values, including the 32-hex signature route identifier and arbitrary scanner paths, in the `endpoint` metric label. It also emitted arbitrary HTTP extension methods verbatim. Both dimensions could create unbounded Prometheus cardinality and eventually drive the OpenTelemetry SDK into its overflow series.

## Changes

- Read Axum's `MatchedPath` extension before calling the next middleware and use its registered route template as the `endpoint` label.
- Emit one bounded `/unmatched` endpoint label for every request that does not match a registered route.
- Preserve the nine canonical HTTP methods (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, `HEAD`, `OPTIONS`, `TRACE`, `CONNECT`) and emit one `OTHER` method label for every extension method.
- Remove raw-path normalization and its UUID/prefix heuristics, so route-label safety no longer depends on anticipating identifier formats.
- Remove the now-unreferenced aggregate `ALL_PREFIXES` constant and its stale metrics-normalization documentation; individual resource prefix constants remain unchanged.
- Add router-surface tests for Axum's nested/fallback `Router::layer` behavior: `/v1/signature/{chat_id}` is recorded as its template, an arbitrary scanner path uses `/unmatched`, and a `PROPFIND` scanner request uses `OTHER` without emitting the raw method.
- Document at the production router assembly point that metrics must remain a `Router::layer`; moving it outside the Axum router would make `MatchedPath` unavailable and collapse all endpoints to `/unmatched`.

The companion dashboard/query/overflow alert changes are in [nearai/cvm-ansible-playbooks#472](https://github.com/nearai/cvm-ansible-playbooks/pull/472).

## Validation

- Failing-first regressions:
  - the signature route emitted its raw 32-hex identifier before the fix;
  - the scanner route emitted its raw path before the fix;
  - the extension-method test emitted raw `PROPFIND` before the method allowlist.
- `cargo check -p api --all-targets`
- `cargo clippy -p api --all-targets -- -D warnings`
- `cargo test -p api middleware::metrics::tests -- --nocapture` (3 passed)
- `cargo test -p api --lib --bins` (255 passed, 2 ignored)
- `cargo test --workspace --lib --bins`
- Workspace-wide search found no remaining `ALL_PREFIXES` consumer before removal.
- Independent exact-head goal/constraint review passed on `5bda1808dcefbdf87532261c80fe2315ac6e9e8f`.
- The explicitly re-triggered Claude review inspected the real production assembly at this head and reported no blocking issues.

## Rollout and risk

- This PR does not deploy Cloud API.
- After a Cloud API release, endpoint-label series intentionally change from concrete/heuristically-normalized paths to registered param-named route templates plus `/unmatched`; saved queries matching the old literal `{id}` value must migrate to the route-specific template.
- Canonical HTTP method series are unchanged. Any extension method series collapse into `OTHER` by design.
- Request handling and status codes are unchanged. The behavioral surface is metric labeling only.
- Existing raw series will age out under normal Prometheus retention.
- The companion monitoring PR makes status-based customer/error queries overflow-safe and adds bounded overflow visibility/alerting.

### DIFF
--- a/crates/api/src/lib.rs
+++ b/crates/api/src/lib.rs
@@ -1413,7 +1413,9 @@ pub fn build_app_with_config(
         .fallback(routes::unsupported::unknown_route)
         .method_not_allowed_fallback(routes::unsupported::method_not_allowed)
         .layer(cors)
-        // Add HTTP metrics middleware to track all requests
+        // Keep metrics on Router::layer: MatchedPath is populated inside Axum's
+        // router, and moving this middleware outside would label every route
+        // as /unmatched.
         .layer(from_fn_with_state(
             metrics_state,
             middleware::http_metrics_middleware,
--- a/crates/api/src/middleware/metrics.rs
+++ b/crates/api/src/middleware/metrics.rs
@@ -4,18 +4,22 @@
 //! - `cloud_api.http.requests` - Count of HTTP requests by method, endpoint, status
 //! - `cloud_api.http.duration` - Histogram of request durations by method, endpoint
 //!
-//! Endpoints are normalized to replace UUIDs with `{id}` to reduce cardinality.
+//! Endpoints use Axum's matched route templates. Requests that do not match a
+//! registered route use a single `/unmatched` bucket.
 
-use axum::{body::Body, extract::State, http::Request, middleware::Next, response::Response};
-use services::{
-    id_prefixes::ALL_PREFIXES,
-    metrics::{
-        consts::{
-            get_environment, METRIC_HTTP_DURATION, METRIC_HTTP_REQUESTS, TAG_ENDPOINT,
-            TAG_ENVIRONMENT, TAG_METHOD, TAG_STATUS_CODE,
-        },
-        MetricsServiceTrait,
+use axum::{
+    body::Body,
+    extract::{MatchedPath, State},
+    http::{Method, Request},
+    middleware::Next,
+    response::Response,
+};
+use services::metrics::{
+    consts::{
+        get_environment, METRIC_HTTP_DURATION, METRIC_HTTP_REQUESTS, TAG_ENDPOINT, TAG_ENVIRONMENT,
+        TAG_METHOD, TAG_STATUS_CODE,
     },
+    MetricsServiceTrait,
 };
 use std::sync::Arc;
 use std::time::Instant;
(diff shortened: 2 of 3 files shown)


lloydmak99:
/ocr

@claude review

PierreLeGuen (approved):
This bounds the HTTP metric `route` and `method` labels: route labels now use Axum's `MatchedPath` templates plus a single `/unmatched` bucket, and methods collapse to 9 canonical verbs plus `OTHER`. The mechanism checks out against the router assembly (`Router::layer` at `lib.rs:1419` covers both the fallback and method-not-allowed fallback, so `MatchedPath` is populated for matched, unmatched, and 405 requests) and axum 0.8's `MatchedPath` extension API. Wildcard routes (`/models/{*model_id}`, `/threads/{*path}`) resolve to fixed template strings, so cardinality stays bounded. Pure metric-label change — request handling and status codes are unchanged, and it removes raw chat IDs and scanner paths from labels.

- `crates/api/src/middleware/metrics.rs:33` — `bounded_method` matches on `method.as_str()` literals rather than `Method::GET | ...` variants; behaviorally identical, but variant matching would make a typo a compile error instead of a silent fall-through to `OTHER`.
- Tests at `crates/api/src/middleware/metrics.rs:87+` build a synthetic router rather than exercising the real assembly, so hoisting the metrics layer outside `Router::layer` would make `MatchedPath` vanish with all tests still green. The guard comment at `lib.rs:1416-1418` documents but doesn't enforce this.
- Operator note: `endpoint` values shift from `{id}`-normalized paths to param-named templates (`{chat_id}`, …); dashboards/alerts keyed on the literal `{id}` will break rather than degrade — ensure the companion monitoring PR accounts for it.

Checks: `cargo test -p api middleware::metrics::tests` (3/3 passed), `cargo clippy -p api -p services` clean, `cargo fmt --check` passed, and a workspace-wide grep confirmed no remaining `ALL_PREFIXES` consumers.

PR #865, opened 2026-07-06
fix(chutes): resync golden measurements — restores GLM-5.2/GLM-5.1 Chutes fallback

Evrard-Nil:
## Context

During tonight's outage (2026-07-06 ~20:49–21:03 UTC, NEAR-fleet backends unreachable), the Chutes fallback for `z-ai/glm-5.2` engaged but **failed closed on every instance**:

```
measurement register-pin: observed measurements match no accepted Chutes config (mrtd=261ce538…, rtmr0=35038cbb…)
```

Chutes promoted the GLM-5.2 chute off `v1.3.1-rc1` onto the **final v1.3.1** release, and re-measured/re-published its hardware rows after our last snapshot (#849, 2026-06-24). The failing row (`8xb200 [10.2.1]`, RTMR0 `35038cbb…`) is exactly the one identified in the 2026-07-02 GLM-5.2 incident; it was never added. GLM-5.1 was partially affected too (its fleet now spans four v1.3.1 h200 rows, two of which were unpinned — observed live: `8xh200 [10.1.0-flat]`).

## Change

Resync `vetted_golden_measurements()` with the currently published `GET https://api.chutes.ai/servers/tee/measurements`:

- **v1.3.0**: add `8xh200-r2` (new published row).
- **v1.3.1**: add the six missing hardware rows (`8xh200 [10.2.1, NVSW0]`, `8xh200 [10.1.0, NVSW0]`, `8xh200 [10.1.0-flat]`, `8xb200 [10.2.1]`, `8xb200 [10.2.1, IB4]`, `8xb200 [10.2.1, XEON6]`); correct `8xb300`'s RTMR0 (re-measured upstream); rename `8xb200-xeon6` → its published name `8xb200 [10.1.0, XEON6, SNC3]` (same RTMR0).
- **Drop the stale plain `8xb200`/`8xb300` rows from the final-v1.3.1 family** — they were `-rc1` hardware values carried over on the assumption they'd persist into the final release; Chutes never published them for final v1.3.1 (fail-closed hygiene: don't accept combinations Chutes never published).
- **v1.3.1-rc1**: unchanged (still published).

Software identities (MRTD/RTMR1/RTMR2/runtime-RTMR3) are unchanged from the published values; only hardware RTMR0 rows changed. Total pinned rows: 14 → 20.

## Vetting evidence

[...]

PR #918, opened 2026-08-12
fix(chutes): resync v1.3.1 golden measurements — 2026-08-12 drift (9 new hardware rows)

lloydmak99:
## RCA

infra-tests' `test_chutes_attestation_path_is_alive[z-ai/glm-5.2]` is failing intermittently (history: 475/1028 runs ≈ 46%) with:

```
all candidate Chutes instances failed attestation; last: instance
e72047ab-c565-4a14-8699-2bb176e1dff6 attestation failed: measurement
register-pin: observed measurements match no accepted Chutes config
```

This is the third resync of `vetted_golden_measurements()` (after #849 and #865). I fetched the live `GET https://api.chutes.ai/servers/tee/measurements` on 2026-08-12 and diffed it against the pinned table. GLM-5.2's previously-fixed row (`8xb200 [10.2.1]`, from #865) still byte-matches correctly — this is the *next* wave of drift: Chutes has published 9 more v1.3.1 hardware rows that aren't pinned yet (new B200 CPU/NUMA variants, a new H200 FLAT variant, and several new RTX_PRO_6000 variants). Some live Chutes instances for GLM-5.1/5.2 get scheduled onto these newer nodes and are rejected, which is why the failure is intermittent rather than a hard outage.

This is exactly the drift class #865's own follow-up note flagged ("continuous detection so published-row drift can't silently kill the fallback again") — the per-model infra-tests probe *is* that follow-up detector, working as designed.

## Change

Purely additive, one file (`crates/services/src/attestation/chutes.rs`):
- Appended 9 new `(name, rtmr0)` hardware rows to the existing `"1.3.1"` `Family` — all share the same already-pinned MRTD/RTMR1/RTMR2/runtime-RTMR3 software identity; only RTMR0 (the per-hardware boot register) differs.
- No existing row, family, or software identity touched. `covers_the_full_v130_hardware_family`'s total-row assertion updated from 20 → 29 to match.
- Added `accepts_the_2026_08_12_observed_new_v131_hardware_rows`, asserting each new RTMR0 verifies against the shared family identity.
- Updated the doc comments to record this as resync #3, with source/date.

One of the 9 (`8xRTX_PRO_6000 [10.1.0]`) is a byte-for-byte duplicate RTMR0 of the already-pinned plain `8xRTX_PRO_6000` row (Chutes appears to be mid-rename, same pattern as the `8xb200-xeon6` → `8xb200 [10.1.0, XEON6, SNC3]` rename in #865). Verified `ChutesMeasurementPolicy::verify()` matches on register bytes, never on `name`, so pinning both labels for the same measurement is harmless.

## Verification

- `cargo fmt --all -- --check` clean.
- `cargo test -p services --lib chutes`: 26 passed, 0 failed (including the new test and the updated full-family count).
- `cargo test -p services --lib attestation`: 107 passed, 0 failed, 1 ignored (pre-existing, unrelated).

### DIFF
--- a/crates/services/src/attestation/chutes.rs
+++ b/crates/services/src/attestation/chutes.rs
@@ -257,6 +257,12 @@ impl ChutesInstanceVerifier for ChutesBackendVerifier {
 /// transparency endpoint — these rows merely enumerate which genuine software
 /// identities we accept.
 ///
+/// Resync #3 (after #849 and #865) fetched the published measurements from
+/// `GET https://api.chutes.ai/servers/tee/measurements` on 2026-08-12. It is
+/// purely additive for newly provisioned hardware: the new rows retain the same
+/// v1.3.1 MRTD/RTMR1/RTMR2/runtime-RTMR3 software identity and differ only in
+/// per-hardware RTMR0.
+///
 /// NOT included (fail-closed): the v1.0.0–v1.2.0 rows publish RTMR3 = all-zeros
 /// (a boot template, never matchable against a live extended RTMR3 — the running
 /// app is unmeasured). Both v1.3.1 and v1.3.1-rc1 publish a non-zero runtime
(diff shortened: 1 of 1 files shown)


PierreLeGuen (approved):
Clean, purely-additive resync of the Chutes golden-measurement allow-list: 9 new hardware rows, count assertion updated 20→29, one new acceptance test. `verify()` matches on register bytes only, so extra rows only ever admit more genuine hardware — fail-closed semantics and existing identities are untouched. No data writes, API contract change, or logging change; values are public transparency-endpoint data.

Non-blocking follow-up (optional):
- `crates/services/src/attestation/chutes.rs:347` — the added `8xRTX_PRO_6000 [10.1.0]` row is a byte-identical RTMR0 duplicate of the pre-existing `8xRTX_PRO_6000` row (line 340). Because `verify()` returns the first `.find()` byte-match, this row is unreachable, and the new test (line 634) asserts the old name — so it exercises the existing row, not the new one. Harmless as merged (registers already match), but consider dropping it or annotating the intentional mid-rename pin.

Checks: `cargo fmt --all -- --check` clean; `cargo test -p services --lib chutes` 26 passed; `cargo test -p services --lib attestation` 107 passed / 1 ignored; count math verified (6+3+20=29) and the only intra-family RTMR0 duplicate is the intentional pair above. RTMR0 values could not be re-fetched from the live Chutes endpoint (no external network), matching the author's sourced snapshot.

PR #919, opened 2026-08-12
fix(build): bump stale ca-certificates pin (reproducible-build guard drift)

lloydmak99:
## RCA

The "Verify Reproducible Build" workflow failed on the main push that merged #918 (a Chutes attestation golden-measurements resync) — **unrelated to that PR's content**. It's the `Guard - committed pins match resolved pins` step in `.github/workflows/verify-reproducible-build.yml` (https://github.com/nearai/cloud-api/actions/runs/31638680304):

```
--- pinned-packages-runtime.txt
+++ pinned-packages-runtime.resolved.txt
@@ -4,7 +4,7 @@
 base-passwd=3.6.1
 bash=5.2.15-2+b9
 bsdutils=1:2.38.1-5+deb12u3
-ca-certificates=20230311+deb12u1
+ca-certificates=20250419~deb12u1
 coreutils=9.1-1
```

This is the same recurring class #843 fixed once before for OpenSSL ("break apt-pin round-trip cycle that kept breaking reproducible builds"). The runtime Dockerfile stage bootstraps `ca-certificates` via `apt-get install -y ca-certificates` against the base image's live/default apt source — needed just to fetch the frozen `snapshot.debian.org` sources.list over HTTPS in the first place — before the pinned-preferences mechanism takes over. That bootstrap version drifts forward over calendar time independent of the committed pin, regardless of what the frozen snapshot itself contains.

The "Build reproducible image" step itself passed — only the guard/diff step failed — so `ca-certificates=20250419~deb12u1` is exactly what the current pipeline already resolves to and successfully installs; this bumps the commit to match a value CI has already proven safe.

## Change

One line in `pinned-packages-runtime.txt`:
```
-ca-certificates=20230311+deb12u1
+ca-certificates=20250419~deb12u1
```

`pinned-packages-builder.txt` is untouched — its guard diff passed cleanly in the same failed run, since the builder stage never reinstalls `ca-certificates` (inherited untouched from the exact-digest-pinned `rust:1.92.0-bookworm` base image).

## Not done here (flagging, not fixing)

I looked at *why* `Pin-Priority: 1001` didn't force a downgrade back to the old committed version during the pinned reinstall (per `apt_preferences(5)`, priority >1000 should force this). I couldn't confirm from outside CI whether the old exact build is still present in the frozen snapshot's `debian-security`/point-release pool. This is now the **second** time this exact guard has broken on pin drift (OpenSSL via #843, now ca-certificates) — a real root-cause fix (e.g. periodic scheduled pin refresh, or fixing the bootstrap-before-pin ordering) would prevent a third recurrence, but that needs its own careful, independently-verified investigation rather than a guess bundled into this hotfix.

## Verification

- `git diff --stat`: 1 file changed, 1 insertion(+), 1 deletion(-).
- `pinned-packages-builder.txt` and `Dockerfile` confirmed byte-identical to main (hash-verified).
- No local build available to re-run the self-hosted CI step; the bumped value is the exact one CI's own guard reported as the correct resolved version.


PierreLeGuen (approved):
Single-line data change bumping the stale `ca-certificates` pin in `pinned-packages-runtime.txt` (`20230311+deb12u1` → `20250419~deb12u1`) to match the value CI's reproducibility guard already resolves and installs. The guard diffs each pin file against its own resolved output independently, so leaving `pinned-packages-builder.txt` untouched is correct, and the "Build reproducible image" step passed in the failing run with this exact version. No correctness, security, or contract risk.

Optional follow-up (non-blocking, pre-existing, out of scope for this hotfix):
- The bumped value postdates the frozen snapshot timestamp, so the guard can drift again when Debian publishes a newer `ca-certificates`. Already documented in the PR body; not introduced here.

Checks: confirmed via `gh pr diff`/`gh pr view` that the diff is exactly one line and the guard (`.github/workflows/verify-reproducible-build.yml`) only runs on `main`/`workflow_dispatch` (not this PR); traced pin usage through `Dockerfile` and `build-image.sh`; `git diff --check` passed on a clean worktree. Local reproducible-build skipped (self-hosted CI infra; no Rust code changed, cargo audit/tests already green).

